PR #13, opened 2020-01-30
Add dag_longest_path function

mtreinish:
Previously retworkx only exposed the dag_longest_path_length() function
which gave the pairwise length of the longest path. However, there are
some cases where that path needs to be exposed itself. This commit adds
a new function dag_longest_path() to address that use case which instead
of returning the pairwise length it returns a list of node indexes for
that path.


mtreinish:
This is failing because the longest path returned is missing the last element in it. I'm not sure where it's going wrong, but it's something that will need to be addressed prior to merging

PR #16, opened 2020-02-07
Add CI jobs for other architectures

mtreinish:
This commit adds ci jobs for testing on non-x86 architectures. Non-x86
architectures will only run a single python 3.7 job. Now that the
manylinux2014 spec [1] exists we can start publishing wheels for non-x86
architectures. Adding test jobs is the first step towards that goal.

[1] https://www.python.org/dev/peps/pep-0599/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,7 @@ jobs:
     - name: Python 3.7 Tests s390x Linux
       python: 3.7
       arch: s390x
+      dist: bionic
       env:
         - RUST_BACKTRACE=full
     - name: Python 3.7 Tests arm64 Linux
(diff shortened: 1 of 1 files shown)


mtreinish:
All the new jobs worked except for s390x, I opened an thread on the travis community forum about this here: https://travis-ci.community/t/installing-rust-nightly-via-rustup-on-s390x-fails/7142 it might be a bug in rustup or an issue with the configuration of the s390x travis environment. Until there is a definitive answer I'm going to mark the jobs as allowed to fail and add the jobs for arm and ppc

mtreinish:
It looks like switching to ubuntu 18.04 fixed the s390x jobs. So I've updated all the new non-x86 jobs to use bionic just to be consistent between all the environments

PR #17, opened 2020-02-09
Add bfs_successors function

mtreinish:
This commit adds a bfs_successors function that mirrors the same
function from networkx. It leverages petgraph's bfs visitor struct which
makes the implementation straightforward. There is an outerloop that
does the graph traversal using the Bfs struct and then an inner loop to
return all the children from that node.


mtreinish:
Still needs tests before this can merge

PR #25, opened 2020-02-11
find adjacent node by edge

gadial:
This PR adds a method to the PyDAG class that is used to traverse the edges from a given node that
match a given function. The goal here was to provide a function for users to find a target node in
a graph given an edge that meets certain conditions. For example, to replace loops like:

https://github.com/Qiskit/qiskit-terra/blob/0.14.1/qiskit/dagcircuit/dagcircuit.py#L1300-L1314


mtreinish:
@gadial thanks for this, it looks like this is failing cargo fmt check. Can you either make the changes suggested here: https://travis-ci.com/mtreinish/retworkx/jobs/286041461#L289 or run `cargo fmt` to have cargo make these changes automatically. 

Also if it's not too much trouble could you add a test for this it'd be good to have coverage for this method.

PR #28, opened 2020-02-18
Make cycle checking opt-in

mtreinish:
Currently the PyDAG class always checks on add_edge() calls that the new
edge would not create a cycle in the dag. However, this comes at a
performance cost, and sometimes depending on the size and complexity of
the dag this can be large. To avoid this performance penalty by default
this commit changes the PyDAG class to not do this checking by default.
It adds a new attribute check_cycle to the class, if this is set to True
then cycle checking is enabled on subsequent calls to add_edge().

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -8,7 +8,18 @@ retworkx API
    the ``StableGraph`` type. The limitations and quirks with this library and
    type dictate how this operates. The biggest thing to be aware of when using
    the PyDAG class is that an integer node and edge index is used for accessing
-   elements on the DAG, not the data/weight of nodes and edges.
+   elements on the DAG, not the data/weight of nodes and edges. By default the
+   PyDAG realtime cycle checking is disabled for performance, however you can
+   opt-in to having the PyDAG class ensure that no cycles are added by setting
+   the ``check_cycle`` attribute to True. For example::
+
+       import retworkx
+       dag = retworkx.PyDAG()
+       dag.check_cycle = True
+
+   With check_cycle set to true any calls to :method:`PyDAG.add_edge` will
+   ensure that no cycles are added, ensuring that the PyDAG class truly
+   represents a directed acyclic graph.
 
      .. note::
           When using ``copy.deepcopy()`` or pickling node indexes are not
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -303,6 +303,11 @@ impl PyDAG {
 
     #[setter]
     fn set_check_cycle(&mut self, value: bool) -> PyResult<()> {
+        if !self.check_cycle && value {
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -418,13 +431,23 @@ impl PyDAG {
...
-        let should_check_for_cycle =
-            must_check_for_cycle(self, p_index, c_index);
-        let state = Some(&mut self.cycle_state);
-        if should_check_for_cycle
-            && algo::has_path_connecting(&self.graph, c_index, p_index, state)
-        {
-            Err(DAGWouldCycle::py_err("Adding an edge would cycle"))
+        if self.check_cycle {
+            let should_check_for_cycle =
+                must_check_for_cycle(self, p_index, c_index);
```
lcapelluto: remind me what this does?

mtreinish: `must_check_for_cycle` basically just looks if the parent node has parents, the child node has children, and that there is not already an edge from parent to child. If any of these are not true we can skip the cycle checking because there is no way to introduce a cycle.

Although, now that I think about it there is a potential bug by making this opt-in where you could add an edge without the cycle checking enabled that that introduces a cycle, then set `self.check_cycle=True` and never catch it because this will skip the cycle detection where there is an existing edge. I'm not sure how to best handle this, maybe run `is_directed_acyclic_graph()` in the setter if it goes from False to True?

docs/source/api.rst
```diff
@@ -8,7 +8,18 @@ retworkx API
...
+   opt-in to having the PyDAG class ensure that no cycles are added by setting
+   the ``check_cycle`` attribute to True. For example::
+
+       import retworkx
+       dag = retworkx.PyDAG()
+       dag.check_cycle = True
+
+   With check_cycle set to true any calls to :method:`PyDAG.add_edge` will
+   ensure that no cycles are added, ensuring that the PyDAG class truly
+   represents a directed acyclic graph.
```
kdk: One possible source of confusion from this structure arises if a user toggles `check_cycle`. e.g. if a user sets `check_cycle=False`, introduces a cycle (on purpose or by mistake), sets `check_cycle=True`, edge additions will (currently) check that no additional cycles are introduced, but it is still possible the graph is cyclic.

If this is the desired behavior, we should document that the user is responsible for checking for existing cycles when setting `check_cycle=True` (and maybe find a more specific name).

Otherwise, the `check_cycle` setter could search for cycles when being set to `True`, or `check_cycle` could be a static property of an instance.

mtreinish: Done in: https://github.com/mtreinish/retworkx/pull/28/commits/2b8e8ef80536f3eb3f750fbbbf57a1d70739d3ff

PR #30, opened 2020-02-20
Add layers function

mtreinish:
This commit adds a function to get layers from a PyDAG object. You
provide it with an inital set of indexes for the first layer and it will
return a list of lists with the node data, where each inner list
represents a layer.


mtreinish:
I still need to add tests and docs to this before it can merge

PR #36, opened 2020-02-28
Add option to cache sort keys in lexicographical_topological_sort

mtreinish:
Calling out to python to get a sort key can be expensive and is easily
the slowest part of the lexicographical_topological_sort function.
However, in some use cases the sort keys are guaranteed to be static for
the life of a node in the DAG. In such cases there is a performance
benefit from caching the result so that we only have to call out to
python once for each node instead. This commit adds a new kwarg for the
lexicographical_topological_sort function, use_cache, which when set to
true will cache the sort key returned for a dag.


mtreinish:
This is superseded by: https://github.com/Qiskit/qiskit-terra/pull/4040 and does not have the limitations on consecutive runs that this would impose.

PR #37, opened 2020-02-28
[WIP] Add lexicographical_topological_sort_py_cmp

mtreinish:
This commit adds a new function lexicographical_topological_sort_py_cmp
that uses a python comparison to determine sort order.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,7 @@ extern crate pyo3;
 mod dag_isomorphism;
 
 use std::cmp::Ordering;
+use std::cmp::{Eq, PartialEq};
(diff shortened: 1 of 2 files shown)


mtreinish:
This is mostly a test to see if doing this or the str comparison makes more sense

mtreinish:
After doing some local testing with this function in qiskit-terra this function is significantly slower than the existing function. This took 265.5 seconds while for the same sorts for the lambda function version took 4 secs, and with #36 (and using that option) it took 3 secs.

mtreinish:
The performance here is too slow to be worth pursuing this any further. Closing since it's not a realistic approach.

PR #44, opened 2020-03-16
Preserve python exceptions from node comparison

mtreinish:
When we call out to python as part of
dag_isomorphism::is_isomorphic_matching() we previously were not
handling any exceptions raised by the python code. This would result in
a panic and backtrace from rust because an unhandled err condition got
returned. This commit addresses this edge case by changing the return
types through the entire is_isomorphic_matching() call path to ensure
any python errors get returned up to the caller so that the pyfunction,
is_isomorphic_node_match(), will be able to properly return the python
error to python. This way we're no longer supressing the python
exception if/when a python callable raises when trying to compare 2
nodes.


mtreinish:
The tests failures on appveyor/pr are because I had to manual restart the jobs because of a networkx timeout. This caused the wheel jobs to trigger which I then had to cancel (since they only should run at tag time). However, if I did not cancel it, then it would have failed because of twine upload failures. They can safely be ignored.

PR #74, opened 2020-06-02
Add support for deepcopy/pickle to PyGraph

mtreinish:
The PyGraph class was missing a __getstate__ and __setstate__ function
which were needed to enable support for pickling and deepcopy. This
commit adds the missing methods so PyGraph objects can be deepcopied.
One note on the test is that there aren't isomorphism functions that
support the pygraph type yet due to typing constraints a new trait will
need to be added for both PyGraph and PyDAG to leverage the existing
function. Until then the tests just assert that deepcopy() doesn't raise
and returns a PyGraph object.


mtreinish:
The loop to remove holes in the node indexes doesn't work yet so marking as on hold until that is fixed

PR #75, opened 2020-06-03
Add adjacency_matrix functions

mtreinish:
This commit adds 2 new functions to the retworkx module,
graph_adjacency_matrix and dag_adjacency_matrix for returing the
adjacency matrix as a numpy array for PyGraph and PyDAG respectively.

- [x] Add tests
- [x] Add docs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,9 +16,9 @@ name = "retworkx"
 crate-type = ["cdylib"]
 
 [dependencies]
-petgraph = "0.5"
+petgraph = "0.5.1"
 fixedbitset = "0.2.0"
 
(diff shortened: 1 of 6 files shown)


mtreinish:
I haven't tested this at all yet, so I assume it doesn't actually work yet. But this is a starting point

mtreinish:
The ci failure here is unrelated. It looks like `pip install numpy` is failing on the s390x environments because of a permissions or space issue so it never finishes the pip install. I'll push up a separate PR to allow failures on that job to unblock this. s390x support isn't critical and we're limited by the ci resources available to actually test it.

PR #77, opened 2020-06-04
Add graph greedy color function

mtreinish:
This commit adds a new function for getting the greedy coloring using a
largest first strategy.

This PR is based on top of #75 and will have to be manually rebased after #75 merges (this was needed to get the changes to enable using PyGraph in a function outside of the graph module didn't want to duplicate them here).

- [x] Add tests (and verify it works)
- [x]  Add docs

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -920,8 +920,8 @@ fn graph_greedy_color(
 ) -> PyResult<PyObject> {
     let mut colors: HashMap<usize, usize> = HashMap::new();
     let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
-    node_vec.sort_by_key(|k| graph.graph.edges(*k).count());
-    node_vec.reverse();
+    let node_count = graph.graph.node_count();
+    node_vec.sort_by_key(|k| node_count - graph.graph.edges(*k).count());
     for u_index in node_vec.iter() {
         let mut neighbor_colors: HashSet<usize> = HashSet::new();
         let u = u_index.index();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -899,22 +899,20 @@ fn graph_greedy_color(
     let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
     node_vec.sort_by_key(|k| graph.graph.edges(*k).count());
     node_vec.reverse();
-
-    for u in node_vec.iter().map(|x| x.index()) {
-        let mut neighbor_colors: HashMap<usize, usize> = HashMap::new();
-        let u_index = NodeIndex::new(u);
-        for edge in graph.graph.edges(u_index) {
+    for u_index in node_vec.iter() {
+        let mut neighbor_colors: HashSet<usize> = HashSet::new();
+        let u = u_index.index();
+        for edge in graph.graph.edges(*u_index) {
             let target = edge.target().index();
-            let source = edge.source().index();
-            if colors.contains_key(&target) && target != u {
-                neighbor_colors.insert(target, *colors.get(&target).unwrap());
-            } else if colors.contains_key(&source) && source != u {
-                neighbor_colors.insert(source, *colors.get(&source).unwrap());
-            }
+            let existing_color = match colors.get(&target) {
+                Some(node) => node,
+                None => continue,
+            };
+            neighbor_colors.insert(*existing_color);
         }
         let mut count: usize = 0;
         loop {
-            if !neighbor_colors.contains_key(&count) {
+            if !neighbor_colors.contains(&count) {
                 break;
             }
             count += 1;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -925,6 +925,8 @@ fn graph_greedy_color(
     let node_count = graph.graph.node_count();
     let mut sort_map: HashMap<NodeIndex, usize> = HashMap::new();
     for k in node_vec.iter() {
+        // The intent behind this sort key is to sort nodes in descending order
+        // of degree
         sort_map.insert(*k, node_count - graph.graph.edges(*k).count());
     }
     node_vec.sort_by_key(|k| sort_map.get(k));
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -887,6 +887,44 @@ fn lexicographical_topological_sort(
     Ok(PyList::new(py, out_list).into())
 }
 
+#[pyfunction]
+fn graph_greedy_color(
+    py: Python,
+    graph: &graph::PyGraph,
+) -> PyResult<PyObject> {
+    let mut colors: HashMap<usize, usize> = HashMap::new();
+    let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
+    node_vec.sort_by_key(|k| graph.graph.edges(*k).count());
+    node_vec.reverse();
+
+    for u in node_vec.iter().map(|x| x.index()) {
+        let mut neighbor_colors: HashMap<usize, usize> = HashMap::new();
+        let u_index = NodeIndex::new(u);
+        for edge in graph.graph.edges(u_index) {
+            let target = edge.target().index();
+            let source = edge.source().index();
+            if colors.contains_key(&target) && target != u {
+                neighbor_colors.insert(target, *colors.get(&target).unwrap());
+            } else if colors.contains_key(&source) && source != u {
+                neighbor_colors.insert(source, *colors.get(&source).unwrap());
+            }
+        }
+        let mut count: usize = 0;
+        loop {
+            if !neighbor_colors.contains_key(&count) {
+                break;
+            }
+            count += 1;
+        }
+        colors.insert(u, count);
+    }
+    let out_dict = PyDict::new(py);
+    for (index, color) in colors {
+        out_dict.set_item(index, color)?;
+    }
+    Ok(out_dict.into())
+}
+
 // Find the shortest path lengths between all pairs of nodes using Floyd's
 // algorithm
 // Note: Edge weights are assumed to be 1
@@ -1087,6 +1125,7 @@ fn retworkx(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
     m.add_wrapped(wrap_pyfunction!(layers))?;
     m.add_wrapped(wrap_pyfunction!(dag_adjacency_matrix))?;
     m.add_wrapped(wrap_pyfunction!(graph_adjacency_matrix))?;
+    m.add_wrapped(wrap_pyfunction!(graph_greedy_color))?;
     m.add_class::<PyDAG>()?;
     m.add_class::<graph::PyGraph>()?;
     Ok(())
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -923,11 +923,14 @@ fn graph_greedy_color(
     let mut colors: HashMap<usize, usize> = HashMap::new();
     let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
     let node_count = graph.graph.node_count();
-    node_vec.sort_by_key(|k| node_count - graph.graph.edges(*k).count());
-    for u_index in node_vec.iter() {
+    let mut sort_map: HashMap<NodeIndex, usize> = HashMap::new();
+    for k in node_vec.iter() {
+        sort_map.insert(*k, node_count - graph.graph.edges(*k).count());
+    }
+    node_vec.sort_by_key(|k| sort_map.get(k));
+    for u_index in node_vec {
         let mut neighbor_colors: HashSet<usize> = HashSet::new();
-        let u = u_index.index();
-        for edge in graph.graph.edges(*u_index) {
+        for edge in graph.graph.edges(u_index) {
             let target = edge.target().index();
             let existing_color = match colors.get(&target) {
                 Some(node) => node,
@@ -942,7 +945,7 @@ fn graph_greedy_color(
             }
             count += 1;
         }
-        colors.insert(u, count);
+        colors.insert(u_index.index(), count);
     }
     let out_dict = PyDict::new(py);
     for (index, color) in colors {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -970,8 +970,6 @@ fn graph_greedy_color(
     let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
     let mut sort_map: HashMap<NodeIndex, usize> = HashMap::new();
     for k in node_vec.iter() {
-        // The intent behind this sort key is to sort nodes in descending order
-        // of degree
         sort_map.insert(*k, graph.graph.edges(*k).count());
     }
     node_vec.sort_by_key(|k| Reverse(sort_map.get(k)));
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,7 +20,7 @@ extern crate pyo3;
 mod dag_isomorphism;
 mod graph;
 
-use std::cmp::Ordering;
+use std::cmp::{Ordering, Reverse};
 use std::collections::{BinaryHeap, HashSet};
 use std::ops::{Index, IndexMut};
 
@@ -968,14 +968,13 @@ fn graph_greedy_color(
 ) -> PyResult<PyObject> {
     let mut colors: HashMap<usize, usize> = HashMap::new();
     let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
-    let node_count = graph.graph.node_count();
     let mut sort_map: HashMap<NodeIndex, usize> = HashMap::new();
     for k in node_vec.iter() {
         // The intent behind this sort key is to sort nodes in descending order
         // of degree
-        sort_map.insert(*k, node_count - graph.graph.edges(*k).count());
+        sort_map.insert(*k, graph.graph.edges(*k).count());
     }
-    node_vec.sort_by_key(|k| sort_map.get(k));
+    node_vec.sort_by_key(|k| Reverse(sort_map.get(k)));
     for u_index in node_vec {
         let mut neighbor_colors: HashSet<usize> = HashSet::new();
         for edge in graph.graph.edges(u_index) {
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -505,6 +505,14 @@ retworkx API
    :return matrix: The adjacency matrix for the input dag as a numpy array
    :rtype: numpy.ndarray
 
+.. py:function:: graph_greedy_color
+   Color a PyGraph object using a largest_first strategy greedy graph coloring.
+
+   :param PyGraph: The input PyGraph to color
+
+   :returns: A dictionary where keys are node indices and the value is the color
+   :rtype: dictionary
+
 .. py:class:: PyGraph
    A class for creating undirected graphs.
 
--- a/tests/graph/test_coloring.py
+++ b/tests/graph/test_coloring.py
@@ -1,0 +1,32 @@
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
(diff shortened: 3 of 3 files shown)


t-imamichi:
Thank you for your invitation of review. I'll check it on the weekend.

src/lib.rs
```diff
@@ -913,6 +913,42 @@ fn lexicographical_topological_sort(
...
 }
 
+#[pyfunction]
+fn graph_greedy_color(
+    py: Python,
+    graph: &graph::PyGraph,
+) -> PyResult<PyObject> {
+    let mut colors: HashMap<usize, usize> = HashMap::new();
+    let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
+    node_vec.sort_by_key(|k| graph.graph.edges(*k).count());
```
t-imamichi, line 923: Python networkx seems to apply stable sort with reverse mode enabled. This retworkx implementation applies a stable sort then reverse it. So, the resulting coloring of retworkx will be OK, but it may be different from that by networkx.
How about `sort_by_key(|k| {num nodes} - {num neighbors of k})`?

t-imamichi:
I tried to modify aqua to use retworkx for coloring with this dataset. By adding a workaround the outputs get identical. But retworkx is still slower than numby-based coloring impl. Do you have any idea?

```
(Comparison of BeH2JW.txt)
-Elasped time: 4.844101860999672 (<- retworkx)
+Elasped time: 0.45734181499983606 (<- numpy)
@@ -1,4 +1,4 @@
-Elasped time: 4.844101860999672
+Elasped time: 0.45734181499983606
 ListOp(
 [AbelianListOp(
 [-12.279604476184712 * IIIIIIIIIIIIII,
```

https://github.com/Qiskit/qiskit-aqua/pull/923
my branch: https://github.com/t-imamichi/qiskit-aqua/tree/rx-color

mtreinish:
Hmm yeah, adding `add_edges_from` is probably a good idea. Individually the `add_edge` calls are fast, but when it's called 477k times it does add up, so iterating in rust instead of python is a good idea.

mtreinish:
I'll have to do some profiling of the rust code to figure out why it's taking 2.8 seconds to run the function. That does seem quite slow for what it's doing, I think your suggestion of changing the initial sorting is a good first step, that will eliminate the need for the `reverse()` call which eliminates a second traversal. From there I'll have to see where it's spending it's time and go from there.

t-imamichi:
I found that your original implementation without my workaround runs very fast. I used `rx.add_edges_from` in the following result.

```
options: Namespace(input='TaperingPaperHamiltonians/BeH2/BeH2JW.txt')
         3579710 function calls (3528368 primitive calls) in 2.845 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.281    0.281    0.338    0.338 abelian_grouper.py:177(_largest_degree_first_coloring)
     2831    0.280    0.000    0.280    0.000 {built-in method marshal.loads}
        1    0.150    0.150    0.150    0.150 {graph_greedy_color}
        1    0.084    0.084    0.084    0.084 abelian_grouper.py:168(<listcomp>)
        2    0.077    0.039    0.143    0.072 abelian_grouper.py:141(_commutation_graph_fast)
        2    0.075    0.037    0.916    0.458 abelian_grouper.py:86(group_subops)
        1    0.068    0.068    0.358    0.358 abelian_grouper.py:162(_networkx_coloring)
```

t-imamichi:
I tried with a large instance.
```
options: Namespace(input='TaperingPaperHamiltonians/NH3/NH3JW.txt')
         22511348 function calls (22456658 primitive calls) in 25.430 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    8.037    8.037    8.037    8.037 {graph_greedy_color}
        1    5.598    5.598    6.671    6.671 abelian_grouper.py:177(_largest_degree_first_coloring)
        2    1.844    0.922   23.326   11.663 abelian_grouper.py:86(group_subops)
        1    1.690    1.690    1.690    1.690 abelian_grouper.py:168(<listcomp>)
        2    1.622    0.811    2.726    1.363 abelian_grouper.py:141(_commutation_graph_fast)
        1    1.178    1.178   12.073   12.073 abelian_grouper.py:162(_networkx_coloring)
        1    1.168    1.168    1.168    1.168 {method 'add_edges_from' of 'retworkx.PyGraph' objects}
```

mtreinish:
What was the workaround you did? I just ran the script from https://github.com/Qiskit/qiskit-aqua/pull/923 with:
```diff
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -920,8 +920,9 @@ fn graph_greedy_color(
 ) -> PyResult<PyObject> {
     let mut colors: HashMap<usize, usize> = HashMap::new();
     let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
-    node_vec.sort_by_key(|k| graph.graph.edges(*k).count());
-    node_vec.reverse();
+    node_vec.sort_by_key(|k| {
+        graph.graph.node_count() - graph.graph.edges(*k).count()
+    });
     for u_index in node_vec.iter() {
         let mut neighbor_colors: HashSet<usize> = HashSet::new();
         let u = u_index.index();
```
applied locally and when I ran `diff  fast-nx.txt fast-numpy.txt` (with #82 and your current branch) and it returned:
```diff
< Elasped time: 0.3222836180066224
---
> Elasped time: 0.2867906879982911
```
so the numpy implementation is still a bit faster. I can probably speed up this function marginally more by calling `node_count()` only once instead of for every entry. Other than that I'll have to do some more profiling to see where it's spending most of the time.

The one thing I've been meaning to investigate is using [rayon](https://github.com/rayon-rs/rayon) to do parallel iteration, which might let us speed things up even further (it was part of the motivation for  #78). But I'll save that for a follow-on PR because that is really outside the scope of this pr.

mtreinish:
I ran a profile on the script with BeH2JW.txt as the input and it seems to be spending most of the time in the merge sort. Here is the relevant section from the flame graph:

Screenshot_2020-06-08_11-12-52

There are 2 ways I can think to speed this up, either switch to an unstable sort https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_unstable_by_key  which we probably don't want to do to be consistent with networkx. Or add rayon usage and leverage `par_sort_by_key`: https://docs.rs/rayon/1.3.0/rayon/slice/trait.ParallelSliceMut.html#method.par_sort_by_key to do sorting in parallel.

src/lib.rs
mtreinish: Fixed in: https://github.com/Qiskit/retworkx/pull/77/commits/03b7872b67379325fd2a1ac86ed9bbec662f0810 although I just read up a bit more on the `sort_by_key` method and it looks like I can use https://doc.rust-lang.org/std/cmp/struct.Reverse.html to reverse the sort without having to add a subtraction to the key func. I'll give that a try and if it's faster I'll switch to that.

mtreinish: Didn't behave as expected the results differed and it wasn't any faster, so lets just stick with what's in https://github.com/Qiskit/retworkx/commit/03b7872b67379325fd2a1ac86ed9bbec662f0810

t-imamichi:
Perhaps the sort is slow due to `graph.graph.edges(*k).count()` in sort_by_key.
I tried to make a vector to compute `graph.graph.node_count() - graph.graph.edges(*k).count()` beforehand, then it gets faster than python impl.
```rust
@@ -950,8 +950,12 @@ fn graph_greedy_color(
 ) -> PyResult<PyObject> {
     let mut colors: HashMap<usize, usize> = HashMap::new();
     let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
-    let node_count = graph.graph.node_count();
-    node_vec.sort_by_key(|k| node_count - graph.graph.edges(*k).count());
+    let mut node_count: HashMap<usize, usize> = HashMap::new();
+    for k in node_vec.iter() {
+        let u = k.index();
+        node_count.insert(u, graph.graph.node_count() - graph.graph.edges(*k).count());
+    }
+    node_vec.sort_by_key(|k| node_count.get(&k.index()));
     for u_index in node_vec.iter() {
         let mut neighbor_colors: HashSet<usize> = HashSet::new();
         let u = u_index.index();
```

```
--- fast-nx.txt	2020-06-09 11:35:26.093397126 +0900
+++ fast-numpy.txt	2020-06-09 11:35:36.001772804 +0900
@@ -1,4 +1,4 @@
-Elasped time: 6.927748423000139 <- rust
+Elasped time: 9.86378538300005 <- python
 ListOp(
 [AbelianListOp(
 [-45.66713521736221 * IIIIIIIIIIIIIIII,
(no output diff)
```

```
options: Namespace(input='TaperingPaperHamiltonians/NH3/NH3JW.txt')
         22521014 function calls (22466324 primitive calls) in 18.880 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    5.403    5.403    6.480    6.480 abelian_grouper.py:177(_largest_degree_first_coloring)
        2    2.054    1.027   16.783    8.391 abelian_grouper.py:86(group_subops)
        1    1.712    1.712    1.712    1.712 abelian_grouper.py:168(<listcomp>)
        2    1.624    0.812    2.728    1.364 abelian_grouper.py:141(_commutation_graph_fast)
        1    1.453    1.453    5.509    5.509 abelian_grouper.py:162(_networkx_coloring)
        1    1.188    1.188    1.188    1.188 {graph_greedy_color}
        1    1.156    1.156    1.156    1.156 {method 'add_edges_from' of 'retworkx.PyGraph' objects}
```

mtreinish:
> Perhaps the sort is slow due to `graph.graph.edges(*k).count()` in sort_by_key.
> I tried to make a vector to compute `graph.graph.node_count() - graph.graph.edges(*k).count()` beforehand, then it gets faster than python impl.
> Because I'm new to rust, I think there might be a better rust impl than mine.

It's hard to argue with what works. :) This looks fine to me (although we trade a bit of memory for run time speed) there might be a more idiomatic rust way to write it, but I'm still not as experienced with rust as I am with other languages either. If it makes that much of a speed difference we can go with it and clean things up later if needed. I'd rather have a good starting point for performance that we can refactor later if needed.

> 
> ```rust
> @@ -950,8 +950,12 @@ fn graph_greedy_color(
>  ) -> PyResult<PyObject> {
>      let mut colors: HashMap<usize, usize> = HashMap::new();
>      let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
> -    let node_count = graph.graph.node_count();
> -    node_vec.sort_by_key(|k| node_count - graph.graph.edges(*k).count());
> +    let mut node_count: HashMap<usize, usize> = HashMap::new();
> +    for k in node_vec.iter() {
> +        let u = k.index();
> +        node_count.insert(u, graph.graph.node_count() - graph.graph.edges(*k).count());
> +    }
> +    node_vec.sort_by_key(|k| node_count.get(&k.index()));
>      for u_index in node_vec.iter() {
>          let mut neighbor_colors: HashSet<usize> = HashSet::new();
>          let u = u_index.index();
> ```

The only little thing I would change in this is probably change the `HashMap` key to be a `NodeIndex` instead of a usize, that way we don't need to run `.index()` on everything all the time. It's not a lot of overhead and the compiler probably just optimizes it away anyway, but we really only need to use `.index()` when passing to or from python because usize gets converted to a python int automatically but `NodeIndex` doesn't.

> 
> ```
> --- fast-nx.txt	2020-06-09 11:35:26.093397126 +0900
> +++ fast-numpy.txt	2020-06-09 11:35:36.001772804 +0900
> @@ -1,4 +1,4 @@
> -Elasped time: 6.927748423000139 <- rust
> +Elasped time: 9.86378538300005 <- python
>  ListOp(
>  [AbelianListOp(
>  [-45.66713521736221 * IIIIIIIIIIIIIIII,
> (no output diff)
> ```
> 
> ```
> options: Namespace(input='TaperingPaperHamiltonians/NH3/NH3JW.txt')
>          22521014 function calls (22466324 primitive calls) in 18.880 seconds
> 
>    Ordered by: internal time
> 
>    ncalls  tottime  percall  cumtime  percall filename:lineno(function)
>         1    5.403    5.403    6.480    6.480 abelian_grouper.py:177(_largest_degree_first_coloring)
>         2    2.054    1.027   16.783    8.391 abelian_grouper.py:86(group_subops)
>         1    1.712    1.712    1.712    1.712 abelian_grouper.py:168(<listcomp>)
>         2    1.624    0.812    2.728    1.364 abelian_grouper.py:141(_commutation_graph_fast)
>         1    1.453    1.453    5.509    5.509 abelian_grouper.py:162(_networkx_coloring)
>         1    1.188    1.188    1.188    1.188 {graph_greedy_color}
>         1    1.156    1.156    1.156    1.156 {method 'add_edges_from' of 'retworkx.PyGraph' objects}
> ```
> 
> By the way, I proposed to include `graph.graph.node_count()` in sort because `graph.graph.edges(*k).count()` is usize and it does not allow `-graph.graph.edges(*k).count()`. If there is any good way to get a minus of usize value (or convert it to a signed value), could you use it instead of `graph.graph.node_count() - graph.graph.edges(*k).count()`? The idea is that just taking a minus of value to sort the list in the reversed order.

Well you could probably do something like `-(graph.graph.edges(*k).count() as isize)` (or `i64`). I just worry about the overflow case for things like that (not that it's a likely failure for this). I also still think we should be able to leverage https://doc.rust-lang.org/std/cmp/struct.Reverse.html to do the reversed order. I probably was just using it wrong earlier.

t-imamichi:
Thank you for your information.
I tried your recommendation to use NodeIndex and add_edges_from_no_data.
https://github.com/t-imamichi/retworkx/commit/279c2cc1c464c1443af2f4b3b05292656865b9b4
https://github.com/t-imamichi/qiskit-aqua/commit/371e4da8b22993bf98efd3217c6486002056fd04
The result gets a little faster than before. I think it looks good as the first version.

```
--- fast-nx.txt	2020-06-09 22:36:05.638027225 +0900
+++ fast-numpy.txt	2020-06-09 22:36:15.662389597 +0900
@@ -1,4 +1,4 @@
-Elasped time: 6.3002666709999176 <- rust
+Elasped time: 9.983879566000041 <- python
 ListOp(
 [AbelianListOp(
 [-45.66713521736221 * IIIIIIIIIIIIIIII,
```

```
options: Namespace(input='TaperingPaperHamiltonians/NH3/NH3JW.txt')
         22521753 function calls (22467062 primitive calls) in 18.410 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    4.859    4.859    5.995    5.995 abelian_grouper.py:177(_largest_degree_first_coloring)
        2    3.894    1.947    5.007    2.504 abelian_grouper.py:141(_commutation_graph_fast)
        1    1.327    1.327    6.300    6.300 pauli_grouping.py:40(<lambda>)
        1    1.290    1.290    9.984    9.984 pauli_grouping.py:41(<lambda>)
        1    1.206    1.206    1.206    1.206 {graph_greedy_color}
        1    1.132    1.132    1.132    1.132 {method 'add_edges_from_no_data' of 'retworkx.PyGraph' objects}
```

t-imamichi:
I'm happy to hear that retworkx coloring gets faster in your environment as well. It looks good to me.

src/lib.rs
```diff
@@ -915,6 +915,45 @@ fn lexicographical_topological_sort(
...
+fn graph_greedy_color(
+    py: Python,
+    graph: &graph::PyGraph,
+) -> PyResult<PyObject> {
+    let mut colors: HashMap<usize, usize> = HashMap::new();
+    let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
+    let node_count = graph.graph.node_count();
+    let mut sort_map: HashMap<NodeIndex, usize> = HashMap::new();
+    for k in node_vec.iter() {
+        sort_map.insert(*k, node_count - graph.graph.edges(*k).count());
```
t-imamichi, line 928: It might be good to add a comment "It is intended to sort nodes in the descending order of degree" because `node_count - graph.graph.edges(*k).count()` does not look intuitive.

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/77/commits/b63f6bc0bf30ba3e5e51cf92a3df4dfb01cf20f5

t-imamichi:
Thank you. It looks good!

src/lib.rs
```diff
@@ -961,6 +961,47 @@ fn lexicographical_topological_sort(
...
+    graph: &graph::PyGraph,
+) -> PyResult<PyObject> {
+    let mut colors: HashMap<usize, usize> = HashMap::new();
+    let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
+    let node_count = graph.graph.node_count();
+    let mut sort_map: HashMap<NodeIndex, usize> = HashMap::new();
+    for k in node_vec.iter() {
+        // The intent behind this sort key is to sort nodes in descending order
+        // of degree
+        sort_map.insert(*k, node_count - graph.graph.edges(*k).count());
```
lcapelluto, line 976: comment is definitely helpful. How does `node_count` factor into this?

mtreinish: `node_count` is just the number of nodes in the graph so we get a descending order by subtracting the degree from the max value. But now that you mention this there is a potential bug here with multigraphs I think. If the degree for 1 node is greater than the number of nodes this will overflow and have the opposite of the intended effect. This should be easy to fix, just change the value type in the sort_map from `usize` to `isize`. I'll push a fix and add a test

mtreinish: So, my original idea of using isze would have too many tradeoffs for super large graphs (we lose too much between `usize::max` and `isize::max`). So, I tried the `Reverse` struct I mentioned eariler, by itself on `sort_by_key(|k| Reverse(graph.graph.edges(*k).count()))` I managed to get it working this time (I believe I made a typo the last time I tried it in my comments from a couple weeks ago) but this ended up being significantly slower the hash map approach @t-imamichi introduced. This is likely because `sort_by_key()` isn't caching the results from the key func, so it's calling `edges().count()` more than once for nodes. So I added back the hash map, but just store the degree for each node instead of subtracting and this gets us back to more or less parity with the numbers we had before:

```
--- fast-nx.txt	2020-06-22 14:07:58.020319946 -0400
+++ fast-numpy.txt	2020-06-22 14:08:04.167015153 -0400
@@ -1,4 +1,4 @@
-Elasped time: 5.515628322958946
+Elasped time: 6.124710976029746
```

Anyway, that's a long way of saying fixed in: https://github.com/Qiskit/retworkx/pull/77/commits/40c197fa71c2e14e3f67e5fa31adeb364d8cd667

src/lib.rs
```diff
@@ -961,6 +961,46 @@ fn lexicographical_topological_sort(
...
+fn graph_greedy_color(
+    py: Python,
+    graph: &graph::PyGraph,
+) -> PyResult<PyObject> {
+    let mut colors: HashMap<usize, usize> = HashMap::new();
+    let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
+    let mut sort_map: HashMap<NodeIndex, usize> = HashMap::new();
+    for k in node_vec.iter() {
+        // The intent behind this sort key is to sort nodes in descending order
+        // of degree
```
t-imamichi, line 974: It would be good to remove the comment because the code is revised.

t-imamichi:
Thanks. `Reverse` looks more intuitive.

src/lib.rs
mtreinish: Oops I thought that I deleted that when I changed the loop. Yeah I agree it's just confusing now. Fixed in: https://github.com/Qiskit/retworkx/pull/77/commits/2049ddf2746ab3f68aaa14721f730acf613d432b

t-imamichi:
It looks good!

src/lib.rs
lcapelluto: thanks 👍  :)

PR #78, opened 2020-06-05
Switch to hashbrown hashmap implementation

mtreinish:
The hashbrown library provides a drop-in replacement for the stdlib
HashMap implementation with improved performance. Since rust 1.36.0 the
stdlib implementation has been using the same algorithm as hasbrown
(it was ported from hashbrown) but hashbrown is still faster because it
uses the AHash hashing algorithm which is a tradeoff for speed vs
hash DOS resistance. Since the use case for HashMaps in retworkx doesn't
really need protection against denial of service collisions picking performance
makes sense here. In the future if a use case arises we can always make the
 hashmap lib configurable or add the use of the 'ahash-compile-time' feature
which increases the DOS-resistance of the algorithm.

The other potential speedup that hasbrown can provide in the future is
support for parallel iterators. Right now this commit doesn't use it
because adding and testing parallel iteration is a separate task, but
that can potentially provide a useful speedup for several functions
(including lexicographical topological sort) which rely on iterating
over hashmaps currently.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,8 @@ stages:
 
 jobs:
   fast_finish: true
+  allow_failures:
+    - name: Python 3.7 Tests s390x Linux
   include:
     - name: Compile and rustfmt
       language: rust
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ install:
   - virtualenv test-venv
   - test-venv/bin/pip install -U .
 script:
+  - rustc --print cfg
   - cd tests && ../test-venv/bin/python -m unittest discover .
 stages:
   - name: Compile and lint
@@ -44,6 +45,7 @@ jobs:
         - rustup update
         - rustup override set nightly
       script:
+        - rustc --print cfg
         - cargo build
         - cargo +stable fmt -- --check
     - name: Python lint
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ install:
   - virtualenv test-venv
   - test-venv/bin/pip install -U .
 script:
-  - rustc --print cfg
   - cd tests && ../test-venv/bin/python -m unittest discover .
 stages:
   - name: Compile and lint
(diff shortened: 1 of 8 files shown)


mtreinish:
Doing some high level performance benchmarks with qiskit-terra's transpile (so not targetted synthetic benchmarks of functions that use a hashmap): with hasbrown the transpile time for a 53x53 quantum volume circuit  is `11.534` seconds vs `11.941` seconds with the stdlib HashMap. So not a drastic speedup, but a small consistent improvement. It's also worth it if we can parallelize iteration in places in the future.

The raw data is here:

**hashbrown:**
```
11.512400388717651
11.522578477859497
11.477795124053955
11.54568076133728
11.513428449630737
11.527084350585938
11.528125762939453
11.55392575263977
11.624308586120605
11.537017107009888
11.409081220626831
11.656954288482666
```
**stdlib:**
```
11.78228497505188
11.999693155288696
11.789955377578735
11.968479871749878
11.916764259338379
11.964802980422974
12.033486604690552
12.02527666091919
12.041775941848755
12.055840253829956
11.875913619995117
11.835525512695312
```

Running:
```python3
from qiskit import *
from qiskit.quantum_info.random import random_unitary
from qiskit.test import mock
import numpy as np

def build_qv_model_circuit(width, depth, seed=None):
    """
    The model circuits consist of layers of Haar random
    elements of SU(4) applied between corresponding pairs
    of qubits in a random bipartition.
    """
    np.random.seed(seed)
    circuit = QuantumCircuit(width)
    # For each layer
    for _ in range(depth):
        # Generate uniformly random permutation Pj of [0...n-1]
        perm = np.random.permutation(width)
        # For each pair p in Pj, generate Haar random SU(4)
        for k in range(int(np.floor(width/2))):
            U = random_unitary(4)
            pair = int(perm[2*k]), int(perm[2*k+1])
            circuit.append(U, [pair[0], pair[1]])
    return circuit

backend = mock.FakeRochester()
circ = build_qv_model_circuit(53, 53, seed=1203421)
start = time.time()
transpile(circ, backend)
stop = time.time()
run_time = stop - start
print(run_time)
```

PR #82, opened 2020-06-08
Add add_nodes_from and add_edges_from methods

mtreinish:
This commit adds 2 methods to PyDAG and PyGraph, add_nodes_from() and
add_edges_from(). These methods take in lists of objects to add to the
graph to minimize the back and forth calling when adding many objects to
the graph at once. add_nodes_from() takes in a list of objects which
will be the data for the node, and add_nodes_from() takes in a list of
tuples with 2 node indexes for where the edge should be added and the
3rd tuple element is the data for the edge.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -136,6 +136,25 @@ retworkx API
         :returns index: The index of the newly created node
         :rtype: int
 
+    .. py:method:: add_nodes_from(self, obj_list):
+        Add new nodes to the dag.
+
+        :param list obj_list: A list of python object to attach to the graph.
+
+        :returns indexes: A list of int indexes of the newly created nodes
+        :rtype: list
+
+    .. py:method:: add_edges_from(self, obj_list):
+        Add new edges to the dag.
+
+        :param list obj_list: A list of tuples of the form
+            ``(parent, child, obj)`` to attach to the graph. ``parent`` and
+            ``child`` are integer indexes describing where an edge should be
+            added, and obj is the python object for the edge data.
+
+        :returns indexes: A list of int indexes of the newly created edges
+        :rtype: list
+
     .. py:method:: add_child(self, parent, obj, edge):
         Add a new child node to the dag.
 
@@ -611,12 +630,31 @@ retworkx API
         :raises: When the new edge will create a cycle
 
     .. py:method:: add_node(self, obj):
-        Add a new node to the dag.
+        Add a new node to the graph.
 
         :param obj: The python object to attach to the node
 
         :returns index: The index of the newly created node
         :rtype: int
+
+    .. py:method:: add_nodes_from(self, obj_list):
+        Add new nodes to the graph.
+
+        :param list obj_list: A list of python object to attach to the graph.
+
+        :returns indexes: A list of int indexes of the newly created nodes
+        :rtype: list
+
+    .. py:method:: add_edges_from(self, obj_list):
+        Add new edges to the graph.
+
+        :param list obj_list: A list of tuples of the form
+            ``(parent, child, obj)`` to attach to the graph. ``parent`` and
+            ``child`` are integer indexes describing where an edge should be
+            added, and obj is the python object for the edge data.
+
+        :returns indexes: A list of int indexes of the newly created edges
+        :rtype: list
 
     .. py:method:: adj(self, node):
         Get the index and data for the neighbors of a node.
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -141,7 +141,7 @@ retworkx API
 
         :param list obj_list: A list of python objects to attach to the graph.
 
-        :returns indexes: A list of int indexes of the newly created nodes
+        :returns indices: A list of int indices of the newly created nodes
         :rtype: list
 
     .. py:method:: add_edges_from(self, obj_list):
@@ -152,11 +152,11 @@ retworkx API
             ``child`` are integer indexes describing where an edge should be
             added, and obj is the python object for the edge data.
 
-        :returns indexes: A list of int indexes of the newly created edges
+        :returns indices: A list of int indices of the newly created edges
         :rtype: list
 
     .. py:method:: add_edges_from_no_data(self, obj_list):
-        Add new edges to the dag without data.
+        Add new edges to the dag without python data.
 
         :param list obj_list: A list of tuples of the form
             ``(parent, child)`` to attach to the graph. ``parent`` and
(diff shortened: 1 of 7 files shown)


t-imamichi:
Is it easy to support 2-tuple (u, v) as well as 3-tuple (u, v, d)? networkx supports both and I use 2-tuple for coloring.
https://networkx.github.io/documentation/networkx-1.10/reference/generated/networkx.Graph.add_edges_from.html#networkx.Graph.add_edges_from

mtreinish:
Hmm, it'll get tricky to support both 2 or 3, especially dynamically like python would normally allow. The problem is that rust needs to know at compile time the size of the Vec elements so it has to be a fixed length tuple. We probably could add a kwarg to do 2-tuple inputs or doing it as a separate function.

Another alternative would be to call out to python to make it a more dynamic structure, but that would add a lot more overhead than what this has now so I'd probably want to avoid that.

What do you think?

t-imamichi:
I see. Then, let's leave it as it is. If the conversion of 2-tuple to 3-tuple takes a long time in coloring, let's discuss it again.

mtreinish:
Well I'm not sure if it'll be any faster but I wasn't happy with the profiles in #77 showing >1sec to do the list comprehension to add None to the tuples. I added a second function for using 2-tuples in: https://github.com/Qiskit/retworkx/pull/82/commits/a20e420a909b3bfba2c0006517790cfb5392d65b we can always revert it if it's not any faster

src/lib.rs
```diff
@@ -234,6 +234,38 @@ impl GetAdjacencyMatrix for PyDAG {
...
+            {
+                Err(DAGWouldCycle::py_err("Adding an edge would cycle"))
+            } else {
+                let edge = self.graph.add_edge(p_index, c_index, edge);
+                Ok(edge.index())
+            }
+        } else {
+            let edge = self.graph.add_edge(p_index, c_index, edge);
+            Ok(edge.index())
+        }
```
lcapelluto, line 265: Could these be combined by merging the logic again once an error isn't raised?

src/lib.rs
```diff
@@ -234,6 +234,38 @@ impl GetAdjacencyMatrix for PyDAG {
...
+impl PyDAG {
+    fn _add_edge(
+        &mut self,
+        p_index: NodeIndex,
+        c_index: NodeIndex,
+        edge: PyObject,
+    ) -> PyResult<usize> {
+        if self.check_cycle {
+            let should_check_for_cycle =
+                must_check_for_cycle(self, p_index, c_index);
```
lcapelluto, line 247: I remember looking through this and understanding it at some point, but it's confusing to see "check cycle" then "should check" then "must check". Maybe you could add a note about why these different check booleans are all necessary?

src/graph.rs
```diff
@@ -362,6 +362,35 @@ impl PyGraph {
         Ok(edge.index())
     }
 
+    pub fn add_edges_from(
+        &mut self,
+        obj_list: Vec<(usize, usize, PyObject)>,
```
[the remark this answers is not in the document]
mtreinish: `py: Python` is only needed if the code is going to call python, it represents a GIL lock basically. Normally this is only to interact with an existing python object or to create a new python object. In this case it's not needed because PyO3 will handle the conversion of `list(int, int, object)` -> `Vec<(usize, usize, PyObject)` on the input and `Vec<usize>` -> `list(int)` on the return. So we don't need to acquire the GIL while this function is running.

docs/source/api.rst
```diff
@@ -599,13 +630,44 @@ retworkx API
...
     .. py:method:: add_node(self, obj):
-        Add a new node to the dag.
+        Add a new node to the graph.
 
         :param obj: The python object to attach to the node
 
         :returns index: The index of the newly created node
         :rtype: int
 
+    .. py:method:: add_nodes_from(self, obj_list):
```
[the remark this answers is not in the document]
mtreinish: Yeah, I need to go through a redo/reorganize the docs. Right now everything is in a single file and it makes it difficult to parse. Also the rendered formatting is kind of a mess: https://retworkx.readthedocs.io/en/latest/api.html

src/lib.rs
mtreinish: Sure, I'll do that now. But to quickly summarize `check_cycle` is an instance attribute that says whether the cycle checking is run or not. It's basically a performance optimization because cycle checking gets increasingly expensive as the graph grows so we default it off (and for terra it's impossible for dagcircuit to introduce a cycle so this is safe) and make cycle checking opt-in. `must_check_for_cycle` and `should_check_for_cycle` should probably be renamed for clarity they're bad names. `must_check_for_cycle` is the function to check the relative position of parent and child in the graph: https://github.com/Qiskit/retworkx/blob/master/src/lib.rs#L661-L671 and says if the parent node has no parents or the child node has no children or there is an existing edge between them then it is safe to skip cycle checking because adding an edge couldn't introduce a cycle. `should_check_for_cycle` just is the output from that function in a variable. Bu I agree this is needlessly confusing

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/82/commits/98d316eeaff8db546fbec33f80370279de193cd9

src/lib.rs
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/82/commits/98d316eeaff8db546fbec33f80370279de193cd9

docs/source/api.rst
```diff
@@ -136,6 +136,37 @@ retworkx API
...
+        :param list obj_list: A list of tuples of the form
+            ``(parent, child, obj)`` to attach to the graph. ``parent`` and
+            ``child`` are integer indexes describing where an edge should be
+            added, and obj is the python object for the edge data.
+
+        :returns indexes: A list of int indexes of the newly created edges
+        :rtype: list
+
+    .. py:method:: add_edges_from_no_data(self, obj_list):
+        Add new edges to the dag without data.
```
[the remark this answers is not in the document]
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/82/commits/98d316eeaff8db546fbec33f80370279de193cd9

docs/source/api.rst
```diff
@@ -136,6 +136,37 @@ retworkx API
         :returns index: The index of the newly created node
         :rtype: int
 
+    .. py:method:: add_nodes_from(self, obj_list):
+        Add new nodes to the dag.
+
+        :param list obj_list: A list of python object to attach to the graph.
+
+        :returns indexes: A list of int indexes of the newly created nodes
```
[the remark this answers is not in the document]
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/82/commits/98d316eeaff8db546fbec33f80370279de193cd9

PR #83, opened 2020-06-09
Add graph_floyd_warshall_numpy function for graph classes

mtreinish:
This commit adds a new function `floyd_warshall_numpy()` which is used to
find all-pairs of shortest path lengths. It returns the output as a
numpy matrix with shortest path distances between nodes. If there is no
path between nodes in the graph the value will be `np.inf`.

TODO:

- [x] Fix test failures
- [x] Add more tests
- [x] Add docs


itoko:
Thank you for adding this new feature. I've looked through and found no problem in logic/algorithm.
I have one comment on an internal interface, `_graph_adjacency_matrix`, especially, I don't really like the `min_multigraph` flag.
How about having a private function that returns edge (with weight) list (or iterator) instead?
If we have such a function (say `_edges_with_weights` here), I think we would follow what happens in both of `graph_adjacency_matrix` and `graph_floyd_warshall_numpy` more easily.
For example, `graph_floyd_warshall_numpy` would look like
```
# initialize n x n matrix with infs
# set zeros to diagonal elements of the matrix
for i, j, weight in graph._edges_with_weights(...) {
    mat[[i, j]] = mat[[i, j]].min(weight);
}
# main three for-loops
```
I'm also anticipating someone would request a sort of `floyd_warshall_predecessor_and_distance` function, which returns not only the shortest paths' distances but also the paths them selves, in the future. The above implementation would make such an extension easier.

mtreinish:
I like that idea, I agree it would make this a bit cleaner. It shouldn't be too hard to implement, it might not even need to be a separate private function I can probably just do an inline iterator map over `graph.graph.edge_references()` to get a float from the weight `PyObject`, probably something like:

```rust
for i, j, weight in graph.graph_edge_references().map(|(i, j, weight)| i, j, edge_weight_callable(weight)?) {
     mat[[i, j]] = mat[[i, j]].min(weight);
}
```

I'll play with this and see what I come up with.

FWIW, I went common `_graph_adjacency_matrix` common private function approach originally because in the first draft of this PR I was returning the same matrix for both functions, but I needed to switch to the `min()`/`inf` matrix for floyd warshall and didn't change it back. But I agree it doesn't make as much sense as a common function like this where there are basically 2 separate return types.

itoko:
I also recommend you to add a test case for a graph with negative edges. Floyd-Warshall algorithm is occasionally used for checking if there is any negative cycle in a graph. (If returned matrix has any negative diagonal element, that means there exists some negative cycle in the graph.)

mtreinish:
@itoko ok, I reworked the internal api in: https://github.com/Qiskit/retworkx/pull/83/commits/ca36754e545c1ee4c0995e3b96ab968a07e011a9 and added a simple test for a graph with a negative cycle in: https://github.com/Qiskit/retworkx/pull/83/commits/bb41532bd9e7491f457e2fae32ebbeddf87e0068

mtreinish:
I was able to de-duplicate some more code with the patch below, but there was a **minor** ~1% average performance hit with that (I assume it's the `edge.weight().clone()`) although that may be within my error margin on my desktop. So I'm not sure if it's worth that tradeoff, what do you think?
```diff
diff --git a/src/lib.rs b/src/lib.rs
index be32bab..8b4d2a4 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,7 +44,10 @@ use pyo3::Python;
 use petgraph::algo;
 use petgraph::graph::NodeIndex;
 use petgraph::prelude::*;
-use petgraph::visit::{Bfs, IntoEdgeReferences, NodeIndexable, Reversed};
+use petgraph::visit::{
+    Bfs, Data, GraphBase, GraphProp, IntoEdgeReferences, IntoNodeReferences,
+    NodeIndexable, Reversed,
+};
 
 use ndarray::prelude::*;
 use numpy::IntoPyArray;
@@ -572,6 +575,39 @@ fn floyd_warshall(py: Python, dag: &digraph::PyDiGraph) -> PyResult<PyObject> {
     Ok(out_dict.into())
 }
 
+fn get_edge_iter_with_weights<G>(
+    graph: G,
+    node_map: Option<HashMap<NodeIndex, usize>>,
+) -> impl Iterator<Item = (usize, usize, PyObject)>
+where
+    G: GraphBase
+        + IntoEdgeReferences
+        + IntoNodeReferences
+        + NodeIndexable
+        + GraphProp,
+    G: Data<NodeWeight = PyObject, EdgeWeight = PyObject>,
+{
+    graph.edge_references().map(move |edge| {
+        let i: usize;
+        let j: usize;
+        match &node_map {
+            Some(map) => {
+                let source_index =
+                    NodeIndex::new(graph.to_index(edge.source()));
+                let target_index =
+                    NodeIndex::new(graph.to_index(edge.target()));
+                i = *map.get(&source_index).unwrap();
+                j = *map.get(&target_index).unwrap();
+            }
+            None => {
+                i = graph.to_index(edge.source());
+                j = graph.to_index(edge.target());
+            }
+        }
+        (i, j, edge.weight().clone())
+    })
+}
+
 /// Find all-pairs shortest path lengths using Floyd's algorithm
 ///
 /// Floyd's algorithm is used for finding shortest paths in dense graphs
@@ -628,24 +664,8 @@ fn graph_floyd_warshall_numpy(
     };
 
     // Build adjacency matrix
-    for (source, target, weight) in graph
-        .graph
-        .edge_references()
-        .map(|edge| (edge.source(), edge.target(), edge.weight()))
-    {
-        let i: usize;
-        let j: usize;
-        let edge_weight = weight_callable(weight)?;
-        match &node_map {
-            Some(map) => {
-                i = *map.get(&source).unwrap();
-                j = *map.get(&target).unwrap();
-            }
-            None => {
-                i = source.index();
-                j = target.index();
-            }
-        }
+    for (i, j, weight) in get_edge_iter_with_weights(graph, node_map) {
+        let edge_weight = weight_callable(&weight)?;
         mat[[i, j]] = mat[[i, j]].min(edge_weight);
         mat[[j, i]] = mat[[j, i]].min(edge_weight);
     }
@@ -726,24 +746,8 @@ fn digraph_floyd_warshall_numpy(
     };
 
     // Build adjacency matrix
-    for (source, target, weight) in graph
-        .graph
-        .edge_references()
-        .map(|edge| (edge.source(), edge.target(), edge.weight()))
-    {
-        let i: usize;
-        let j: usize;
-        let edge_weight = weight_callable(weight)?;
-        match &node_map {
-            Some(map) => {
-                i = *map.get(&source).unwrap();
-                j = *map.get(&target).unwrap();
-            }
-            None => {
-                i = source.index();
-                j = target.index();
-            }
-        }
+    for (i, j, weight) in get_edge_iter_with_weights(graph, node_map) {
+        let edge_weight = weight_callable(&weight)?;
         mat[[i, j]] = mat[[i, j]].min(edge_weight);
     }
     // 0 out the diagonal
@@ -892,27 +896,12 @@ fn digraph_adjacency_matrix(
     }
     let mut matrix = Array::<f64, _>::zeros((n, n).f());
 
-    let weight_callable = |a: &PyObject| -> PyResult<PyObject> {
+    let weight_callable = |a: &PyObject| -> PyResult<f64> {
         let res = weight_fn.call1(py, (a,))?;
-        Ok(res.to_object(py))
+        res.extract(py)
     };
-    for edge in graph.graph.edge_references() {
-        let edge_weight_raw = weight_callable(&edge.weight())?;
-        let edge_weight: f64 = edge_weight_raw.extract(py)?;
-        let source = edge.source();
-        let target = edge.target();
-        let i: usize;
-        let j: usize;
-        match &node_map {
-            Some(map) => {
-                i = *map.get(&source).unwrap();
-                j = *map.get(&target).unwrap();
-            }
-            None => {
-                i = source.index();
-                j = target.index();
-            }
-        }
+    for (i, j, weight) in get_edge_iter_with_weights(graph, node_map) {
+        let edge_weight = weight_callable(&weight)?;
         matrix[[i, j]] += edge_weight;
     }
     Ok(matrix.into_pyarray(py).into())
@@ -963,27 +952,12 @@ fn graph_adjacency_matrix(
     }
     let mut matrix = Array::<f64, _>::zeros((n, n).f());
 
-    let weight_callable = |a: &PyObject| -> PyResult<PyObject> {
+    let weight_callable = |a: &PyObject| -> PyResult<f64> {
         let res = weight_fn.call1(py, (a,))?;
-        Ok(res.to_object(py))
+        res.extract(py)
     };
-    for edge in graph.graph.edge_references() {
-        let edge_weight_raw = weight_callable(&edge.weight())?;
-        let edge_weight: f64 = edge_weight_raw.extract(py)?;
-        let source = edge.source();
-        let target = edge.target();
-        let i: usize;
-        let j: usize;
-        match &node_map {
-            Some(map) => {
-                i = *map.get(&source).unwrap();
-                j = *map.get(&target).unwrap();
-            }
-            None => {
-                i = source.index();
-                j = target.index();
-            }
-        }
+    for (i, j, weight) in get_edge_iter_with_weights(graph, node_map) {
+        let edge_weight = weight_callable(&weight)?;
         matrix[[i, j]] += edge_weight;
         matrix[[j, i]] += edge_weight;
     }
```

mtreinish:
I decided to go for it in: https://github.com/Qiskit/retworkx/pull/83/commits/4a3fd9b81e72d435b21952c752596e32a599465f the performance hit seems to be in the error margin. But if you test it locally and or have a problem with that approach let me know and we can easily revert it.

itoko:
Thank you for quick response. LGTM. I like the latest de-duplicated code (4a3fd9b). It's very easy to read for me, so I think it's worth accepting a minor performance hit.

src/lib.rs
```diff
@@ -588,6 +595,182 @@ fn floyd_warshall(py: Python, dag: &digraph::PyDiGraph) -> PyResult<PyObject> {
...
+                    NodeIndex::new(graph.to_index(edge.target()));
+                i = *map.get(&source_index).unwrap();
+                j = *map.get(&target_index).unwrap();
+            }
+            None => {
+                i = graph.to_index(edge.source());
+                j = graph.to_index(edge.target());
+            }
+        }
+        (i, j, edge.weight().clone())
```
itoko, line 639: Probably a very basic question, Why do we need `clone()` (or clone_ref) here? What happens if we omit `clone()`?

src/lib.rs
```diff
@@ -588,6 +595,182 @@ fn floyd_warshall(py: Python, dag: &digraph::PyDiGraph) -> PyResult<PyObject> {
...
+    if graph.nodes_removed() {
+        let mut node_hash_map: HashMap<NodeIndex, usize> = HashMap::new();
+        for (count, node) in graph.node_identifiers().enumerate() {
+            let index = NodeIndex::new(graph.to_index(node));
+            node_hash_map.insert(index, count);
+        }
+        node_map = Some(node_hash_map);
+    } else {
+        node_map = None;
+    }
```
itoko, line 620: In contrast to these lines, why don't we need to care about `node_removed` in `PyGraph::weighted_edge_list()`?

mtreinish: You can see the rationale behind why this is needed here: https://github.com/Qiskit/retworkx/pull/80 but basically it comes down to the `StableGraph` type used from petgraph to build the graph data structure in `PyGraph` (and `PyDiGraph`) The `StableGraph` doesn't reuse node indices when there are removals (unless it was the end of the list, which might be a bug actually). So for example, if you had nodes 0, 1, 2, and 3 then removed node 1 you'd have node indices 0, 2, 3. Without mapping 0, 2, 3 to 0, 1, 2 when we build the adjacency matrix we'd either end up with an extra column and row for each removed node or an index error (depending on how we initialized the matrix). The first version of `graph_adjacency_matrix()` had the extra entries in the matrix for missing nodes, but that was decided to be too confusing so this was the compromise to fix it.

For `PyGraph::weighted_edge_list()` this isn't a concern because it just loops over all edges in the graph it doesn't matter if there is a hole in the index list since we'll still get the same list output regardless of whether we're missing nodes or not.

src/lib.rs
mtreinish: Oh, this actually a really good (and pretty involved) question and is actually the one thing in the PR I'm not happy with, but it was the only way I managed to make it work. The `clone()` is necessary because the return type from this function for weight is of type `PyObject` but `edge.weight()` returns a `&PyObject`. If you remove the `clone()` you get a type mismatch error:

```
error[E0271]: type mismatch resolving `<[closure@src/lib.rs:622:33: 640:6 node_map:_, graph:_] as std::ops::FnOnce<(<G as petgraph::visit::IntoEdgeReferences>::EdgeRef,)>>::Output == (usize, usize, pyo3::PyObject)`
   --> src/lib.rs:600:6
    |
600 | ) -> impl Iterator<Item = (usize, usize, PyObject)>
    |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `&pyo3::PyObject`, found struct `pyo3::PyObject`
    |
    = note: expected tuple `(_, _, &pyo3::PyObject)`
               found tuple `(_, _, pyo3::PyObject)`
    = note: required because of the requirements on the impl of `std::iter::Iterator` for `std::iter::Map<<G as petgraph::visit::IntoEdgeReferences>::EdgeReferences, [closure@src/lib.rs:622:33: 640:6 node_map:_, graph:_]>`
    = note: the return type of a function must have a statically known size

error: aborting due to previous error

For more information about this error, try `rustc --explain E0271`.
error: could not compile `retworkx`.
```

There are 2 paths to fix the type mismatch besides a `clone()` here, either dereference the borrow or change the return type to a reference. The approach which I actually wanted to do at first was change the return type to a reference. But the problem here is the [lifetime](https://doc.rust-lang.org/1.9.0/book/lifetimes.html) of the reference. The borrow checker doesn't know how long the reference will live which results in an error:

```error[E0106]: missing lifetime specifier
   --> src/lib.rs:600:42
    |
600 | ) -> impl Iterator<Item = (usize, usize, &PyObject)>
    |                                          ^ expected named lifetime parameter
    |
    = help: this function's return type contains a borrowed value with an elided lifetime, but the lifetime cannot be derived from the arguments
help: consider using the `'static` lifetime
    |
600 | ) -> impl Iterator<Item = (usize, usize, &'static PyObject)>
    |                                          ^^^^^^^^

error[E0310]: the associated type `<G as petgraph::visit::IntoEdgeReferences>::EdgeRef` may not live long enough
   --> src/lib.rs:639:16
    |
639 |         (i, j, edge.weight())
    |                ^^^^
    |
    = help: consider adding an explicit lifetime bound `<G as petgraph::visit::IntoEdgeReferences>::EdgeRef: 'static`...
    = note: ...so that the type `<G as petgraph::visit::IntoEdgeReferences>::EdgeRef` is not borrowed for too long

error[E0310]: the associated type `<G as petgraph::visit::IntoEdgeReferences>::EdgeRef` may not live long enough
   --> src/lib.rs:639:21
    |
639 |         (i, j, edge.weight())
    |                     ^^^^^^
    |
    = help: consider adding an explicit lifetime bound `<G as petgraph::visit::IntoEdgeReferences>::EdgeRef: 'static`...
    = note: ...so that the reference type `&<G as petgraph::visit::IntoEdgeReferences>::EdgeRef` does not outlive the data it points at

error: aborting due to 3 previous errors

Some errors have detailed explanations: E0106, E0310.
For more information about an error, try `rustc --explain E0106`.
```

Normally you can fix this by setting an explicit lifetime to tell the compiler how long the reference will live so it can ensure that the operations on the reference are safe or not. But, in this case the compiler error message is pointing out that the lifetime of the reference is defined in the trait definition for `IntoEdgeReferences`, which already has a lifetime set on it. I wasn't able to find a way to specify that the lifetime of the returned `PyObject` reference from the iterator was the same as the lifetime for the `PyGraph` object (which is specified here: https://github.com/Qiskit/retworkx/blob/master/src/graph.rs#L132 ). I think this is why the compiler is suggestion that I use a [static lifetime](https://doc.rust-lang.org/stable/rust-by-example/scope/lifetime/static_lifetime.html#reference-lifetime) for the returned `&PyObject` even though that wouldn't work (the edge reference doesn't have a static lifetime).

So without a way to pass a reference we can look at dereferencing the return from `edge.weight()`. The problem is if we dereference the borrow like `*edge.weight()` the lifetime of the borrow will expire when the function exits and the value will be dropped. The compiler errors telling us this with:

```
error[E0507]: cannot move out of a shared reference
   --> src/lib.rs:639:16
    |
639 |         (i, j, *edge.weight())
    |                ^^^^^^^^^^^^^^ move occurs because value has type `pyo3::PyObject`, which does not implement the `Copy` trait

error: aborting due to previous error

For more information about this error, try `rustc --explain E0507`.
error: could not compile `retworkx`.
```

Using `clone()` here is just how we can avoid that data ownership issue. By returning a "copy" of the data instead of value from a reference. However, `clone()` of a PyObject isn't actually a memory copy it just increases the refcount in python: https://docs.rs/pyo3/0.11.1/src/pyo3/object.rs.html#311-318 so it typically shouldn't be that slow. It just needs a GIL context to call python to do that though, which will be the bottleneck on `clone()`.

It's also worth pointing out that this wasn't an issue in the duplicated version because the scope (each of the functions) was the same so we didn't have to worry about the lifetime of the data.

itoko: Thank you for your detailed answer. I understand I have to learn more about lifetime and ownership:-) I've played with tweaking the code a bit and getting error messages like you showed above, and found there is no other way than using `clone()` since we have no means to change the lifetime of `EdgeRef` defined in the `petgraph` library. So now `clone()` makes sense to me.

src/lib.rs
itoko: I see. Thank you.

src/lib.rs
mtreinish: FWIW, borrowing, lifettimes, etc are personally what I find to be the hardest to get used to in Rust. I've personally only recently felt like I'm comfortable dealing with the concepts relatively recently. When I first started working in rust I felt like I was just brute forcing my way through compiler errors. But eventually after working with the language for a while I started to understand more of it. The compiler messages and `rustc --explain` help a lot normally actually. It's just in this case petgraph (and some other libs) just make the error messages harder to read, because they're written to be super generic the error messages don't make a lot of sense without understanding how petgraph works.

Issue #20, opened 2020-02-09
Add a PyDiGraph class


mtreinish:
There are some networkx uses cases in qiskit-terra that are outside of the `DAGCircuit` class. Mainly the coupling map and the noise aware layout class. Exposing a `PyDiGraph` class will be straightforward because it's essentially identical to `PyDAG` except we don't need or want any cycle checks as part of adding new nodes (and a few places in the logic where things are shortcut because we assume no cycles).

This should be achievable by making the inner graph struct on `PyDAG` a `PyDiGraph` and then for the method impl just call the method on the inner `PyDiGraph` and return that. The exception being where we need dag specifics.

PR #85, opened 2020-06-10
Add all_simple_paths for dag and graph

mtreinish:
This commit adds 2 new functions dag_all_simple_paths() and
graph_all_simple_paths() to find all simple paths between 2 nodes in a
PyDAG and PyGraph object.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -505,6 +505,36 @@ retworkx API
    :return matrix: The adjacency matrix for the input dag as a numpy array
    :rtype: numpy.ndarray
 
+.. py:function:: graph_all_simple_paths(graph, from, to, cutoff=-1)
+   Return all simple paths between 2 nodes in a PyGraph object
+
+   A simple path is a path with no repeated nodes.
+
+   :param PyGraph graph: The graph to find the path in
+   :param int from: The node index to find the paths from
+   :param int to: The node index to find the paths to
+   :param int cutoff: The maximum depth of path to include in the output list
+       of paths. Defaults to -1 which includes all paths regardless of length.
+       Also, any integer value < 1 will include all found paths.
+
+   :returns paths: A list of lists where each inner list is a path path
+   :rtype: list
+
+.. py:function:: dag_all_simple_paths(graph, from, to, cutoff=-1)
+   Return all simple paths between 2 nodes in a PyGraph object
+
+   A simple path is a path with no repeated nodes.
+
+   :param PyDAG graph: The graph to find the path in
+   :param int from: The node index to find the paths from
+   :param int to: The node index to find the paths to
+   :param int cutoff: The maximum depth of path to include in the output list
+       of paths. Defaults to -1 which includes all paths regardless of length.
+       Also, any integer value < 1 will include all found paths.
+
+   :returns paths: A list of lists where each inner list is a path path
+   :rtype: list
+
 .. py:class:: PyGraph
    A class for creating undirected graphs.
 
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -505,7 +505,7 @@ retworkx API
    :return matrix: The adjacency matrix for the input dag as a numpy array
    :rtype: numpy.ndarray
 
-.. py:function:: graph_all_simple_paths(graph, from, to, cutoff=-1)
+.. py:function:: graph_all_simple_paths(graph, from, to, min_depth=None, cutoff=None)
    Return all simple paths between 2 nodes in a PyGraph object
 
    A simple path is a path with no repeated nodes.
@@ -513,14 +513,17 @@ retworkx API
    :param PyGraph graph: The graph to find the path in
    :param int from: The node index to find the paths from
    :param int to: The node index to find the paths to
+   :param int min_depth: The minimum depth of the path to include in the output
+       list of paths. By default all paths are included regardless of depth,
+       sett to 0 will behave like the default.
    :param int cutoff: The maximum depth of path to include in the output list
-       of paths. Defaults to -1 which includes all paths regardless of length.
-       Also, any integer value < 1 will include all found paths.
+       of paths. By default includes all paths regardless of depth, setting to
+       0 will behave like default.
 
    :returns paths: A list of lists where each inner list is a path path
    :rtype: list
 
-.. py:function:: dag_all_simple_paths(graph, from, to, cutoff=-1)
+.. py:function:: dag_all_simple_paths(graph, from, to, min_depth=None, cutoff=None)
    Return all simple paths between 2 nodes in a PyGraph object
 
    A simple path is a path with no repeated nodes.
@@ -528,9 +531,12 @@ retworkx API
    :param PyDAG graph: The graph to find the path in
    :param int from: The node index to find the paths from
    :param int to: The node index to find the paths to
+   :param int min_depth: The minimum depth of the path to include in the output
+       list of paths. By default all paths are included regardless of depth,
+       sett to 0 will behave like the default.
    :param int cutoff: The maximum depth of path to include in the output list
-       of paths. Defaults to -1 which includes all paths regardless of length.
-       Also, any integer value < 1 will include all found paths.
+       of paths. By default includes all paths regardless of depth, setting to
+       0 will behave like default.
 
    :returns paths: A list of lists where each inner list is a path path
    :rtype: list
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -546,16 +546,16 @@ retworkx API
    :param int to: The node index to find the paths to
    :param int min_depth: The minimum depth of the path to include in the output
        list of paths. By default all paths are included regardless of depth,
-       sett to 0 will behave like the default.
+       setting to 0 will behave like the default.
    :param int cutoff: The maximum depth of path to include in the output list
        of paths. By default includes all paths regardless of depth, setting to
        0 will behave like default.
 
-   :returns paths: A list of lists where each inner list is a path path
+   :returns paths: A list of lists where each inner list is a path
    :rtype: list
 
 .. py:function:: dag_all_simple_paths(graph, from, to, min_depth=None, cutoff=None)
-   Return all simple paths between 2 nodes in a PyGraph object
+   Return all simple paths between 2 nodes in a PyDAG object
 
    A simple path is a path with no repeated nodes.
 
@@ -569,7 +569,7 @@ retworkx API
        of paths. By default includes all paths regardless of depth, setting to
        0 will behave like default.
 
-   :returns paths: A list of lists where each inner list is a path path
+   :returns paths: A list of lists where each inner list is a path
    :rtype: list
 
 .. py:class:: PyGraph
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1149,49 +1149,66 @@ fn graph_adjacency_matrix(
     Ok(matrix.into_pyarray(py).into())
 }
 
-#[pyfunction(cutoff=-1)]
+#[pyfunction]
 fn graph_all_simple_paths(
     graph: &graph::PyGraph,
     from: usize,
     to: usize,
-    cutoff: isize,
+    min_depth: Option<usize>,
+    cutoff: Option<usize>,
 ) -> PyResult<Vec<Vec<usize>>> {
     let from_index = NodeIndex::new(from);
     let to_index = NodeIndex::new(to);
-    let cutoff_petgraph: Option<usize>;
-    if cutoff < 1 {
-        cutoff_petgraph = None;
-    } else {
-        cutoff_petgraph = Some(cutoff as usize - 2);
-    }
-    let result: Vec<Vec<usize>> =
-        algo::all_simple_paths(graph, from_index, to_index, 0, cutoff_petgraph)
-            .map(|v: Vec<NodeIndex>| v.into_iter().map(|i| i.index()).collect())
-            .collect();
+    let min_intermediate_nodes: usize = match min_depth {
+        Some(depth) => depth - 2,
+        None => 0,
+    };
+    let cutoff_petgraph: Option<usize> = match cutoff {
+        Some(depth) => Some(depth - 2),
+        None => None,
+    };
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -1149,6 +1149,49 @@ fn graph_adjacency_matrix(
...
+    let from_index = NodeIndex::new(from);
+    let to_index = NodeIndex::new(to);
+    let cutoff_petgraph: Option<usize>;
+    if cutoff < 1 {
+        cutoff_petgraph = None;
+    } else {
+        cutoff_petgraph = Some(cutoff as usize - 2)
+    }
+    let result: Vec<Vec<usize>> =
+        algo::all_simple_paths(graph, from_index, to_index, 0, cutoff_petgraph)
```
ewinston, line 1190: It might be nice to expose `min_intermediate_nodes` as well.

mtreinish: Sure, I was just trying to mirror the networkx call which only has the cutoff parameter. But it's simple to add a kwarg to set this too. Should I do the same thing I did for cutoff and subtract 2 to make the python parameter depth?

ewinston: When I use this method I'm actually looking for all simply connected paths of length m on a device so this is useful. Depth seems good for the application I had in mind but maybe some algorithm could use the default too. Either is ok with me.

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/85/commits/df4761ef36e9c117dfb084500a04583260412db3

docs/source/api.rst
```diff
@@ -505,6 +505,42 @@ retworkx API
...
+   Return all simple paths between 2 nodes in a PyGraph object
+
+   A simple path is a path with no repeated nodes.
+
+   :param PyGraph graph: The graph to find the path in
+   :param int from: The node index to find the paths from
+   :param int to: The node index to find the paths to
+   :param int min_depth: The minimum depth of the path to include in the output
+       list of paths. By default all paths are included regardless of depth,
+       sett to 0 will behave like the default.
```
lcapelluto, line 549:
```suggestion
       setting to 0 will behave like the default.
```

docs/source/api.rst
```diff
@@ -505,6 +505,42 @@ retworkx API
...
+   :param int cutoff: The maximum depth of path to include in the output list
+       of paths. By default includes all paths regardless of depth, setting to
+       0 will behave like default.
+
+   :returns paths: A list of lists where each inner list is a path path
```
lcapelluto, line 554:
```suggestion
   :returns paths: A list of lists where each inner list is a path
```

docs/source/api.rst
```diff
@@ -505,6 +505,42 @@ retworkx API
...
+   :rtype: list
+
+.. py:function:: dag_all_simple_paths(graph, from, to, min_depth=None, cutoff=None)
+   Return all simple paths between 2 nodes in a PyGraph object
```
lcapelluto, line 558:
```suggestion
   Return all simple paths between 2 nodes in a PyDAG object
```

docs/source/api.rst
```diff
@@ -536,6 +536,42 @@ retworkx API
...
+       list of paths. By default all paths are included regardless of depth,
+       sett to 0 will behave like the default.
+   :param int cutoff: The maximum depth of path to include in the output list
+       of paths. By default includes all paths regardless of depth, setting to
+       0 will behave like default.
+
+   :returns paths: A list of lists where each inner list is a path path
+   :rtype: list
+
+.. py:function:: dag_all_simple_paths(graph, from, to, min_depth=None, cutoff=None)
```
lcapelluto, line 557: In a DAG, aren't all paths simple? Should this just be `all_paths`, or `dag_all_paths`?

docs/source/api.rst
```diff
@@ -536,6 +536,42 @@ retworkx API
...
+   :param int from: The node index to find the paths from
+   :param int to: The node index to find the paths to
+   :param int min_depth: The minimum depth of the path to include in the output
+       list of paths. By default all paths are included regardless of depth,
+       sett to 0 will behave like the default.
+   :param int cutoff: The maximum depth of path to include in the output list
+       of paths. By default includes all paths regardless of depth, setting to
+       0 will behave like default.
+
+   :returns paths: A list of lists where each inner list is a path path
```
lcapelluto, line 572:
```suggestion
   :returns paths: A list of lists where each inner list is a path
```

src/lib.rs
```diff
@@ -1195,6 +1195,66 @@ fn graph_adjacency_matrix(
...
+        min_intermediate_nodes,
+        cutoff_petgraph,
+    )
+    .map(|v: Vec<NodeIndex>| v.into_iter().map(|i| i.index()).collect())
+    .collect();
+    Ok(result)
+}
+
+#[pyfunction]
+fn dag_all_simple_paths(
```
lcapelluto, line 1229: It's not possible for one function to implement the feature for both PyDAG and PyGraph? It looks like these functions are the same.

src/lib.rs
```diff
@@ -1195,6 +1195,66 @@ fn graph_adjacency_matrix(
...
+    let to_index = NodeIndex::new(to);
+    let min_intermediate_nodes: usize = match min_depth {
+        Some(depth) => depth - 2,
+        None => 0,
+    };
+    let cutoff_petgraph: Option<usize> = match cutoff {
+        Some(depth) => Some(depth - 2),
+        None => None,
+    };
+    let result: Vec<Vec<usize>> = algo::all_simple_paths(
```
lcapelluto, line 1246: this is provided by petgraph?

tests/test_all_simple_paths.py
```diff
@@ -0,0 +1,347 @@
...
+            dag.add_node(i)
+        edges = [
+            (0, 1),
+            (0, 2),
+            (0, 3),
+            (1, 2),
+            (1, 3),
+            (2, 3),
+            (2, 4),
+            (3, 2),
```
lcapelluto, line 31: 2->3 and 3->2 makes a cycle, so this shouldn't be a DAG... I assume DAG checking is off by default?

tests/test_all_simple_paths.py
```diff
@@ -0,0 +1,347 @@
...
+
+import retworkx
+
+
+class TestDAGAllSimplePaths(unittest.TestCase):
+    def test_all_simple_paths(self):
+        dag = retworkx.PyDAG()
+        for i in range(6):
+            dag.add_node(i)
+        edges = [
```
lcapelluto, line 23: maybe use a setup func and set this to `self.edges` since it looks like they're all the same (and would be helpful for reviewing if I can assume it's the same graph)

tests/test_all_simple_paths.py
```diff
@@ -0,0 +1,347 @@
...
+        ]
+        self.assertEqual(len(expected), len(paths))
+        for i in expected:
+            self.assertIn(i, paths)
+
+    def test_all_simple_path_no_path(self):
+        dag = retworkx.PyDAG()
+        dag.add_node(0)
+        dag.add_node(1)
+        self.assertEqual([], retworkx.dag_all_simple_paths(dag, 0, 5))
```
lcapelluto, line 143:
```suggestion
        self.assertEqual([], retworkx.dag_all_simple_paths(dag, 0, 1))
```
? Should it raise an error if `to` or `from` is not in the graph?

docs/source/api.rst
mtreinish: Yeah in a true dag they are, but with the cycle check disabled PyDAG could be just used as a directional graph, and all the functions that require a DAG have a cycle check in them. What I've been meaning to do is make `PyDAG` a subclass of a new class `PyDiGraph` and remove the entire cycle checking bit from `PyDiGraph` (see #20). The problem is subclassing in rust/PyO3 is a bit weird and I haven't been able to find a working pattern for it yet. It's partially because rust doesn't actually do OO like python, polymorphism is done via shared traits between types and I'm too inexperienced with rust an pyo3 to work my way out of that conflict sp far. But until that's ready I'm just treating PyDAG as a directed graph here since unless cycle check is set it really is one.

src/lib.rs
mtreinish: Yeah, this is a built in algorithm in petgraph: https://docs.rs/petgraph/0.5.1/petgraph/algo/fn.all_simple_paths.html

src/lib.rs
mtreinish: I think it should be possible, but I haven't found a good pattern for doing this yet. The way it would have to be done is defining a trait (or a set of traits) which are defined for both `PyDAG` and `PyGraph` that cover all the required methods and attributes that these functions required. But I've not been able to find a way to do this yet. It would be something like what is done here: https://github.com/Qiskit/retworkx/blob/master/src/dag_isomorphism.rs#L263 Until I figure that out, I've just been duplicating all the functions over both types. I figured when I do come up with a pattern that works I can just deprecate the class specific functions and replace these with `all_simple_paths()` and replace both of these implementations with a call to that.

tests/test_all_simple_paths.py
mtreinish: Yeah, see my earlier comment. By default `PyDAG` isn't actually a DAG. This was done for performance reasons in https://github.com/Qiskit/retworkx/pull/28 so unless we set it to true manually this is just a directed graph

tests/test_all_simple_paths.py
mtreinish: Oh, good catch. Yeah we probably should raise here, I guess the petgraph function doesn't check this and just says there are no paths because it can't ever find a non-existent node index. I'll add a check for this before calling petgraph and update the test (and add a new one for this case).

docs/source/api.rst
mtreinish: I pushed up https://github.com/Qiskit/retworkx/pull/88 to try and resolve the confusing api around this.

src/lib.rs
mtreinish: I tried to play with this a while today, and haven't made much progress. There seems to be a conflict between all the input traits required from petgraph's all_simple_paths and the pyfunction macro that pyo3 uses. The closest I've found is to have a function signature like:

```rust
#[pyfunction]
fn all_simple_paths<'a, T>(
    graph: T,
    from: usize,
    to: usize,
    min_depth: Option<usize>,
    cutoff: Option<usize>,
) -> PyResult<Vec<Vec<usize>>>
where
    T: FromPyObject<'a>
        + petgraph::visit::IntoNeighborsDirected<
            EdgeId = EdgeIndex,
            NodeId = NodeIndex,
            Neighbors = petgraph::stable_graph::Neighbors<'a, PyObject>,
            NeighborsDirected = petgraph::stable_graph::Neighbors<'a, PyObject>,
        > + GraphBase<NodeId = NodeIndex, EdgeId = EdgeIndex>
        + petgraph::visit::NodeCount,
    T::NodeId: Eq + std::hash::Hash,
{
```

but then rust complains about the bounds traits not being implemented for the input `()`. 

```
error[E0277]: the trait bound `(): petgraph::visit::IntoNeighborsDirected` is not satisfied
    --> src/lib.rs:1199:4
     |
1199 |   fn all_simple_paths<'a, T>(
     |      ^^^^^^^^^^^^^^^^
     |      |
     |      the trait `petgraph::visit::IntoNeighborsDirected` is not implemented for `()`
     |      required by a bound in this
...
1208 |           + petgraph::visit::IntoNeighborsDirected<
     |  ___________-
1209 | |             EdgeId = EdgeIndex,
1210 | |             NodeId = NodeIndex,
1211 | |             Neighbors = petgraph::stable_graph::Neighbors<'a, PyObject>,
1212 | |             NeighborsDirected = petgraph::stable_graph::Neighbors<'a, PyObject>,
1213 | |         > + GraphBase<NodeId = NodeIndex, EdgeId = EdgeIndex>
     | |_________- required by this bound in `all_simple_paths`

error[E0277]: the trait bound `(): petgraph::visit::NodeCount` is not satisfied
    --> src/lib.rs:1199:4
     |
1199 | fn all_simple_paths<'a, T>(
     |    ^^^^^^^^^^^^^^^^
     |    |
     |    the trait `petgraph::visit::NodeCount` is not implemented for `()`
     |    required by a bound in this
...
1214 |         + petgraph::visit::NodeCount,
     |           -------------------------- required by this bound in `all_simple_paths`

error[E0277]: the trait bound `(): pyo3::PyClass` is not satisfied
    --> src/lib.rs:1199:4
     |
1199 | fn all_simple_paths<'a, T>(
     |    ^^^^^^^^^^^^^^^^
     |    |
     |    the trait `pyo3::PyClass` is not implemented for `()`
     |    required by a bound in this
...
1207 |     T: FromPyObject<'a>
     |        ---------------- required by this bound in `all_simple_paths`
     |
     = note: required because of the requirements on the impl of `pyo3::FromPyObject<'_>` for `()`

error[E0277]: the trait bound `(): pyo3::PyClass` is not satisfied
    --> src/lib.rs:1198:1
     |
1198 | #[pyfunction]
     | ^^^^^^^^^^^^^ the trait `pyo3::PyClass` is not implemented for `()`
     |
     = note: required because of the requirements on the impl of `pyo3::FromPyObject<'_>` for `()`
     = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to 4 previous errors
```

I'll probably  have to reach out for help from the pyo3 maintainers to figure out how to do this or if it's even possible. I'm probably just overlooking something simple, but it might be a while before I can come up with an answer for this. That being said I think just living with 2 funcs for now doesn't seem like a big deal until we can figure this out.

tests/test_all_simple_paths.py
mtreinish: Fixed in: https://github.com/Qiskit/retworkx/pull/85/commits/0cc66afae85e70b67096306e4900137a1476a166

tests/test_all_simple_paths.py
mtreinish: Fixed in: https://github.com/Qiskit/retworkx/pull/85/commits/0cc66afae85e70b67096306e4900137a1476a166

src/lib.rs
lcapelluto: yeah, that seems fine. thanks for looking into it and explaining :))

PR #88, opened 2020-06-16
Rename PyDAG to PyDiGraph and reorganize modules

### LINKED ISSUE #20: Add a PyDiGraph class

mtreinish:
This commit makes 2 changes, the first is that the rust class PyDAG is
renamed to be PyDiGraph. This makes it more clear that the class can be
used for any directed graph, not just a DAG. The PyDAG class is then
added as a subclass without any changes on the python side of the
module. This will enable a seamless transition for current users where
the only class exposed is PyDAG. Additionally, it will let users
differentiate between a PyDAG and PyDiGraph. Idealling we would want to
only have the cycle checking options exposed for the PyDAG class, but
there isn't any harm if PyDiGraph has it too.

Corresponding with this change the class code for PyDiGraph was moved
to a separate module to make it easier to follow. After this commit
lib.rs only contains the algorithms function code and the module configuration.
This makes it easier to follow as lib.rs has been growing as new
algorithms are added.

Fixes #20

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -557,12 +557,12 @@ retworkx API
    :returns paths: A list of lists where each inner list is a path
    :rtype: list
 
-.. py:function:: dag_all_simple_paths(graph, from, to, min_depth=None, cutoff=None)
-   Return all simple paths between 2 nodes in a PyDAG object
+.. py:function:: digraph_all_simple_paths(graph, from, to, min_depth=None, cutoff=None)
+   Return all simple paths between 2 nodes in a PyDiGraph object
 
    A simple path is a path with no repeated nodes.
 
-   :param PyDAG graph: The graph to find the path in
+   :param PyDiGraph graph: The graph to find the path in
    :param int from: The node index to find the paths from
    :param int to: The node index to find the paths to
    :param int min_depth: The minimum depth of the path to include in the output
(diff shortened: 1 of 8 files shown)


mtreinish:
Once this merges we'll have to update the in progress PRs that add dag_* functions to be digraph_* to  better reflect how they work (or rename them post-merge but before the 0.4.0 release)

docs/source/api.rst
```diff
@@ -700,3 +703,310 @@ retworkx API
...
+   can do this by enabling realtime cycle checking. For example::
+
+       import retworkx
+       dag = retworkx.PyDiGraph()
+       dag.check_cycle = True
+
+   or at object creation::
+
+       import retworkx
+       dag = retworkx.PyDiGraph(True)
```
lcapelluto, line 726: weird to see a flag with no kwarg / more readable as `check_cycle=True`

retworkx/__init__.py
```diff
@@ -1 +1,5 @@
 from .retworkx import *
+
+
+class PyDAG(PyDiGraph):
```
lcapelluto, line 4: This wasn't checking for cycles previously, right?

mtreinish: Right, in #28 we made the default behavior (for performance reasons) to not check for cycles by default. I probably would have preferred to make the `PyDAG` class just set `check_cycle` to true by default, so it's actuallya DAG. But, that would be a breaking change for any current 0.3.4 users out there so I didn't want to do it yet. (maybe for `0.5.0`, or `1.0.0`)

docs/source/api.rst
mtreinish: Sure, done in: https://github.com/Qiskit/retworkx/pull/88/commits/34875a73502eb4adee6b2ab46e1db9f3aadb5152

docs/source/api.rst
```diff
@@ -30,6 +30,9 @@ retworkx API
           When using ``copy.deepcopy()`` or pickling node indexes are not
           guaranteed to be preserved.
 
+    PyDAG, is a subclass of the PyDiGraph class and behaves identically to
```
lcapelluto, line 33:
```suggestion
    PyDAG is a subclass of the PyDiGraph class and behaves identically to
```

PR #91, opened 2020-06-29
Bump PyO3 version and support Stable Rust

mtreinish:
This commit updates retworkx to use the latest version of PyO3 which
enables support for compiling with stable rust (rustc >= 1.39). While
updating the PyO3 version supported this takes the opportunity to update
all the documentation and CI builds to rely on stable rust instead of
nightly. The nightly step was always an extra manual step that detracted
from the contribution process and prevented use in environments where
rustup wasn't allowed to be used.


mtreinish:
This is on hold until rust-numpy releases it's new version which uses PyO3 0.11.0. Until rust numpy updates this won't work because there would be a version mismatch between the PyO3 API used by retworkx and rust-numpy

mtreinish:
Once #90 and this merge that should be it for the `0.4.0`  release.

mtreinish:
Also before this merges I will push up a DNM test commit to verify the wheel builds still work with the new version of cibuidlwheel. I'm not expecting any issues, but we should verify it before release just to be sure.

Cargo.toml
```diff
@@ -19,9 +19,9 @@ crate-type = ["cdylib"]
 petgraph = "0.5.1"
 fixedbitset = "0.2.0"
 hashbrown = "0.7"
-numpy = "0.9.0"
+numpy = "0.10.0"
```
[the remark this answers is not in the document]
mtreinish: It's not up yet, they've merged the version bump: https://github.com/PyO3/rust-numpy/commit/60bdaad175b5dc42588f54a7ac450807be517bd8 but haven't pushed the release yet. This is blocked until its released because without it the pyo3 version it uses doesn't match what retworkx uses.

drewrisinger:
Any possibility for switching to ``maturin`` for building too (instead of ``setuptools-rust``)? https://github.com/PyO3/maturin
Maturin is under active development, while setuptools-rust seems stagnant.

mtreinish:
> Any possibility for switching to `maturin` for building too (instead of `setuptools-rust`)? https://github.com/PyO3/maturin
> Maturin is under active development, while setuptools-rust seems stagnant.

@drewrisinger there aren't any current plans to switch. Was there a particular reason you were looking to switch over to maturin, or just that it is a more active project?

Back when we started retworkx maturin (then called pyo3-pack) was missing a lot of features, most importantly sdist support, so it wasn't viable. Those have been mostly addressed (although not completely) and the main advantage it really provides over setuptools-rust now is that it allows for building the python packages without having to create boilerplate config files in the repo. But, that work has already been done so there isn't really a reason to remove it. Also, in my experience having the setup.py and other files in repo ends up working in more environments (especially those with older pip versions).

drewrisinger:
@mtreinish thanks for the response. I was mostly just wondering for re-packaging this for NixOS. I have at least one or two examples of using ``maturin``, and none with ``setuptools-rust`` due to its reliance on the Rust Nightly toolchain , though I think that is changing with the latest ``pyo3``.

I have no objections, and that makes sense about the existing system working. 👍

README.md
```diff
@@ -25,28 +33,25 @@ pip install retworkx
...
-running:
+### Installing on a platform without precompiled binaries
 
-```bash
-pip install setuptools-rust
-```
-
-prior to running:
+If there are no precompiled binaries published for your system you'll have to
+build the package from source for your system. A source package is also
```
elfrocampeador, line 39: -> "build the package from source." or "build the package from source yourself."  The extra "for your system" is redundant

README.md
```diff
@@ -25,28 +33,25 @@ pip install retworkx
...
+published on pypi, so you still can also run the above `pip` command to install
+it. However, to be able able to build the package from the published source
```
elfrocampeador, line 41: If you want to use however, then maybe put this sentence next to the one that you have to build from source (in which case, i'd rephrase it to "However, to do so..."

README.md
```diff
@@ -59,56 +64,45 @@ git clone https://github.com/Qiskit/retworkx.git
...
-python interface. Unfortunately, this means you need to use nightly rust
-because PyO3 only works with nightly at this point. You can use
-[rustup](https://rustup.rs/) to install rust nightly.
-
-Once you have nightly rust and cargo installed you can easily install
-retworkx into your python environment using pip. Once you have a local
-clone of the repo, change your current working directory to the root of
-the repo. To set the compiler for `retworkx` to nightly rust using
-`rustup`, run the following from this directory:
+python interface. Which enables using standard python tooling to work. So,
```
elfrocampeador, line 67: -> "python interface which enables the use of standard python tooling"

README.md
```diff
@@ -59,56 +64,45 @@ git clone https://github.com/Qiskit/retworkx.git
...
+assuming you have rust installed then you can easily install retworkx into your
```
elfrocampeador, line 68: -> "assuming you have rust installed, you can easily install retworkx into your"

README.md
mtreinish: I think probably just dropping "yourself" so it's just `build the package from source.` is probably better

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/91/commits/504ab0fc206d8e6c7c27474f89c223e00b139969

README.md
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/91/commits/504ab0fc206d8e6c7c27474f89c223e00b139969

README.md
mtreinish: Done in https://github.com/Qiskit/retworkx/pull/91/commits/504ab0fc206d8e6c7c27474f89c223e00b139969

README.md
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/91/commits/504ab0fc206d8e6c7c27474f89c223e00b139969

mtreinish:
Wheel builds all succeeded. The azure jobs which are macOS and windows have them in the artifacts store: https://dev.azure.com/qiskit-ci/retworkx/_build/results?buildId=16278&view=artifacts&type=publishedArtifacts (the travis jobs don't host anywhere). I'll revert the DNM test commit now and this should be good to merge.

PR #92, opened 2020-06-29
Attempt to fix readthedocs build

mtreinish:
Since #90 the docs are now autogenerated from the installed retworkx
package. However, the readthedocs builds are not working because they
do not have rustc installed in their build env so it is unable to build
retworkx. This commit attempts to configure readthedocs to install rust
via conda and then use that to build retworkx.

### DIFF
--- a/.conda-rtd-environment.yml
+++ b/.conda-rtd-environment.yml
@@ -5,3 +5,4 @@ channels:
 dependencies:
   - rust=1.40.0=1
   - m2r=0.2.1=py_0
+  - sphinx=3.0.4=py_0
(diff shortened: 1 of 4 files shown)


mtreinish:
This is blocked on #91 since the only way I could install rust in the readthedocs env was via conda (which seems a bit weird) and that is rust 1.41.0 not nightly. Before we release 0.4.0 we'll have to get the docs builds working again here and then push the release

mtreinish:
Now that #91 has merged and we support building with a stable release of rust this is unblocked. The rtd jobs have been fixed by this pr you can see the build worked here: https://readthedocs.org/projects/retworkx/builds/11468394/

mtreinish:
You can see the updated docs after this fix here: https://retworkx.readthedocs.io/en/latest/

PR #94, opened 2020-07-10
Add *_gnp_random_graph functions

mtreinish:
This commit adds 2 new functions directed_gnp_random_graph and
undirected_gnp_random_graph. These 2 functions are used for generating
a random Erdős-Rényi/binomial graph. The implementation of these 2
functions is based on the networkx function fast_gnp_random_graph. [1]

[1] https://github.com/networkx/networkx/blob/networkx-2.4/networkx/generators/random_graphs.py#L49

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1113,13 +1113,11 @@ pub fn directed_gnp_random_graph(
     }
     let mut v: isize = 0;
     let mut w: isize = -1;
-    let mut lp: f64 = 1.0 - probability;
-    lp = lp.ln();
+    let lp: f64 = (1.0 - probability).ln();
 
     while v < num_nodes {
         let random: f64 = rng.gen_range(0.0, 1.0);
-        let mut lr = 1.0 - random;
-        lr = lr.ln();
+        let lr: f64 = (1.0 - random).ln();
         let ratio: isize = (lr / lp) as isize;
         w = w + 1 + ratio;
         // avoid self loops
@@ -1199,13 +1197,11 @@ pub fn undirected_gnp_random_graph(
     }
     let mut v: isize = 1;
     let mut w: isize = -1;
-    let mut lp: f64 = 1.0 - probability;
-    lp = lp.ln();
+    let lp: f64 = (1.0 - probability).ln();
 
     while v < num_nodes {
         let random: f64 = rng.gen_range(0.0, 1.0);
-        let mut lr = 1.0 - random;
-        lr = lr.ln();
+        let lr = (1.0 - random).ln();
         let ratio: isize = (lr / lp) as isize;
         w = w + 1 + ratio;
         while w >= v && v < num_nodes {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,6 +16,8 @@ extern crate ndarray;
 extern crate numpy;
 extern crate petgraph;
 extern crate pyo3;
+extern crate rand;
+extern crate rand_pcg;
 
 mod astar;
 mod dag_isomorphism;
@@ -28,7 +30,7 @@ use std::collections::{BinaryHeap, HashSet};
 use hashbrown::HashMap;
 
 use pyo3::create_exception;
-use pyo3::exceptions::Exception;
+use pyo3::exceptions::{Exception, IndexError};
 use pyo3::prelude::*;
 use pyo3::types::{PyDict, PyList};
 use pyo3::wrap_pyfunction;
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -1061,17 +1065,177 @@ fn digraph_astar_shortest_path(
...
+    }
+    let mut v: isize = 0;
+    let mut w: isize = -1;
+    let mut lp: f64 = 1.0 - probability;
+    lp = lp.ln();
+
+    while v < num_nodes {
+        let random: f64 = rng.gen_range(0.0, 1.0);
+        let mut lr = 1.0 - random;
+        lr = lr.ln();
```
Cryoris, line 1122: Is there a reason for having these in two separate lines instead of one? (Maybe this is just a rust convention)

tests/test_random.py
```diff
@@ -0,0 +1,62 @@
...
+
+import retworkx
+
+
+class TestGNPRandomGraph(unittest.TestCase):
+
+    def test_random_gnp_directed(self):
+        graph = retworkx.directed_gnp_random_graph(20, .5, 10)
+        self.assertEqual(len(graph), 20)
+        self.assertEqual(len(graph.edges()), 104)
```
Cryoris, line 23: Correct me if im wrong, but if the number of edges is random how can this reproduce the same number without setting a random seed? Is there a seed set somewhere else?

src/lib.rs
```diff
@@ -1061,17 +1065,177 @@ fn digraph_astar_shortest_path(
...
+#[pyfunction]
+#[text_signature = "(num_nodes, probability, seed=None, /)"]
+pub fn directed_gnp_random_graph(
+    py: Python,
+    num_nodes: isize,
+    probability: f64,
+    seed: Option<u64>,
+) -> PyResult<digraph::PyDiGraph> {
+    if num_nodes <= 0 {
+        return Err(IndexError::py_err("num_nodes must be > 0"));
```
Cryoris, line 1099: This is a minor thing, but wouldn't a ValueError make more sense since `num_nodes` is not really used as index?

tests/test_random.py
```diff
@@ -0,0 +1,62 @@
...
+    def test_random_gnp_undirected_invalid_num_nodes(self):
+        with self.assertRaises(IndexError):
+            retworkx.undirected_gnp_random_graph(-23, .5)
+
+    def test_random_gnp_undirected_invalid_probability(self):
+        with self.assertRaises(IndexError):
+            retworkx.undirected_gnp_random_graph(23, 123.5)
+
+    def test_random_gnp_directed_edges_close_to_probability(self):
+        edges = 0
```
Cryoris, line 47: I'm not quite sure I understand this test.

The expected number of edges for a connection probability `p` and `n` nodes is `E = p n(n - 1) / 2`. Wouldn't it make sense to fix a random seed and check that the average number of edges is close to this expected value?

src/lib.rs
mtreinish: Oh, yeah that's probably a better error to return I'll change it. I was struggling to think of what a good exception would be for this. `ValueError`  is better choice

tests/test_random.py
mtreinish: Yeah, the seed is 10 it's the 3rd arg on the random call in L21, I just passed the kwarg in positionally. I'll add the name to avoid confusion.

Cryoris: Oh right! Thanks 👍

tests/test_random.py
mtreinish: I honestly just copied this from the networkx tests: https://github.com/networkx/networkx/blob/master/networkx/generators/tests/test_random_graphs.py#L248-L253 for the equivalent func over there. I thought it was validating that we got at least a reasonable number of edges averaged over 100 run when the probability was near 1 (~45 for graph). But looking at it more closely I'm confused what the point of these are too because it's not doing that.

src/lib.rs
mtreinish: There was no real reason, I'll combine it 1 line.

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/94/commits/25c0a1416e8fad86ba84207a62c77f289ef0245d

tests/test_random.py
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/94/commits/ec119572439c08a284a03b2f8315320ace9d57ff

src/lib.rs
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/94/commits/05464c6bb6207dc0e243c63d22c9ae698e95a099

tests/test_random.py
mtreinish: I just removed the tests in: https://github.com/Qiskit/retworkx/pull/94/commits/7bc46a07061cabf45925fb2d7d58441f4fd203f1 no point in keeping them around if it's not clear what we're actually verifying with the test. We can always add them back if we figure out the point of them.

PR #97, opened 2020-07-12
Add s390x wheels

mtreinish:
This commit adds an s390x wheel job to CI that will run at release time
so we publish wheels for s390x too. We haven't been publishing or
building wheels for pypi since an issue with the manylinux2014 s390x
images and installing rust via rustup combined with disk size limits in
the s390x travis environment preventing use of the offline installer.
The travis issues have now been resolved so we now have enough space
to use the offline rust installer. For the next release we'll be
building s390x wheels.


mtreinish:
The successful build output from the wheel job is here: https://travis-ci.com/github/Qiskit/retworkx/builds/175316878

PR #103, opened 2020-07-19
Add to_dot method to PyGraph and PyDiGraph

mtreinish:
This commit adds a new method to PyGraph and PyDiGraph, to_dot(), which
is used to generate a graphviz dot file for a graph object. The
implementation of this is influenced by petgraph's built-in dot module
[1].

[1] https://github.com/petgraph/petgraph/blob/0.5.1/src/dot.rs


src/digraph.rs
```diff
@@ -971,6 +973,103 @@ impl PyDiGraph {
...
+    ///   dot = pydot.graph_from_dot_data(dot_str.decode('utf8'))[0]
+    ///
+    ///   with tempfile.TemporaryDirectory() as tmpdirname:
+    ///       tmp_path = os.path.join(tmpdirname, 'dag.png')
+    ///       dot.write_png(tmp_path)
+    ///       image = Image.open(tmp_path)
+    ///       os.remove(tmp_path)
+    ///   image
+    ///
+    #[text_signature = "(node_attr, edge_attr, /, graph_attr=None, filename=None)"]
```
lcapelluto, line 1030: What does this line do?

src/dot_utils.rs
```diff
@@ -0,0 +1,85 @@
...
+where
+    T: Write,
+    G: GraphBase
+        + IntoEdgeReferences
+        + IntoNodeReferences
+        + NodeIndexable
+        + GraphProp,
+    G: Data<NodeWeight = PyObject, EdgeWeight = PyObject>,
+    N: FnMut(&PyObject) -> PyResult<BTreeMap<String, String>>,
+    E: FnMut(&PyObject) -> PyResult<BTreeMap<String, String>>,
```
lcapelluto, line 41: I don't follow what's going on with G N E T here

src/dot_utils.rs
```diff
@@ -0,0 +1,85 @@
...
+        )?;
+    }
+    writeln!(file, "}}")?;
+    Ok(())
+}
+
+/// Convert an attr map to an output string
+fn attr_map_to_string(attrs: BTreeMap<String, String>) -> String {
+    if attrs.is_empty() {
+        return "".to_string();
```
lcapelluto, line 77: is `to_string` here a python conversion? or is `""` not a `String`?

tests/test_dot.py
```diff
@@ -0,0 +1,114 @@
...
+        graph.add_node({'color': 'black', 'fillcolor': 'green',
+                        'label': "a", 'style': 'filled'})
+        graph.add_node({'color': 'black', 'fillcolor': 'red',
+                        'label': "a", 'style': 'filled'})
+        graph.add_edge(0, 1, dict(label='1', name='1'))
+        expected = (
+            "digraph {\n0 [color=black, fillcolor=green, label=a, style=filled"
+            "];\n1 [color=black, fillcolor=red, label=a, style=filled];"
+            "\n0 -> 1 [label=1, name=1];\n}\n")
+        res = graph.to_dot(lambda node: node, lambda edge: edge)
```
lcapelluto, line 51: this could be replaced so that node data is not the same as dot attribute options, right?

tests/test_dot.py
```diff
@@ -0,0 +1,114 @@
...
+        expected = (
+            "graph {\n0 [color=black, fillcolor=green, label=a, style=filled"
+            "];\n1 [color=black, fillcolor=red, label=a, style=filled];"
+            "\n0 -- 1 [label=1, name=1];\n}\n")
+        res = graph.to_dot(lambda node: node, lambda edge: edge).decode('utf8')
+        self.assertEqual(expected, res)
+
+    def test_digraph_to_dot(self):
+        graph = retworkx.PyDiGraph()
+        graph.add_node({'color': 'black', 'fillcolor': 'green',
```
lcapelluto, line 42: How much would users normally interact with these dot options?

src/digraph.rs
mtreinish: This line is a macro that defines the text signature for python function over the c api. It basically lets python's `inspect` and other tools that need a signature know the parameters on the function via a string. https://pyo3.rs/v0.11.1/function.html#making-the-function-signature-available-to-python

src/dot_utils.rs
mtreinish: These are [bounds](https://doc.rust-lang.org/stable/rust-by-example/generics/bounds.html) for generic types on this function. Basically it's saying that the parameter `graph` is of generic type `G` where any object passed in via graph implements a bunch of traits. A [trait](https://doc.rust-lang.org/book/ch10-02-traits.html) defines a set of common methods and attributes that can be implemented for a type. The specific list of traits for `G` are from petgraph and basically are just shared functions for graph traversal and access (I got most of them from compiler errors and the similar function in petgraph that generates dot). The definitions of them are here: https://docs.rs/petgraph/0.5.1/petgraph/visit/index.html#traits but they basically just enable us to pass in either a `PyGraph` or `PyDiGraph` and have the function work as expected for both (the trait definitions for either type are at the top of both `digraph.rs` and `graph.rs`).

`T`, `N`, and `E` are easier to follow because they're much simpler. Parameter `file` is a mutable borrow of generic type `T`, where `T` is anything that implements the [Write](https://doc.rust-lang.org/std/io/trait.Write.html) trait. This can be a file object, a buffer, etc. Parameter `node_attrs` is a generic type `N` whiere `N` is any function that takes in a `&PyObject` and returns a `PyResult<BTreeMap<String, String>` and parameter `edge_attrs` is a generic type `E` where it's also a function like that (`E` and `N` are different though because they are not necessarily going to be the same function).

src/dot_utils.rs
mtreinish: `""` isn't a `String` in rust it's a `&str` (https://doc.rust-lang.org/std/primitive.str.html), `to_string()` is being used to convert it to a `String` object (https://doc.rust-lang.org/std/string/struct.String.html).

tests/test_dot.py
mtreinish: Right, that was my reasoning behind making them callables. The node data could be anything so you could define functions to adjust the formatting based on the data. A simple example would be something like:
```python3
import retworkx
def node_color(data):
    if data > 4:
        return {'color': 'black'}
    else:
        return {'color': 'red'}
graph = retworkx.PyDiGraph
a = graph.add_node(52)
graph.add_child(a, 2, {})
graph.to_dot(node_color, lambda edge: edge)
```

What I was thinking of was for being able to replace https://github.com/Qiskit/qiskit-terra/blob/master/qiskit/visualization/dag_visualization.py#L97-L111 with a function that could be passed to the dot method here.

tests/test_dot.py
mtreinish: It depends on how much they want to format the output. If you just did `to_dot(lambda _: {}, lambda _: {})` it works fine but there wouldn't be any formatting just simple lines and no annotation except for the node ids. Something like:

test_dot

Are you thinking that we should make the `*attrs` parameters default to None and make formatting opt-in?

mtreinish: @lcapelluto I modified the methods so that `node_attr` and `edge_attr` are optional and default to None in https://github.com/Qiskit/retworkx/pull/103/commits/c4536b863577412e7bf05c0abda24de01e0ff74a so now you can just call `PyGraph.to_dot()` to generate a dot file without any formatting.

lcapelluto: Cool! I like that a lot. Yeah, it felt like it might be nice to be able to separate building a graph and formatting a graph. I'd usually build one, do some operations, and maybe later want to display it. But I see now that it's just because the function passed is `lambda node: node`, and the user could do something different, like you showed in the other comment thread. Neat!

PR #104, opened 2020-07-21
cargo: add lockfile

drewrisinger:
Per the instructions in the [Cargo Book](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html)
best practices guide, Rust applications should always commit the `Cargo.lock`
file in git to ensure a reproducible build. This adds the most recent
`Cargo.lock` file, generated using ``cargo generate-lockfile``.

This allows reproducible builds of retworkx from source.

Specifically, I'm packaging this for Nixpkgs (https://nixos.org), and this allows me to reproducibly build ``retworkx`` from source vs having to download the PyPi wheel.


drewrisinger:
Huh, interesting. I didn't know that difference @mtreinish. Thanks for the info.

PR #107, opened 2020-07-30
Configure dependabot to ignore fixedbitset for now

mtreinish:
The fixedbitset dependency has to be kept in sync with the version
petgraph is using since we need to interoperate with pegraph using
fixedbitset objects a version mismatch will cause a compilation failure.
Right now dependabot is trying to update the fixedbitset verision
independently which will not pass CI. This commit sets fixedbitset as an
ignored package in the dependabot configuration.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,5 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "daily"
+    ignore:
+      - dependency-name: "fixedbitset"


mtreinish:
Wow dependabot is smarter than I thought, https://github.com/Qiskit/retworkx/pull/106#issuecomment-666323717 I don't think we need to ignore fixedbitset (since it doesn't release often). I'm going to close this for now, if it gets to be a problem we can always explore this again in the future.

PR #110, opened 2020-08-12
Downgrade jupyter-sphinx to fix readthedocs build

mtreinish:
Since the introduction of the `to_dot()` methods in #103 and the use of
jupyter-sphinx in the docs, the read the docs builds have been failing.
It looks like this is due to:
https://github.com/jupyter/jupyter-sphinx/issues/126
This commit fixes this issue by pinning the jupyter sphinx version to
a version that is known working on read the docs.

### DIFF
--- a/.conda-rtd-environment.yml
+++ b/.conda-rtd-environment.yml
@@ -8,6 +8,6 @@ dependencies:
   - pip
   - pip:
     - m2r2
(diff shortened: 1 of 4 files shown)


mtreinish:
This fixes the build: https://readthedocs.org/projects/retworkx/builds/11653485/ now

PR #111, opened 2020-08-13
Add strongly connected components function

mtreinish:
This commit adds a function for computing the strongly connected
components from a PyDiGraph (and PyDAG) object. It returns a list of
lists of node ids of the strongly connected components.


tests/test_strongly_connected.py
```diff
@@ -0,0 +1,63 @@
...
+import unittest
+
+import retworkx
+
+
+class TestStronglyConnected(unittest.TestCase):
+
+    def test_number_strongly_connected_all_strong(self):
+        G = retworkx.PyDiGraph()
+        node_a = G.add_node(1)
```
lcapelluto, line 22: It's confusing that this says `add_node(1)` but then it appears as `[0]` below. Is that right?

mtreinish: It's because 0 is the node id/index while the weight/data payload is 1. The new function is returning the indexes not the data. I can change the assert to be:

```python3
self.assertEqual(retworkx.strongly_connected_components(G),
                         [[node_c], [node_b], [node_a]])
```

if you think that'll be clearer.

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/111/commits/092badfae7a66f8f400bc62bc3930d55474b3e3e

PR #115, opened 2020-08-18
Workaround isomorphism failure with node removals

### LINKED ISSUE #27: dag_isomorphism::is_isomorphic_matching checks don't work on PyDAG with removed node
retworkx::dag_isomorphism::is_isomorphic_matching was written by copy and pasting petgraph's isomorphism functions and changing the input type to be PyDAG instead of petgraph's `Graph`. However, the code wasn't 100% compatible as originally assumed because of node removals. The function assumes that an isomorphic graph will always have the same number of node indexes, however in the case of petgraph's `StableGraph`(which PyDAG uses internally) this is not the case. A stable graph's number of indexes increases on each node add, and if a node is removed the old index is not used again, unlike `Graph`. This will cause a panic like: `thread '<unnamed>' panicked at 'index out of bounds: the len is 4 but the index is 4'` when trying to compare a graph with removals to a graph without any.

A temporary workaround for this is to use deepcopy (or pickle and unpickle) the pydags. This will reset the index counters so they'll be the same. This is not a great answer, it's just a workaround until the function is updated.

mtreinish:
The dag_isomorphism module was forked from upstream petgraph to handle
the PyDiGraph type and also to enable handling exceptions in the python
check functions gracefully. However, when this was done it neglected
that here were limitations with that module which causes failures in
certain scenarios after node removals. This was because the upstream
petgraph implementation was built on the Graph type instead of the
StableGraph type. The only difference between these types is that
StableGraph does not reuse indexes on removals but Graph does. This
can cause there to be holes in the list of node ids. This breaks
assumptions in multiple places of the VF2 implementation causing a
panic if isomorphism checks are run on a PyDiGraph that has nodes
removed. This commit worksaround this limitation by checking if we've
removed nodes from the PyDiGraph object and if we have it iterates over
the graph and clones it into a copy with a condensed set of node ids.

This fix is less than ideal in that it results in a copy of the graph
which will potentially have performance implications, especially for
larger graphs. But after attempting to fix the VF2 implementation that
seems to be a more involved project than I originally hoped. This will
at least workaround the bug until a better more robust VF2
implementation can be written (and likely should be contributed back
upstream to petgraph).

Fixes #27


mtreinish:
@georgesbarron this should fix the issue and remove the need to always deepcopy in `QuantumCircuit.__eq__` If you could test this for the case where you were seeing performance impact I'm curious to see how much this mitigates any issues. This is slower than the ideal fix but since it's self contained in the rust code (except to increase the python ref counter as part of the copy) but should still be faster than unconditionally deepcopying. Also, if there are no removals on the dags being compared then this will be as fast as the ideal fix.

mtreinish:
I did a quick synthetic test of the performance of this and removing the deepcopy from terra's dagcircuit by adding together to aqua operators:
```python3
import time
from qiskit.aqua.operators import X, Y, Z, I, CX, T, H

c_op = (((CX ^ 3) ^ X) @
        (H ^ 7) @
        (X ^ Y ^ Z ^ I ^ X ^ X ^ X) @
        (Y ^ (CX ^ 3)) @
        (X ^ Y ^ Z ^ I ^ X ^ X ^ X))
newop = (((CX ^ 3) ^ X) @
         (H ^ 7) @
         (X ^ Y ^ Z ^ I ^ X ^ X ^ X) @
         (Y ^ (CX ^ 3)) @
         (X ^ Y ^ Z ^ I ^ X ^ X ^ X))
start = time.time()
c_op + newop
stop = time.time()
print(stop - start)
```

With this PR (and the deepcopy removal from terra) the timed line took `0.005 seconds` with qiskit 0.20.0 and retworkx 0.4.0 the timed line took `0.015 seconds`.  So it's definitely going to fix the issue reported in #27 (although it'd be good to confirm that with a more realistic example).

If you'd like to try and recreate this locally, the terra patch:

```diff
diff --git a/qiskit/dagcircuit/dagcircuit.py b/qiskit/dagcircuit/dagcircuit.py
index b98b4092..81c958dd 100644
--- a/qiskit/dagcircuit/dagcircuit.py
+++ b/qiskit/dagcircuit/dagcircuit.py
@@ -773,12 +773,8 @@ class DAGCircuit:
         return full_pred_map, full_succ_map
 
     def __eq__(self, other):
-        # TODO remove deepcopy calls after
-        # https://github.com/mtreinish/retworkx/issues/27 is fixed
-        slf = copy.deepcopy(self._multi_graph)
-        oth = copy.deepcopy(other._multi_graph)
-
-        return rx.is_isomorphic_node_match(slf, oth,
+        return rx.is_isomorphic_node_match(self._multi_graph,
+                                           other._multi_graph,
                                            DAGNode.semantic_eq)
 
     def topological_nodes(self):
```

then I ran:

```
virtualenv test-no-deepcopy
test-no-deepcopy/bin/pip install -e $PATH_TO_TERRA_WITH_DIFF
test-no-deepcopy/bin/pip install git+https://github.com/mtreinish/retworkx@fix-isomorphism
test-no-deepcopy/bin/pip install qiskit-aer qiskit-ignis qiskit-aqua
test-no-deepcopy/bin/python time_circuitop_add.py
```
(this does require rust be installed)

src/dag_isomorphism.rs
```diff
@@ -190,13 +194,44 @@ pub fn is_isomorphic(dag0: &PyDiGraph, dag1: &PyDiGraph) -> PyResult<bool> {
     Ok(res.unwrap_or(false))
 }
 
+fn clone_graph(py: Python, dag: &PyDiGraph) -> PyDiGraph {
```
lcapelluto, line 197: Maybe `reindex_graph`? When I got 247 and 253 it read like `if a node was removed, assign to copy(g), else assign to g`

mtreinish: yep, I agree that's a better name I'll fix it now

mtreinish: Fixed in: https://github.com/Qiskit/retworkx/pull/115/commits/a1da585a06acd81b20f346ab666263e6c9aa2075

PR #116, opened 2020-08-18
Fix handling of index holes in PyDiGraph pickling

mtreinish:
This commit fixes an issue with pickling (and by extension deepcopy)
PyDiGraph or PyDAG objects that have holes in their node id lists.
Previously the node ids would not be preserved across pickling leading
to a compacted list instead of the original node ids. For example, if
you had a PyDiGraph with node ids `[0, 1, 3]` after pickling/deepcopy the
node ids would be `[0, 1, 2]` but otherwise identical. This commit fixes
this issue by adding a check for holes to `__setstate__` method and
incrementing the node id to reproduce a 1:1 mapping with the original
node ids prior to pickling.


mtreinish:
I'm marking this as on hold because this depends on #115 before we can safely remove it. Currently this bug is actually being used by users to workaround #27, so the reordering is a feature. Once there is a fix in place for #27 users will no longer need to run `deepcopy()`  or to pickle objects to compact node indices for doing an isomorphism check. But we don't want to accidentally merge this until that is fixed

PR #117, opened 2020-08-19
Add cycle_basis() function

mtreinish:
This commit adds a new algorithm function cycle_basis() for getting a
list of cycles from a PyGraph object. The implementation is based on
networkx's equivalent function. [1]

[1] https://github.com/networkx/networkx/blob/networkx-2.4/networkx/algorithms/cycles.py#L33


Cryoris (approved):
Thanks for adding the comments, it makes things a lot clearer 😄 LGTM 👍

PR #118, opened 2020-08-19
Add option to set multigraph on PyGraph and PyDiGraph

mtreinish:
This commit adds a new kwarg, multigraph, to the PyGraph and PyDiGraph
constructors. When this is set to `False` the created graph object will not be a
multigraph and only one edge will exist between nodes. If a parallel
edge is attempted to be added it will be ignored since the edge already
exists.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -23,7 +23,7 @@ use hashbrown::{HashMap, HashSet};
 use pyo3::class::PyMappingProtocol;
 use pyo3::exceptions::PyIndexError;
 use pyo3::prelude::*;
-use pyo3::types::{PyDict, PyList, PyLong, PyString, PyTuple};
+use pyo3::types::{PyBool, PyDict, PyList, PyLong, PyString, PyTuple};
 use pyo3::Python;
 
 use super::dot_utils::build_dot;
@@ -76,7 +76,8 @@ use petgraph::visit::{
 ///
 /// :param bool multigraph: When this is set to ``False`` the created PyGraph
 ///     object will not be a multigraph (which is the default behavior). When
-///     ``False`` if parallel edges are added they will be ignored.
+///     ``False`` if parallel edges are added the weight/weight from that
+///     method call will be used to update the existing edge in place.
 ///
 #[pyclass(module = "retworkx")]
 #[text_signature = "()"]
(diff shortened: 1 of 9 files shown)


mtreinish:
Marking this as on hold because there are some rebase conflicts that will need to be resolved (also all the rust side PyGraph struct creation needs to be updated to set the new parameter). Also if we're going to do this we should also have the same support in PyDiGraph. So I don't think this is ready yet until that's done.

tests/graph/test_edges.py
```diff
@@ -190,6 +190,29 @@ def test_add_edge_from_empty_no_data(self):
         res = graph.add_edges_from_no_data([])
         self.assertEqual([], res)
 
+    def test_extend_from_weighted_edge_list_empty(self):
+        graph = retworkx.PyGraph()
```
lcapelluto, line 194: don't need DAG tests?

src/graph.rs
```diff
@@ -290,6 +301,17 @@ impl PyGraph {
             dict_state.get_item("nodes").unwrap().downcast::<PyDict>()?;
         let edges_list =
             dict_state.get_item("edges").unwrap().downcast::<PyList>()?;
+        let nodes_removed_raw = dict_state
+            .get_item("nodes_removed")
+            .unwrap()
+            .downcast::<PyBool>()?;
+        self.node_removed = nodes_removed_raw.extract()?;
```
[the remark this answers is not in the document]
mtreinish: Yeah, it was me being sloppy again. When I went to go add the multigraph attribute to the getstate output I realized we weren't preserving the node_removed flag. I added both here at the same time. I should have split this out into a separate PR thought

tests/graph/test_edges.py
mtreinish: This PR only added the option to `PyGraph`  so there are no changes on `PyDiGraph`/`PyDAG`. There wasn't as pressing a need for this feature on directed graphs so I figured we could just do it later if people wanted it.

mtreinish: I added PyDiGraph/PyDAG support in https://github.com/Qiskit/retworkx/pull/118/commits/0220e88d56573e4aee432622361943f95e605279 it felt odd to omit it (and also makes the interface consistent in #252 for documenting directed and undirected graphs).

lcapelluto (approved):

PR #119, opened 2020-08-19
Implement mapping protocol methods for graph classes

mtreinish:
This commit adds implementations for the `__getitem__`, `__setitem__`, and
`__delitem__` mapping protocol methods to the PyGraph and PyDiGraph
classes. This enables accessing nodes from the graph by using standard
dict/mapping access patterns (ie `graph[2]`) with the node index. This
makes it easier to work with methods and functions that return node
indexes without having to call helper methods.

### DIFF
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -1045,7 +1045,7 @@ impl PyDiGraph {
 
 #[pyproto]
 impl PyMappingProtocol for PyDiGraph {
-    /// Return the nmber of nodes in the graph
+    /// Return the number of nodes in the graph
     fn __len__(&self) -> PyResult<usize> {
         Ok(self.graph.node_count())
     }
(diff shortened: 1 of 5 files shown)


src/digraph.rs
```diff
@@ -1045,10 +1045,35 @@ impl PyDiGraph {
 
 #[pyproto]
 impl PyMappingProtocol for PyDiGraph {
-    /// Return the number of nodes in the graph
+    /// Return the nmber of nodes in the graph
```
mtreinish, line 1048:
```suggestion
    /// Return the number of nodes in the graph
```

tests/graph/test_nodes.py
```diff
@@ -108,3 +108,57 @@ def test_add_node_from_empty(self):
...
+        graph.add_edge(node_a, node_b, "Edgy")
+        with self.assertRaises(IndexError):
+            graph[42]
+
+    def test_set_node_data_setitem(self):
+        graph = retworkx.PyGraph()
+        node_a = graph.add_node('a')
+        node_b = graph.add_node('b')
+        graph.add_edge(node_a, node_b, "Edgy")
+        graph[node_b] = 'Oh so cool'
```
lcapelluto, line 132: :D

PR #121, opened 2020-08-20
Add generators module

mtreinish:
This commit adds a new module to retworkx, generators, which contains a
bunch of helper functions to quickly create various types of graphs. To
start it adds generators for a star graph, path graph, and cycle graph.
We can continue to grow this set out over time.

### DIFF
--- a/tests/generators/test_cycle.py
+++ b/tests/generators/test_cycle.py
@@ -35,7 +35,7 @@ class TestCycleGraph(unittest.TestCase):
             self.assertEqual(graph.out_edges(i), [(i, i + 1, None)])
         self.assertEqual(graph.out_edges(19), [(19, 0, None)])
 
-    def test_cycle_directed_both_weights_and_num(self):
+    def test_cycle_directed_no_weights_or_num(self):
         with self.assertRaises(IndexError):
             retworkx.generators.directed_cycle_graph()
 
--- a/tests/generators/test_cycle.py
+++ b/tests/generators/test_cycle.py
@@ -1,0 +1,55 @@
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import unittest
+
+import retworkx
+
+
+class TestCycleGraph(unittest.TestCase):
+
+    def test_directed_cycle_graph(self):
+        graph = retworkx.generators.directed_cycle_graph(20)
+        self.assertEqual(len(graph), 20)
+        self.assertEqual(len(graph.edges()), 20)
+        for i in range(19):
+            self.assertEqual(graph.out_edges(i), [(i, i + 1, None)])
+        self.assertEqual(graph.out_edges(19), [(19, 0, None)])
+
+    def test_directed_cycle_graph_weights(self):
+        graph = retworkx.generators.directed_cycle_graph(
+            weights=list(range(20)))
+        self.assertEqual(len(graph), 20)
+        self.assertEqual([x for x in range(20)], graph.nodes())
+        self.assertEqual(len(graph.edges()), 20)
+        for i in range(19):
+            self.assertEqual(graph.out_edges(i), [(i, i + 1, None)])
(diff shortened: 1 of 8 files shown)


mtreinish:
Hmm, I think coveralls is a little confused. There was no change in coverage from the previous commit (it was just a merge commit pulling in a requirements bump). Looking at https://coveralls.io/builds/32982740 it shows the new generator module having 97% coverage  but files not changed decreasing. I'm not sure how to retrigger it, rerunning the job didn't work.

mtreinish:
> Hmm, I think coveralls is a little confused. There was no change in coverage from the previous commit (it was just a merge commit pulling in a requirements bump). Looking at https://coveralls.io/builds/32982740 it shows the new generator module having 97% coverage but files not changed decreasing. I'm not sure how to retrigger it, rerunning the job didn't work.

Yeah looking a bit more closely at the coverage reporting there it's missing `src/lib.rs` I'm not sure where it went, but that drops ~1k total lines so the total ratio of covered lines goes down. This actually is increasing coverage by about ~0.6%. I'm not sure why grcov/coveralls isn't reporting the coverage for `src/lib.rs` anymore.

tests/generators/test_path.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class TestPathGraph(unittest.TestCase):
+
+    def test_directed_path_graph(self):
+        graph = retworkx.generators.directed_path_graph(20)
+        self.assertEqual(len(graph), 20)
+        self.assertEqual(len(graph.edges()), 19)
+        for i in range(19):
+            self.assertEqual(graph.out_edges(i), [(i, i + 1, None)])
```
lcapelluto, line 25: Check that there isn't one from 19 to 0?

tests/generators/test_cycle.py
```diff
@@ -0,0 +1,55 @@
...
+        graph = retworkx.generators.directed_cycle_graph(20)
+        self.assertEqual(len(graph), 20)
+        self.assertEqual(len(graph.edges()), 20)
+        for i in range(19):
+            self.assertEqual(graph.out_edges(i), [(i, i + 1, None)])
+        self.assertEqual(graph.out_edges(19), [(19, 0, None)])
+
+    def test_directed_cycle_graph_weights(self):
+        graph = retworkx.generators.directed_cycle_graph(
+            weights=list(range(20)))
```
lcapelluto, line 30: Weights are attached to nodes, rather than edges? And it's also separate from node data (which seems to be `None`, here)?

tests/generators/test_path.py
```diff
@@ -0,0 +1,53 @@
...
+    def test_directed_path_graph_weights(self):
+        graph = retworkx.generators.directed_path_graph(
+            weights=list(range(20)))
+        self.assertEqual(len(graph), 20)
+        self.assertEqual([x for x in range(20)], graph.nodes())
+        self.assertEqual(len(graph.edges()), 19)
+        for i in range(19):
+            self.assertEqual(graph.out_edges(i), [(i, i + 1, None)])
+
+    def test_path_directed_both_weights_and_num(self):
```
lcapelluto, line 38: some of these are called `both_weights_and_num` and should probably be `no_weights_or_num`

tests/generators/test_path.py
mtreinish: Oh, yeah that's a good call. I'll add that assert after the loop now

mtreinish: fixed in: https://github.com/Qiskit/retworkx/pull/121/commits/bce9502291d5817ef44b081d75ec68bb6b4401a6

tests/generators/test_cycle.py
mtreinish: Yeah, the weights here are attached to the nodes, the edge weights end up as `None` in the generator. I use data and weights interchangeably in retworkx (which is probably pretty sloppy on my part). The `None`s below are actually the edge weight/data `PyDiGraph.out_edges()` returns a list of tuples of the form `(source node index, target node index, edge weight)` https://github.com/Qiskit/retworkx/blob/master/src/digraph.rs#L852-L871. The `assertEqual` at L32 is checking the node weights, `PyDigraph.nodes()` returns a list of all the node data https://github.com/Qiskit/retworkx/blob/master/src/digraph.rs#L391-L401

tests/generators/test_path.py
mtreinish: yep, good catch. This was a leftover from an earlier local revision when I was trying to decide on what the api would be. I'll update this now.

mtreinish: Fixed in https://github.com/Qiskit/retworkx/pull/121/commits/745dcafad941b9dcbb4f52bdd28065a39ed9a71e

tests/generators/test_cycle.py
lcapelluto: thank you:)

PR #122, opened 2020-08-20
Add edge list methods to get edgelist from graph objects

mtreinish:
This commit adds 2 new methods, edge_list() and weighted_edge_list(),
to the PyDiGraph (and PyDAG) and PyGraph classes for getting an edge
list from a graph object. They return a list of tuples that represent
all the edges in the graph.


src/digraph.rs
```diff
@@ -543,6 +543,42 @@ impl PyDiGraph {
         }
     }
 
+    /// Get edge list
+    ///
+    /// Returns a list of tuples of the form ``(source, target)`` where
+    /// ``source`` and ``target are the node indices.
```
lcapelluto, line 549:
```suggestion
    /// ``source`` and ``target`` are the node indices.
```

src/digraph.rs
```diff
@@ -543,6 +543,42 @@ impl PyDiGraph {
...
+    pub fn edge_list(&self) -> Vec<(usize, usize)> {
+        self.edge_references()
+            .map(|edge| (edge.source().index(), edge.target().index()))
+            .collect()
+    }
+
+    /// Get edge list with weights
+    ///
+    /// Returns a list of tuples of the form ``(source, target, weight)`` where
+    /// ``source`` and ``target are the node indices and ``weight`` is the
```
lcapelluto, line 562:
```suggestion
    /// ``source`` and ``target`` are the node indices and ``weight`` is the
```

lcapelluto: In the other file, too

lcapelluto (approved):

PR #112, opened 2020-08-13
Use rayon for parallel sort in graph_greedy_color

mtreinish:
This commit replaces the use of sort_by_key() with par_sort_by_key for
a parallelized version of the sort function. This should improve performance
of the sort which is one of the slower components of the function.

Partially addresses #86

Issue #86, opened 2020-06-11
Investigate using Rayon for parallelization


mtreinish:
### What is the expected enhancement?

In #78 we changed the hashmap implementation to use hashbrown for performance, but also to get
potential access to parallel iterators. In general though using rayon https://github.com/rayon-rs/rayon could provide a speed-up for places where we're in pure rust code. Especially for things like sorts where there are drop in parallel replacements from rayon (ie replace `.sort_by_key()` with `.par_sort_by_key` https://docs.rs/rayon/1.3.0/rayon/slice/trait.ParallelSliceMut.html#method.par_sort_by_key.  We should investigate using rayon  to speed up execution where it makes sense.

mtreinish:
Now that #112 has merged rayon is now a dependency for retworkx. This issue is more about exploring how we can leverage rayon to speedup other more parts of retworkx (instead of just a single sort call).

mtreinish:
It's worth pointing out that I have a benchmark suite here https://github.com/mtreinish/retworkx-bench however, there aren't good benchmarks for scaling tests. The benchmarks that do parameter sweeps are not good example graphs for testing scaling (they tend to be pretty pathological cases because I didn't test the generators I used). The large circuit examples are good for testing large networks but part of what we need to test for this is the overhead for smaller graphs and see how a parallel version scales. The new generators module in retworkx is useful for this.

mtreinish:
Another use case of rayon in retworkx is being added in #166 which uses a parallel iterator from ndarray to parallelize over rows in the output matrix to calculate the distance matrix. I'm going to leave this open because it's really a short project to determine where in the existing code base we can get speedups from using rayon for parallelization.

mtreinish:
I'm going to close this issue now, we've been using rayon for parallelization pretty regularly now and in a lot of different places. There hasn't been a unified effort to try to parallelize everything, but I think we can continue to do it in an ad-hoc fashion as it makes sense or we encounter bottlenecks in a function. So there isn't anything left to do on this issue.

PR #126, opened 2020-08-25
Use HashSet from hashbrown instead of std::collections

mtreinish:
This commit switches the HashSet usage in retworkx to use hashbrown's
version of HashSet instead of the version in std::collections. This
provides a performance boost primarily from the use of the AHash default
hash algorithm. It also opens up potential future performance
improvements with hashbrown's rayon support to use parallel iterators.

At the same time since we're already listing rayon as a dependency
in #112 this adds the rayon feature to hashbrown in the cargo.toml
to expose the ParallelIterator traits for hashbrown's HashMap and
HashSet types. While this isn't being used yet it's potentially
useful in the future for #86.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,9 +104,9 @@ checksum = "37ab347416e802de484e4d03c7316c48f1ecb56574dfd4a46a80f173ce1de04d"
 
 [[package]]
 name = "getrandom"
-version = "0.1.14"
+version = "0.1.15"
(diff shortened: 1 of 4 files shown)


mtreinish:
I need to do some more benchmarking on this to see if this is actually a performance benefit for real world use cases. In microbenchmarks it looks promising, but there are a couple of performance edge cases that I haven't been able to figure out where they come from (I'm assuming overhead with the gil and set type conversion).

PR #134, opened 2020-09-04
Add remove_node_retain_edges method to PyDiGraph

mtreinish:
This commit adds a new method remove_node_retain_edges() which is used
to remove a node but add edges from all predecessors to all descendants.


src/graph.rs
```diff
@@ -427,6 +427,41 @@ impl PyGraph {
...
+    pub fn remove_node_retain_edges(
+        &mut self,
+        py: Python,
+        node: usize,
+    ) -> PyResult<()> {
+        let index = NodeIndex::new(node);
+        let mut edge_list: Vec<(NodeIndex, NodeIndex, PyObject)> = Vec::new();
+        for (source, weight) in self
+            .graph
+            .edges_directed(index, petgraph::Direction::Incoming)
```
kdk, line 446: I haven't looked at the graph class much, but does petgraph ignore direction if using a graph instead of a digraph?

src/digraph.rs
```diff
@@ -556,6 +556,41 @@ impl PyDiGraph {
         Ok(())
     }
 
+    /// Remove a node from the graph and add edges from all predecessors to all
+    /// descendants
```
kdk, line 560:
```suggestion
    /// successors
```

kdk (commented):
Nice! This looks good. I would point out in the documentation that the labels of the inbound edges of the removed node will be preserved, and the labels of the outbound edges will be lost.

This adds, for each inbound edge on the removed node, a new edge to the target of every outbound edge. For the case I had in mind in https://github.com/Qiskit/retworkx/pull/66#discussion_r483744084 (again, specific to terra's usage), it would be useful to be able to add an edge only between sources and targets of inbound and outbound edges that shared the same edge label. e.g. If I had a graph like
image
and removed the central CX, to end up with
image
instead of
image

src/graph.rs
mtreinish: Yeah, if you call `edges_directed` on an undirected graph it returns all edges with the target or source set based on the direction: https://docs.rs/petgraph/0.5.1/petgraph/stable_graph/struct.StableGraph.html#method.edges_directed

But, I misread that doc and I need to tweak this so it only loops once because right now it's adding 2 edges between all neighbors and it should just be one.

mtreinish:
> Nice! This looks good. I would point out in the documentation that the labels of the inbound edges of the removed node will be preserved, and the labels of the outbound edges will be lost.
> 
> This adds, for each inbound edge on the removed node, a new edge to the target of every outbound edge. For the case I had in mind in [#66 (comment)](https://github.com/Qiskit/retworkx/pull/66#discussion_r483744084) (again, specific to terra's usage), it would be useful to be able to add an edge only between sources and targets of inbound and outbound edges that shared the same edge label.

Ok, that makes sense. I'm thinking to add 2 optional boolean flags then and documenting this. One to retain the edge weight/data from the outgoing edge instead of the incoming edge. The second is a boolean flag to only add the edge if the edge data between the incoming edge and outgoing edge are the same (this will have overhead since we'll need to go to python to do the comparison. One question on this, should I use a callback function that returns `bool` for this or should I just do python `==` between the edge weights?

kdk:
>  One question on this, should I use a callback function that gets 2 edge weights and returns `bool` for this or should I just do python `==` between the edge weights?

A callback is more general, but `==` would be sufficient for our use case.

src/graph.rs
mtreinish: I decided to just remove the `PyGraph` implementation for now: https://github.com/Qiskit/retworkx/pull/134/commits/bb82c45649f044d88e26dbca090acabaf063b6f5

The only use case we have is for the directed graphs. If we need it for undirected graphs in the future we can add it.

mtreinish:
Ok I added the 2 new kwargs in https://github.com/Qiskit/retworkx/pull/134/commits/b22e478a001f1c10cef66a9e2bd4610e417bc20e

kdk (approved):

PR #136, opened 2020-09-10
Add extend_from_edge_list methods

mtreinish:
This commit adds methods to extend a graph from an edge list. This works
like the add_edges_from methods but also will add nodes if there is an
index in the edge list not present in the graph already.


src/digraph.rs
```diff
@@ -752,6 +752,64 @@ impl PyDiGraph {
...
     }
 
+    /// Extend graph from an edge list
+    ///
+    /// This method differs from :meth:`add_edges_from_no_data` in that it will
+    /// add nodes if a node index is not present in the edge list.
+    ///
+    /// :param list edge_list: A list of tuples of the form ``(source, target)``
+    ///     where source and target are integer node indices. If the node index
+    ///     is no present in the edge list nodes will be added (with a node
```
lcapelluto, line 762:
```suggestion
    ///     is not present in the graph, nodes will be added (with a node
```

src/digraph.rs
```diff
@@ -752,6 +752,64 @@ impl PyDiGraph {
...
+    }
+
+    /// Extend graph from a weighted edge list
+    ///
+    /// This method differs from :meth:`add_edges_from` in that it will
+    /// add nodes if a node index is not present in the edge list.
+    ///
+    /// :param list edge_list: A list of tuples of the form
+    ///     ``(source, target, weight)`` where source and target are integer
+    ///     node indices. If the node index is no present in the edge list
```
lcapelluto, line 791:
```suggestion
    ///     node indices. If the node index is not present in the graph
```

src/graph.rs
```diff
@@ -545,6 +545,62 @@ impl PyGraph {
...
     }
 
+    /// Extend graph from an edge list
+    ///
+    /// This method differs from :meth:`add_edges_from_no_data` in that it will
+    /// add nodes if a node index is not present in the edge list.
+    ///
+    /// :param list edge_list: A list of tuples of the form ``(source, target)``
+    ///     where source and target are integer node indices. If the node index
+    ///     is no present in the edge list nodes will be added (with a node
```
lcapelluto, line 555:
```suggestion
    ///     is not present in the graph, nodes will be added (with a node
```

src/graph.rs
```diff
@@ -545,6 +545,62 @@ impl PyGraph {
...
+    }
+
+    /// Extend graph from a weighted edge list
+    ///
+    /// This method differs from :meth:`add_edges_from` in that it will
+    /// add nodes if a node index is not present in the edge list.
+    ///
+    /// :param list edge_list: A list of tuples of the form
+    ///     ``(source, target, weight)`` where source and target are integer
+    ///     node indices. If the node index is no present in the edge list
```
lcapelluto, line 583:
```suggestion
    ///     node indices. If the node index is not present in the graph,
```

lcapelluto (approved):

PR #143, opened 2020-09-22
Add option to digraph_floyd_warshall to treat input as undirected

mtreinish:
This commit adds a new option to the digraph_floyd_warshall function to
treat the input directed graph as undirected. When set each edge in the
directed graph will be treated as bidirection/undirected for the
generated output distance matrix.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -825,7 +825,7 @@ fn graph_floyd_warshall_numpy(
 ///
 ///     to cast the edge object as a float as the weight.
 /// :param as_undirected: If set to true each directed edge will be treated as
-///     bidirection/undirected.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -824,17 +824,20 @@ fn graph_floyd_warshall_numpy(
 ///         graph_floyd_warshall_numpy(graph, weight_fn: lambda x: float(x))
 ///
 ///     to cast the edge object as a float as the weight.
+/// :param as_undirected: If set to true each directed edge will be treated as
+///     bidirection/undirected.
```
mtreinish, line 828:
```suggestion
///     bidirectional/undirected.
```

PR #144, opened 2020-09-22
Add subgraph method to PyGraph and PyDiGraph

mtreinish:
This commit adds a new method, subgraph(), to PyGraph and PyDiGraph.
When called it takes in a list of node indices and will return a new
object of the same type representing a subgraph containing the node
indices in that list.


itoko:
LGTM. Just one comment, I found two same test files, `tests/test_subgraph.py` and `tests/graph/test_subgraph.py`. I suppose either of them would be sufficient.

mtreinish:
> LGTM. Just one comment, I found two same test files, `tests/test_subgraph.py` and `tests/graph/test_subgraph.py`. I suppose either of them would be sufficient.

Thanks for taking a look. The 2 test files are for the different classes. tests/graph/test_subgraph.py is for `PyGraph` and tests/test_subgraph.py is for `PyDiGraph`. Most of the tests for the 2 graph classes' methods are split this way. But, I can can combine the tests into a single module if you think that is clearer.

itoko:
Oh, I see they are different. I don't think you don't need to combine them in this PR. I couldn't find any convention in retworkx to follow: `PyGraph` and `PyDiGraph` test cases are combined in `test_edgelist.py` while they are separated in `test_nodes.py` and `test_edges.py`. This might be an issue but not urgent at all.

mtreinish:
Ok, I opened about the test organization in #183

src/graph.rs
```diff
@@ -1111,6 +1112,47 @@ impl PyGraph {
...
+    ///     it will silently be ignored.
+    ///
+    /// :returns: A new PyGraph object representing a subgraph of this graph.
+    ///     It is worth noting that node and edge weight/data payloads are
+    ///     passed by reference so if you update (not replace) an object used
+    ///     as the weight in graph or the subgraph it will also be updated in
+    ///     the other.
+    /// :rtype: PyGraph
+    ///
+    #[text_signature = "(other, node_map, /, node_map_func=None, edge_map_func=None)"]
```
mtreinish, line 1129:
```suggestion
    #[text_signature = "(nodes, /)"]
```

PR #146, opened 2020-09-23
Add graph_dfs_edges and digraph_dfs_edges

mtreinish:
This commit adds a new function to retworkx dfs edges for returning a
depth first ordered edge list from a graph object.


tests/test_dfs_edges.py
```diff
@@ -0,0 +1,45 @@
...
+        graph.extend_from_edge_list([(0, 1), (1, 2), (1, 3), (2, 4), (3, 4)])
+        edges = retworkx.graph_dfs_edges(graph, 0)
+        expected = [(0, 1), (1, 2), (2, 4), (4, 3)]
+        self.assertEqual(expected, edges)
+
+    def test_digraph_dfs_edges(self):
+        graph = retworkx.PyDiGraph()
+        graph.extend_from_edge_list([(0, 1), (1, 2), (1, 3), (2, 4), (3, 4)])
+        edges = retworkx.digraph_dfs_edges(graph, 0)
+        expected = [(0, 1), (1, 2), (2, 4), (4, 3)]
```
lcapelluto, line 30: I'm confused. Is `4 -> 3` in this graph? I'd have expected `[(0, 1), (1, 2), (2, 4), (1, 3)]`

mtreinish: Yeah, you're right this looks like a bug in the function. There shouldn't be an edge from 4->3, it should be 1->3. I guess I just blindly copy and pasted the results here when I originally wrote this and didn't think about it.

mtreinish: Ok, I just pushed a fix for this here: https://github.com/Qiskit/retworkx/pull/146/commits/ae22f229acbd1356109e13fe5eb98328b7f1ef01.

The previous algorithm only worked if there was a linear series of edges, but would fall apart when the dfs order had nodes at a different level next to each other, as this showed. Basically it was incorrectly assuming that the previous entry in the dfs node list was always the parent for the edge, which wasn't always true. That commit manually does the dfs traversal and tracks the parents as it does so we can build the correct edge list.

src/lib.rs
```diff
@@ -342,15 +343,43 @@ where
...
+            let children = temp_parent.1.clone();
+            let count = *index_map.get(&parent).unwrap();
+            let mut found = false;
+            let mut index = 0;
+            for child in &children {
+                // Skip children for a parent already inspected.
+                if index < count {
+                    index += 1;
+                    continue;
+                }
```
lcapelluto, line 365: Could you set `index` to `count` instead?

src/lib.rs
```diff
@@ -342,15 +343,43 @@ where
...
+                    grandchildren.reverse();
+                    stack.push((*child, grandchildren));
+                    index_map.insert(*child, 0);
+                    *index_map.get_mut(&parent).unwrap() = index;
+                    found = true;
+                    break;
+                }
+            }
+            if !found || children.is_empty() {
+                stack.pop();
```
lcapelluto, line 381: What does this do?

src/lib.rs
```diff
@@ -342,15 +343,43 @@ where
...
+                index += 1;
+                if !visited.contains(&child) {
+                    out_vec.push((parent.index(), child.index()));
+                    visited.insert(*child);
+                    let mut grandchildren: Vec<NodeIndex> =
+                        graph.neighbors(*child).collect();
+                    grandchildren.reverse();
+                    stack.push((*child, grandchildren));
+                    index_map.insert(*child, 0);
+                    *index_map.get_mut(&parent).unwrap() = index;
```
lcapelluto, line 375: Does this update the value of `index_map` keyed on `parent`?

src/lib.rs
mtreinish: I could, the tricky bit is because this is iterating over the list. To do that I'd probably have to change the for loop to be a while loop an manually access each element. Something like:

```rust
while index < children.len() {
    child = children[index]
    if index < count {
        index = count;
        continue;
...
}
```

src/lib.rs
mtreinish: Yeah, that's what this does. This is basically saying get a mutable reference from the `index_map` HashMap for the key `parent`, unwrap it (ie assuming it's always present) then dereference it (as mutable) and set it's value to index.

src/lib.rs
mtreinish: This removes the last element from `stack`: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.pop

src/lib.rs
mtreinish: Oh, as you mentioned in slack using a slice here is a better path than switching to a while loop. I tested this locally and it works fine. I've pushed this up in: https://github.com/Qiskit/retworkx/pull/146/commits/1f4caafbf65b6ba3aa5db7e4152d4f02d91d7d87 which removes this if statement.

lcapelluto (approved):

PR #147, opened 2020-09-23
Add neighbors methods to graph classes

mtreinish:
This commit adds neighbors methods to the graph classes. It returns a
list of neighbors node indices for a given node in the graph.


src/digraph.rs
```diff
@@ -1073,6 +1073,45 @@ impl PyDiGraph {
...
+
+    /// Get the directed neighbors of a node.
+    ///
+    /// This with return a list of directed neighbor node indices
+    ///
+    /// :param int node: The index of the node to get the neibhors of
+    ///
+    /// :returns: A list of the neighbor node indicies
+    /// :rtype: list
+    pub fn neighbors_directed(
```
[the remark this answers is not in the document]
mtreinish: That's a good point, I will change it to be `successor_indices` and `predecessor_indices` those are better function names. The use case here specifically was for: https://github.com/Qiskit/qiskit-terra/blob/master/qiskit/transpiler/passes/layout/noise_adaptive_layout.py#L124-L126 where the nodes don't have a payload/weight, so the `PyObject` is `None`. We need to return the the indices so that pass can do a lookup on the graph matrices there.

The overhead for returning a list of `PyObject`s isn't actually that high since a `PyObject` is basically just a pointer/reference to an object in the Python heap (https://pyo3.rs/master/types.html#pyt-and-pyobject ). The issue is that there really isn't a mapping from `PyObject` -> graph node (for a qiskit `DAGCircuit` we have this because the weight `PyObject`, is a `DAGNode` object which contains a node_id attribute).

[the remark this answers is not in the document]
mtreinish: Ok, the function was split and renamed in: https://github.com/Qiskit/retworkx/pull/147/commits/75da922d4da7c0444b391a8c4cedc10344b5f49f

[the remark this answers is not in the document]
itoko: I understand in what case current `successors` (returning PyObject) is not sufficient, i.e. why we need `successor_indices`. Thank you.

itoko (changes requested):

src/digraph.rs
```diff
@@ -1076,6 +1076,61 @@ impl PyDiGraph {
         Ok(out_map)
     }
 
+    /// Get the neighbors of a node.
```
itoko, line 1079: LGTM. Just one comment. I think we should note that PyDiGraph.neigbors is the same as PyDiGraph.successor_indices (and different from PyDiGraph.to_directed().neighbors). 
E.g. Get the neighbors (i.e. successors) of a node.

mtreinish: That's a good point, added in: https://github.com/Qiskit/retworkx/pull/147/commits/37c8785860b82120fd3d8dc6eff01fb80f4be4e4

src/digraph.rs
```diff
@@ -1076,6 +1076,62 @@ impl PyDiGraph {
         Ok(out_map)
     }
 
+    /// Get the neighbors (i.e. successors) of a node.
+    ///
+    /// This will return a list of neighbor node indices. This function
+    /// is equivalent to :meth:`get_successor_nodes`.
```
itoko, line 1082: Typo: :meth:`successor_indices`?

itoko, line 1082:
```suggestion
    /// is equivalent to :meth:`successor_indices`.
```

itoko (approved):

PR #156, opened 2020-09-30
Add is_weakly_connected and weakly_connected_components

### LINKED ISSUE #152: Add is_weakly_connected function

Right now retworkx doesn't have an `is_weakly_connected()` function, there is a `number_weakly_connected_components` function which returns the number of weakly connected components in a PyDiGraph, but we don't have a function to return whether the entire graph is weakly connected or not. As part of this it will probably be necessary to add a function to actually get the weakly connected components too, since the petgraph function `number_weakly_connected_components` uses only returns a count.

mtreinish:
This commit adds two new functions to retworkx, is_weakly_connected and
weakly_connected_components, for checking if a PyDiGraph is weakly
connected and to get a list of the weakly connected components of a
PyDiGraph.

Fixes #152


src/lib.rs
```diff
@@ -153,6 +153,67 @@ fn number_weakly_connected_components(graph: &digraph::PyDiGraph) -> usize {
...
+/// :rtype: bool
+///
+/// :raises NullGraph: If an empty graph is passed in
+#[pyfunction]
+#[text_signature = "(graph, /)"]
+pub fn is_weakly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
+    if graph.graph.node_count() == 0 {
+        return Err(NullGraph::new_err("Invalid operation on a NullGraph"));
+    }
+    Ok(weakly_connected_components(graph)[0].len() == graph.graph.node_count())
```
itoko, line 214: Is this line equivalent with `Ok(number_weakly_connected_components(graph) == 1)`?

mtreinish: It might be, I'm not sure if there is a case where you can have >1 component and still have the graph fully connected. I based this check on the networkx equivalent function: https://github.com/networkx/networkx/blob/master/networkx/algorithms/components/weakly_connected.py#L144

itoko (approved):
LGTM. Thanks!

PR #158, opened 2020-10-01
Make weight_fn optional in adjacency matrix and fw numpy

mtreinish:
This commit makes the weight_fn argument for graph_adjacency_matrix,
digraph_adjacency_matrix, graph_floyd_warshall_numpy, and
digraph_floyd_warshall_numpy optional. A new kwarg is added
default_weight (which defaults to 1.0) which can be used instead of
passing a callable. If weight_fn is not set the value of default_weight
will be used for all edges. Previously, a function returning a fixed
value would have to be used to accomplish this. In practice there was
not much overhead to just using something like 'lambda _: 1' as the
weight fn, but it was a bit of a clumsy interface.

TODO:

- [x] Add tests


src/lib.rs
```diff
@@ -818,27 +815,24 @@ fn graph_floyd_warshall_numpy(
 ///     path between two nodes then the corresponding matrix entry will be
 ///     ``np.inf``.
 /// :rtype: numpy.ndarray
-#[pyfunction(as_undirected = "false")]
-#[text_signature = "(graph, weight_fn, /, as_undirected=False)"]
+#[pyfunction(as_undirected = "false", default_weight = "1.0")]
+#[text_signature = "(graph, weight_fn, /, as_undirected=False, default_weight=1.0)"]
```
itoko, line 819: #[text_signature = "(graph, /, weight_fn=None, as_undirected=False, default_weight=1.0)"]
as well as the others?

itoko (commented):
LGTM. Just one nit pick.

src/lib.rs
mtreinish: Good catch, yeah I missed that. Updated in https://github.com/Qiskit/retworkx/pull/158/commits/f2d3b7792e8fc925920c5d0a62772a4f05a173e9

itoko (approved):

PR #161, opened 2020-10-02
Add to_undirected method for PyDiGraph

### LINKED ISSUE #153: Add `to_undirected` method for PyDiGraph

There is no method (short of going over the edge list manually in python) to take a `PyDiGraph` object with directed edges and generating a `PyGraph` object from it with undirected edges. We should have a `to_undirected` method for a `PyDiGraph` that will treat every edge in the graph as undirected and generate a new `PyGraph` object from it.

mtreinish:
This commit adds a new method to the PyDiGraph class, to_undirected(),
which will generate an undirected PyGraph object from the PyDiGraph
object.

Fixes #153


tests/test_to_undirected.py
```diff
@@ -0,0 +1,55 @@
...
+    def test_single_direction_graph(self):
+        digraph = retworkx.generators.directed_path_graph(5)
+        graph = digraph.to_undirected()
+        self.assertEqual(digraph.weighted_edge_list(),
+                         graph.weighted_edge_list())
+
+    def test_bidirectional_graph(self):
+        digraph = retworkx.generators.directed_path_graph(5)
+        for i in range(0, 4):
+            digraph.add_edge(i + 1, i, None)
```
1ucian0, line 34: I'm a bit lost here. `directed_path_graph(5)` creates a path dag with 5 nodes. But in this for-loop you are connecting 6 nodes (from 0 to 5). I'm missing something...

1ucian0 (approved):

src/digraph.rs
```diff
@@ -1561,6 +1563,37 @@ impl PyDiGraph {
         }
         edges.is_empty()
     }
+
+    /// Generate a new PyGraph object from this graph
+    ///
+    /// This will create a new :class:`~retworkx.PyGraph` object from this
+    /// graph. All edges in this graph will be created as undirected edges in
+    /// the new graph obejct.
```
mtreinish, line 1571:
```suggestion
    /// the new graph object.
```

mtreinish:
@1ucian0 thanks for the review

PR #162, opened 2020-10-05
Add Dijkstra shortest path functions

### LINKED ISSUE #151: Add shortest path function

Right now retworkx doesn't have a `shortest_path` function to return the shortest path between 2 nodes in a graph. There a several shortest path length functions that return the shortest path length between nodes. We should add a `retworkx.graph_shortest_path` and `retworkx.digraph_shortest_path` functions that will return the path as node indices from a source to the target node. One thing to mention here is that for the directed graph function there should be an option to treat edges as undirected/bidirectional so that a user can find the undirected shortest path between nodes.

mtreinish:
This commit adds 2 new functions, digraph_dijkstra_shortest_paths() and
graph_dijkstra_shortest_path(), which is a function to get the shortest
path from a node in a graph. It leverages the same dijkstra's algorithm
module which has been modified to get a path in addition to the path
length.
    
Depends on #161
Fixes #151


mtreinish:
Labeling this as on hold because it's based on top of #161. So blocking until #161 merges

mtreinish:
Now that https://github.com/Qiskit/retworkx/pull/161 has merged this is unblocked, so removing the on hold label.

src/dijkstra.rs
```diff
@@ -97,6 +97,7 @@ pub fn dijkstra<G, F, K>(
     start: G::NodeId,
     goal: Option<G::NodeId>,
     mut edge_cost: F,
+    mut path: Option<&mut HashMap<G::NodeId, Vec<G::NodeId>>>,
```
lcapelluto, line 100: Don't need to update docs?

src/lib.rs
```diff
@@ -1384,6 +1369,165 @@ fn digraph_all_simple_paths(
     Ok(result)
 }
 
+fn weight_callable(
+    py: Python,
+    weight_fn: &Option<PyObject>,
+    weight: &PyObject,
```
lcapelluto, line 1375: My intuition tells me this `&` addition makes sense, and it must be related to using `e.weight()` below, but I'm not sure why. Can you explain to me why you added it?

lcapelluto: Does `e.weight()` not return a value?

src/lib.rs
```diff
@@ -1384,6 +1369,165 @@ fn digraph_all_simple_paths(
...
+}
+
+/// Find the shortest path from a node
+///
+/// This function will generate the shortest path from a source node using
+/// Dijkstra's algorithm.
+///
+/// :param PyGraph graph:
+/// :param int source: The node index to find paths from
+/// :param int target: An optional target path to find the path
```
lcapelluto, line 1394:
```suggestion
/// :param int target: An optional target to find a path to
```
?

src/lib.rs
```diff
@@ -1384,6 +1369,165 @@ fn digraph_all_simple_paths(
...
+    weight_fn: Option<PyObject>,
+    default_weight: f64,
+) -> PyResult<PyObject> {
+    let start = NodeIndex::new(source);
+    let goal_index: Option<NodeIndex> = match target {
+        Some(node) => Some(NodeIndex::new(node)),
+        None => None,
+    };
+    let mut paths: HashMap<NodeIndex, Vec<NodeIndex>> = HashMap::new();
+    paths.insert(start, vec![start]);
```
lcapelluto, line 1422: I don't quite follow the format of the `paths` mapping. Is it a dict of "targets" and their paths to that node? So for graph = `0 -> 1 -> 2` and `start = 0`, would it end up: `{0: [0], 1: [0, 1], 2: [0, 1, 2]`?

tests/test_dijkstra.py
```diff
@@ -39,6 +39,240 @@ def test_dijkstra(self):
...
+            (e, f, 6),
+        ]
+        g.add_edges_from(edge_list)
+        paths = retworkx.digraph_dijkstra_shortest_paths(g, a)
+        expected = {
+            1: [0, 1],
+            2: [0, 3, 2],
+            3: [0, 3],
+            4: [0, 3, 4],
+            5: [0, 1, 5],
```
lcapelluto, line 68:
```suggestion
        expected = {
            # a -> b
            1: [0, 1],
            # a -> c: a, d, c
            2: [0, 3, 2],
            # a -> d
            3: [0, 3],
            # a -> e: a, d, e
            4: [0, 3, 4],
            # a -> f: a, b, f
            5: [0, 1, 5],
```

lcapelluto: This helped me review, up to you if you want to add it. My alpha->zero indexed numeric mental conversion isn't that sharp

tests/test_dijkstra.py
```diff
@@ -39,6 +39,240 @@ def test_dijkstra(self):
...
+            (b, f, 15),
+            (c, f, 11),
+            (e, f, 6),
+        ]
+        g.add_edges_from(edge_list)
+        paths = retworkx.digraph_dijkstra_shortest_paths(
+            g, a, weight_fn=lambda x: x)
+        expected = {
+            1: [0, 1],
+            2: [0, 1, 2],
```
lcapelluto, line 96: The other one doesn't find this path because of the ordering of the edges, right? And here this one is found because it takes the weights into account?

tests/test_dijkstra.py
```diff
@@ -39,6 +39,240 @@ def test_dijkstra(self):
         expected = {4: 23.0}
         self.assertEqual(expected, path)
 
+    def test_dijkstra_path(self):
+        g = retworkx.PyDiGraph()
```
lcapelluto, line 43: If this is the same graph for every test, could it be moved into a `setUp`? Or is there a diff that I'm not catching?

src/lib.rs
mtreinish: `e.weight()` returns a `&PyObject` a reference/borrow of a PyObject. So to avoid needing to increment the python ref counter and creating a separate object I made this change so we can just pass the reference to python directly.

src/lib.rs
mtreinish: Yeah, that's it. But we remove the 0 field being added here in the output dict below at L1434. This needs to be in the dict we pass to `dijkstra::dijkstra` because the dijkstra function uses the values in the paths dict on each iteration so without it there it would fail. 

Basically this line is just a starter for building the path dict for the children of the source node so that in: https://github.com/Qiskit/retworkx/pull/162/files/42eb1656062122528ab2051f6cfc23126d630aa5#diff-35666815bfec005ea9e5e2774a0cee5e97ea7af1ee9db63f254f8bfe7d59256aR138-R147 the paths include 0.

src/dijkstra.rs
mtreinish: Yeah, I probably should, even though it's internal for retworkx we should still keep it documented

src/lib.rs
mtreinish: Actually writing this and adding docstrings for this function. I'm thinking I should just move this line into the `dijkstra::dijkstra()` function. So instead of expecting that users add an entry to the HashMap before passing it to the function we add it before starting the traversal. What do you think?

src/dijkstra.rs
mtreinish: Added in: https://github.com/Qiskit/retworkx/pull/162/commits/c16f64b2d1b5c90bb7b9ebb5f729fdf1e503e740

tests/test_dijkstra.py
mtreinish: Yeah, the weight value is set to an default of `1.0` in the above test method, so the path it returns is based on the binary heap insertion order (LIFO iirc). While here the weight is being taken into account so a-> b has a weight of 7 vs a->e having a weight of 14.

tests/test_dijkstra.py
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/162/commits/06731356ccbfd37a408b4c3bbe9dfc7fe3e7bbda

src/lib.rs
lcapelluto: Yes that makes sense to me. It seems that this piece is part of the dijsktra algorithm itself, rather than a required input.

mtreinish: Ok, done in: https://github.com/Qiskit/retworkx/pull/162/commits/7eab33db66f81daad36223ee2730265660c8c7e7

lcapelluto (approved):

PR #166, opened 2020-10-06
Add distance matrix functions

mtreinish:
This commit adds 2 new functions, `graph_distance_matrix()` and
`digraph_distance_matrix()`, which will generate a distance matrix for a
graph object. It is for use cases when all edges are unweighted or have
the same weight and is significantly faster than other algorithms
because of this assumption. It also operates in parallel if the size of
the graph justifies it.


src/lib.rs
```diff
@@ -1010,6 +1010,162 @@ fn layers(
...
+            level += 1
+        }
+    };
+    if n < parallel_threshold {
+        matrix
+            .axis_iter_mut(Axis(0))
+            .enumerate()
+            .for_each(|(index, row)| bfs_traversal(index, row));
+    } else {
+        // Parallelize by row and iterate from each row index in BFS order
```
lcapelluto, line 1159: This is a helpful comment. Consider adding it above, as well

lcapelluto (approved):

src/lib.rs
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/166/commits/f5d4a784ab2dcae4ce6a9c8f4d72c2c21c3793e4

PR #168, opened 2020-10-07
Add support for Python 3.9

mtreinish:
Python 3.9.0 was released on 10-05-2020, this commits marks the support
of Python 3.9 in qiskit-ignis. It adds the supported python version in
the package metadata and updates the CI configuration to run test jobs
on Python 3.9 and build wheels for Python 3.9.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,7 @@ jobs:
     strategy:
       matrix:
         rust: [stable]
-        python-version: [3.5, 3.6, 3.7, 3.8]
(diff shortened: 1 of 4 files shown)


mtreinish:
Adding on hold here until numpy publishes wheels for python 3.9 support. Right now the macOS build is failing because it's linking against whatever blas lib is installed in the macOS CI node image and causing internal numpy errors. While this can probably be fixed by installing openblas locally in the CI job it's going to be a potential issue for users until numpy publishes a release with 3.9 support.

PR #173, opened 2020-10-17
Added find cycle support for PyDiGraph

### LINKED ISSUE #155: Add find_cycle functions

Retworkx is missing functionality equivalent to networkx's `find_cycle` method: https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.cycles.find_cycle.html
while retworkx does have `cycle_basis` functions it doesn't have a method to find cycles from arbitrary nodes. However, to start the retworkx functions added don't need to have the same level of flexibility as networkx's. The immediate use case for this is: https://github.com/Qiskit/qiskit-terra/blob/779fa14d4452e35306ca849d98e56f8aa40d14be/qiskit/transpiler/passes/routing/algorithms/token_swapper.py#L160 which only cares about finding a cycle from a source node without an orientation. If the need arises after that is added we can always expand it later.

darknight009:
Fixes #155

- Feature support for finding a cycle in a PyDiGraph from a starting node

- Picks an arbitrary node if starting node is not defined

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1696,6 +1696,86 @@ pub fn strongly_connected_components(
         .collect()
 }
 
+/// Return the first cycle encountered during DFS of a given PyDiGraph
+///
+/// :param PyDiGraph graph: The graph to find the cycle in
+/// :param int root: Optional index for starting node for cycle
+///
+/// :returns: A list describing the cycle. The index of node ids which
+///     forms a cycle (loop) in the input graph
+/// :rtype: list
+#[pyfunction]
+#[text_signature = "(graph, /, root=None)"]
+pub fn digraph_find_cycle(
+    graph: &digraph::PyDiGraph,
+    root: Option<usize>,
+) -> Vec<(usize, usize)> {
+    let mut root_node = root;
+    let mut graph_nodes: HashSet<NodeIndex> =
+        graph.graph.node_indices().collect();
+    let mut cycle: Vec<(usize, usize)> = Vec::new();
+    let temp_value: NodeIndex;
+    // If root_node is not set get an arbitrary node from the set of graph
+    // nodes we've not "examined"
+    let root_index = match root_node {
+        Some(root_value) => NodeIndex::new(root_value),
+        None => {
+            temp_value = *graph_nodes.iter().next().unwrap();
+            graph_nodes.remove(&temp_value);
+            temp_value
+        }
+    };
+
+    // Stack (ie "pushdown list") of vertices already in the spanning tree
+    let mut stack: Vec<NodeIndex> = Vec::new();
+    stack.push(root_index);
+    // map to store parent of a node
+    let mut pred: HashMap<NodeIndex, NodeIndex> = HashMap::new();
+    let mut visiting = HashSet::new();
+    let mut visited = HashSet::new();
+    while !stack.is_empty() {
+        let mut z = stack.pop().unwrap();
+        stack.push(z);
+        visiting.insert(z);
+
+        let children = graph
+            .graph
+            .neighbors_directed(z, petgraph::Direction::Outgoing);
+
+        for child in children {
+            //cycle is found
+            if visiting.contains(&child) {
+                cycle.push((z.index(), child.index()));
+                //backtrack
+                loop {
+                    if z == child {
+                        cycle.reverse();
+                        break;
+                    }
+                    cycle.push((pred[&z].index(), z.index()));
+                    z = pred[&z];
+                }
+                return cycle;
+            }
+            //if an unexplored node is encountered
+            if !visited.contains(&child) {
+                stack.push(child);
+                pred.insert(child, z);
+            }
+        }
+
+        let mut top = stack.pop().unwrap();
+        stack.push(top);
+        //if no further children and explored, move to visited
+        if top.index() == z.index() {
+            stack.pop();
+            visiting.remove(&z);
+            visited.insert(z);
+        }
+    }
+    cycle
+}
+
 // The provided node is invalid.
 create_exception!(retworkx, InvalidNode, PyException);
 // Performing this operation would result in trying to add a cycle to a DAG.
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -1696,6 +1696,86 @@ pub fn strongly_connected_components(
...
+    let root_index = match root_node {
+        Some(root_value) => NodeIndex::new(root_value),
+        None => {
+            temp_value = *graph_nodes.iter().next().unwrap();
+            graph_nodes.remove(&temp_value);
+            temp_value
+        }
+    };
+
+    // Stack (ie "pushdown list") of vertices already in the spanning tree
```
mtreinish, line 1729: Heh, this comment was kind of a joke and mostly made made sense in the context of `cycle_basis` functions. The paper describing the algorithm for `cycle_basis` is quite old and called a stack a pushdown list, which is a descriptive name but not one commonly used because it was before CS terminology was standardized. But I'm glad to see it catching on :)

src/lib.rs
```diff
@@ -1696,6 +1696,86 @@ pub fn strongly_connected_components(
...
+    let mut stack: Vec<NodeIndex> = Vec::new();
+    stack.push(root_index);
+    // map to store parent of a node
+    let mut pred: HashMap<NodeIndex, NodeIndex> = HashMap::new();
+    let mut visiting = HashSet::new();
+    let mut visited = HashSet::new();
+    while !stack.is_empty() {
+        let mut z = stack.pop().unwrap();
+        stack.push(z);
```
mtreinish, line 1738: I'm a bit confused by this, you pop `z` off the stack and then just push it back on? Are you trying to just access the last element of the Vec? If so could you replace these 2 lines with:

```suggestion
        let mut z = *stack.last().unwrap();
```

src/lib.rs
```diff
@@ -1696,6 +1696,86 @@ pub fn strongly_connected_components(
...
+        }
+    };
+
+    // Stack (ie "pushdown list") of vertices already in the spanning tree
+    let mut stack: Vec<NodeIndex> = Vec::new();
+    stack.push(root_index);
+    // map to store parent of a node
+    let mut pred: HashMap<NodeIndex, NodeIndex> = HashMap::new();
+    let mut visiting = HashSet::new();
+    let mut visited = HashSet::new();
```
mtreinish, line 1735: Could you add a comment on how these 2 are being used? The names are very similar so it's potentially confusing to understand how they're used.

src/lib.rs
```diff
@@ -1696,6 +1696,86 @@ pub fn strongly_connected_components(
         .collect()
 }
 
+/// Return the first cycle encountered during DFS of a given PyDiGraph
+///
```
mtreinish, line 1700: We probably should add a note here that the list will be empty if there is no cycle found. I think networkx raises an exception which we could do too. It doesn't really matter which way we go about it, just that we document the behavior.

src/lib.rs
```diff
@@ -1696,6 +1696,86 @@ pub fn strongly_connected_components(
...
+            }
+            //if an unexplored node is encountered
+            if !visited.contains(&child) {
+                stack.push(child);
+                pred.insert(child, z);
+            }
+        }
+
+        let mut top = stack.pop().unwrap();
+        stack.push(top);
```
mtreinish, line 1768: Same question as above, is this just to get the last element in the Vec without removing it? If so we should probably do
```suggestion
        let top = *stack.last().unwrap();
```
instead.

mtreinish (commented):
At a high level this LGTM, thanks for getting this started. This is a bit different from the networkx algorithm for their `find_cycles` method, but I think it will work the same way. Just a few inline comments on some things that I think we can clean up.

Also besides the python lint error that needs to be fixed in the unittests there are also a couple of rust compiler warnings we should fix too:
https://github.com/Qiskit/retworkx/pull/173/checks?check_run_id=1269322152#step:6:120

src/lib.rs
darknight009: Yes I was trying to access the last element. I wrote it this way to preserve the idea of stack. I'd change it to your suggested way

src/lib.rs
darknight009: Hah! Let's keep it then

src/lib.rs
```diff
@@ -1721,6 +1721,87 @@ pub fn strongly_connected_components(
...
+/// :param int root: Optional index for starting node for cycle
+///
+/// :returns: A list describing the cycle. The index of node ids which
+///     forms a cycle (loop) in the input graph
+/// :rtype: list
+#[pyfunction]
+#[text_signature = "(graph, /, root=None)"]
+pub fn digraph_find_cycle(
+    graph: &digraph::PyDiGraph,
+    root: Option<usize>,
```
mtreinish, line 1737:
```suggestion
    source: Option<usize>,
```

src/lib.rs
```diff
@@ -1721,6 +1721,87 @@ pub fn strongly_connected_components(
         .collect()
 }
 
+/// Return the first cycle encountered during DFS of a given PyDiGraph,
+/// empty list is returned if no cycle is found
+///
+/// :param PyDiGraph graph: The graph to find the cycle in
+/// :param int root: Optional index for starting node for cycle
```
mtreinish, line 1728:
```suggestion
/// :param int source: Optional index to find a cycle for. If not specified an
        arbitrary node will be selected from the graph.
```

src/lib.rs
```diff
@@ -1721,6 +1721,87 @@ pub fn strongly_connected_components(
...
+/// :returns: A list describing the cycle. The index of node ids which
+///     forms a cycle (loop) in the input graph
+/// :rtype: list
+#[pyfunction]
+#[text_signature = "(graph, /, root=None)"]
+pub fn digraph_find_cycle(
+    graph: &digraph::PyDiGraph,
+    root: Option<usize>,
+) -> Vec<(usize, usize)> {
+    let root_node = root;
```
mtreinish, line 1739:
```suggestion
```

src/lib.rs
```diff
@@ -1721,6 +1721,87 @@ pub fn strongly_connected_components(
...
+    let mut graph_nodes: HashSet<NodeIndex> =
+        graph.graph.node_indices().collect();
+    let mut cycle: Vec<(usize, usize)> = Vec::new();
+    let temp_value: NodeIndex;
+    // If root_node is not set get an arbitrary node from the set of graph
+    // nodes we've not "examined"
+    let root_index = match root_node {
```
mtreinish, line 1746:
```suggestion
    let root_index = match root {
```

or really after we change `root` -> `source` above

```suggestion
    let root_index = match source {
```

mtreinish (approved):
Ok, awesome thanks for making the update. There are 2 small changes I think we should make before merging, but otherwise I think this is good to go. The first is we should rename `root` to `source`.  This just matches the api a bit more to what networkx is doing so it makes the migration a bit easier for people (especially in the qiskit-terra case where it does `source=`). 

The second change is just a small code deletion we don't need to assign `root` to `root_node` in the function, that was necessary in the `cycle_basis` function because we were going to get multiple roots while iterating over the full graph, but since this function just finds a single cycle that doesn't happen so we don't need an intermediate variable to avoid mutating the parameter input.

darknight009:
Thank you for the support! Can I delete this branch now?

mtreinish:
Yes, you can delete your branch for this PR now. Your new function has been merged into retworkx's master branch now. Thanks for handling this.

PR #178, opened 2020-10-19
Switch gen_range() to Uniform in random graph functions

mtreinish:
This commit updates the gnp random graph functions to internally use
Uniform.sample() instead of rng.gen_range(). The rand docs for
gen_range() recommend the use of Uniform if you're going to getting
random values in a range more than once, as the gen_range() function is
optimized for the single sample case. [1][2]

[1] https://docs.rs/rand/0.7.3/rand/trait.Rng.html#method.gen_range
[2] https://docs.rs/rand/0.7.3/rand/distributions/uniform/struct.Uniform.html

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,6 +42,7 @@ use petgraph::visit::{
 
 use ndarray::prelude::*;
 use numpy::IntoPyArray;
+use rand::distributions::{Distribution, Uniform};
 use rand::prelude::*;
 use rand_pcg::Pcg64;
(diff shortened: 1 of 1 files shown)


mtreinish:
@MoAllabbad Does this look good to you?

MoAllabbad (approved):

MoAllabbad:
> 
> 
> @MoAllabbad Does this look good to you?

LGTM. Thanks.

PR #179, opened 2020-10-19
Add remove_edges_from method to PyGraph

### LINKED ISSUE #177: Missing method to remove multiple edges simultaneously in PyGraph class
A method will be added to the PyGraph class which can simultaneously remove multiple edges.

The PyGraph class has the method `remove_nodes_from` which accepts multiple indices to remove in a single call. However, there is no corresponding `remove_edges_from` to remove multiple edges. Are there any implementation details preventing this method from existing?

seatonullberg:
This commit adds a new method remove_edges_from() to remove multiple edges in a single call.
Fixes #177

### DIFF
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -897,6 +897,35 @@ impl PyDiGraph {
         Ok(())
     }
 
+    /// Remove edges from the graph.
+    ///
+    /// Note if there are multiple edges between the specified nodes only one
+    /// will be removed.
+    ///
+    /// :param list index_list: A list of node index pairs to remove from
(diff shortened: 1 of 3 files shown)


mtreinish (approved):
This LGTM. thanks for pushing this. Before we merge it though could you add the same method to `PyDiGraph` too? (it should just be a copy and paste)

seatonullberg:
This commit adds the same method to `PyDiGraph` as you suggested.

mtreinish (approved):
LGTM, thanks for adding it to `PyDiGraph` too

PR #181, opened 2020-10-20
Add insert_node_on_in_edges and insert_node_on_out_edges methods

mtreinish:
This commit adds 2 new methods to the PyDiGraph class,
insert_node_between() and insert_node_between_multiple. These methods
are for inserting edges between a node and another node that has edges.
You specify a node and which node to insert it between and it will loop
over either the incoming or outgoing edges and add edges between the
existing edges.

### DIFF
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -325,6 +325,52 @@ impl PyDiGraph {
         let edge = self.graph.add_edge(p_index, c_index, edge);
         Ok(edge.index())
     }
+
+    fn insert_between(
+        &mut self,
+        py: Python,
+        node: usize,
+        node_between: usize,
+        direction: bool,
+    ) -> PyResult<()> {
+        let dir = if direction {
+            petgraph::Direction::Outgoing
+        } else {
+            petgraph::Direction::Incoming
+        };
+        let index = NodeIndex::new(node);
+        let node_between_index = NodeIndex::new(node_between);
+        let edges: Vec<(NodeIndex, EdgeIndex, PyObject)> = self
+            .graph
+            .edges_directed(node_between_index, dir)
+            .map(|edge| {
+                if direction {
+                    (edge.target(), edge.id(), edge.weight().clone_ref(py))
+                } else {
+                    (edge.source(), edge.id(), edge.weight().clone_ref(py))
+                }
+            })
+            .collect::<Vec<(NodeIndex, EdgeIndex, PyObject)>>();
+        for (other_index, edge_index, weight) in edges {
+            if direction {
+                self._add_edge(
+                    node_between_index,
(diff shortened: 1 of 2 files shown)


mtreinish:
Marking as on hold because I need to add more unittests to improve the coverage, and also add more than the simple cases.

itoko (commented):
Similar to #147, I think it would be better to have `insert_node_on_in_edges(node, node_ref)` and `insert_node_on_out_edges` rather than unified `insert_node_between(node, node_ref, direction)`. What do you think?

mtreinish:
> Similar to #147, I think it would be better to have `insert_node_on_in_edges(node, node_ref)` and `insert_node_on_out_edges` rather than unified `insert_node_between(node, node_ref, direction)`. What do you think?

Yeah that's a good idea. I'll update it shortly to split the functions

itoko:
@mtreinish Thank you. I have one more comment on docstrings.

I think we can describe the spec more clearly without parent/child notation:
```
/// Insert a node between a reference node and all its predecessor nodes
///
/// This essentially iterates over all edges into the reference node specified
/// in the ``ref_node`` parameter removes those edges and then adds 2 edges,
/// one from the predecessor of ``ref_node`` to ``node`` and the other from
/// ``node`` to ``ref_node``. The edge payloads for the newly created edges
/// are copied by reference from the original edge that gets removed.
///
/// :param int node: The node index to insert between
/// :param int ref_node: The reference node index to insert ``node`` between
pub fn insert_node_on_in_edges(..., node, ref_node)

/// Insert a node between a reference node and all its successor nodes
///
/// This essentially iterates over all edges out of the reference node specified
/// in the ``ref_node`` parameter removes those edges and then adds 2 edges,
/// one from ``ref_node`` to ``node`` and the other from ``node`` to the
/// successor of ``ref_node``. The edge payloads for the newly created edges
/// are copied by reference from the original edge that gets removed.
pub fn insert_node_on_out_edges(..., node, ref_node)
```
(I like the term `reference node` (ref_node) while I'm afraid someone might confuse with "reference of node object" (node_ref). Anyway, I think we could have better name than `node_between`. Any good idea?)

BTW, I personally prefer `start -> end` notation to `parent -> child` or `tail -> head` notation.
Honestly, I'm always confused with which is the "head" of parent and child...
(I know the importance of consistency, so if we have already used parent/child notations in many places, I think we should follow the convention.)

mtreinish:
@itoko thank you for the suggestions. I was actually hoping you'd leave a comment like that. I thought the names were awkward in this function, but couldn't come up with anything better so just left it as is. I'll update it now per your suggestions.

itoko (approved):

Issue #71, opened 2020-05-02
iterators?


ajavadia:
I don't think retworkx uses iterators, the same way networkx does? This might lead to very large memory usage for large graphs (and possibly hurt time too).

For example often you want a shallow view into the immediate successors of a node, not all the nodes that may succeed it until the end. Same for layers of a dag.

Networkx[ implements this an an iterator](https://networkx.github.io/documentation/stable/reference/classes/generated/networkx.DiGraph.successors.html). Terra also uses this as an iterator.

But I think retworkx [constructs the full list](https://github.com/Qiskit/retworkx/blob/181c98f62912873ffecf325ee217973033c43795/src/lib.rs#L341), and only artificially returns [an iterator in terra](https://github.com/Qiskit/qiskit-terra/blob/b41891aaffec0896294096cdcc24bf8c8ec6e540/qiskit/dagcircuit/retworkx_dagcircuit.py#L116).

For an example of how this can be problematic for large circuits, see this PR in terra: https://github.com/Qiskit/qiskit-terra/pull/371

mtreinish:
The main reason was the ease of implementation, iterators behave a bit differently between rust and python and building a rust function that returns a python iterator wasn't as straightforward as just doing `yield data` from inside a function and probably would require creating a custom type that implements the python iter protocol and then returning an object of that new type. At least for the first version to get something working it was easier to just return a list.

I agree that for large results implementing it with iterators could be more efficient. But at least so far I haven't seen any scaling limits or memory growth issues with it from the terra dagcircuit. In general memory consumption went down [1] [2] and all graph operations became about an order of magnitude faster in aggregate compared to the previous networkx implementation. Although this is likely just the benefit of using a compiled language and there could still be a scaling limit as things increase.

I think we should leave this open as a lower priority future enhancement though and if the need arises (like we start hitting scaling limits) we can prioritize it.

[1] https://qiskit.github.io/qiskit/#transpiler_levels.TranspilerLevelBenchmarks.peakmem_quantum_volume_transpile_50_x_20?commits=95328adb
[2] https://qiskit.github.io/qiskit/#transpiler_levels.TranspilerLevelBenchmarks.peakmem_transpile_from_large_qasm_backend_with_prop?commits=95328adb (this is an eoh qasm example from aqua loaded from qasm: https://github.com/Qiskit/qiskit/blob/master/test/benchmarks/qasm/test_eoh_qasm.qasm )

mtreinish:
[...]

PR #185, opened 2020-10-28
Add custom iterator class for BFS successors return

mtreinish:
This commit changes the return type of the bfs_successors function to be
a custom class BFSSuccessors. This new return class implements both the
sequence protocol and iterator protocol. This means that aside from
explicit type checking it should be backwards compatible with the list
being previously returned. It can be used with either index based access
or iterated over.

This should be more efficient for large graphs because instead of doing
the copy and type conversion and iterating over the entire nested Vec of
results it instead does it per access (either via `__getitem__` or
`__next__`). It does add a small amount of overhead for smaller graphs but
it is minimal since the function returns in microseconds in such cases
so a 10-20% overhead is not a big dea).

It's worth noting while this defers the type conversion, it does not
defer execution like most python iterators normally do. When
bfs_successors is called it will still always fully traverse the graph.
However, in practice the bottleneck for the bfs_successor function
wasn't actually the graph traversal, but instead the type conversion.

Related to #71

### DIFF
--- a/src/iterators.rs
+++ b/src/iterators.rs
@@ -1,0 +1,81 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License. You may obtain
+// a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations
+// under the License.
+
+use std::convert::TryInto;
+
+use pyo3::class::iter::{IterNextOutput, PyIterProtocol};
+use pyo3::class::PySequenceProtocol;
+use pyo3::exceptions::PyIndexError;
+use pyo3::prelude::*;
+
+/// A custom iterator class for the return from :func:`retworkx.bfs_successors`
+///
+/// This class is a container class for the results of the
+/// :func:`retworkx.bfs_successors` function. It implements both the Python
+/// iterator protocol and sequence protocol. So you can treat the return as
+/// either a read-only sequence/list that is integer indexed or use it as an
+/// iterator that will yield the results in order.
+///
+/// For example::
+///
+///     import retworkx
+///
+///     graph = retworkx.generators.directed_path_graph(5)
+///     bfs_succ = retworkx.bfs_successors(0)
+///     # Index based access
+///     third_element = bfs_succ[2]
+///     # Use as iterator
+///     first_element = next(bfs_succ)
+///     second_element = nex(bfs_succ)
+///
+#[pyclass(module = "retworkx")]
+pub struct BFSSuccessors {
+    pub bfs_successors: Vec<(PyObject, Vec<PyObject>)>,
+    pub index: usize,
+}
+
+#[pyproto]
+impl PySequenceProtocol for BFSSuccessors {
+    fn __len__(&self) -> PyResult<usize> {
+        Ok(self.bfs_successors.len())
+    }
+
+    fn __getitem__(
+        &'p self,
+        idx: isize,
+    ) -> PyResult<(PyObject, Vec<PyObject>)> {
+        if idx >= self.bfs_successors.len().try_into().unwrap() {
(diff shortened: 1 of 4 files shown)


tests/test_pred_succ.py
```diff
@@ -124,4 +124,68 @@ def test_bfs_succesors(self):
...
+        node_d = dag.add_child(node_c, 3, {})
+        node_e = dag.add_child(node_d, 4, {})
+        node_f = dag.add_child(node_e, 5, {})
+        dag.add_child(node_f, 6, {})
+        node_h = dag.add_child(node_c, 7, {})
+        node_i = dag.add_child(node_h, 8, {})
+        node_j = dag.add_child(node_i, 9, {})
+        dag.add_child(node_j, 10, {})
+        res = retworkx.bfs_successors(dag, node_b)
+        with self.assertRaises(IndexError):
```
kdk, line 171: I was expecting this to be a `TypeError` (for trying to index into an iterator) until I read the API docs for the `BFSSuccessors` return type.

src/iterators.rs
```diff
@@ -0,0 +1,81 @@
...
+impl PySequenceProtocol for BFSSuccessors {
+    fn __len__(&self) -> PyResult<usize> {
+        Ok(self.bfs_successors.len())
+    }
+
+    fn __getitem__(
+        &'p self,
+        idx: isize,
+    ) -> PyResult<(PyObject, Vec<PyObject>)> {
+        if idx >= self.bfs_successors.len().try_into().unwrap() {
```
kdk, line 56: Would it make sense to check that `idx` is non-negative here? (Is `isize` signed?)

kdk (approved):
This looks good. Out of curiosity, would it be more natural to implement either the sequence protocol (keeping the lazy copy and type conversion) or the iterator protocol? Implementing both seems better for backwards compatibility, but it seems like the differences in the two protocols might lead non-obvious behavior (see one comment below.)

mtreinish:
> This looks good. Out of curiosity, would it be more natural to implement either the sequence protocol (keeping the lazy copy and type conversion) or the iterator protocol? Implementing both seems better for backwards compatibility, but it seems like the differences in the two protocols might lead non-obvious behavior (see one comment below.)

That was my first thought on this PR too, and I started with just that. But IIRC the issue wass that just the sequence protocol doesn't actually work with `iter(BFSSuccessors)`. But I was trying a couple of different things at once, and I might be misremembering. Let me try dropping the iter protocol and see what happens.

src/iterators.rs
mtreinish: Oh, actually this is a good point. This will totally fall apart if you pass a negative number in. I think I need to add a bit more logic here to handle negatives, because I assume the expectation from users is if you were do `bfs_successors(graph)[-1]` you'd get the last entry, which definitely won't work like this.

Also yes [`isize`](https://doc.rust-lang.org/std/primitive.isize.html) is a pointer sized signed integer. (compared to [`usize`](https://doc.rust-lang.org/std/primitive.usize.html) which is an unsigned pointer sized integer

mtreinish:
Ok, yeah I took a look at this locally and yeah we can remove the iter protocol implementation and it will still work. You'll need to call `iter()` around the return to use it as an iterator (which is probably where I got confused before). With that it works fine. I'll make this change in a sec (just need to clean up the docs).

src/iterators.rs
mtreinish: So I actually tested this with `bfs_successors(graph)[-1]` and it works as expected (which I'm surprised about because indices for Vec are unsigned) so I think it's something python is doing for us. I added a test to assert negative indices work as expected. So I don't think we need to change anything here. What do you think?

kdk: :+1: Agree, the python behavior looks good. Thanks for adding the test.

src/lib.rs
```diff
@@ -429,31 +430,41 @@ fn graph_dfs_edges(
 /// :param PyDiGraph graph: The DAG to get the bfs_successors from
 /// :param int node: The index of the dag node to get the bfs successors for
 ///
-/// :returns: A list of nodes's data and their children in bfs order
-/// :rtype: list
+/// :returns: A list of nodes's data and their children in bfs order. The
+///     BFSSuccessors class that is returned is a custom container class that
+///     implements the iterator protocol and sequence protocol. This can
+///     be used as either a list with index based access or as an iterator.
```
kdk, line 436:
```suggestion
///     implements the sequence protocol. This can be used as a python list
///     with index based access.
```

PR #186, opened 2020-10-29
DNM: Test wheel builds with new cibuildwheel

mtreinish:
This commit just triggers a build of the release wheels to verify that
everythin still works with the new cibuildwheel version and python 3.9.


mtreinish:
Everything seems to work (except for #187 which this caught), the s390x is failing because of typical travis flakiness. We should be good to release since all the jobs seem to work still and build 3.9 wheels

PR #189, opened 2020-11-02
Added Bidirectional parameters for directed graph generators

### LINKED ISSUE #184: Add bidirectional flag to existing generator functions

The existing directed_* generator functions in https://github.com/Qiskit/retworkx/blob/master/src/generators.rs only have an option to add a single edge. But it could be useful to add edges in both directions. While this is essentially the same as a undirected version of the function there are use cases where you could want bidirectional edges added to a directed graph generatore. For example a bidirectional path graph: `a <-> b <-> c` instead of just `a -> b -> c` that exists now. This would be to add a new boolean kwarg to the existing (and potential future) directed generator functions to set adding bidirectional edges.

This is related to #150 but self contained enhancement for the existing generator functions rather than adding new functions.

darknight009:
Fixes #184 .
> Allows users to pass optional parameter 'bidirectional' to add edges in both directions
> For directed_star_graph generator if 'bidirectional' is set to 'True', the 'inward' parameter is ignored.

### DIFF
--- a/src/generators.rs
+++ b/src/generators.rs
@@ -70,20 +70,14 @@ where
 ///       os.remove(tmp_path)
 ///   image
 ///
-#[pyfunction]
+#[pyfunction(bidirectional = "false")]
 #[text_signature = "(/, num_nodes=None, weights=None, bidirectional=False)"]
 pub fn directed_cycle_graph(
     py: Python,
     num_nodes: Option<usize>,
     weights: Option<Vec<PyObject>>,
-    bidirectional: Option<bool>,
+    bidirectional: bool,
 ) -> PyResult<digraph::PyDiGraph> {
-    // boolean flag to indicate if edges should be added in both directions
-    let bidirection = match bidirectional {
-        Some(x) => bidirectional.unwrap(),
-        None => false,
-    };
-
     let mut graph = StableDiGraph::<PyObject, PyObject>::default();
     if weights.is_none() && num_nodes.is_none() {
         return Err(PyIndexError::new_err(
@@ -111,7 +105,7 @@ pub fn directed_cycle_graph(
     for (node_a, node_b) in pairwise(nodes) {
         match node_a {
             Some(node_a) => {
-                if bidirection {
+                if bidirectional {
                     graph.add_edge(node_b, node_a, py.None());
                 }
                 graph.add_edge(node_a, node_b, py.None());
@@ -122,7 +116,7 @@ pub fn directed_cycle_graph(
     let last_node_index = NodeIndex::new(node_len - 1);
     let first_node_index = NodeIndex::new(0);
     graph.add_edge(last_node_index, first_node_index, py.None());
-    if bidirection {
+    if bidirectional {
         graph.add_edge(first_node_index, last_node_index, py.None());
     }
     Ok(digraph::PyDiGraph {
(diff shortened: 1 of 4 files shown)


src/generators.rs
```diff
@@ -69,12 +71,19 @@ where
 ///   image
 ///
 #[pyfunction]
-#[text_signature = "(/, num_nodes=None, weights=None)"]
+#[text_signature = "(/, num_nodes=None, weights=None, bidirectional=False)"]
 pub fn directed_cycle_graph(
     py: Python,
     num_nodes: Option<usize>,
     weights: Option<Vec<PyObject>>,
+    bidirectional: Option<bool>,
```
mtreinish, line 79: This is a bit redundant since we're basically saying this can be None (which is treated as false) or true or false. I think when you couple this with my comment above above about the macro you can just set this to be a `bool`

src/generators.rs
```diff
@@ -69,12 +71,19 @@ where
 ///   image
 ///
 #[pyfunction]
```
mtreinish, line 73: I think you should add the kwarg to the definition here so the Python C API FFI function this macro generates sets a default. This will avoid the need to use an `Option<bool>` below. Something like:

```suggestion
#[pyfunction(bidirectional = "false")]
```

This basically makes it so Python callers get an implicit default of `false`.

mtreinish (commented):
Thanks for pushing this up so quickly! The logic here looks good, just one comment on the parameter handling. We can set a default for the python api using the `#[pyfunction]` macro which is a bit cleaner than wrapping the bool with an option. Once that change is made I think this should be good to go.

src/generators.rs
```diff
@@ -437,10 +462,16 @@ pub fn directed_star_graph(
...
+        if bidirectional {
             graph.add_edge(*node, nodes[0], py.None());
-        } else {
             graph.add_edge(nodes[0], *node, py.None());
+        } else {
+            if inward {
+                graph.add_edge(*node, nodes[0], py.None());
+            } else {
+                graph.add_edge(nodes[0], *node, py.None());
+            }
```
mtreinish, line 474: I think that this if can be collapsed:

```suggestion
        } else if inward {
                graph.add_edge(*node, nodes[0], py.None());
        } else {
                graph.add_edge(nodes[0], *node, py.None());
        }
```

src/generators.rs
```diff
@@ -409,12 +433,13 @@ pub fn path_graph(
...
-#[pyfunction(inward = "false")]
-#[text_signature = "(/, num_nodes=None, weights=None, inward=False)"]
+#[pyfunction(inward = "false", bidirectional = "false")]
+#[text_signature = "(/, num_nodes=None, weights=None, inward=False, bidirectional=False)"]
 pub fn directed_star_graph(
     py: Python,
     num_nodes: Option<usize>,
     weights: Option<Vec<PyObject>>,
+    bidirectional: bool,
     inward: bool,
```
mtreinish, line 443: Could you flip these? I just think it'll make theupgrade path a bit simpler for any potential rust callers of this function.

```suggestion
    inward: bool,
    bidirectional: bool,
```

mtreinish (approved):
LGTM, thanks for making the changes so quickly. I highlighted 2 small nits inline. After those are updated I'll go ahead and merge this.

darknight009:
@mtreinish Can you help? It fails with the message: No module named 'setuptools_rust'

mtreinish:
> @mtreinish Can you help? It fails with the message: No module named 'setuptools_rust'

Unfortunately it was poor timing, setuptools-rust pushed a new release this morning that dropped python 3.5 support but didn't raise the minimum python version. I opened an issue on this: https://github.com/PyO3/setuptools-rust/issues/85 and in the meantime I'll push up a PR to try and set a version cap for python 3.5.

Issue #150, opened 2020-09-30
Expand Generators Module


mtreinish:
### What is the expected enhancement?

In #121 we added a new module for graph generators which will create graphs of with a certain layout easily. #121 only added a few different types of graphs with the intention of adding more in the future. We should add more generators to the graph to offer more options for quickly creating graphs. Networkx has a long list of generators that can be used for inspiration on different types of generators to add:

https://networkx.github.io/documentation/stable/reference/generators.html

Also, specifically for a qiskit use case we should make sure that we have generators to cover `qiskit.transpiler.CouplingMap.from_line`, `qiskit.transpiler.CouplingMap.from_ring`, `qiskit.transpiler.CouplingMap.from_grid`, and `qiskit.transpiler.CouplingMap.from_full`

mtreinish:
Some generators that would be useful to do as part of this would be to add generators for some layouts that are used for error correction codes like heavy hex and heavy square. (see: https://journals.aps.org/prx/abstract/10.1103/PhysRevX.10.011022 ).

nahumsa:
I will work on the heavy hex generator.

mtreinish:
There has been a lot of good work here since this was initially opened and we've added a bunch of new generators. But, when you compare the number of generators in networkx: https://networkx.org/documentation/stable/reference/generators.html to list in networkx: https://qiskit.org/documentation/retworkx/api.html#generators there is still a large functionality gap. 

I think before we can close this issue we need to close this gap much more. Each generator function is a really good starting PR that's self contained and a good project for getting your feet wet with the library.

e-eight:
Added a generator for undirected lollipop graphs in PR #454. Not sure if there should also be a directed version. Networkx does not seem to have any.

k-tamuraphys:
I am working on `lattice` module, https://github.com/Qiskit/qiskit-nature/pull/365 , in which retworkx is used. I hope I can generate translational invariant graphs (lattice) with uniform weights on edges using retworkx. Standard examples of such are hypercubic (hyperrectangle, triangular, hexagonal lattices.
Furthermore, is such a graph has edges several directions like a square (or grid graph) and a triangular lattice , it is desirable to be able to assign uniform weights to each direction. For example, I may want to generate a square lattice with a weight of 1.0 in the horizontal direction and a weight of 2.0 in the vertical direction.

PR #191, opened 2020-11-08
Added Mesh and Grid graph generator

darknight009:
#150 Adds both directed and undirected versions of Mesh and Grid graph generators. 
This PR doesn't add test cases for Grid generator as it is unclear how the weights will be passed for the grid. As a multidimensional array or a flattened array?
@mtreinish Please let me know how to proceed with that

### DIFF
--- a/src/generators.rs
+++ b/src/generators.rs
@@ -655,17 +655,24 @@ pub fn directed_mesh_graph(
     })
 }
 
-/// Generate an undirected grid graph
+/// Generate an undirected grid graph.
 ///
-/// :param int rows: The number of rows to generate the graph with. Value
-///     required
-/// :param list cols: The number of rows to generate the graph with. Value
-///     required
-/// :param list weights: A list of node weights.
+/// :param int rows: The number of rows to generate the graph with.
+///     If specified, cols also need to be specified
+/// :param list cols: The number of rows to generate the graph with.
+///     If specified, rows also need to be specified. rows*cols 
+///     defines the number of nodes in the graph
+/// :param list weights: A list of node weights. If rows and cols 
+///     are not specified, then a linear graph containing all the 
+///     values in weights list is created. 
+///     If number of nodes(rows*cols) is less than length of
+///     weights list, the trailing weights are ignored.
+///     If number of nodes(rows*cols) is greater than length of
+///     weights list, extra nodes with None weight are appended
 ///
 /// :returns: The generated star graph
 /// :rtype: PyGraph
-/// :raises IndexError: If neither ``rows`` and ``cols`` or ``weights`` are
+/// :raises IndexError: If neither ``rows`` or ``cols`` and ``weights`` are
 ///      specified
 ///
 #[pyfunction]
@@ -683,14 +690,39 @@ pub fn grid_graph(
         ));
     }
 
-    let rowlen = rows.unwrap();
-    let collen = cols.unwrap();
-    let num_nodes = rowlen * collen;
+    let mut rowlen = match rows {
+        Some(rows) => rows,
+        None => 0,
+    };
+
+    let mut collen = match cols {
+        Some(cols) => cols,
+        None => 0,
+    };
+
+    let mut num_nodes = rowlen * collen;
+
     let nodes: Vec<NodeIndex> = match weights {
         Some(weights) => {
             let mut node_list: Vec<NodeIndex> = Vec::new();
+            if num_nodes < weights.len() && rowlen == 0 {
+                collen = weights.len();
+                rowlen = 1;
+                num_nodes = collen;
+            }
+
+            let mut node_cnt = num_nodes;
+            
             for weight in weights {
+                if node_cnt <= 0 {
+                    break;
+                }
                 let index = graph.add_node(weight);
+                node_list.push(index);
+                node_cnt -= 1;
+            }
+            for _i in 0..node_cnt {
+                let index = graph.add_node(py.None());
                 node_list.push(index);
             }
             node_list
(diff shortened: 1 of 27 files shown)


mtreinish:
@darknight009 thanks for pushing this. Hmm, I hadn't thought about that, but I'd probably just go with a flattened array. It's consistent with the other generator methods and easier to deal with. We just need to make sure it's documented the order it's used, row major or column major.

src/generators.rs
```diff
@@ -551,6 +551,329 @@ pub fn star_graph(
...
+    Ok(graph::PyGraph {
+        graph,
+        node_removed: false,
+    })
+}
+
+/// Generate a directed grid graph. The edges propagate towards right and
+///     bottom direction if ``bidirectional`` is ``false``
+///
+//// :param int rows: The number of rows to generate the graph with.
```
mtreinish, line 760: This should fix the failing docs build I think:

```suggestion
/// :param int rows: The number of rows to generate the graph with.
```

src/generators.rs
```diff
@@ -551,6 +551,329 @@ pub fn star_graph(
...
+///     defines the number of nodes in the graph
+/// :param list weights: A list of node weights. If rows and cols
+///     are not specified, then a linear graph containing all the
+///     values in weights list is created.
+///     If number of nodes(rows*cols) is less than length of
+///     weights list, the trailing weights are ignored.
+///     If number of nodes(rows*cols) is greater than length of
+///     weights list, extra nodes with None weight are appended
+///
+/// :returns: The generated star graph
```
mtreinish, line 673:
```suggestion
/// :returns: The generated grid graph
```

src/generators.rs
```diff
@@ -551,6 +551,329 @@ pub fn star_graph(
...
+///     are not specified, then a linear graph containing all the
+///     values in weights list is created.
+///     If number of nodes(rows*cols) is less than length of
+///     weights list, the trailing weights are ignored.
+///     If number of nodes(rows*cols) is greater than length of
+///     weights list, extra nodes with None weight are appended.
+/// :param bidirectional: A parameter to indicate if edges should exist in
+///     both directions between nodes
+///
+/// :returns: The generated star graph
```
mtreinish, line 775:
```suggestion
/// :returns: The generated grid graph
```

src/generators.rs
```diff
@@ -551,6 +551,329 @@ pub fn star_graph(
...
+        Some(rows) => rows,
+        None => 0,
+    };
+
+    let mut collen = match cols {
+        Some(cols) => cols,
+        None => 0,
+    };
+
+    let mut num_nodes = rowlen * collen;
```
mtreinish, line 806: What happens if num_nodes is 0? Also what happens if weights.len() != num_nodes? I think we should probably add error handling to cover the case where someone calls: `grid_graph(0, 0)` or `grid_graph(100, 100, ['a', 'b'])`

mtreinish (commented):
Thanks for pushing the update. From a quick pass through the code looks good, I left a few initial comments inline mostly around docs cleanups and some edge case handling. But, I haven't dug through each line in depth yet. Besides the docs failure (which I outlined how to fix inline I think). Can you also add an example with the graphviz visualization to the docs? It's kind of nice for the generator documentation to visualize the graph so people can see them in the documentation. Basically something like: https://github.com/Qiskit/retworkx/blob/master/src/generators.rs#L50-L72

src/generators.rs
darknight009: I have handled the case grid_graph(100, 100, ['a', 'b']) by appending Py.None() weights. So in this case 100*100 - 2 nodes are added into the graph. For grid_graph(0, 0) an empty graph is returned. Do we need to error out both the cases?

mtreinish: Nah, I think it's fine like this as long as we document this behavior and assert it with tests.

darknight009: I had added the documentation and tests. Let me know if they need to be updated

src/generators.rs
```diff
@@ -551,6 +551,421 @@ pub fn star_graph(
...
+            if num_nodes < weights.len() && rowlen == 0 {
+                collen = weights.len();
+                rowlen = 1;
+                num_nodes = collen;
+            }
+
+            let mut node_cnt = num_nodes;
+
+            for weight in weights {
+                if node_cnt <= 0 {
```
mtreinish, line 786: I ran `cargo clippy` on the PR locally and it flagged this. Since `node_cnt` is a usize it can't ever be less than 0, so they suggested replacing this with:

```suggestion
                if node_cnt == 0 {
```

src/generators.rs
```diff
@@ -551,6 +551,421 @@ pub fn star_graph(
...
```
mtreinish, line 912: I ran `cargo clippy` on the PR locally and it flagged this. Since `node_cnt` is a usize it can't ever be less than 0, so they suggested replacing this with:

```suggestion
                if node_cnt == 0 {
```

src/generators.rs
mtreinish: Thanks, yeah it looks fine now.

mtreinish (approved):
This LGTM, there are 2 inline suggestions that were basically just errors that cargo clippy reported. I should probably add a CI job with it to catch things like this in the future. But after those are fixed this should be ready to go. Thanks for updating the docs.

mtreinish (approved):
Great, thanks for the update

mtreinish:
@darknight009 just an fyi thanks to your work on this PR we were able to speed up qiskit-terra's coupling map construction in: https://github.com/Qiskit/qiskit-terra/pull/5469

PR #199, opened 2020-11-16
Implement rich comparison for BFSSuccessors

mtreinish:
This commit implements the rich comparison method of the python object
protocol for the BFSSuccessors class. This enables doing an equals and
not equals comparison between a BFSSuccessors object and any other
python sequence. If will compare all the elements of the 2 and return
appropriate result. Previously this was not implemented so you would
have to cast a BFSSuccessors object as a list first to compare it which
would iterate over the object. Implementing the method avoids the need
to do this.

### DIFF
--- a/src/iterators.rs
+++ b/src/iterators.rs
@@ -12,9 +12,10 @@
 
 use std::convert::TryInto;
 
-use pyo3::class::PySequenceProtocol;
-use pyo3::exceptions::PyIndexError;
+use pyo3::class::{PyObjectProtocol, PySequenceProtocol};
+use pyo3::exceptions::{PyIndexError, PyNotImplementedError};
 use pyo3::prelude::*;
+use pyo3::types::PySequence;
 
 /// A custom class for the return from :func:`retworkx.bfs_successors`
 ///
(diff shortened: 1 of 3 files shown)


src/iterators.rs
```diff
@@ -44,6 +45,52 @@ pub struct BFSSuccessors {
...
+                    other_raw.extract()?;
+                if self.bfs_successors[i].0.as_ref(py).compare(other_value.0)?
+                    != std::cmp::Ordering::Equal
+                {
+                    return Ok(false);
+                }
+                for (index, obj) in self.bfs_successors[i].1.iter().enumerate()
+                {
+                    if obj.as_ref(py).compare(&other_value.1[index])?
+                        != std::cmp::Ordering::Equal
```
lcapelluto, line 73: this is funny `if this is not equal to equal` (in a brain teasing way; I think it's good)

lcapelluto (approved):

src/iterators.rs
mtreinish: Yeah, the api around comparisons is weird. It's because they're trying to map the python c api to rust's ordering https://docs.rs/pyo3/0.12.3/pyo3/struct.PyAny.html#method.compare it's unfortunately the easiest way to do 'a == b' for python objects. it would be nice if pyo3 had a `eq` method or something to do this and return a bool

lcapelluto: yeah, a bit roundabout.

PR #200, opened 2020-11-16
[WIP] Add profile guided optimization to wheels

mtreinish:
This commit adds scripts to build profile data for profile guided
optimization to the retworkx repo and then updates the release jobs to
use the script and the subsequently generated profile data into the
release binaries in the built wheels. The scripts are based on the
official rust docs for using PGO. [1]

The profile data is built using qiskit-terra's unit tests, a dedicated
100 x 20 quantum volume transpile run, transpiling a randomized
benchmarking experiment, and running qiskit-aqua's abelian grouper.

[1] https://doc.rust-lang.org/rustc/profile-guided-optimization.html


mtreinish:
Before we go to merge this we should need to test that the cibuildwheel jobs work with PGO enabled. The other thing missing here is benchmarking to see if this is worth the effort or not. In local testing I haven't really seen any performance impact when running a QV transpile (the only thing I've tested so far). We'll need to do some real benchmarking to determine what the tradeoffs are with PGO and whether it's worth it.

mtreinish:
In local testing, I'm not really seeing any improvement for binaries built with PGO. I'm not sure if this is going to be worth the extra effort. I'll keep playing with this eventually to see if it's just an instance of not having enough (or good enough) profile data or something else that can be tuned.

mtreinish:
I've been doing some testing and when building with PGO enabled it's about 2-5% faster than without PGO enabled. For release builds I think the one time cost in extra build time and complexity are worth it, even if the gains are modest. I'm gonna leave this as WIP because I need to push a DNM commit to the branch to run cibuildwheel and verify that the PGO builds work in the release jobs.

mtreinish:
I'm going to close this, I've only been able to get it to build on macOS (using the clang tools), the complexity of the build and ensuring all the llvm tools are present and working is more effort than this is worth right now. I might reopen it in the future, but for a small 2-5% performance increase it doesn't seem worth the effort if it's only on a handful of release binaries.

PR #204, opened 2020-11-19
Add EdgeList and WeightedEdgeList return types

mtreinish:
This commit adds 2 new return types `EdgeList` and `WeightedEdgeList`
that are used as the return type for functions that return
`Vec<(usize, usize)>` and `Vec<(usize, usize, PyObject)>` respectively.
This custom type implements the Python sequence protocol and can be
used as the list which was previously returned except for where
explicit type checking was used. Just as in #185 and #198 the
advantage here is that for large edge lists the conversion from
Rust to a Python type becomes a large bottleneck. This avoids that
conversion and only does it per element on access.

### DIFF
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -41,7 +41,7 @@ use petgraph::visit::{
 };
 
 use super::dot_utils::build_dot;
-use super::iterators::NodeIndices;
+use super::iterators::{EdgeList, NodeIndices, WeightedEdgeList};
 use super::{
     is_directed_acyclic_graph, DAGHasCycle, DAGWouldCycle, NoEdgeBetweenNodes,
     NoSuitableNeighbors, NodesRemoved,
@@ -661,11 +661,14 @@ impl PyDiGraph {
     /// ``source`` and ``target`` are the node indices.
     ///
     /// :returns: An edge list with weights
-    /// :rtype: list
-    pub fn edge_list(&self) -> Vec<(usize, usize)> {
-        self.edge_references()
-            .map(|edge| (edge.source().index(), edge.target().index()))
-            .collect()
+    /// :rtype: EdgeList
+    pub fn edge_list(&self) -> EdgeList {
+        EdgeList {
+            edges: self
+                .edge_references()
+                .map(|edge| (edge.source().index(), edge.target().index()))
+                .collect(),
+        }
     }
 
     /// Get edge list with weights
@@ -675,20 +678,20 @@ impl PyDiGraph {
     /// payload of the edge.
     ///
     /// :returns: An edge list with weights
-    /// :rtype: list
-    pub fn weighted_edge_list(
-        &self,
-        py: Python,
-    ) -> Vec<(usize, usize, PyObject)> {
-        self.edge_references()
-            .map(|edge| {
-                (
-                    edge.source().index(),
-                    edge.target().index(),
-                    edge.weight().clone_ref(py),
-                )
-            })
-            .collect()
+    /// :rtype: WeightedEdgeList
+    pub fn weighted_edge_list(&self, py: Python) -> WeightedEdgeList {
+        WeightedEdgeList {
+            edges: self
+                .edge_references()
+                .map(|edge| {
+                    (
+                        edge.source().index(),
+                        edge.target().index(),
+                        edge.weight().clone_ref(py),
+                    )
+                })
+                .collect(),
+        }
     }
 
     /// Remove a node from the graph.
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -1163,21 +1163,22 @@ impl PyDiGraph {
 
             let mut edges_to_add: Vec<(usize, usize, PyObject)> = Vec::new();
             for dir in &DIRECTIONS {
-                let edges;
-                if dir == &petgraph::Direction::Outgoing {
-                    edges = self.out_edges(u);
-                } else {
-                    edges = self.in_edges(u);
-                }
-
-                for edge in edges {
+                for edge in self.graph.edges_directed(NodeIndex::new(u), *dir) {
                     let s = edge.source();
                     let d = edge.target();
 
-                    if s == u {
-                        edges_to_add.push((v, d, edge.weight().clone()));
+                    if s.index() == u {
+                        edges_to_add.push((
+                            v,
+                            d.index(),
+                            edge.weight().clone_ref(py),
+                        ));
                     } else {
-                        edges_to_add.push((s, v, edge.weight().clone()));
+                        edges_to_add.push((
+                            s.index(),
+                            v,
+                            edge.weight().clone_ref(py),
+                        ));
                     }
                 }
             }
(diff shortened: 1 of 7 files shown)


src/union.rs
```diff
@@ -62,7 +62,7 @@ pub fn digraph_union(
         node_map.insert(node.index(), node_index);
     }
 
-    for edge in b.weighted_edge_list(py) {
+    for edge in b.weighted_edge_list(py).edges {
```
lcapelluto, line 65: Why is this the only place where `.edges` is needed? Is this the only place where `weighted_edge_list()` was used?

src/iterators.rs
```diff
@@ -183,3 +183,160 @@ impl PySequenceProtocol for NodeIndices {
         }
     }
 }
+
+/// A custom class for the return of edge lists
+///
+/// This class is a container class for the results of functions that
+/// return a list of node indices. It implements the Python sequence
```
lcapelluto, line 190:
```suggestion
/// return a list of edges. It implements the Python sequence
```

src/iterators.rs
```diff
@@ -183,3 +183,160 @@ impl PySequenceProtocol for NodeIndices {
...
+/// that is integer indexed. If you want to use it as an iterator you
+/// can by wrapping it in an ``iter()`` that will yield the results in
+/// order.
+///
+/// For example::
+///
+///     import retworkx
+///
+///     graph = retworkx.generators.directed_path_graph(5)
+///     edges = retworkx.edge_list()
```
lcapelluto, line 201: should this be
```suggestion
///     edges = graph.edge_list()
```

src/iterators.rs
```diff
@@ -183,3 +183,160 @@ impl PySequenceProtocol for NodeIndices {
...
```
lcapelluto, line 276: Here, too?

src/iterators.rs
```diff
@@ -183,3 +183,160 @@ impl PySequenceProtocol for NodeIndices {
...
+        } else {
+            Ok(self.edges[idx as usize])
+        }
+    }
+}
+
+/// A custom class for the return of edge lists
+///
+/// This class is a container class for the results of functions that
+/// return a list of node indices. It implements the Python sequence
```
lcapelluto, line 265:
```suggestion
/// return a list of edges. It implements the Python sequence
```

lcapelluto (changes requested):

src/union.rs
mtreinish: Yeah, this is the only place `weighted_edge_list()` is called from inside rust. Normally it's only used from python but this function uses it instead of calling `b.graph..edge_references()`. We probably should update this to just use the petgraph api directly, but it's not a big deal and an easy followup thing we can change later.

itoko:
LGTM. Just out of curiosity, I found `pub fn in_edges(&self, node: usize) -> Vec<(usize, usize, &PyObject)>`, which is not touched in this PR. Although I think it's ok because the type `&PyObject` is different from `PyObject` (weight type in EdgeWeightList), where does this difference in return types come from?

mtreinish:
> LGTM. Just out of curiosity, I found `pub fn in_edges(&self, node: usize) -> Vec<(usize, usize, &PyObject)>`, which is not touched in this PR. Although I think it's ok because the type `&PyObject` is different from `PyObject` (weight type in EdgeWeightList), where does this difference in weight types come from?

Oh I missed that, I just searched for the exact matches on the return and did a find replace on them. The difference is really just between where reference counting increments happen. I'll update that when I fix the remaining docstring issues @lcapelluto pointed out.

That function retuning `&PyObject` (it originally comes from the petgraph method since it's returning a reference to the weight object in the graph not moving it out from the graph struct) means for any rust code callers they get a borrowed `PyObject` and have to managed the lifetime of that borrow. But for python it doesn't matter because pyo3's `#[pymethods]` macro generates a wrapper for us to handle the borrow and return the pyobject properly to python. The `WeightedEdgeList` struct can only have a `Vec` with owned `PyObject` it's a current limitation with pyo3. The change here is simple I think, we should just need to call `clone()`/`clone_ref()` on the weight to increment the python refcount when creating the `Vec`.

src/lib.rs
```diff
@@ -2437,7 +2439,7 @@ pub fn strongly_connected_components(
 pub fn digraph_find_cycle(
```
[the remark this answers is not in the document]
mtreinish: Fixed in: https://github.com/Qiskit/retworkx/pull/204/commits/a8d71c9968fbc1749fd4569004c4b2d40e31c684

src/iterators.rs
mtreinish: Fixed in: https://github.com/Qiskit/retworkx/pull/204/commits/a8d71c9968fbc1749fd4569004c4b2d40e31c684

mtreinish:
I've updated `in_edges` and `out_edges` to use the new return type in: https://github.com/Qiskit/retworkx/pull/204/commits/407bd594154883351a3a776a262239b86dce23e4 there was a small complication with `merge_nodes` which was using those functions directly. I had to update that to use the petgraph api directly because of the rust side changes to those methods.

src/digraph.rs
```diff
@@ -1163,21 +1163,22 @@ impl PyDiGraph {
...
             for dir in &DIRECTIONS {
-                let edges;
-                if dir == &petgraph::Direction::Outgoing {
-                    edges = self.out_edges(u);
-                } else {
-                    edges = self.in_edges(u);
-                }
-
-                for edge in edges {
+                for edge in self.graph.edges_directed(NodeIndex::new(u), *dir) {
```
lcapelluto, line 1166: 👍

src/digraph.rs
```diff
@@ -1163,21 +1163,22 @@ impl PyDiGraph {
...
                     let s = edge.source();
                     let d = edge.target();
 
-                    if s == u {
-                        edges_to_add.push((v, d, edge.weight().clone()));
+                    if s.index() == u {
+                        edges_to_add.push((
+                            v,
+                            d.index(),
+                            edge.weight().clone_ref(py),
```
lcapelluto, line 1174: Why `clone_ref`? Also, why does `index()` have to be added here but not before?

lcapelluto (approved):

src/digraph.rs
mtreinish: This was working before because it was using `clone()` instead of `clone_ref()` Before this PR `in_edges()` and `out_edges()` were returning a `Vec(usize, usize, &PyObject)` and `edges_to_add` was a `Vec(usize, usize, &PyObject)` (it's just being used as a temp storage Vec before its used later to add edges). So prior to the changing `in_edges()` and `out_edges()` this worked by the 2 usizes implement copy so can efficiently be used in place without moving, but to store `&PyObject` as a `PyObject` (which is neededt for creating a new edge in the graph) it needs to take ownership of the object being borrowed. We can't do `*edge.weight()` since it would move the weight out of the graph (and would fail) so we need to clone/copy it which for PyObject isn't a big deal because it just increments the ref counter.

The change I made here (besides switching to petgraph method `edges_directed` instead of using in_edges and out_edges) is to switch just switch from `clone()` to `clone_ref(py)`. There isn't a functional difference between them as they both increment the python reference counter for the object, except that you give a gil handle to clone_ref if you have it: https://github.com/PyO3/pyo3/blob/v0.12.4/src/instance.rs#L196-L200 while `clone()` will queue it for when there is one it doesn't already have it: https://github.com/PyO3/pyo3/blob/v0.12.4/src/instance.rs#L517-L524. In the context of this method I don't think there is any performance difference since it has a gil reference, I just prefer it to be explicit and use clone_ref when we already have py

mtreinish: Oh I missed the second half of the question. The `index()` calls are needed because the `edges_to_add` vec is defined as `Vec<(usize, usize, PyObject)>` and `self.graph.edges_directed()` returns an iterator yielding petgraph `EdgeReference` objects. Those objects return `NodeIndex` objects for `source()` and `target()` instead of usize ints. So to get the usize from the `NodeIndex` I had to call index() on it which returns a usize (since NodeIndex is just a wrapper around a usize anyay). This wasn't necessary before because in_edges() and out_edges returned a `Vec<usize, usize, &PyObject)>` and had already done the conversion from NodeIndex -> usize for us.

lcapelluto: got it!:)

PR #208, opened 2020-11-29
Switch coverage job to use source based coverage

mtreinish:
This commit adjust the coverage job to use source based coverage [1][2]
instead of gcov. This should result in more accurate coverage reporting
since it's done via rust instead of llvm.

[1] https://blog.rust-lang.org/inside-rust/2020/11/12/source-based-code-coverage.html
[2] https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/source-based-code-coverage.html


mtreinish:
Now that this finally working the source based coverage method is definitely what I was hoping for. Looking at the results that coveralls just posted the coverage reports make a lot more sense now. There aren't weird artifacts like lines in the middle of a larger block being shown as uncovered or comments being shown as covered. So I'm going to merge this as how we collect coverage moving forward.

PR #212, opened 2020-12-03
Fix wheel builds

mtreinish:
This commit fixes the wheel builds. The jobs hadn't been updated in a
while and have become stale. The github actions x86 (linux, mac, and
windows) were still trying to build 35 wheel even though python 3.5
support was been dropped in #196. Then the docker image used for the
manylinux2010 build needed to be updated because the pinned version in
cibuildwheel is failing yum because of the centos 6 eol, which has been
fixed on the latest docker image.


mtreinish:
I've run these changes in ci: https://github.com/Qiskit/retworkx/pull/211 (which is this commit plus a dnm change enabling release jobs everywhere and disabling upload) and it fixes everything.

PR #213, opened 2020-12-03
Add pickle support to custom return types

mtreinish:
In recent releases we have started to add custom return types instead of
returning lists of of objects to avoid the overhead of converting large
lists from rust to python. However, these custom return types were
missing support for pickling, which was causing issue for places where
they were being used with Python's multiprocessing or deepcopy. This has
caused issues in places where a list was returned before. This commit
fixes this by adding the necessary methods to the custom return classes
to enable python to pickle them so they can avoid these issues.


mtreinish:
Once this PR merges I'm planning to push out a 0.7.1 release to fix this. I bumped the version strings as part of this PR to save time.

lcapelluto (approved):

PR #214, opened 2020-12-03
Update documentation and add release notes

mtreinish:
This commit updates the documentation theme to match the current theme
that the qiskit project is using. It also takes the opportunity to add
the release notes to the documentation. We previously were only
publishing the release documentation on the github release page which
has limitations. Moving forward we'll use reno to automate the
generation of release notes.

TODO:

- [x] Finish release notes conversion to sphinx rst


CONTRIBUTING.md
```diff
@@ -1,10 +1,14 @@
 # Contributing
 
-First read the overall project contributing guidelines. These are all
-included in the qiskit documentation:
+First read the overall Qiskit project contributing guidelines. These are all
```
elfrocampeador, line 3:
```suggestion
First read the overall Qiskit project contribution guidelines. These are all
```

CONTRIBUTING.md
```diff
@@ -28,6 +32,48 @@ the root of the repo, this is because retworkx packaging shim will conflict
...
+```
+
+locally. This will automatically restyle the rust code in retworkx to match
+what CI is checking.
+
+##### Lint
+
+An additional step is to run [clippy](https://github.com/rust-lang/rust-clippy)
+on your changes. While this is not run in CI (because it's very dependent on
+the rust/cargo version) it often catches issues in your code. You can run it
```
elfrocampeador, line 56:
```suggestion
the rust/cargo version) it can often catch issues in your code. You can run it
```

CONTRIBUTING.md
```diff
@@ -38,3 +84,131 @@ tox -edocs
...
+we document that when we release a new version of retworkx. The expectation
+is that if your code contribution has user facing changes that you will write
+the release documentation for these changes. This documentation must explain
+what was changed, why it was changed, and how users can either use or adapt
+to the change. The idea behind release documentation is that when a naive
+user with limited internal knowledge of the project is upgrading from the
+previous release to the new one, they should be able to read the release notes,
+understand if they need to update their program which uses retworkx, and how they
+would go about doing that. It ideally should explain why they need to make
+this change too, to provide the necessary context.
```
elfrocampeador, line 100:
```suggestion
It is important to document any end user facing changes when we release a new 
version of retworkx.  The expectation is that if your code contribution has 
user facing changes that you will write the release documentation for these 
changes. This documentation must explain what was changed, why it was changed, 
and how users can either use or adapt to the change. The idea behind release 
documentation is that when a naive user with limited internal knowledge of the 
project is upgrading from the previous release to the new one, they should be  
able to read the release notes, understand if they need to update their 
program which uses retworkx, and how they would go about doing that. It 
ideally should explain why they need to make this change too, to provide the 
necessary context.
```

CONTRIBUTING.md
```diff
@@ -38,3 +84,131 @@ tox -edocs
...
+    reno report
+
+but you can also use the ``--version`` argument to view a single release (after
+it has been tagged::
+
+    reno report --version 0.8.0
+
+#### Building release notes locally
+
+Building The release notes are part of the standard retworkx documentation
```
elfrocampeador, line 210:
```suggestion
Building the release notes is part of the standard retworkx documentation
```

elfrocampeador (changes requested):
General copy-editing, as usual.

PR #217, opened 2020-12-10
Switch is_directed_acyclic_graph to use toposort internally

mtreinish:
Previously the retworkx.is_directed_acyclic_graph() function was
internally using the petgraph::algo::is_cyclic_directed function to look
for a potential cycle in the graph. However, as noted in the docs[1]
this function is recursive and as reported in Qiskit/qiskit-terra#5502
with very large graphs this is causing segfaults. To avoid this
potential issue this commit changes the internal implementation to use
petgraph::algo::toposort which isn't recursive and should be more
reliable.

Fixes: Qiskit/qiskit-terra#5502

[1] https://docs.rs/petgraph/0.5.1/petgraph/algo/fn.is_cyclic_directed.html

### DIFF
--- a/releasenotes/notes/fix-segfault-is_dag_fn-8335e900ffca6af9.yaml
+++ b/releasenotes/notes/fix-segfault-is_dag_fn-8335e900ffca6af9.yaml
@@ -1,0 +1,8 @@
+---
+fixes:
+  - |
+    Fixed a potential segfault that could occur when calling
+    :func:`~retworkx.is_directed_acyclic_graph` with a a very deep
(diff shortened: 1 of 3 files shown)


mtreinish:
I ran some benchmarks on this locally, and using `petgraph::algo::toposort` ends up being roughly 30% slower for the cases they both worked, but does not segfault for large graphs. While using the `petgraph::algo::is_cyclic_directed` is segfaulting at >=100000 nodes in a graph. So it's a necessary performance hit and one most people probably won't see because because for a 1000 node graph it still only takes 21μs.

kdk (approved):

PR #218, opened 2020-12-11
Fix the read the docs builds

mtreinish:
In #214 we updated the documentation configuration to use the new
qiskit sphinx theme to be consistent with qiskit's documentation and
also add the use of reno for release notes. However, that PR did not
update the conda environment file used by read the docs for buidling the
documetation to include the new requiremetns. This commit corrects that
oversight and adds the missing dependencies to the docs builds.

### DIFF
--- a/.conda-rtd-environment.yml
+++ b/.conda-rtd-environment.yml
@@ -11,3 +11,5 @@ dependencies:
     - jupyter-sphinx==0.2.3
     - pydot
     - pillow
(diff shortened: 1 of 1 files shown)


mtreinish:
This is still failing read the docs builds because reno requires the full git history to compile the release notes. I've submitted a support request to get the flag set to do full clones hopefully this will be quick and we can unblock this and get out a quick 0.7.2 release with #217 to fix the potential segfault issue.

Issue #215, opened 2020-12-05
Make API uniform


eddieschoute:
### What is the expected enhancement?
Currently, many methods require prefixing the name with the type of the graph (i.e. `digraph_`). Would it be possible to dispatch the right method based on the type of the graph automatically? This arose in a discussion on a prior PR (https://github.com/Qiskit/qiskit-terra/pull/5471#discussion_r536334895).

One pythonic way to implement this would be to switch on the type, if implementing this in Rust is a problem. Then the right Rust method may be called by the Python code.

mtreinish:
I think that using python would be the easier path forward here. I've tried to do this on the rust side several ways and the issue I have is that python binding macros `#[pyfunction]` which generate the functions which wraps the retworkx functions and actually passes the objects from python aren't able to tell us what traits the python object it passes to the rust definitions has implemented, just that it implements `FromPyObject` so we get a compiler type error (that's pretty cryptic because it originates in the expanded macro generated function). It'll be easier to just do something like:

```python3
def dfs_edges(self, graph, source=None):
    """Get edge list in depth first order

    :param (PyGraph|PyDiGraph|PyDAG) graph: The graph to get the DFS edge list from
    :param int source: An optional node index to use as the starting node
        for the depth-first search. The edge list will only return edges in
        the components reachable from this index. If this is not specified
        then a source will be chosen arbitrarly and repeated until all
        components of the graph are searched.
    :returns: A list of edges as a tuple of the form ``(source, target)`` in
        depth-first order
    :rtype: EdgeList
    """
    if isinstance(graph, PyGraph):
       graph_dfs_edges(graph, source)
    else:
       digraph_dfs_edges(graph, source)
```

in the main package `__init__`:  https://github.com/Qiskit/retworkx/blob/master/retworkx/__init__.py
and just do that for every split function in the api. Although, given the number of these there will be though (since it's most of the algorithm functions, all the generators, and all the random graph functions) it might make more sense to create a new `ufunc.py` or something like that and do this in there and then just re-export all of them from the package root.

PR #220, opened 2020-12-13
Implementation of betweenness centrality

### LINKED ISSUE #195: Add betweenness functions

Add function(s) to retworkx's algorithm library for calculating the betweenness centrality for nodes and edges. This is similar to networkx's `betweenness_centrality` https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.betweenness_centrality.html and `edge_betweenness_centrality` https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.edge_betweenness_centrality.html
or graph-tool's `betweenness` https://graph-tool.skewed.de/static/doc/centrality.html#graph_tool.centrality.betweenness

jlapeyre:
This PR implements betweenness centrality for graphs and digraphs.
It supports boolean options `endpoints` and `normalized`, with the
same names and meanings as their networkx counterparts.

Closes #195

- [x] Needs documentation beyond the minimal list of parameters.
- [x] May need additional comments in code.
- [x] Centralities are returned in a `list` embedded in another list. This is an accident of the way it is constructed. The return type should probably be a `dict` to agree with networkx.

- [x] May need to use `Vec` in some places rather than hash maps for performance. Or, if we can't assume that all graph nodes can be referenced by an int, we need to change everything to hash maps.

- [x] A higher level function that checks whether the graph is of type `PyGraph` or `PyDigraph` and calls the corresponding function might be nice. (See #215)

- [x] There are some tests. ~~One test written, but I need to learn to run the tests. I have run standalone tests in scripts.~~

- [x] Release note.


src/centrality.rs
```diff
@@ -0,0 +1,217 @@
...
+        + NodeCount
+        + GraphProp
+        + GraphBase<NodeId = NodeIndex>,
+    // rustfmt deletes the following comments if placed inline above
+    // + IntoNodeIdentifiers // for node_identifiers()
+    // + IntoNeighborsDirected // for neighbors()
+    // + NodeCount // for node_count
+    // + GraphProp // for is_directed
+{
+    let mut betweenness = vec![0.0; graph.node_count()];
```
mtreinish, line 42: So I assume based on this betweeness is being returned as a `Vec` where the index in the Vec matches the node index in the graph. The problem with this is that there is no guarantee with the stable graph struct we use for the node indices to be a contiguous set, we can end up with holes in the indices list if nodes are removed. So there are situations where `max(node_indices) != max(node_count)`.

So I'm thinking it might make sense to change the output type to be a `HashMap<usize, f64>` instead. You can use an iterator and collect to intialize it in the same way. Something like:

```rust
let mut betweeness: HashMap<usize, f64> = graph.node_identifiers().map(|iden| (graph.to_index(iden), 0.0)).collect();
```
```

File: src/centrality.rs [hunk 2]
```diff
@@ -0,0 +1,217 @@
...
+use petgraph::graph::NodeIndex;
+use petgraph::visit::{
+    GraphBase,
+    GraphProp, // allows is_directed
+    IntoNeighborsDirected,
+    IntoNodeIdentifiers,
+    NodeCount,
+    NodeIndexable,
+};
+use std::collections::HashMap;
```
mtreinish, line 12: We typically use the [Hashbrown](https://github.com/rust-lang/hashbrown/) HashMap type instead of the one from stdlib, mostly because it is faster out of the box because it uses a different hasher which is not DoS resistant and also adds support for multithreaded iterators via rayon. It's a drop in replacement (and actually the upstream of the stdlib implementation), so just:
```suggestion
use hashbrown::HashMap;
```

src/centrality.rs
```diff
@@ -0,0 +1,217 @@
...
+                    graph.to_index(w),
+                    *(sigma.get(&graph.to_index(w)).unwrap())
+                        + *(sigma.get(&graph.to_index(v)).unwrap()),
+                );
+                let e_p = predecessors.get_mut(&w).unwrap();
+                e_p.push(v);
+            }
+        }
+    }
+    return (sorted_by_distance, predecessors, sigma);
```
mtreinish, line 216: You don't actually need the return here, the last line in a function without a semicolon is an implicit return (`cargo clippy` will complain about this).

```suggestion
    (sorted_by_distance, predecessors, sigma)
```

src/centrality.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+    _rescale(
+        &mut betweenness,
+        graph.node_count(),
+        normalized,
+        graph.is_directed(),
+        endpoints,
+    );
+
+    return Ok(betweenness);
```
mtreinish, line 76: You don't actually need the return here, the last line in a function without a semicolon is an implicit return (cargo clippy will complain about this).

```suggestion
    Ok(betweenness)
```

src/centrality.rs
```diff
@@ -0,0 +1,217 @@
...
+        } else {
+            scale = 1.0 / ((node_count - 1) * (node_count - 2)) as f64;
+        }
+    } else {
+        if !directed {
+            scale = 0.5;
+        } else {
+            do_scale = false;
+        }
+    }
```
mtreinish, line 106: Could this be collapsed to?

```suggestion
    } else if !directed {
        scale = 0.5;
    } else {
        do_scale = false;
    }
```

src/centrality.rs
```diff
@@ -0,0 +1,217 @@
...
+        sigma.insert(i, 0);
+        distance.insert(index, -1);
+    }
+    sigma.insert(index_s.index(), 1);
+    distance.insert(index_s, 0);
+    Q.push(index_s);
+    while !Q.is_empty() {
+        let v = Q.remove(0);
+        sorted_by_distance.push(v);
+        //        let neighbors = graph.neighbors_directed(v, Incoming);
```
mtreinish, line 198: Probably isn't needed anymore
```suggestion
```

src/centrality.rs
```diff
@@ -0,0 +1,217 @@
...
+        for w in graph.neighbors(v) {
+            if *(distance.get(&w).unwrap()) < 0 {
+                Q.push(w);
+                distance.insert(w, *(distance.get(&v).unwrap()) + 1);
```
mtreinish, line 202: One thing I've been using more is `distance[&v]` in cases like this where we know v is in `distance` just to avoid the need to unwrap since it fails if v isn't in the hashmap instead of returning `Option<T>`

src/centrality.rs
```diff
@@ -0,0 +1,217 @@
...
+            }
+            if *(distance.get(&w).unwrap()) == *(distance.get(&v).unwrap()) + 1
```
mtreinish, line 204: Since we use distance[w] and distance[v] more than once do you think it makes sense to store them in a temp so we only need to access once. It likely won't have a performance impact but will definitely clean things up a bit so this line a bit so it'll be something like:

```rust
if distance_w == distance_v + 1 {
```

tests/test_centrality.py
```diff
@@ -0,0 +1,79 @@
...
+            self.graph, endpoints=True)
+        expected = [0.25, 0.41666666666666663, 0.41666666666666663, 0.25]
+        self.assertEqual(expected, betweenness[0])
+
+    def test_betweenness_centrality_unnormalized(self):
+        betweenness = retworkx.digraph_betweenness_centrality(
+            self.graph, endpoints=False, normalized=False)
+        expected = [0.0, 2.0, 2.0, 0.0]
+        self.assertEqual(expected, betweenness[0])
+
```
mtreinish, line 48: It probably would be good to have a test case here for holes in the node indices. https://github.com/Qiskit/retworkx/pull/229/files#diff-abab60754038a6bbd6ef6b34128658a3bedbdbccc6743e87c5f8a2d6660f5614R113 is an example where I just did this in an open PR.

src/lib.rs
```diff
@@ -1620,6 +1621,32 @@ fn weight_callable(
     }
 }
 
+#[pyfunction(normalized = "true", endpoints = "false")]
```
mtreinish, line 1624: This will need a docstring (with `///`) in the sphinx rst format. The pyfunction macro will set this string to `__doc__` on the  python function it generates. Also since these are lib functions we'll need to add them to the `docs/source/api.rst` so they get included in the docs build.

src/lib.rs
```diff
@@ -1620,6 +1621,32 @@ fn weight_callable(
...
+#[text_signature = "(graph, /, normalized=True, endpoints=False)"]
+pub fn graph_betweenness_centrality(
+    py: Python,
+    graph: &graph::PyGraph,
+    normalized: bool,
+    endpoints: bool,
+) -> PyResult<PyObject> {
+    let betweenness =
+        centrality::betweenness_centrality(&graph, endpoints, normalized);
+    Ok(PyList::new(py, betweenness).into())
```
mtreinish, line 1634: One thing here is you can return `Ok(betweeness)` directly, the `pyfunction` macro will understand how to convert that into a python list (because there is an implementation of the `ToPythonObject` trait for Vec<T> where T also implements it). But it's probably worth benchmarking first because I've seen situations where constructing a python type explicitly in the rust code is faster than relying on the macro to do it for us (and vice versa).

mtreinish (commented):
Thanks for pushing this up, sorry it took me so long to take a look at it. 

This is a very good start, I did a quick pass and just highlighted some quick things inline. Most of them are minor style things or quick fixes, the only thing which I think is a real issue is using a `Vec<f64>` for the main return type because node indices aren't necessarily a contiguous range with the StableGraph type, they're more integer keys (which is why the PyGraph and PyDiGraph implement the python mapping protocol instead of the sequence protocol). So this would potentially have some weird bugs in graphs where there are removed nodes.

mtreinish:
I tagged this with the 0.8.0 milestone because I think it would be good to include it in the next release, but there's no pressure we can always hold it off until 0.9.0 if it's not ready in time.

src/centrality.rs
jlapeyre: (lost this comment earlier?) I am usually in the always-use-explicit-return camp for languages that allow both. One reason is I sometimes see a big blob of computation at the bottom of a function and it's not clear whether its for the side effect or a return value. A strong convention helps clarify it. But, it's more important to follow conventions.

src/centrality.rs
jlapeyre: I might guess that the compiler would optimize this. But, it would be more clear to store them in temps.

mtreinish: Yeah, the compiler would most likely optimize this for us, sometimes I get too used to working in python (where you would need to do this optimization by hand) :)

src/centrality.rs
jlapeyre: OMG. I just successfully replaced `*(distance.get(&w).unwrap())` with `distance[&w]`. I am now more positively disposed towards rust.

src/centrality.rs
mtreinish: Yeah I tend to agree with you I like to be explicit instead of implicit behavior, but I was deferring to the default linter rules that `cargo clippy` has and using an implicit return where possible seems to be the normal rust convention.

src/centrality.rs
jlapeyre: Is there a simpler way to write `distance.insert(w, distance[&v] + 1)`? I'd like to write `distance[&w] = distance[&v] + 1`.

src/centrality.rs
jlapeyre: I now have `if distance[&w] == distance[&v] + 1`, which is ok for verbosity. `distance[&w]` may be mutated. That leaves `distance[&v]` which is read twice. I guess the compiler can see that `distance[&v]` is never written to inside the while loop.

tests/test_centrality.py
jlapeyre: For the same graph you build in your test, my current implementation results in `panicked at 'index out of bounds`. I definitely need to add the test and then fix the implementation.

src/centrality.rs
mtreinish: I don't think HashMap implements the [`IndexMut` trait](https://doc.rust-lang.org/std/ops/trait.IndexMut.html) which allows mutating like that. So the only way I can see it working is if you know that `distance[w]` is always defined you can do something like:

```rust
*distance.get_mut(&w).unwrap() = distance[&v] + 1
```

If it's not guaranteed to be defined (so `unwrap()` could panic) you use an entry. so do something like:

```rust
distance.entry(&w).and_modify(|_| {distance[&v] + 1}).or_insert(distance[&v] + 1);
```

(I guess you can use the entry to do `distance.entry(&w).and_modify(|_| {distance[&v] + 1});` too to avoid the `get_mut` and `unwrap()`)

src/centrality.rs
mtreinish: One thing here I should have mentioned is that if you don't want to use a hashmap for performance you could use `graph.node_bound()` here to get the max node index in the graph and use that instead. But if you were I would probably switch this to be `Vec<Option<f64>>` instead so we can set any potential holes in the indices to `None`. We probably should still return a `HashMap` for ease of accessing from python, but that's as easy as something like: `out_map: HashMap<usize, f64> = betweeness.iter().enumerate().filter(|(index, v)| v.is_some()).collect()` to filter out None and return a HashMap.

src/centrality.rs
jlapeyre: This is probably resolved with recent edits.

src/centrality.rs
jlapeyre: This is probably resolved with recent edits

src/centrality.rs
jlapeyre: > you could use `graph.node_bound()` here to get the max node

I don't understand how this solves the problem of holes in the node index set.

> I would probably switch this to be `Vec<Option<f64>>`

This seems to amount to looping over all integers, checking whether each one is in the index set. Is it reasonable that someone would have, say,  five nodes only with contiguous indices starting at `10^6`? In this case, using `Vec<Option<f64>>` far less space and time efficient than using a hashmap.

I am in the process of implementing the `Vec<Option<f64>>` idea. In fact it's done, but buggy; betweenness values are wrong. But, as I am finishing, it occurs to me that this may not be  a good solution.

In any case, I agree that returning a `dict` is a good idea.

jlapeyre: I have changed the implementation to use `Vec<Option<f64>>` and pushed the commits.

tests/test_centrality.py
jlapeyre: The link above for the test for holes is broken (don't know if it is temporary github disruption or not). In any case, I added another test to check that the algorithm works with a hole in the index set.

jlapeyre:
I suppressed a clippy error, "type_complexity". This may not be the best solution, but it might allow the CI to pass.

src/centrality.rs
```diff
@@ -0,0 +1,307 @@
...
+    // + GraphProp // for is_directed
+{
+    let mut max_index = 0;
+    for node_s in graph.node_identifiers() {
+        let is: usize = graph.to_index(node_s);
+        if is > max_index {
+            max_index = is;
+        }
+    }
+    max_index += 1;
```
mtreinish, line 65: You can replace this with:

```suggestion
    let max_index = graph.node_bound();
```

The result of `node_bound()` is the current max index in the graph: https://docs.rs/petgraph/0.6.0/petgraph/stable_graph/struct.StableGraph.html#method.node_bound It does require the `NodeIndexable` trait but you already have that as a trait bound for `G`.

tests/test_centrality.py
mtreinish: Ah yeah that test file moved around since things reorganized https://github.com/Qiskit/retworkx/blob/main/tests/graph/test_max_weight_matching.py#L242-L279 but yeah you've got the coverage below in `TestCentralityDiGraphDeletedNode`

tests/test_centrality.py
```diff
@@ -0,0 +1,113 @@
...
+        betweenness = retworkx.graph_betweenness_centrality(
+            self.graph, endpoints=True)
+        expected = {0: 0.5, 1: 0.8333333333333333, 2: 0.8333333333333333, 3: 0.5}
+        self.assertEqual(expected, betweenness)
+
+    def test_betweenness_centrality_unnormalized(self):
+        betweenness = retworkx.graph_betweenness_centrality(
+            self.graph, endpoints=False, normalized=False)
+        expected = {0: 0.0, 1: 2.0, 2: 2.0, 3: 0.0}
+        self.assertEqual(expected, betweenness)
```
mtreinish, line 79: We've moved to splitting the test files up by graph class they work with. So for this do you mind moving it to a new file `tests/graph/test_centrality.py` and move the remaining digraph cases in this file to `tests/digraph/test_centrality.py`? (we probably should update the contributing guide to outline this test dir structure it was an oversight in #296 that we didn't do that then)

Also do you want to add a test class for `PyGraph` with holes too to match up  the coverage for `PyDiGraph` below?

mtreinish (commented):
Thanks for updating this again. This is getting close and I think we definitely have enough time to include in 0.10.0. I just did a quick initial pass through the code I have a few inline comments on some higher level things there.

Besides those the only other thing missing here is adding the functions to `docs/source/api.rst` so they get included in the docs build and also the dispatch python function to `retworkx/__init__.py` for `retworkx.betweenness_centrality()` so there is a single entrypoint (I realize this wasn't a thing when you first opened that PR but we've since closed #215 and been adding these for all new functions that work with either graph type).

src/centrality.rs
jlapeyre: I knew there must be something like this. I was grepping for things with "max" in it.

tests/test_centrality.py
jlapeyre: I don't know if you saw my comment above: I'm concerned about the possibility that someone creates a graph with, say, five nodes, but the indices are all greater than `10^6`. In the current implementation, the algorithm would create and iterate over vecs of length greater than `10^6`. I wonder if this is some that is possible and if so should we account for it, i.e. give a more efficient implementation.

mtreinish: Oh I must have missed that comment. It is technically possible but a pretty pathological edge case. Mostly because stable graph maintains a list of free nodes and reuses indices on new nodes after removal. So to get this state you'd have to create a graph with a million nodes, then delete all but 5, and then run centrality on it. We probably could use a hashmap or btreemap instead of a vec and then we'd only iterate over n nodes, but the tradeoff there is the iteration performance is lower than a vec. Personally I'm fine either way and don't have a strong preference.

jlapeyre:
I don't know if you saw my comment above: I'm concerned about the possibility that someone creates a graph with, say, five nodes, but the indices are all greater than 10^6. In the current implementation, the algorithm would create and iterate over vecs of length greater than 10^6. I wonder if this is some that is possible and if so should we account for it, i.e. give a more efficient implementation.

jlapeyre:
I noticed in `__init__.py` that we sometimes use a single python entry point for both the graph and digraph versions of an algorithm. Also, in `api.rst` it appears that there are three entry points for some algorithms: one for graph, one for digraph, and one that accepts either and then dispatches to one of the first two.  Doesn't it make more sense to provide either one or two functions, rather than three? At the moment, I am planning on simply following the pattern by providing three entry points for betweenness.

mtreinish:
> I noticed in `__init__.py` that we sometimes use a single python entry point for both the graph and digraph versions of an algorithm. Also, in `api.rst` it appears that there are three entry points for some algorithms: one for graph, one for digraph, and one that accepts either and then dispatches to one of the first two. Doesn't it make more sense to provide either one or two functions, rather than three? At the moment, I am planning on simply following the pattern by providing three entry points for betweenness.

It's an unfortunate aspect of how the python binding is built. There is no way to use a generic on the python interface function (using the `pyfunction` macro) because the macro can't resolve the conversion into the generated ffi functions. So we need to have an explicit variant for each graph type exposed off the python module. The dispatch functions in `__int_.py`  are just there to provide a unified api for users (see #215). It was the best pattern I could find to provide a single graph type neutral function.

jlapeyre:
> It was the best pattern I could find to provide a single graph type neutral function.

I think I understand that dispatching, either with a generic or just `isinstance` has to be done in pure python. But, is it possible to avoid exposing the interface functions for graph and digraph? Maybe prepend underscores, omit them from the api documentation in api.rst, things like that. Then, the pure-python generic would be the only thing advertised to users.

mtreinish:
> > It was the best pattern I could find to provide a single graph type neutral function.
> 
> I think I understand that dispatching, either with a generic or just `isinstance` has to be done in pure python. But, is it possible to avoid exposing the interface functions for graph and digraph? Maybe prepend underscores, omit them from the api documentation in api.rst, things like that. Then, the pure-python generic would be the only thing advertised to users.
> 
> EDIT: In any case, I have pushed edits following the existing pattern in retworkx.

That would be an option for new functions. We couldn't do that for places where we were already exposing the strictly typed variants which are still needed for backwards compatibility (which is most of the API surface since the python dispatch functions are relatively new). The only other reason I could think of is additional overhead. Although this is probably not a big concern in most cases, dispatching does add a second python function call. While a python function call is relatively quick (scale of 100s of ns) it can add a noticeable overhead in some cases.

src/centrality.rs
```diff
@@ -0,0 +1,296 @@
...
+
+/// Compute the betweenness centrality of all nodes in a PyGraph.
+///
+/// :param PyGraph graph: The input graph
+/// :param bool normalized: Whether to normalize the betweenness scores by the number of distinct
+///    paths between all pairs of nodes.
+/// :param bool endpoints: Whether to include the endpoints of paths in pathlengths used to
+///    compute the betweenness.
+///
+/// :returns a dict whose keys are the node indices and values are the betweenness score
```
mtreinish, line 250:
```suggestion
/// :returns: a dict whose keys are the node indices and values are the betweenness score
```

src/centrality.rs
```diff
@@ -0,0 +1,296 @@
...
+
+/// Compute the betweenness centrality of all nodes in a PyDiGraph.
+///
+/// :param PyDiGraph graph: The input graph
+/// :param bool normalized: Whether to normalize the betweenness scores by the number of distinct
+///    paths between all pairs of nodes.
+/// :param bool endpoints: Whether to include the endpoints of paths in pathlengths used to
+///    compute the betweenness.
+///
+/// :returns a dict whose keys are the node indices and values are the betweenness score
```
mtreinish, line 278:
```suggestion
/// :returns: a dict whose keys are the node indices and values are the betweenness score
```

mtreinish (commented):
Sphinx error message are super difficult to parse, mostly because they come from translating docutils parser errors in my experience (which is a really crusty old package). I left an inline comment to what I think the issue which was a missing colon at the end of the returns directive

jlapeyre:
>  I left an inline comment to what I think the issue which was a missing colon at the end of the returns directive

I made the change locally and did not seem to fix the doc build locally; same error message.

src/centrality.rs
```diff
@@ -0,0 +1,296 @@
...
+    normalized: bool,
+    endpoints: bool,
+) -> PyResult<HashMap<usize, f64>> {
+    let betweenness = betweenness_centrality(&graph, endpoints, normalized);
+    let out_map: HashMap<usize, f64> = betweenness
+        .into_iter()
+        .enumerate()
+        .filter_map(|(i, v)| v.map(|x| (i, x)))
+        .collect();
+    Ok(out_map)
```
mtreinish, line 295: I'm wondering if it makes sense to use a custom return type for this. This would avoid the python dict construction cost from the function and return a read-only dict like mapping instead. Off the shelf `PathLengthMapping` wraps a `HashMap<usize, f64>` but the name doesn't make any sense here, but you could add an equivalent new return type by just copying the macro call in `iterators.rs` for that and give it a better name.

src/centrality.rs
```diff
@@ -0,0 +1,296 @@
...
+        let index = NodeIndex::new(i);
+        predecessors.insert(index, Vec::new());
+        sigma.insert(i, 0);
+        distance.insert(index, -1);
+    }
+    sigma.insert(index_s.index(), 1);
+    distance.insert(index_s, 0);
+    Q.push(index_s);
+    while !Q.is_empty() {
+        let v = Q.remove(0);
```
mtreinish, line 216: If we're using `Q` as a FIFO I'm wondering if makes more sense to use a `VecDeque` here https://doc.rust-lang.org/std/collections/struct.VecDeque.html because doing `Vec.remove(0)` requires shifting every element in the queue  while `VecDeque` is a ring buffer and avoids that.

src/centrality.rs
```diff
@@ -0,0 +1,296 @@
...
+use petgraph::visit::{
+    GraphBase,
+    GraphProp, // allows is_directed
+    IntoNeighborsDirected,
+    IntoNodeIdentifiers,
+    NodeCount,
+    NodeIndexable,
+};
+
+// Correspondence to notation in Brandes 2001
```
mtreinish, line 30: I think it might make sense to add a link to this or a citation to the docstring we export to python.

src/centrality.rs
```diff
@@ -0,0 +1,296 @@
...
+        } else {
+            scale = 1.0 / ((node_count - 1) * (node_count - 2)) as f64;
+        }
+    } else if !directed {
+        scale = 0.5;
+    } else {
+        do_scale = false;
+    }
+    if do_scale {
+        for i in 0..betweenness.len() as usize {
```
mtreinish, line 123: hmm, this needs a cast? That's kind of surprising, `Vec.len()` should return a `usize`

src/centrality.rs
```diff
@@ -0,0 +1,296 @@
...
+            if betweenness[i].is_some() {
+                betweenness[i] = Some(betweenness[i].unwrap() * scale);
+            }
```
mtreinish, line 126: I think this can be simplified a bit with:

```suggestion
             betweenness[i] = betweenness[i].map(|x| x* scale);
```

src/centrality.rs
```diff
@@ -0,0 +1,296 @@
...
+        let iw = w.index();
+        let coeff = (1.0 + delta[iw]) / (path_calc.sigma[&iw] as f64);
+        let p_w = path_calc.predecessors.get_mut(w).unwrap();
+        for v in p_w {
+            let iv = (*v).index();
+            delta[iv] += (path_calc.sigma[&iv] as f64) * coeff;
+        }
+        if iw != is && betweenness[iw].is_some() {
+            betweenness[iw] = Some(betweenness[iw].unwrap() + delta[iw]);
+        }
```
mtreinish, line 148:
```suggestion
        if iw != is {
            betweenness[iw] = betweenness[iw].map(|x| x + delta[iw]);
        }
```

src/centrality.rs
```diff
@@ -0,0 +1,296 @@
...
+        let iw = w.index();
+        let coeff = (1.0 + delta[iw]) / (path_calc.sigma[&iw] as f64);
+        let p_w = path_calc.predecessors.get_mut(w).unwrap();
+        for v in p_w {
+            let iv = (*v).index();
+            delta[iv] += (path_calc.sigma[&iv] as f64) * coeff;
+        }
+        if iw != is && betweenness[iw].is_some() {
+            betweenness[iw] = Some(betweenness[iw].unwrap() + delta[iw] + 1.0);
+        }
```
mtreinish, line 175:
```suggestion
        if iw != is {
            betweenness[iw] = betweenness[iw].map(|x| x + delta[iw] + 1.0);
        }
```

mtreinish (commented):
LGTM, thanks for updating this and keeping with it. I left a few inline comments on some small nits we can fix to be a bit more idiomatic. But besides those (which aren't really blockers) there are 2 comments I think we probably want to think about before approving. The first is the data structure we use for the queue in `shortest_path_for_centrality`, because depending on the number of neighbors a `Vec` might not be the most efficient there.

The second is the return type. I'm thinking it might be good to return a custom read only mapping that wraps a `HashMap<usize, f64>` instead of a dict. You can basically just copy: https://github.com/Qiskit/retworkx/blob/main/src/iterators.rs#L1074-L1111 and change the class names to be more about betweeness and then wrap the output hashmap in that struct (like: https://github.com/Qiskit/retworkx/blob/main/src/shortest_path/mod.rs#L223-L235). This will avoid the overhead of converting from a hashmap to a dict (which requires iterating over every entry in python) and make the function return faster. This isn't a blocker though, we can always do this in a followup too as it's mostly backwards compatible.

src/centrality.rs
jlapeyre: If I remove `as usize`, tests still pass. But then clippy wants me to use a more idiomatic form. I did not insert `as usize` for that reason. I don't know why I did it; I'm sure in response to some failure. In any case, clippy suggests ways around it all. I will make the edit.

docs/source/api.rst
```diff
@@ -37,6 +37,16 @@ Shortest Paths
    retworkx.num_shortest_paths_unweighted
    retworkx.unweighted_average_shortest_path_length
 
+.. _centrality:
+
+Centrality
```
georgios-ts, line 42: We can name this section something like `Graph Measures` and include here `retworkx.transitivity` and `retworkx.core_number` that are currently placed in the more general section `Other Algorithm Functions`.

docs/source/api.rst
```diff
@@ -37,6 +37,16 @@ Shortest Paths
...
+--------------
+
+.. autosummary::
+   :toctree: stubs
+
+   retworkx.graph_betweenness_centrality
```
georgios-ts, line 48:
```suggestion
   retworkx.betweenness_centrality
```

src/centrality.rs
```diff
@@ -0,0 +1,298 @@
...
+    betweenness: &mut Vec<Option<f64>>,
+    max_index: usize,
+    path_calc: &mut ShortestPathData,
+    is: usize,
+) {
+    let mut delta = vec![0.0; max_index];
+    for w in &path_calc.verts_sorted_by_distance {
+        let iw = w.index();
+        let coeff = (1.0 + delta[iw]) / (path_calc.sigma[&iw] as f64);
+        let p_w = path_calc.predecessors.get_mut(w).unwrap();
```
georgios-ts, line 143: here we don't really need a mutable ref.
```suggestion
        let p_w = path_calc.predecessors.get(w).unwrap();
```

src/centrality.rs
```diff
@@ -0,0 +1,298 @@
...
+    max_index: usize,
+    path_calc: &mut ShortestPathData,
+    is: usize,
+) {
+    if betweenness[is].is_some() {
+        betweenness[is] = Some(
+            betweenness[is].unwrap()
+                + ((path_calc.verts_sorted_by_distance.len() - 1) as f64),
+        );
+    }
```
georgios-ts, line 165:
```suggestion
    betweenness[is] = betweenness[is]
        .map(|x| x + ((path_calc.verts_sorted_by_distance.len() - 1) as f64));
    }
```

src/centrality.rs
```diff
@@ -0,0 +1,298 @@
...
+    let mut delta = vec![0.0; max_index];
+    for w in &path_calc.verts_sorted_by_distance {
+        let iw = w.index();
+        let coeff = (1.0 + delta[iw]) / (path_calc.sigma[&iw] as f64);
+        let p_w = path_calc.predecessors.get_mut(w).unwrap();
```
georgios-ts, line 170:
```suggestion
        let p_w = path_calc.predecessors.get(w).unwrap();
```

src/centrality.rs
```diff
@@ -0,0 +1,298 @@
...
+        + IntoNodeIdentifiers
+        + IntoNeighborsDirected
+        + GraphBase<NodeId = NodeIndex>, // for get() and get_mut()
+{
+    let mut verts_sorted_by_distance: Vec<NodeIndex> = Vec::new(); // a stack
+    #[allow(non_snake_case)]
+    let mut Q: Vec<NodeIndex> = Vec::new(); // a queue
+    let mut predecessors = HashMap::<G::NodeId, Vec<G::NodeId>>::new();
+    let mut sigma = HashMap::<usize, i64>::new();
+    let mut distance = HashMap::<G::NodeId, i64>::new();
```
georgios-ts, line 202: Since we will add `node_count` elements in this dictionaries we can initially create them with this capacity so we don't have to reallocate. (note that we will need type `G` to implement 
 `NodeCount` trait).

```suggestion
    let c = graph.node_count();
    let mut predecessors =
        HashMap::<G::NodeId, Vec<G::NodeId>>::with_capacity(c);
    let mut sigma = HashMap::<usize, i64>::with_capacity(c);
    let mut distance = HashMap::<G::NodeId, i64>::with_capacity(c);
```

I am thinking also if it's better in terms of performance to use here a `Vec` of size `node_bound` (kind of what you did with `betweeness` in `betweenness_centrality`).

georgios-ts (commented):
Some inline suggestions and a general remark that we may hit in `shortest_path_for_centrality` the same overflow issues as in https://github.com/Qiskit/retworkx/pull/355#pullrequestreview-682446520

jlapeyre:
Thanks for the comments, @georgios-ts 

> Some inline suggestions and a general remark that we may hit in shortest_path_for_centrality the same overflow issues as in #355 (review)

You are correct. This [issue in the Julia graph library](https://github.com/JuliaGraphs/LightGraphs.jl/issues/1405) is concerned with exactly this problem. There, it is argued that using `f64` for the path counts is preferable to an arbitrary precision integer type because it offers much better performance with, in this case, no loss of precision. Using the Julia library, I computed betweenness centrality for a grid graph using both `f64` and `BigInt`. There was no difference in precision, but the former was five times faster. But, it's worth testing both here.

docs/source/api.rst
jlapeyre: I intentionally followed the [networkx docs](https://networkx.org/documentation/networkx-1.10/reference/algorithms.html) in choosing this heading. My thinking was that, if retworkx were to develop anything like feature parity (a big job), then following their doc organization, which seems to be done very well, would be a good idea.

src/centrality.rs
jlapeyre: Hmm. Don't know why I did this. I'm surprised the linting tools didn't complain about this.

src/centrality.rs
jlapeyre: I find preallocating often doesn't really help. But, this optimization doesn't obfuscate the code, in fact, it signals some intent. A good idea.

src/centrality.rs
```diff
@@ -0,0 +1,298 @@
...
+        let index = NodeIndex::new(i);
+        predecessors.insert(index, Vec::new());
+        sigma.insert(i, 0);
+        distance.insert(index, -1);
+    }
+    sigma.insert(index_s.index(), 1);
+    distance.insert(index_s, 0);
+    Q.push_back(index_s);
+    while !Q.is_empty() {
+        let v = Q.pop_front().unwrap(); // remove(0);
```
georgios-ts, line 218:
```suggestion
    while let Some(v) = Q.pop_front() {
```

src/centrality.rs
```diff
@@ -0,0 +1,298 @@
...
+        + IntoNodeIdentifiers
+        + IntoNeighborsDirected
+        + NodeCount
+        + GraphBase<NodeId = NodeIndex>, // for get() and get_mut()
+{
+    let mut verts_sorted_by_distance: Vec<NodeIndex> = Vec::new(); // a stack
+    let c = graph.node_count();
+    let mut predecessors =
+        HashMap::<G::NodeId, Vec<G::NodeId>>::with_capacity(c);
+    let mut sigma = HashMap::<usize, i64>::with_capacity(c);
```
georgios-ts, line 199: It's kind of weird that `sigma` has keys of type `usize` and the others have `G::NodeId`. Is there a reason for this?

docs/source/api.rst
georgios-ts: Yeah, that's a good idea. It's just right now that retworkx lacks a ton of functionality comparing to networkx so having more general sections where we can place a few different functions seemed more appropriate. But it's totally fine to leave it as it is.

src/centrality.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    let i_s = graph.to_index(*node_s);
+    let index_s = NodeIndex::new(i_s);
+
+    for node in graph.node_identifiers() {
+        let i = graph.to_index(node);
+        let index = NodeIndex::new(i);
+        predecessors.insert(index, Vec::new());
+        sigma.insert(i, 0);
+        distance.insert(index, -1);
```
georgios-ts, line 212: there is no need to create a new `index` since `node` is the same.
```suggestion
        let i = graph.to_index(node);
        predecessors.insert(node, Vec::new());
        sigma.insert(i, 0);
        distance.insert(node, -1);
```

georgios-ts (commented):
Ok then we might get away with `f64`. `BigUint` seems not to be that much slower (see https://github.com/Qiskit/retworkx/pull/355#discussion_r652184306) but in this case we make `node_count` calls in `shortest_path_for_centrality`, so the time difference might start accumulating. Once you make this change, it should be ready to go.

Oh, it might be good also to add a note in the docs saying that this function supports multigraphs and (if i get this right) the paths that pass through different parallel edges are considered to be unique.

src/centrality.rs
jlapeyre: I don't recall. I'll look into it.

jlapeyre: It looks like using `G::NodeId`  (or `NodeIndex`) is no less efficient than `usize`.

src/centrality.rs
jlapeyre: Done

jlapeyre:
I think all the issues raised have been addressed.

src/iterators.rs
```diff
@@ -1130,6 +1130,45 @@ impl PyDisplay for PathLengthMapping {
...
+        # Target node access
+        third_element = edges[2]
+        # Use as iterator
+        edges_iter = iter(edges)
+        first_target = next(edges_iter)
+        first_path = edges[first_target]
+        second_target = next(edges_iter)
+        second_path = edges[second_target]
+
+    "
```
mtreinish, line 1168: Just some small docstring tweaks:

```suggestion
    "A custom class for the return of centralities at target nodes

    This class is a container class for the results of functions that
    return a mapping of node index to the betweenness score for that node.
    It implements the Python mapping protocol so you can treat the return
    as a read-only mapping/dict.
    "
```

Also we should probably add this class to the exported list in `src/lib.rs` and the docs for custom return types in `docs/source/api.rst`. Without including `CentralityMapping` there we're basically returning a private undocumented type that users can't import.

jlapeyre: Oh yeah, thanks for this. I completely forgot to change the docstring, I just did one word. I also just pushed an edit to export `CentralityMapping` that is neccessary for tests to pass.

mtreinish (approved):
LGTM thanks for sticking with this and making all the updates. I left one inline comment on a spot where I think we might be able to investigate making the implementation multithreaded, but that's definitely a follow on. The only thing I see missing is adding the new custom return type to the docs, but I'll just do this in #414 when I rebase it.

PR #229, opened 2021-01-17
Add max weight matching function

### LINKED ISSUE #216: Add max_weight_matching function

retworkx is missing a `max_weight_matching()` function, similar to networkx's function: https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.matching.max_weight_matching.html?highlight=max_weight_matching#networkx.algorithms.matching.max_weight_matching

This is currently a blocker from removing the last networkx usage from qiskit in qiskit-ignis. Specifically this function is used here:: https://github.com/Qiskit/qiskit-ignis/blob/master/qiskit/ignis/verification/topological_codes/fitters.py#L298

mtreinish:
This commit adds a new python function max_weight_matching() for
computing the maximum-weighted matching of a PyGraph object. The
implementation of this function is based on “Efficient Algorithms for
Finding Maximum Matching in Graphs”, Zvi Galil, ACM Computing Surveys,
1986. [1] It is basically a porting of the networkx implementation of
the algorithm [2] with some additional inspiration from the prototype for
the networkx version (it's basically the same code but some aspects were
a bit clearer to figure out from the prototype rather than networkx's
copy of it). [3][4]

Fixes #216

[1] https://dl.acm.org/doi/10.1145/6462.6502
[2] https://github.com/networkx/networkx/blob/3351206a3ce5b3a39bb2fc451e93ef545b96c95b/networkx/algorithms/matching.py
[3] http://jorisvr.nl/article/maximum-matching
[4] http://jorisvr.nl/files/graphmatching/20130407/mwmatching.py

TODO:

- [x] Fix test failures
- [x] Change output from function to be set of tuples for each matching pair instead of dict?

### DIFF
--- a/src/max_weight_matching.rs
+++ b/src/max_weight_matching.rs
@@ -102,7 +102,6 @@ fn assign_label(
     labels[b] = Some(t);
     label_ends[b] = p;
     label_ends[w] = p;
-    label_ends[b] = p;
     best_edge[w] = None;
     best_edge[b] = None;
     if t == 1 {
@@ -356,7 +355,7 @@ fn expand_blossom(
     label_ends: &mut Vec<Option<usize>>,
     best_edge: &mut Vec<Option<usize>>,
     queue: &mut Vec<usize>,
-    blossom_base: &[Option<usize>],
+    blossom_base: &mut Vec<Option<usize>>,
     endpoints: &[usize],
     mate: &[Option<usize>],
     blossom_endpoints: &mut Vec<Vec<usize>>,
@@ -563,6 +562,7 @@ fn expand_blossom(
     label_ends[blossom] = None;
     blossom_children[blossom].clear();
     blossom_endpoints[blossom].clear();
+    blossom_base[blossom] = None;
     best_edge[blossom] = None;
     unused_blossoms.push(blossom);
     Ok(())
@@ -1249,7 +1249,7 @@ pub fn max_weight_matching(
                     &mut label_ends,
                     &mut best_edge,
                     &mut queue,
-                    &blossom_base,
+                    &mut blossom_base,
                     &endpoints,
                     &mate,
                     &mut blossom_endpoints,
@@ -1283,7 +1283,7 @@ pub fn max_weight_matching(
                     &mut label_ends,
                     &mut best_edge,
                     &mut queue,
-                    &blossom_base,
+                    &mut blossom_base,
                     &endpoints,
                     &mate,
                     &mut blossom_endpoints,
--- a/src/max_weight_matching.rs
+++ b/src/max_weight_matching.rs
@@ -291,7 +291,8 @@ fn add_blossom(
         in_blossoms[node] = blossom;
     }
     // Compute blossom_best_edges[blossom]
-    let mut best_edge_to: Vec<Option<usize>> = vec![None; 2 * num_nodes];
+    let mut best_edge_to: HashMap<usize, usize> =
+        HashMap::with_capacity(2 * num_nodes);
     for bv_ref in &blossom_children[blossom] {
         let bv = *bv_ref;
         // This sub-blossom does not have a list of least-slack edges;
@@ -317,15 +318,15 @@ fn add_blossom(
                 let blossom_j = in_blossoms[j];
                 if blossom_j != blossom
                     && labels[blossom_j] == Some(1)
-                    && (best_edge_to[blossom_j].is_none()
+                    && (best_edge_to.get(&blossom_j).is_none()
                         || slack(edge_index, &dual_var, &edges)
                             < slack(
-                                best_edge_to[blossom_j].unwrap(),
+                                best_edge_to[&blossom_j],
                                 &dual_var,
                                 &edges,
                             ))
                 {
-                    best_edge_to[blossom_j] = Some(edge_index);
+                    best_edge_to.insert(blossom_j, edge_index);
                 }
             }
         }
@@ -333,11 +334,7 @@ fn add_blossom(
         blossom_best_edges[bv].clear();
         best_edge[bv] = None;
     }
-    blossom_best_edges[blossom] = best_edge_to
-        .iter()
-        .filter(|x| x.is_some())
-        .map(|x| x.unwrap())
-        .collect();
+    blossom_best_edges[blossom] = best_edge_to.values().map(|x| *x).collect();
     //select best_edge[blossom]
     best_edge[blossom] = None;
     for edge_index in &blossom_best_edges[blossom] {
--- a/src/max_weight_matching.rs
+++ b/src/max_weight_matching.rs
@@ -868,8 +868,8 @@ pub fn max_weight_matching(
     // the blossom.
     let mut labels: Vec<Option<usize>>;
     // If b is a labeled top-level blossom,
-    // label_ends[b] is the remote endpoint of the edge through which b is
-    // obtained. Its label, or None if b's base vertex is single.
+    // label_ends[b] is the remote endpoint of the edge through which b
+    // obtained its label, or None if b's base vertex is single.
     // If v is a vertex inside a T-blossom and label[v] == 2,
     // label_ends[v] is the remote endpoint of the edge through which v is
     // reachable from outside the blossom
--- a/src/max_weight_matching.rs
+++ b/src/max_weight_matching.rs
@@ -19,6 +19,7 @@
 
 use std::cmp::max;
 use std::mem;
+use std::panic;
 
 use hashbrown::{HashMap, HashSet};
 
@@ -26,6 +27,7 @@ use petgraph::graph::NodeIndex;
 use petgraph::visit::{EdgeRef, IntoEdgeReferences};
 
 use crate::graph::PyGraph;
+use pyo3::exceptions::PyException;
 use pyo3::prelude::*;
 
 fn weight_callable(
@@ -394,7 +396,7 @@ fn expand_blossom(
             }
         }
     }
-    // if we expand a T-blossom during a stage, its sub-blossoms must be
+    // if we expand a T-blossom during a stage, its a sub-blossoms must be
     // relabeled
     if !end_stage && labels[blossom] == Some(2) {
         // start at the sub-blossom through which the expanding blossom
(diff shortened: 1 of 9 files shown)


mtreinish:
This is ready to review now. The only open question is on the return format. Right now it is returning a dict but networkx returns a  set of the  matching nodes. It's not hard to adjust it, but I'm not sure what people think is a better choice.

t-imamichi:
It might be good to add a unit test to compare results of networkx and retworkx with random graphs.

mtreinish:
> It might be good to add a unit test to compare results of networkx and retworkx with random graphs.

Sure, I added a couple of tests to do that in: https://github.com/Qiskit/retworkx/pull/229/commits/308c82a831ed73d963925af22545e199db00b895

itoko (commented):
Thank you for this great job. I've just started to read but not yet completed. I have one question so far. Is there any reason to use `endpoints` data structure instead of vertex pairs? (Is it just for performance?)

src/max_weight_matching.rs
```diff
@@ -0,0 +1,1306 @@
...
+    let mut queue: Vec<usize> = Vec::with_capacity(num_nodes);
+
+    // Main loop: continue until no further improvement is possible
+    for _ in 0..num_nodes {
+        // Each iteration of this loop is a "stage".
+        // A stage finds an augmenting path and uses that to improve
+        // the matching.
+
+        // Removal labels from top-level blossoms/vertices
+        labels = vec![Some(0); 2 * num_nodes];
```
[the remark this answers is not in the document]
mtreinish: Yeah, I think that's good to do probably, I'll add this.

I think it might not be strictly necessary here because we end up reset `label_ends` every time we update `labels`. But to be more thorough we should reset the endpoints when we reset labels.

mtreinish:
> Thank you for this great job. I've just started to read but not yet completed. I have one question so far. Is there any reason to use `endpoints` data structure instead of vertex pairs? (Is it just for performance?)

It was actually to keep the data structures a bit simpler, i haven't done any performance on it. When I first went to port the function from networkx I found keeping track of the more complex (and dynamic) data structures that networkx is using hard to do (mostly because there are a lot of different things to keep track of) in a way that is statically defined. This was actually the piece that I borrowed from the prototype version (which looks like itself was ported from old C code) that I mentioned in the commit message vs the networkx version, is using lists/`Vec`s. My thinking was we could start with this flatter but more verbose approach and then make iterative improvements to simplify things with more complex types over time.

src/max_weight_matching.rs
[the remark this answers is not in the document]
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/229/commits/81a59a491d637f3600456b23f84a3662f0eccb2f

src/max_weight_matching.rs
```diff
@@ -0,0 +1,1307 @@
...
+                            // w is inside a T-blossom, but w itself has not
+                            // yet been reached from outside the blossom;
+                            // mark it as reached (we need this to relabel
+                            // during T-blossom expansion).
+                            assert!(labels[in_blossoms[w]] == Some(2));
+                            labels[w] = Some(2);
+                            label_ends[w] = Some(*p ^ 1);
+                        }
+                    } else if labels[in_blossoms[w]] == Some(1) {
+                        // Leep track of the least-slack non-allowable edge to
```
itoko, line 1093:
```suggestion
                        // Keep track of the least-slack non-allowable edge to
```

src/max_weight_matching.rs
```diff
@@ -0,0 +1,1307 @@
...
+    // S-blossoms, or None if no such list has been computed yet.
+    // This is used for efficient computation of delta3.
+    let mut blossom_best_edges: Vec<Vec<usize>> =
+        (0..2 * num_nodes).map(|_| Vec::new()).collect();
+    let mut unused_blossoms: Vec<usize> = (num_nodes..2 * num_nodes).collect();
+    // If v is a vertex,
+    // dual_var[v] = 2 * u(v) where u(v) is the v's variable in the dual
+    // optimization problem (multiplication by two ensures integer values
+    // throughout the algorithm if all edge weights are integers).
+    // If b is a non-trivial blossom,
```
itoko, line 922: Commenting `v in 0..num_nodes` and `b in num_nodes..2 * num_nodes` somewhere may help readers a lot.

mtreinish: Yep I agree, it's definitely a bit confusing without explaining the split in the values. I added this to the bottom of the comment in: https://github.com/Qiskit/retworkx/pull/229/commits/37b518da230c49a08f8173607f6bdb49f2fe5103

mtreinish:
I think there is a subtle bug in here. I've been playing with adding more tests comparing retworkx's output with networkx's per @t-imamichi's suggestion and I'm seeing failures with some cases of different seeds. For example, I modified `test_gnp_random_against_networkx` to sweep the seed from 42 to 1066 with:
```diff
     def test_gnp_random_against_networkx(self):
-        rx_graph = retworkx.undirected_gnp_random_graph(10, .75, seed=42)
-        nx_graph = networkx.Graph(list(rx_graph.edge_list()))
-        nx_matches = networkx.max_weight_matching(nx_graph)
-        rx_match_dict = retworkx.max_weight_matching(rx_graph)
-        # Convert retworkx dict output to networkx set of tuples
-        rx_matches = {
-            (u, v) for (u, v) in set(map(frozenset, rx_match_dict.items()))}
-        self.assertEqual(nx_matches, rx_matches)
+        for i in range(1024):
+            rx_graph = retworkx.undirected_gnp_random_graph(10, .75,
+                                                            seed=42 + i)
+            nx_graph = networkx.Graph(list(rx_graph.edge_list()))
+            nx_matches = networkx.max_weight_matching(nx_graph)
+            rx_match_dict = retworkx.max_weight_matching(rx_graph)
+            # Convert retworkx dict output to networkx set of tuples
+            for (u, v) in set(map(frozenset, rx_match_dict.items())):
+                if (u, v) not in nx_matches:
+                    if (v, u) not in nx_matches:
+                        self.fail("seed %s failed. Element %s and it's "
+                                  " reverse %s not found "
+                                  "in networkx output.\nretworkx output: %s"
+                                  "\nnetworkx output: %s" % (
+                                      42 + i, (u, v), (v, u), rx_match_dict,
+                                      nx_matches))
```

and it fails first on seed 604. I also tried using the gnm random graph method instead of the gnp one and hit similar failures.
I'm also seeing issues in my attempt to use this in ignis (see: Qiskit/qiskit-ignis#552) which are likely related to the implementation here.

I'm going to push these test cases up so we block this on fixing whatever is causing things to fail.

itoko (commented):
I've looked through all lines comparing with [4] http://jorisvr.nl/files/graphmatching/20130407/mwmatching.py, and I couldn't found any mistakes by my eyes.

I also checked the seed 604 test case, and confirmed that retworkx just select a different maximum matching from that by networkx (there are several maximum matchings in the graph instance). So maybe the difference comes from the difference between implementation in [4] and networkx.

Regarding output, I personally think set of tuples for each matching pair would be preferable.

mtreinish:
> I've looked through all lines comparing with [4] http://jorisvr.nl/files/graphmatching/20130407/mwmatching.py
> Excuse me, I couldn't find the lines that causes the test failure.

Yeah, I tested `mwatching.py` for one of the failure cases and it has identical output to this PR. So I think I ported the bug from there, I'll have to compare it to networkx to see where things are diverging.

> 
> Regarding output, I personally think set of tuples for each matching pair would be preferable.

Ok, I'll convert it to a set of tuples then. I have a local patch that does this. The only issue I had when doing this was the reverse ordering `(u, v)` vs `(v, u)` compared to networkx.

itoko:
I've checked the seed 604 test case, and confirmed that retworkx just select a different maximum matching from that by networkx (there are several maximum matchings in the graph instance). So this may not be a bug. Maybe it comes from the subtle difference between [4] and networkx.
```
retworkx output: {3: 9, 7: 2, 0: 4, 6: 8, 9: 3, 1: 5, 2: 7, 4: 0, 5: 1, 8: 6}
= {(0, 4), (2, 7), (5, 1), (6, 8), (9, 3)}
networkx output: {(0, 4), (2, 7), (9, 5), (8, 6), (1, 3)}
edge list: [(1, 0), (3, 1), (3, 2), (4, 0), (4, 1), (4, 3), (5, 0), (5, 1), (5, 2), (6, 1), (7, 0), (7, 1), (7, 2), (7, 4), (7, 5), (8, 1), (8, 3), (8, 4), (8, 6), (8, 7), (9, 0), (9, 1), (9, 2), (9, 3), (9, 4), (9, 5), (9, 6), (9, 7)]
```

mtreinish:
> I've checked the seed 604 test case, and confirmed that retworkx just select a different maximum matching from one by networkx (there are several maximum matchings in this unweighted graph instance). So this may not be a bug. Maybe it comes from the subtle difference between [4] and networkx.
> 
> ```
> retworkx output: {3: 9, 7: 2, 0: 4, 6: 8, 9: 3, 1: 5, 2: 7, 4: 0, 5: 1, 8: 6}
> = {(0, 4), (2, 7), (5, 1), (6, 8), (9, 3)}
> networkx output: {(0, 4), (2, 7), (9, 5), (8, 6), (1, 3)}
> edge list: [(1, 0), (3, 1), (3, 2), (4, 0), (4, 1), (4, 3), (5, 0), (5, 1), (5, 2), (6, 1), (7, 0), (7, 1), (7, 2), (7, 4), (7, 5), (8, 1), (8, 3), (8, 4), (8, 6), (8, 7), (9, 0), (9, 1), (9, 2), (9, 3), (9, 4), (9, 5), (9, 6), (9, 7)]
> ```

Oh, that's interesting I hadn't considered that. I didn't really look closely at it yet and just assumed that retworkx was generating an incorrect matching. I wonder how we should proceed here then?

src/max_weight_matching.rs
```diff
@@ -0,0 +1,1309 @@
...
+            } else {
+                j_index = j as usize;
+            }
+        }
+    }
+    // Recycle the blossom number.
+    labels[blossom] = None;
+    label_ends[blossom] = None;
+    blossom_children[blossom].clear();
+    blossom_endpoints[blossom].clear();
```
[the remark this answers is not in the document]
mtreinish: Yep, good catch. Fixed in: https://github.com/Qiskit/retworkx/pull/229/commits/59bbda2ea4a565ef506dc49cf8c821ac41a74947

src/max_weight_matching.rs
```diff
@@ -0,0 +1,1309 @@
...
+    blossom_children: &[Vec<usize>],
+    blossom_base: &[Option<usize>],
+    endpoints: &[usize],
+    mate: &[Option<usize>],
+) -> PyResult<()> {
+    let b = in_blossoms[w];
+    assert!(labels[w] == Some(0) && labels[b] == Some(0));
+    labels[w] = Some(t);
+    labels[b] = Some(t);
+    label_ends[b] = p;
```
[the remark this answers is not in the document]
mtreinish: Yep, fixed in: https://github.com/Qiskit/retworkx/pull/229/commits/59bbda2ea4a565ef506dc49cf8c821ac41a74947

itoko:
How about checking the equality of the optimal values and the validity of the matched edges (those would be sufficient to check the correctness of the solution) in the cases when the matchings from retworkx and networkx are not equal in the test? Also it would be nice to add a similar randomized test with random edge weights in order to cover cases for weighted graphs.

tests/test_max_weight_matching.py
```diff
@@ -0,0 +1,423 @@
...
+            rx_graph, max_cardinality=True)
+        for (u, v) in rx_matches:
+            if (u, v) not in nx_matches:
+                if (v, u) not in nx_matches:
+                    self.fail("seed %s failed. Element %s and it's "
+                              " reverse %s not found "
+                              "in networkx output.\nretworkx output: %s"
+                              "\nnetworkx output: %s\nedge list: %s" % (
+                                  42, (u, v), (v, u), rx_matches,
+                                  nx_matches, list(rx_graph.edge_list())))
```
t-imamichi, line 423: It might be good to make a method of this part of code because it is used in other unit tests.

mtreinish: Yep, that's a good call. I did this as part of: https://github.com/Qiskit/retworkx/pull/229/commits/cfd9121f0de4a282943adba878f3719c0376c5ba

mtreinish:
Ok, sorry for the delay I've added a fallback check for the cases where networkx and retworkx return differing matchings to verify the retworkx matching is still correct (I hope I implemented it correctly). I also added a test that assigns a random integer weight to each edge.

mtreinish:
As an aside, these tests are the slowest in the suite (taking 3-4 seconds) so I was curious where the bottleneck was and ran a profile of the test class. The slowdown is from running networkx's `max_weight_matching()`, but it actually gives us a good performance comparison against the retworkx version of the function being added here:

| library | cumulative time (sec) | number of calls | time per call (sec) |
| -------- | --------------------- | -------------------- | -----------------|
| retworkx | 0.1376 | 5150 | 2.672e-05 |
| networkx | 7.146 | 5123 | 0.001395 |

These numbers are over the entire test class including the test methods which do not call networkx (also they have cProfile overhead which will be a bit worse for networkx because it has a deeper python call stack).

tests/test_max_weight_matching.py
```diff
@@ -0,0 +1,552 @@
...
+def match_dict_to_set(match):
+    return {(u, v) for (u, v) in set(map(frozenset, match.items()))}
+
+
+def is_matching(matching):
+    return all(
+        len(set(e1) & set(e2)) == 0 for e1, e2 in combinations(matching, 2))
+
+
+def is_maximal_matching(graph, rx_matches):
```
t-imamichi, line 33: `is_matching` and `is_maximal_matching` might be useful if they are included as part of retworkx. What do you think?

mtreinish: Yeah I agree, it'll be good thing to add as a retworkx function in a follow up PR (this one is already big enough :) ). I've opened an issue on this here: #255.

itoko (approved):
Thank you for updating the tests! LGTM

t-imamichi (approved):
LGTM. Great job!

PR #232, opened 2021-01-19
Use full LTO for release builds

mtreinish:
This commit changes the release build configuration to use full LTO for
retworkx release builds. The tradeoff this makes is compile time vs
optimization. It basically serializes the compilation of retworkx's
dependencies so that it can do link time optimization. From benchmarking
the performance gains are modest (in the best case ~2-5%), but it
still seems worth it since most users don't constantly rebuild retworkx
and just use the prebuilt binaries that are included in the wheels on
PyPI so tuning the release builds at the cost of compile time seems
worth it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,3 +35,7 @@ features = ["rayon"]
 [dependencies.ndarray]
 version = "^0.14.0"
 features = ["rayon"]
+
+[profile.release]
+lto = 'fat'
+codegen-units = 1


mtreinish:
Looking at CI jobs this adds roughly a 2x slowdown in compile time, but I think that's fine for a best case ~2-5% performance gain. Aside from CI and local development use cases with `pip install` (which includes `tox`) we `cargo build --release` should be pretty uncommon, especially weighed against the number of downloads from PyPI.

PR #235, opened 2021-01-21
Fix read the docs builds by updating rust compiler version

mtreinish:
The read the docs builds are using conda to install rust and other
system dependencies required to install retworkx and build the docs.
However, the version of rust was pinned at 1.40 which is now too old
after #224 merged which bumped the minimum supported rust version from
1.39.0 to 1.45.0 because of a newer version of pyo3. This commit updates
the read the docs conda env to use the latest stable rust release 1.49.0
so that we can build retworkx for docs builds again.

### DIFF
--- a/.conda-rtd-environment.yml
+++ b/.conda-rtd-environment.yml
@@ -3,7 +3,7 @@ channels:
   - defaults
   - conda-forge
 dependencies:
-  - rust=1.40.0=1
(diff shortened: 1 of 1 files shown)


mtreinish:
This fixes the builds: https://readthedocs.org/projects/retworkx/builds/12822350/

PR #236, opened 2021-01-21
Add APIs to update an edge's data payload inplace

mtreinish:
This commit adds 2 new methods to the graph classes, update_edge and
update_edge_by_index, which are used to update the data payload/weight
of an edge in the graph in place. Prior to this the only way to update
an edge would be to remove it and add a new edge with the new data
payload which may change the edge index (assuming something was using
it).

### DIFF
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -612,7 +612,7 @@ impl PyDiGraph {
     ///
     /// If there are parallel edges in the graph only one edge will be updated.
     /// if you need to update a specific edge or need to ensure all parallel
-    /// edges get update you should use
+    /// edges get updated you should use
     /// :meth:`~retworkx.PyDiGraph.update_edge_by_index` instead.
     ///
     /// :param int source: The index for the first node
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -608,6 +608,60 @@ impl PyDiGraph {
         Ok(data)
     }
 
+    /// Update an edge's weight/payload inplace
+    ///
+    /// If there are parallel edges in the graph only one edge will be updated.
+    /// if you need to update a specific edge or need to ensure all parallel
+    /// edges get update you should use
+    /// :meth:`~retworkx.PyDiGraph.update_edge_by_index` instead.
+    ///
+    /// :param int source: The index for the first node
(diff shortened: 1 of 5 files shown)


src/digraph.rs
```diff
@@ -608,6 +608,60 @@ impl PyDiGraph {
         Ok(data)
     }
 
+    /// Update an edge's weight/payload inplace
+    ///
+    /// If there are parallel edges in the graph only one edge will be updated.
+    /// if you need to update a specific edge or need to ensure all parallel
+    /// edges get update you should use
```
lcapelluto, line 615:
```suggestion
    /// edges get updated you should use
```

src/graph.rs
```diff
@@ -413,6 +413,60 @@ impl PyGraph {
         Ok(data)
     }
 
+    /// Update an edge's weight/payload in place
+    ///
+    /// If there are parallel edges in the graph only one edge will be updated.
+    /// if you need to update a specific edge or need to ensure all parallel
+    /// edges get update you should use
```
lcapelluto, line 420:
```suggestion
    /// edges get updated you should use
```

tests/graph/test_edges.py
```diff
@@ -42,6 +42,35 @@ def test_no_edge(self):
...
+        node_b = graph.add_node('b')
+        self.assertRaises(retworkx.NoEdgeBetweenNodes, graph.update_edge,
+                          node_a, node_b, None)
+
+    def test_update_edge_by_index(self):
+        graph = retworkx.PyGraph()
+        node_a = graph.add_node('a')
+        node_b = graph.add_node('b')
+        edge_index = graph.add_edge(node_a, node_b, 'not edgy')
+        graph.update_edge_by_index(edge_index, 'Edgy')
```
lcapelluto, line 65: It may be worth having a test for `by_index` that has multiple edges to choose from? Esp multiple edges between the same nodes

mtreinish: Ooh, yep that's a good idea, I added a test in: https://github.com/Qiskit/retworkx/pull/236/commits/de9289040f176942cb220acd806e926e8930a081

lcapelluto (approved):

PR #241, opened 2021-01-25
Add universal methods

### LINKED ISSUE #215: Make API uniform

mtreinish:
In the rust generated Python API we need to have fixed class inpurts to
satisfy the traits used by the pyo3 macro generated FFI functions. This
results in duplicate methods like digraph_dfs_edges and graph_dfs_edges
with the same implementation just differing input types. To simplify the
API for users this commit adds universal functions to the python side of
the retworkx package to take in any retworkx graph object and dispatch to
the proper function in the rust generated api that relies on strict
input types.

Fixes #215

- [x] Add release notes

### DIFF
--- a/retworkx/__init__.py
+++ b/retworkx/__init__.py
@@ -7,6 +7,7 @@
 # that they have been altered from the originals.
 
 import sys
+import functools
 
 from .retworkx import *
 sys.modules['retworkx.generators'] = generators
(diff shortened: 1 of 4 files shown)


retworkx/__init__.py
```diff
@@ -81,3 +82,356 @@ class PyDAG(PyDiGraph):
...
+        output distance matrix.
+
+    :returns: The distance matrix
+    :rtype: numpy.ndarray
+    """
+    raise TypeError("Invalid Input Type %s for graph" % type(graph))
+
+
+@distance_matrix.register(PyDiGraph)
+def _digraph_distance_matrix(graph, parallel_threshold=300):
```
lcapelluto, line 116: I'm guessing this matches the default for `digraph_distance_matrix`?

lcapelluto (approved):
This is so awesome!! 🎉 
The only "oof" I have is: if you have to fix a typo in a docstring, or if you want to change a default value, you have to change it now in _three_ places 😬

retworkx/__init__.py
mtreinish: Yeah: https://github.com/Qiskit/retworkx/blob/master/src/lib.rs#L1285-L1287

mtreinish: but that does point out I'm missing `as_undirected` here. I'll fix that now

mtreinish:
> This is so awesome!! tada
> The only "oof" I have is: if you have to fix a typo in a docstring, or if you want to change a default value, you have to change it now in _three_ places grimacing

Yeah I couldn't figure out a way around that. :( We still need to export the per type functions both so they're accessible for the dispatch functions and for backwards compat. If we're going to be exporting the functions we'll want them to be documented.

retworkx/__init__.py
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/241/commits/fa90541e841f0c9957ec2e6783fb0d77c1529da2

PR #243, opened 2021-01-27
Add constructor to build graphs from adjacency matrices

mtreinish:
This commit adds a new constructor method from_adjacency_matrix() to
PyGraph and PyDiGraph classes. This enables creating a new graph from an
input adjacency matrix.

### DIFF
--- a/tests/test_adjacency_matrix.py
+++ b/tests/test_adjacency_matrix.py
@@ -118,7 +118,7 @@ class TestDAGAdjacencyMatrix(unittest.TestCase):
             [[0, 1, 0], [1, 0, 1], [0, 1, 0]],
             dtype=np.int64)
         with self.assertRaises(TypeError):
-            graph = retworkx.PyDiGraph.from_adjacency_matrix(input_matrix)
+            retworkx.PyDiGraph.from_adjacency_matrix(input_matrix)
 
     def test_random_graph_float_dtype(self):
         input_matrix = np.array(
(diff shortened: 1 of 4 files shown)


tests/test_adjacency_matrix.py
```diff
@@ -97,6 +97,37 @@ def test_digraph_with_index_holes(self):
...
+        adjacency_matrix = retworkx.digraph_adjacency_matrix(graph)
+        new_graph = retworkx.PyDiGraph.from_adjacency_matrix(adjacency_matrix)
+        new_adjacency_matrix = retworkx.digraph_adjacency_matrix(new_graph)
+        self.assertTrue(np.array_equal(adjacency_matrix,
+                                       new_adjacency_matrix))
+
+    def test_random_graph_different_dtype(self):
+        input_matrix = np.array(
+            [[0, 1, 0], [1, 0, 1], [0, 1, 0]],
+            dtype=np.int64)
```
lcapelluto, line 119: Why not ints?

mtreinish: It's a limitation with the python <-> rust interface. To take in a numpy array directly without any data copies or conversions we need to specify a data type for that input numpy array. That way pyo3 can construct the ffi function to python's C interface and handle the type conversion. When I tried to use a generic like:

```rust
 pub fn from_adjacency_matrix<'p, T>(
        py: Python<'p>,
        matrix: PyReadonlyArray2<'p, T>,
    ) -> PyDiGraph 
    where
        T: Into<f64>,
    {
```
it raised an error:

```
error: A Python method can't have a generic type parameter
    --> src/digraph.rs:1738:38
     |
1738 |     pub fn from_adjacency_matrix<'p, T>(
     |    
```
This also kind of makes sense (at least to me) since we're accessing the memory from numpy/python's heap directly in a read only fashion so we need to know the data type for the rust side reference to be able to access things correctly. There may be a way to handle this but I wanted to avoid a copy or working with a python object in rust and this seemed like an ok tradeoff. Especially since on the calling python side you can just do:

```python
graph = PyDiGraph.from_adjacency_matrix(input_matrix.astype(np.float64, copy=False))
```

I specifically picked `f64` for the data type because that's what retworkx's adjacency_matrix function returns.

lcapelluto (approved):

PR #250, opened 2021-02-16
Ensure uniqueness of neighbor indices returned by `neighbors()`

### LINKED ISSUE #249: `neighbors()` function should return an iterator producing unique indices
### Information

- **retworkx version**: 0.7.2
- **Python version**: any
- **Rust version**: any
- **Operating system**: any

Function `neighbors()` returns an iterator producing repeated indices in multigraphs.

Function `neighbors()` should return an iterator generating unique indices.

Issue qiskit/qiskit-terra#5530 illustrates the problem.

delapuente:
Fixes #249 

petgraph's underlaying `neighbors()` method returns duplicated
nodes in multigraphs (on a per-edge basis). The desired behavior
for the public API is not to include these repetitions though.

This commit collects the returned indices in a HashSet to remove
duplications, then it exhausts the set with `drain()` and
collect it into a list.

This commit also adds new tests to check there is no duplication
in the list of neighbors indices and adapt the existing tests to
ignore the order.

### DIFF
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -1391,7 +1391,7 @@ impl PyDiGraph {
     ///
     /// :param int node: The index of the node to get the neighbors of
     ///
-    /// :returns: A list of the neighbor node indicies
+    /// :returns: A list of the neighbor node indices
     /// :rtype: NodeIndices
     #[text_signature = "(self, node, /)"]
     pub fn neighbors(&self, node: usize) -> NodeIndices {
@@ -1400,6 +1400,8 @@ impl PyDiGraph {
                 .graph
                 .neighbors(NodeIndex::new(node))
                 .map(|node| node.index())
+                .collect::<HashSet<usize>>()
+                .drain()
                 .collect(),
         }
     }
--- a/releasenotes/notes/neighbors-fn-returns-unique-indices-00109eda7e3c9ee7.yaml
+++ b/releasenotes/notes/neighbors-fn-returns-unique-indices-00109eda7e3c9ee7.yaml
@@ -1,0 +1,11 @@
+---
+upgrade:
+  - |
+    If you are using the method :meth:`~retworkx.PyDiGraph.neighbors`, notice
+    the returned iterator now removes repeated node indices. See
+    `#250 <https://github.com/Qiskit/retworkx/issues/250>`__ for more details.
+fixes:
+  - |
+    The iterator returned by the method :meth:`~retworkx.PyDiGraph.neighbors`
+    now goes throw all the node indices without repetition. See
+    `#250 <https://github.com/Qiskit/retworkx/issues/250>`__ for more details.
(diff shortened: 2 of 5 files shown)


mtreinish (commented):
+1 on the approach, the rust side changes lgtm. The only thing missing is a release note. If you get a sec can you add a reno release note (see https://github.com/Qiskit/retworkx/blob/master/CONTRIBUTING.md#release-notes for instructions on that)? If not I can follow up and add one as prep for the 0.8.0 release, so not blocking.

The only thing that needs to be fixed here before we can merge it is the undirected graph tests in `tests/graph/test_neighbors.py` need to be updated to not be sensitive to ordering (like what you already did for the digraph). Once that's fixed this should be good to go.

src/digraph.rs
```diff
@@ -1400,6 +1400,8 @@ impl PyDiGraph {
                 .graph
                 .neighbors(NodeIndex::new(node))
                 .map(|node| node.index())
+                .collect::<HashSet<usize>>()
+                .drain()
```
mtreinish, line 1404: One thing I was just wondering here is whether using a set here is faster or just sorting and calling dedup on the sorted vec is faster. something like:

```rust
let nodes: Vec<usize> =  self.graph.neighbors(NodeIndex::new(node)).map(|node| node.index()).collect();
nodes.sort_unstable();
nodes.dedup();
NodeIndices {
    nodes,
};
```
(we also have rayon's parallel sort functions too which can be faster depending on how large the input is, so probably not useful here)

Let me benchmark it quickly and see if there is a difference.

delapuente: I was looking at the `dedup()` option too. In the end, it was [this answer in Stack Overflow](https://stackoverflow.com/questions/47636618/vecdedup-does-not-work-how-do-i-deduplicate-a-vector-of-strings#answer-47648303) and the fact we were [already using it in the codebase for similar purposes](https://github.com/Qiskit/retworkx/blob/master/src/digraph.rs#L547) that pushed me in the HashSet direction.

mtreinish: Yeah, that makes sense. It's also consistent with the quick benchmarks I ran to test it too:

```python
import retworkx
import time

graph = retworkx.directed_gnp_random_graph(1024, .95, seed=42)
graph.add_edges_from_no_data(list(graph.edge_list()))
start = time.time()
graph.neighbors(512)
stop = time.time()
print("rand time: %s" % str(stop - start))

star = retworkx.generators.directed_star_graph(1024)
star.add_edges_from_no_data(list(star.edge_list()))
start = time.time()
star.neighbors(0)
stop = time.time()
print("star time: %s" % str(stop - start))
```

which returned:
```
Master:
rand time: 3.0517578125e-05
star time: 5.7220458984375e-06

This PR:
rand time: 3.3855438232421875e-05
star time: 3.4332275390625e-05

Vec dedup:
rand time: 4.9114227294921875e-05
star time: 4.029273986816406e-05

Vec parallel sort dedup:
rand time: 4.029273986816406e-05
star time: 4.2438507080078125e-05
```

either way it's not going to huge difference, but this is definitely the better way to handle it for our use case.

mtreinish (approved):

PR #258, opened 2021-02-25
Add GC support to graph objects

mtreinish:
Right now the 2 graph classes in retworkx, PyGraph and PyDiGraph, did
not integrate with the Python garbage collector. This would result in
memory not being cleared in a running python process even if retworkx
objects were only created in temporary context. This commit attempts to
fix this by implementing the garbage collection protocol [1] which
should enable python to clean up retworkx objects when they're no longer
being used.

[1] https://pyo3.rs/v0.12.4/class/protocols.html?highlight=garbage#garbage-collector-integration

### DIFF
--- a/releasenotes/notes/add-gc-integration-87f0d20a92b3e614.yaml
+++ b/releasenotes/notes/add-gc-integration-87f0d20a92b3e614.yaml
@@ -7,4 +7,4 @@ fixes:
     To fix this issue, the :class:`~retworkx.PyDiGraph` and
     :class:`~retworkx.PyGraph` classes now are integrated with Python's garbage
     collector so they'll properly be cleared when there are no more references
-    to an graph object.
+    to a graph object.
--- a/releasenotes/notes/add-gc-integration-87f0d20a92b3e614.yaml
+++ b/releasenotes/notes/add-gc-integration-87f0d20a92b3e614.yaml
@@ -1,0 +1,10 @@
+---
+fixes:
+  - |
+    The :class:`~retworkx.PyDiGraph` and :class:`~retworkx.PyGraph` classes
+    now are integrated with Python's garbage collector so they'll properly
+    be cleared when there are no more Python references to an object.
+    In previous releases the Python garbage collector did not know how to
+    interact with :class:`~retworkx.PyDiGraph` or :class:`~retworkx.PyGraph`
+    objects and as a result they would never be freed until Python exited.
+    This has been fixed by adding the garbage collection integration.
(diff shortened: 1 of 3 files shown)


mtreinish:
I'm marking this on hold because I want to do some performance benchmarks on this before we merge it. But this will be a release blocker because it's causing a memory leak in qiskit-terra.

mtreinish:
Removing the on hold because I think there really isn't any performance impact here at least from the retworkx side. I ran the transpile benchmark in https://github.com/mtreinish/retworkx-bench and there was no real difference and performance.

src/digraph.rs
```diff
@@ -2167,3 +2169,29 @@ fn weight_transform_callable(
         None => Ok(value.clone_ref(py)),
     }
 }
+
+#[pyproto]
```
lcapelluto, line 2173: Maybe a short docstring, so it's obvious that this is for garbage collection?

mtreinish: I added some comments for this in: https://github.com/Qiskit/retworkx/pull/258/commits/32f7c64d00ea6e59abc324666d3a3adee7d2009a (I didn't use docstrngs ,`///` because these don't get really exposed to the user so we don't need to include it in the exported objects)

itoko:
LGTM as a fix for `PyGraph` and `PyDiGraph`, but don't we have any other classes to implement `PyGCProtocol` trait (maybe a separate PR)? What about `WeightedEdgeList` and `BFSSuccessors`? (I've grepped "PyObject" and "pub", and found these two.)

mtreinish:
> LGTM as a fix for `PyGraph` and `PyDiGraph`, but I wonder if we might have other classes that should implement `PyGCProtocol` trait. What about `WeightedEdgeList` and `BFSSuccessors`? (I've grepped "PyObject" and "pub", and found these two.)
> ==> Oops, sorry. I've just found you added them in #260

Yep, I did it for the custom return types separately in #260 because I had been wanting `__str__` for those return types for some time, so I handled both at the same time there for all 4 of the custom return types and kept it separate from the graph classes.

releasenotes/notes/add-gc-integration-87f0d20a92b3e614.yaml
```diff
@@ -0,0 +1,10 @@
+---
+fixes:
+  - |
+    In previous releases the Python garbage collector did not know how to
+    interact with :class:`~retworkx.PyDiGraph` or :class:`~retworkx.PyGraph`
+    objects and as a result they would never be freed until Python exited.
+    To fix this issue, the :class:`~retworkx.PyDiGraph` and
+    :class:`~retworkx.PyGraph` classes now are integrated with Python's garbage
+    collector so they'll properly be cleared when there are no more references
+    to an graph object.
```
kdk, line 10:
```suggestion
    to a graph object.
```

kdk (approved):

mtreinish:
It looks like there is an infrastructure issue with travis CI on ppc64le nodes (the logs just stop without error) I'm going to merge this

PR #260, opened 2021-02-26
Add more object protocol methods to custom return types

mtreinish:
This commit adds a few more methods from the Python object protocol to the
custom return types. All the classes now have a `__str__` method so you can
see the contents of the custom returns with a `print()`. Additionally, a
`__hash__` is added to each class since the return types are immutable.
However, for return types that contain `PyObject`s (`WeightedEdgeList` and
`BFSSuccessors`) if any of the objects contained are not hashable this
will raise an exception (just as if you called `hash()` on that object).
The last methods are related to garbage collection in python which
enables the Python garbage collector to cleanup the custom return types.

### DIFF
--- a/releasenotes/notes/return-class-object-methods-d8b19d952639118a.yaml
+++ b/releasenotes/notes/return-class-object-methods-d8b19d952639118a.yaml
@@ -10,7 +10,7 @@ features:
   - |
     The custom return types :class:`~retworkx.BFSSuccessors`,
     :class:`~retworkx.NodeIndices`, :class:`~retworkx.EdgeList`, and
-    :class:`~retworkx.WeightedEdgeList` now implement ``_str__`` so that
+    :class:`~retworkx.WeightedEdgeList` now implement ``__hash__`` so that
     running ``hash()`` (for example when insert them into a ``dict``) will
     return a valid hash for the object. The only exception to this is for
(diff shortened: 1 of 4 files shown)


releasenotes/notes/return-class-object-methods-d8b19d952639118a.yaml
```diff
@@ -0,0 +1,29 @@
...
+    The custom return types :class:`~retworkx.BFSSuccessors`,
+    :class:`~retworkx.NodeIndices`, :class:`~retworkx.EdgeList`, and
+    :class:`~retworkx.WeightedEdgeList` now implement ``_str__`` so that
+    running ``str()`` (for example when calling ``print()`` on the object) it
+    will return a human readable string with the contents of the custom return
+    type.
+  - |
+    The custom return types :class:`~retworkx.BFSSuccessors`,
+    :class:`~retworkx.NodeIndices`, :class:`~retworkx.EdgeList`, and
+    :class:`~retworkx.WeightedEdgeList` now implement ``_str__`` so that
```
itoko, line 13:
```suggestion
    :class:`~retworkx.WeightedEdgeList` now implement ``__hash__`` so that
```

itoko (commented):
LGTM (except for a typo).

kdk (approved):

PR #262, opened 2021-03-02
Fix CI wheel jobs

mtreinish:
For the recent 0.8.0 release all the wheel jobs failed because we added
networkx as a test requirement (for comparing the outputs of
max_weight_matching()) but were not installing networkx before running
unit tests. This made all the wheels come back as broken because
unittests failed. This commit fixes that issue by adding networkx as an
explicit test requirement in all the cibuildwheel job configurations.

Additionally, the ppc64le wheel jobs started failed for an additional
reason, it appears that the travis CI images aren't up to date and need
the rust compiler to install cryptography (which is required for pip and
other packages). This commit fixes those jobs so we can install
cibuildwheel and also build the packages.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -1,9 +1,11 @@
 ---
 name: Wheel Builds
 on:
-  push:
-    tags:
-      - '*'
+  pull_request:
+    branches: [master]
+#  push:
+#    tags:
+#      - '*'
 jobs:
   sdist:
     name: Build sdist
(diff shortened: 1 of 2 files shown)


mtreinish:
Still working though the ppc64le jobs which are proving to be problematic. After I get that working I'll update this PR with the pattern that works. We'll also need to push a DNM commit to the PR branch to trigger a run (without uploads) to verify all the jobs work after this PR so we know that the next release will work.

mtreinish:
All the wheel jobs appear to work now I'm going to revert: https://github.com/Qiskit/retworkx/pull/262/commits/753a739f55b42176a50fe5f359d904c20008e843 and merge this.

PR #264, opened 2021-03-08
add transitivity function

### LINKED ISSUE #192: Add transitivity function

Add functions to get the transitivity of a graph. Similar to networkx's `transitivity` method: https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.cluster.transitivity.html or graph-tool's `global_clustering`: https://graph-tool.skewed.de/static/doc/clustering.html?highlight=global_clustering#graph_tool.clustering.global_clustering

georgios-ts:
Resolves #192.

For directed graphs i used the definition of local clustering coefficient in Eq.8 of [1] and define the transitivity coefficient as the sum of all numerators divided by the sum of all denominators.

[1]: https://arxiv.org/abs/physics/0612169


src/lib.rs
```diff
@@ -2753,6 +2753,155 @@ pub fn is_maximal_matching(
...
+///
+/// :returns: Transitivity.
+/// :rtype: float
+#[pyfunction]
+#[text_signature = "(graph, /)"]
+fn graph_transitivity(graph: &graph::PyGraph) -> f64 {
+    let (triangles, triples) =
+        graph
+            .graph
+            .node_indices()
```
mtreinish, line 2801: One thought I just had here is there any advantage to doing this iteration in parallel? It doesn't look like there is any dependence on order (as it's all addition) and there is no interaction with python so we don't need to worry about the gil. To make this multithreaded we can just create a Vec and use rayon's parallel iterator here, so it would be something like:

```suggestion
    let node_indices: Vec<NodeIndex> = graph.graph.node_indices();
    let (triangles, triples) = node_indices
            .par_iter()
```

I'm not sure if there will be a benefit to running this in parallel or not, but my feeling is that it would be faster (although I'm often wrong) so it's probably worth benchmarking to checking.

IvanIsCoding (commented):
LGTM, overall clean implementation from the transivity definition! I have two questions to make though.

Multigraphs: the function does not work with multigraphs (and maybe the transivity definition doesn't make sense for multigraphs too). Should we accept multigraphs as an input? Currently we do, and we return a number that is not correct. Do you think we should explicitly raise an error?

Sparse Graphs and Time Complexity: at least for the undirected case, an algorithm with better time complexity exists for graphs when the number of edges is small (see Theorem 3.1 in https://theory.stanford.edu/~tim/s14/l/l1.pdf). Did you rule out implementing an algorithm with a better time complexity?

georgios-ts:
Thanks @IvanIsCoding for the review!

Since we might have `multigraph=true` but no parallel edges, we would have to pay an overhead to detect them and raise an error. So i decided to let the user "decide" if the function makes sense for the input graph. If you feel that we should instead raise an error, i will add it.

Great suggestion! I will implement this faster algorithm.

IvanIsCoding:
> Thanks @IvanIsCoding for the review!
> 
> Since we might have `multigraph=true` but no parallel edges, we would have to pay an overhead to detect them and raise an error. So i decided to let the user "decide" if the function makes sense for the input graph. If you feel that we should instead raise an error, i will add it.
> 
> Great suggestion! I will implement this faster algorithm.

Fair point, I think someone could use the function on  a graph with `mutligraph=true` that is not a true multigraph. Honestly, I think you can just add a note in the docs that we assume that the graph has no parallel edges or self-loops. That should be enough.

I am approving it, we just need to change documentation. The improved algorithm can come on a later PR.

IvanIsCoding (approved):

src/lib.rs
georgios-ts: Good suggestion. I did some benchmarks locally and it's indeed faster

georgios-ts:
After some thought, I am not sure there is a benefit to implement the improved algorithm decribed [here]() since it's assumed constant-time edge queries. But we have an adjacency list representation of the graph and we would have to pay an O(n^2) overhead to build the adjacency matrix (so will lose the runtime benefit e.g in star graph that the paper describes).

Another argument regarding performance that one could make is that my current implementation takes O(n^3 logn) for complete graphs. We can reach the O(n^3) optimal with extra O(n) memory for each node.  The downside is the extra memory even for sparse graphs. 

To conclude, i prefer to leave it as it is. especially, after the parallel version i am ok with the performance.

IvanIsCoding:
> After some thought, I am not sure there is a benefit to implement the improved algorithm decribed [here](https://theory.stanford.edu/~tim/s14/l/l1.pdf) since it's assumed constant-time edge queries. But we only have an adjacency list representation of the graph and we would have to pay an O(n^2) overhead to build the adjacency matrix (so will lose the runtime benefit e.g in star graph that the paper describes).
> 
> Another argument regarding performance that one could make is that my current implementation takes O(n^3 logn) for complete graphs. We can reach the O(n^3) optimal with extra O(n) memory for each node. The downside is the extra memory even for sparse graphs.
> 
> To conclude, i prefer to leave it as it is. especially, after the parallel version i am ok with the performance.

We can build a structure to have O(1) look up for edges in O(|E|). Indeed, you're already doing it in your code without knowing!

```
let mut neighbors: HashSet<NodeIndex> =
        graph.graph.neighbors(index).collect();
    neighbors.remove(&index);
```

You're doing this for each node in your inner function. If you group them together you'd have a hashet for each adjacency list, which let's you query in O(1).

Anyway, I'd merge the code with the parallel iteration as is and refine the algorithm later if needed.

src/lib.rs
```diff
@@ -2777,6 +2777,161 @@ pub fn is_maximal_matching(
...
+/// The transitivity of a directed graph is defined in [Fag]_, Eq.8:
+///
+/// .. math::
+///     `c=3 \times \frac{\text{number of triangles}}{\text{number of all possible triangles}}`
+///
+/// A triangle is a connected triple of nodes.
+/// Different edge orientations counts as different triangles.
+///
+/// It's worh noting that the function implicitly assumes that there is
+/// no parallel edges or self loops.
```
mtreinish, line 2908:
```suggestion
/// .. note::
///
///     The function implicitly assumes that there are no parallel edges
///     or self loops.it may produce incorrect/unexpected results if the
///     input graph has self loops or parallel edges.
```

src/lib.rs
```diff
@@ -2777,6 +2777,161 @@ pub fn is_maximal_matching(
...
+/// The transitivity of a graph is defined as:
+///
+/// .. math::
+///     `c=3 \times \frac{\text{number of triangles}}{\text{number of connected triples}}`
+///
+/// A “connected triple” means a single vertex with
+/// edges running to an unordered pair of others.
+///
+/// It's worh noting that the function implicitly assumes that there is
+/// no parallel edges or self loops.
```
mtreinish, line 2816:
```suggestion
/// .. note::
///
///     The function implicitly assumes that there are no parallel edges
///     or self loops.it may produce incorrect/unexpected results if the
///     input graph has self loops or parallel edges.
```

retworkx/__init__.py
```diff
@@ -525,3 +525,30 @@ def _digraph_is_isomorphic_node_match(first, second, matcher):
...
 def _graph_is_isomorphic_node_match(first, second, matcher):
     return graph_is_isomorphic(first, second, matcher)
+
+
+@functools.singledispatch
+def transitivity(graph):
+    """Compute the transitivity of a graph.
+
+    It's worh noting that the function implicitly assumes that there is
+    no parallel edges or self loops.
```
mtreinish, line 535:
```suggestion
/// .. note::
///
///     The function implicitly assumes that there are no parallel edges
///     or self loops.it may produce incorrect/unexpected results if the
///     input graph has self loops or parallel edges.
```

src/lib.rs
```diff
@@ -2777,6 +2777,161 @@ pub fn is_maximal_matching(
...
+
+/// Compute the transitivity of an undirected graph.
+///
+/// The transitivity of a graph is defined as:
+///
+/// .. math::
+///     `c=3 \times \frac{\text{number of triangles}}{\text{number of connected triples}}`
+///
+/// A “connected triple” means a single vertex with
+/// edges running to an unordered pair of others.
```
mtreinish, line 2813: I think we also need to document something about the multithreaded nature of this function. I would add something here about the thread pool usage in rayon and how to adjust it. Basically just copy and adapt what I used here:

https://github.com/Qiskit/retworkx/pull/278/files#diff-b8d56121fa60f981d2a35720517aadf6cd87bd0ade23585819d1f84b3262eb54R343-R347

src/lib.rs
```diff
@@ -2777,6 +2777,161 @@ pub fn is_maximal_matching(
...
+    let (triangles, triples) = node_indices
+        .par_iter()
+        .map(|node| _digraph_triangles(graph, node.index()))
+        .reduce(
+            || (0, 0),
+            |(sumx, sumy), (resx, resy)| (sumx + resx, sumy + resy),
+        );
+
+    match triangles {
+        0 => 0.0,
```
mtreinish, line 2930: Coverage reporting showed that there wasn't any coverage for this line: https://coveralls.io/builds/38209715/source?filename=src%2Flib.rs#L2930

Not critical because it's hard to imagine this line misbehaving, but it might be good to add an empty graph test case for digraph too so we execute this line too.

mtreinish (commented):
The code here LGTM, great work. Just some small documentation comments inline, after hose are addressed I think this will be ready to go.

mtreinish (approved):
LGTM, thanks for updating.

PR #267, opened 2021-03-09
improve vf2 to handle StableGraph node index holes

### LINKED ISSUE #125: Reimplement VF2 to handle StableGraph node index holes

The VF2 implementation in `dag_isomorphism.rs` was forked from the upstream petgraph project and modified to handle `PyDiGraph`. As was previously reported in #27 and worked around in #115 that implementation didn't work with the petgraph `StableGraph` type used in retworkx on node removals because there could be holes in the list of node ids (so that `node_bound()` on 2 isomorphic graphs aren't the same) which isn't possible in the petgraph `Graph` type but is in `StableGraph`. This was causing a hard failure in the isomorphism functions if a node was removed from a `PyDiGraph` obejct and was worked around in #115. However, this workaround introduces overhead in the case of node removals because it is cloning and reindexing the graph to not have holes in the index list. The next step here is to reimplement the vf2 algorithm to natively handle holes in the node index so that we don't need this cloning and reindexing step.

It also would probably be good to push an improved vf2 back upstream to petgraph after it is implemented here. We'll still need the local fork in retworkx, but if we fix VF2 support for StableGraph we should make sure to contribute that back to petgraph too.

georgios-ts:
Resolves #125.
Improves `Vf2State` to use `HashMap` instead of `Vec` that was assuming contiguous node indices.
As it turns out, `is_adjacent` implementation of `StableGraph` has a bug in petgraph (in [L69](https://github.com/petgraph/petgraph/blob/master/src/traits_graph.rs#L69)). This commit fixes this in retworkx.

### DIFF
--- a/src/dag_isomorphism.rs
+++ b/src/dag_isomorphism.rs
@@ -22,26 +22,24 @@ use super::digraph::PyDiGraph;
 
 use pyo3::prelude::*;
 
-use petgraph::algo;
 use petgraph::stable_graph::NodeIndex;
-use petgraph::stable_graph::StableDiGraph;
-use petgraph::visit::{EdgeRef, GetAdjacencyMatrix, IntoEdgeReferences};
+use petgraph::visit::GetAdjacencyMatrix;
 use petgraph::{Directed, Incoming};
 
 #[derive(Debug)]
 struct Vf2State {
     /// The current mapping M(s) of nodes from G0 → G1 and G1 → G0,
     /// NodeIndex::end() for no mapping.
-    mapping: Vec<NodeIndex>,
+    mapping: HashMap<NodeIndex, NodeIndex>,
     /// out[i] is non-zero if i is in either M_0(s) or Tout_0(s)
     /// These are all the next vertices that are not mapped yet, but
     /// have an outgoing edge from the mapping.
-    out: Vec<usize>,
+    out: HashMap<NodeIndex, usize>,
     /// ins[i] is non-zero if i is in either M_0(s) or Tin_0(s)
     /// These are all the incoming vertices, those not mapped yet, but
     /// have an edge from them into the mapping.
     /// Unused if graph is undirected -- it's identical with out in that case.
-    ins: Vec<usize>,
+    ins: HashMap<NodeIndex, usize>,
     out_size: usize,
     ins_size: usize,
     adjacency_matrix: FixedBitSet,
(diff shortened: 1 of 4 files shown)


mtreinish (commented):
This LGTM thanks for fixing that bug. Two quick thoughts though:

1. We should definitely contribute this fix upstream to petgraph, especially the fix for the `is_adjacent` function. (looking at the history the upstream petgraph vf2 implementation has been updated recently so some this fix may not apply). Not blocking here, but it would be good to follow up and fix this upstream after this PR merges.
2. For the `mapping`, `out`, and `ins` fields of the `Vf2State` struct I'm wondering if a `BTreeMap` would have better performance than the `HashMap`. The rule of thumb is supposed to be if you're primarily doing inserts and/or lookups a hashmap is better, but if you do anything that iterates over it a btree map is more efficient. In this case there is a good mix of both so I'm not sure what would come out ahead (I'm guessing hashmap will still be faster because hashbrown does a good job optimizing for performance, but it's good to check). Before we merge this we should do some quick bechmarks to see what ends up being faster `hashbrown::HashMap` or `std::collections::BTreeMap`. In python it probably would be something like:

```python
import retworkx
import time

g1 = retworkx.directed_gnm_random_graph(100000, 20000, seed=42)
g2 = retworkx.directed_gnm_random_graph(100000, 20000, seed=4242)

start = time.time()
retworkx.is_isomorphic(g1, g2)
stop = time.time()
print(stop - start)

start = time.time()
retworkx.is_isomorphic(g1, g1)
stop = time.time()
print(stop - start)
```

georgios-ts:
Indeed using `BTreeMap` has better perfomance. In some places (`Vf2State.out` and `Vf2State.ins`) we can simplify the logic and use `BTreeSet`. But still, after doing some quick benchmarks, the algorithm gets 3 times slower, probably due to bad memory layout. So i guess we should abandon this PR.

```python
import retworkx
import time

N = 50
M = 20
g1 = retworkx.directed_gnm_random_graph(N, M, seed=42)
g2 = retworkx.directed_gnm_random_graph(N, M, seed=4242)

start = time.time()
res = retworkx.is_isomorphic(g1, g2)
stop = time.time()
print(stop - start)
---
Output:
Before: 19
After: 60
```

As a sidenote, if we care a lot about the performance of vf2 algorithm, there is an improved version [here](http://bolyai.cs.elte.hu/egres/tr/egres-18-03.pdf) that we could try to implement (the new part of the algorithm is in section 4.1).

mtreinish:
Wow, yeah that is surprising. I was expecting that looping over both graphs and re-indexing and copying everything like we do currently would have been slower. I guess that teaches us not to underestimate the importance of memory locality and CPU caching for performance :)

Yeah, I think we probably don't want to take a ~3x runtime performance hit (which matched up almost exactly for me locally too) for cleaner internal code, so I'm going to close this and the corresponding issue now. Please feel free to reopen it if we need to discuss it more.

As for a faster algorithm yeah that sounds like it would be a good project. Can you open an enhancement issue for using vf2++  so we can track that?

PR #272, opened 2021-03-12
Make vf2 generic

### LINKED ISSUE #268: Expand isomorphism functions

After #267 fixes the issues we had the stable graph and vf2 we should expand the isomorphism apis to cover a wider use case. Right now retworkx has 2 functions `is_isomorphism` and `is_isomorphism_node_match` which only works with the `PyDiGraph` class. We should expand these to include functions for `PyGraph` and also for edge matching too. This will require updating the dag isomorphism module to work with petgraph traits instead of fixed type PyDiGraph. 

It might make sense to combine the functions and just make 2 is_isomorphic functions (one for PyGraph the other for PyDiGraph) that have optional kwargs for node and edge matching functions. The one thing we'll need to keep in mind is backwards compatibility because we can't break existing users passing `PyDiGraph` objects to `retworkx.is_isomorphic` and `retworkx.is_isomorphic_node_matching` so if we change the python functions exposed by rust we'll likely need compatibility wrapper functions in the python to adapt existing users to the new way.

georgios-ts:
Resolves #268.
Introduced changes:
- Vf2 implementation in `dag_isomorphism.rs` now works with `StableGraph` so we can pass the graph that `PyGraph` and `PyDiGraph` store.
- Optional `edge_matcher` argument to compare edge data.
- Unified api in a single `is_isomorphic` call.
- Retain `retworkx.is_isomorphic_node_matching` for backwards compatibility.

### DIFF
--- a/releasenotes/notes/expand-isomorphism-cfb646cfef66fa25.yaml
+++ b/releasenotes/notes/expand-isomorphism-cfb646cfef66fa25.yaml
@@ -1,0 +1,6 @@
+---
+features:
+  - |
+    :func:`~retworkx.is_isomorphic` can take in a :class:`~retworkx.PyGraph`
+    or :class:`~retworkx.PyDiGraph` and two functions to compare node 
+    and edge data.
--- a/releasenotes/notes/expand-isomorphism-cfb646cfef66fa25.yaml
+++ b/releasenotes/notes/expand-isomorphism-cfb646cfef66fa25.yaml
@@ -2,5 +2,8 @@
 features:
   - |
     :func:`~retworkx.is_isomorphic` can take in a :class:`~retworkx.PyGraph`
-    or :class:`~retworkx.PyDiGraph` and two functions to compare node 
-    and edge data.
+    or :class:`~retworkx.PyDiGraph`. You can optionally specify two functions
+    to compare node and edge data.
+  - |
+    :func:`~retworkx.is_isomorphic_node_match` can take in a :class:`~retworkx.PyGraph`
+    or :class:`~retworkx.PyDiGraph`.
--- a/src/isomorphism.rs
+++ b/src/isomorphism.rs
@@ -192,8 +192,8 @@ where
     }
     for edge in graph.edge_references() {
         let edge_w = edge.weight();
-        let p_index = id_map.get(&edge.source()).unwrap();
-        let c_index = id_map.get(&edge.target()).unwrap();
+        let p_index = id_map[&edge.source()];
+        let c_index = id_map[&edge.target()];
         new_graph.add_edge(*p_index, *c_index, edge_w.clone_ref(py));
     }
 
--- a/src/isomorphism.rs
+++ b/src/isomorphism.rs
@@ -194,7 +194,7 @@ where
         let edge_w = edge.weight();
         let p_index = id_map[&edge.source()];
         let c_index = id_map[&edge.target()];
-        new_graph.add_edge(*p_index, *c_index, edge_w.clone_ref(py));
+        new_graph.add_edge(p_index, c_index, edge_w.clone_ref(py));
     }
 
     new_graph
--- a/src/isomorphism.rs
+++ b/src/isomorphism.rs
@@ -256,8 +256,8 @@ pub fn is_isomorphic_matching<Ty, F, G>(
     py: Python,
     g0: &StableGraph<PyObject, PyObject, Ty>,
     g1: &StableGraph<PyObject, PyObject, Ty>,
-    mut node_match: F,
-    mut edge_match: G,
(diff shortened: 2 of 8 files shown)


mtreinish (commented):
This LGTM, the code changes to the isomorphism especially look good. The one thing I'm unsure of is the user facing APIs.

Right now the PR has 6 functions we're exposing (4 from rust 2 universal functions from python):

- `digraph_is_isomorphic`
- `graph_is_isomorphic`
- `is_isomorphic`
- `digraph_is_isomorphic_node_matching`
- `graph_is_isomorphic_node_matching`
- `is_isomorphic_node_match`

I'm wondering if we want to simplify this a bit. For backwards compatibility we need to expose:
`is_isomorphic` and `is_isomorphic_node_match`. What if we made the public facing API:

- `is_isomorphic(first, second, node_matcher=None, edge_matcher=None)`
- `digraph_is_isomorphic(first, second, node_matcher=None, edge_matcher=None)`
- `graph_is_isomorphic(first, second, node_matcher=None, edge_matcher=None)`
- `is_isomorphic_node_matcher(first, second, node_matcher)` (this will be in python and just call `is_isomorphic(first, second, node_matcher=node_matcher)`)

This should maintain backwards compatibility so existing users don't need to change anything but reduce the number of functions we need to expose both in rust and python.

Oh, and one other thing can we rename `src/dag_isomorphism.rs` to `src/isomorphism.rs` to reflect it works for everything now, not just DAGs (that name was just a leftover from the early days when retworkx only had a `PyDAG` class).

src/isomorphism.rs
```diff
@@ -175,54 +180,70 @@ impl Vf2State {
...
+    for node_index in graph.node_indices() {
+        let node_data = graph.node_weight(node_index).unwrap();
         let new_index = new_graph.add_node(node_data.clone_ref(py));
         id_map.insert(node_index, new_index);
     }
-    for edge in dag.graph.edge_references() {
+    for edge in graph.edge_references() {
         let edge_w = edge.weight();
         let p_index = id_map.get(&edge.source()).unwrap();
         let c_index = id_map.get(&edge.target()).unwrap();
```
mtreinish, line 242: A small cleanup here we can do here while we're updating this code is something like:

```suggestion
        let p_index = id_map[&edge.source()];
        let c_index = id_map[&edge.target()];
```

releasenotes/notes/expand-isomorphism-cfb646cfef66fa25.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    :func:`~retworkx.is_isomorphic` can take in a :class:`~retworkx.PyGraph`
+    or :class:`~retworkx.PyDiGraph` and two functions to compare node 
+    and edge data.
```
mtreinish, line 6: We probably also should mention that `is_isomorphic_node_match` will work with PyGraph too now.
 
I also would add a second release note here mentioning that we take an optional node or edge matcher on is_isomorphic now.

mtreinish (commented):
I took a detailed look through the code and the rust changes look great! Especially you cleaned up a lot of baggage and sloppy coding in the isomorphism module that makes it more sustainable and easier to deal with now. I just have a few quick comments inline. 

Also, there are some coverage gaps that I think would be good to fill with some extra tests to close the coverage gaps:

- Running `graph_is_isomorphic` with an edge matcher
- A test case where `second` has `nodes_removed` set to `True` so we need to reindex the nodes.
- A case where `isomorphism::is_isomorphic_matching` is called where `node_count`s or `edge_counts` aren't equal.
- As mentioned inline adding a negative test to `tests/test_dispatch.py` that verifies a `TypeError` is raised if first and second aren't

src/isomorphism.rs
```diff
@@ -175,54 +180,70 @@ impl Vf2State {
 ///
 /// * Luigi P. Cordella, Pasquale Foggia, Carlo Sansone, Mario Vento;
 ///   *A (Sub)Graph Isomorphism Algorithm for Matching Large Graphs*
-pub fn is_isomorphic(dag0: &PyDiGraph, dag1: &PyDiGraph) -> PyResult<bool> {
-    let g0 = &dag0.graph;
-    let g1 = &dag1.graph;
+pub fn is_isomorphic<Ty>(
```
mtreinish, line 183: Is this function used anywhere anymore? Maybe we should just delete it and rely solely on `is_isomorphic_matching` (and maybe rename it to `is_isomorphic`)

georgios-ts: deleted and renamed.

mtreinish (approved):
LGTM, thanks for the quick updates.

PR #273, opened 2021-03-14
add core_number functions

### LINKED ISSUE #193: Add functions to get k-core of a graph

This is to add new functions to retworkx's algorithm library to get a k-core number dictionary from a graph. This is similar to networkx's `core_number` https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.core.core_number.html  or graph_tool's `kcore_decomposition`: https://graph-tool.skewed.de/static/doc/topology.html?highlight=kcore_decomposition#graph_tool.topology.kcore_decomposition

Chriscrosser3310:
Resolves #193; adds core_number functions for both PyGraph and PyDiGraph. The algorithm takes from [[1](https://arxiv.org/pdf/cs/0310049.pdf)].

Currently, it doesn't support graphs with self-loops or parallel edges. Let me know if it should support these two cases or raise an error.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2729,6 +2729,141 @@ pub fn is_maximal_matching(
     })
 }
 
+/// Return the core number for each node
+///
+/// A k-core is a maximal subgraph that contains nodes of degree k or more.
+///
+/// :param PyGraph: The graph to get core numbers
+///
+/// :returns: A dictionary keyed by node index to the core number
+/// :rtype: dict
+#[pyfunction]
+#[text_signature = "(graph, /)"]
+pub fn graph_core_number(
+    py: Python,
+    graph: &graph::PyGraph,
+) -> PyResult<PyObject> {
+    let mut cores: HashMap<NodeIndex, usize> =
+        HashMap::with_capacity(graph.node_count());
+    let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
+    let mut degree_map: HashMap<NodeIndex, usize> =
+        HashMap::with_capacity(graph.node_count());
+    let mut nbrs: HashMap<NodeIndex, Vec<NodeIndex>> =
+        HashMap::with_capacity(graph.node_count());
+    let mut node_pos: HashMap<NodeIndex, usize> =
+        HashMap::with_capacity(graph.node_count());
+
+    for k in node_vec.iter() {
+        let ki = (*k).index();
+        cores.insert(*k, graph.degree(ki));
+        degree_map.insert(*k, graph.degree(ki));
+        nbrs.insert(*k, graph.graph.neighbors(*k).collect());
+    }
+    node_vec.par_sort_by_key(|k| degree_map.get(k));
+
+    let mut bin_boundaries = vec![0];
+    let mut curr_degree = 0;
+    for (i, v) in node_vec.iter().enumerate() {
+        node_pos.insert(*v, i);
+        let v_degree = degree_map[v];
+        if v_degree > curr_degree {
+            bin_boundaries.extend([i].repeat(v_degree - curr_degree));
+            curr_degree = v_degree;
+        }
+    }
+
+    for v_ind in 0..node_vec.len() - 1 {
+        let v = node_vec[v_ind];
+        let v_nbrs = nbrs[&v].clone();
+        for u in v_nbrs {
+            if cores[&u] > cores[&v] {
+                nbrs.get_mut(&u).unwrap().retain(|&n| n == v);
+                let pos = node_pos[&u];
+                let bin_start = bin_boundaries[cores[&u]];
+                *node_pos.get_mut(&u).unwrap() = bin_start;
+                *node_pos.get_mut(&node_vec[bin_start]).unwrap() = pos;
+                node_vec.swap(bin_start, pos);
+                bin_boundaries[cores[&u]] += 1;
+                *cores.get_mut(&u).unwrap() -= 1;
+            }
+        }
+    }
+
+    let out_dict = PyDict::new(py);
+    for (v_index, core) in cores {
+        out_dict.set_item(v_index.index(), core)?;
+    }
+    Ok(out_dict.into())
+}
+
+/// Return the core number for each node
+///
+/// A k-core is a maximal subgraph that contains nodes of degree k or more.
+/// For directed graphs, the degree is calculated as in_degree + out_degree.
+///
+/// :param PyDiGraph: The directed graph to get core numbers
+///
+/// :returns: A dictionary keyed by node index to the core number
+/// :rtype: dict
+#[pyfunction]
+#[text_signature = "(graph, /)"]
+pub fn digraph_core_number(
+    py: Python,
+    graph: &digraph::PyDiGraph,
+) -> PyResult<PyObject> {
+    let mut cores: HashMap<NodeIndex, usize> =
+        HashMap::with_capacity(graph.node_count());
+    let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
+    let mut degree_map: HashMap<NodeIndex, usize> =
+        HashMap::with_capacity(graph.node_count());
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -2729,6 +2729,145 @@ pub fn is_maximal_matching(
...
+            curr_degree = v_degree;
+        }
+    }
+
+    for v_ind in 0..node_vec.len() {
+        let v = node_vec[v_ind];
+        let v_nbrs = nbrs[&v].clone();
+        for u in v_nbrs {
+            if cores[&u] > cores[&v] {
+                nbrs.get_mut(&u).unwrap().retain(|&n| n == v);
```
georgios-ts, line 2782:
```suggestion
                nbrs.get_mut(&u).unwrap().retain(|&n| n != v);
```

src/lib.rs
```diff
@@ -2729,6 +2729,145 @@ pub fn is_maximal_matching(
...
+    let mut nbrs: HashMap<NodeIndex, Vec<NodeIndex>> =
+        HashMap::with_capacity(graph.node_count());
+    let mut node_pos: HashMap<NodeIndex, usize> =
+        HashMap::with_capacity(graph.node_count());
+
+    for k in node_vec.iter() {
+        let ki = (*k).index();
+        cores.insert(*k, graph.in_degree(ki) + graph.out_degree(ki));
+        degree_map.insert(*k, graph.in_degree(ki) + graph.out_degree(ki));
+        nbrs.insert(*k, graph.graph.neighbors(*k).collect());
```
georgios-ts, line 2830:
```suggestion
       nbrs.insert(*k, graph.graph.neighbors_undirected(*k).collect());
```

src/lib.rs
```diff
@@ -2729,6 +2729,145 @@ pub fn is_maximal_matching(
...
+            curr_degree = v_degree;
+        }
+    }
+
+    for v_ind in 0..node_vec.len() {
+        let v = node_vec[v_ind];
+        let v_nbrs = nbrs[&v].clone();
+        for u in v_nbrs {
+            if cores[&u] > cores[&v] {
+                nbrs.get_mut(&u).unwrap().retain(|&n| n == v);
```
georgios-ts, line 2852:
```suggestion
                nbrs.get_mut(&u).unwrap().retain(|&n| n != v);
```

src/lib.rs
Chriscrosser3310: Thanks! Not sure how the tests worked with this mistake... I guess more tests are needed

georgios-ts:
Great! More of a suggestion here, since the function is literally the same for directed and undirected graphs, to avoid duplicating code you could do something like:

```rust
use petgraph::EdgeType;

pub fn _core_number<Ty>(
    py: Python,
    graph: &StableGraph<PyObject, PyObject, Ty>,
) -> PyResult<PyObject>
where
    Ty: EdgeType,
{
    ... blah blah blah ...

    for k in node_vec.iter() {
        let k_nbrs: Vec<NodeIndex> = graph.neighbors_undirected(*k).collect();
        let k_deg = k_nbrs.len();

        nbrs.insert(*k, k_nbrs);
        cores.insert(*k, k_deg);
        degree_map.insert(*k, k_deg);
    }

     ... blah blah blah ...
}

#[pyfunction]
#[text_signature = "(graph, /)"]
pub fn graph_core_number(
    py: Python,
    graph: &graph::PyGraph,
) -> PyResult<PyObject> {
    _core_number(py, &graph.graph)
}
```
Overall, looks good to me

Chriscrosser3310:
@georgios-ts Thanks for the review! The suggestion is good and I'll adopt it. Besides that, I'll add a few more tests.

georgios-ts (approved):

src/lib.rs
```diff
@@ -2950,6 +2951,117 @@ fn digraph_transitivity(graph: &digraph::PyDiGraph) -> f64 {
...
+        let k_nbrs: Vec<NodeIndex> = graph.neighbors_undirected(*k).collect();
+        let k_deg = k_nbrs.len();
+
+        nbrs.insert(*k, k_nbrs);
+        cores.insert(*k, k_deg);
+        degree_map.insert(*k, k_deg);
+    }
+    node_vec.par_sort_by_key(|k| degree_map.get(k));
+
+    let mut bin_boundaries = vec![0];
```
mtreinish, line 2981: Is there a better estimate we can come up with for the size of this Vec? Right now this will end up with a bunch of reallocations because every time we push onto into we'll have to grow (since it's initial allocation will be for 1 element). Not critical just me idly wondering if there is a possible optimization here.

mtreinish (approved):
LGTM, thanks for handling this issue. I have one inline comment/question but it's not a blocker (and something we can easily go back and update in a follow up PR if we come up with an answer for it).

src/lib.rs
Chriscrosser3310: That's a good point, the size of `bin_boundaries` is precisely `max_degree + 1`. I'll push a quick fix

Chriscrosser3310:
Since now the size of `bin_boundaries` depends on `node_num`, an early return is added to prevent the algorithm from failing from an empty graph. Let me know if there are better ways.

src/lib.rs
```diff
@@ -2950,6 +2951,124 @@ fn digraph_transitivity(graph: &digraph::PyDiGraph) -> f64 {
...
+    graph: &StableGraph<PyObject, PyObject, Ty>,
+) -> PyResult<PyObject>
+where
+    Ty: EdgeType,
+{
+    let node_num = graph.node_count();
+    if node_num == 0 {
+        return Ok(PyDict::new(py).into());
+    }
+
```
Chriscrosser3310, line 2965: Since now the size of bin_boundaries depends on node_num, an early return is added to prevent the algorithm from failing from an empty graph. Let me know if there are better ways.

src/lib.rs
```diff
@@ -2950,6 +2951,124 @@ fn digraph_transitivity(graph: &digraph::PyDiGraph) -> f64 {
...
+        let k_nbrs: Vec<NodeIndex> = graph.neighbors_undirected(*k).collect();
+        let k_deg = k_nbrs.len();
+
+        nbrs.insert(*k, k_nbrs);
+        cores.insert(*k, k_deg);
+        degree_map.insert(*k, k_deg);
+    }
+    node_vec.par_sort_by_key(|k| degree_map.get(k));
+
+    let mut bin_boundaries = vec![0; degree_map[&node_vec[node_num - 1]] + 1];
```
georgios-ts, line 2985: Maybe 
```suggestion
    let mut bin_boundaries: Vec<usize> =
        Vec::with_capacity(degree_map[&node_vec[node_num - 1]] + 1);
    bin_boundaries.push(0)
```
so you can still `push` to the vector, i.e no need for the new extra `cur_bin`, `last_bin` variables.

Chriscrosser3310: Oh right, thanks

src/lib.rs
```diff
@@ -2950,6 +2951,123 @@ fn digraph_transitivity(graph: &digraph::PyDiGraph) -> f64 {
...
+            curr_degree = v_degree;
+        }
+    }
+
+    for v_ind in 0..node_vec.len() {
+        let v = node_vec[v_ind];
+        let v_nbrs = nbrs[&v].clone();
+        for u in v_nbrs {
+            if cores[&u] > cores[&v] {
+                nbrs.get_mut(&u).unwrap().retain(|&n| n != v);
```
georgios-ts, line 3005: I was thinking also if it makes sense in terms of permformance to define `nbrs` of type `HashMap<NodeIndex, HashSet<NodeIndex>>` and write this line as:
```suggestion
                nbrs.get_mut(&u).unwrap().remove(&v);
```

Chriscrosser3310: Good catch again!

Chriscrosser3310:
I'm not sure why the build on ppc64le Linux failed. If anyone spotted any error in the changed code pls let me know

mtreinish (approved):
LGTM, thanks for updating

PR #277, opened 2021-03-17
Add hexagonal lattice graph generator

nahumsa:
Related to #150.

Added new functions for the generator module:
- undirected hexagonal lattice graph;
- Directed hexagonal lattice graph;
- ~~undirected complete graph;~~

Currently, the hexagonal generator does not support adding weights, because the algorithm to generate the graph needs to add auxiliary nodes and remove them at the end.

Update: Undirected complete graph was removed because it does the same thing that `mesh_graph` does and the itertools implementation is slower.


src/generators.rs
```diff
@@ -985,6 +985,109 @@ pub fn directed_grid_graph(
     })
 }
 
+/// Generate an undirected hexagonal graph.
+///
+/// :param int rows: The number of rows to generate the graph with.
+/// :param list cols: The number of rows to generate the graph with.
```
georgios-ts, line 991: `
```suggestion
/// :param list cols: The number of columns to generate the graph with.
```

src/generators.rs
```diff
@@ -985,6 +985,109 @@ pub fn directed_grid_graph(
...
+///       image = Image.open(tmp_path)
+///       os.remove(tmp_path)
+///   image
+///
+#[pyfunction(multigraph = true)]
+#[text_signature = "(/, rows=None, cols=None, multigraph=True)"]
+pub fn hexagonal_graph(
+    py: Python,
+    rows: Option<usize>,
+    cols: Option<usize>,
```
georgios-ts, line 1030:
```suggestion
    rows: usize,
    cols: usize,
```

I would not make this optional since they are not allowed to be `None` (you will immediately raise an error).

src/generators.rs
```diff
@@ -985,6 +985,109 @@ pub fn directed_grid_graph(
...
+    collen += 1;
+    let num_nodes = rowlen * collen;
+
+    let nodes: Vec<NodeIndex> =
+        (0..num_nodes).map(|_| graph.add_node(py.None())).collect();
+
+    // Add column edges
+    for i in 0..collen {
+        for j in 0..(rowlen - 1) {
+            if i + 1 < rowlen {
```
georgios-ts, line 1055: I don't believe this check `if i + 1 < rowlen` is needed, e.g `hexagonal_graph(1, 3)` will give wrong result.

georgios-ts (commented):
Just to confirm here that it's indeed intentional: it's a generator for the hexagonal graph and not *heavy* hex (that has some extra nodes in the edges), right?

If `rows=0` or `cols=0` should we return an empty graph as networkx does?

nahumsa:
> Just to confirm here that it's indeed intentional: it's a generator for the hexagonal graph and not _heavy_ hex (that has some extra nodes in the edges), right?

Yep,  I'm currently working on heavy hex.

> If `rows=0` or `cols=0` should we return an empty graph as networkx does?

Yep.

src/generators.rs
nahumsa: Oh, thanks. I forgot that I added `rowlen - 1` on the for loop.

georgios-ts (approved):

src/generators.rs
```diff
@@ -985,6 +985,101 @@ pub fn directed_grid_graph(
...
+                );
+            }
+        }
+    }
+
+    // Remove corner nodes
+    graph.remove_node(nodes[rowlen - 1]);
+    graph.remove_node(
+        nodes[(collen - 1) * rowlen + (rowlen - 1) * ((collen - 1) % 2)],
+    );
```
mtreinish, line 1074: Instead of removing the corner nodes here which will result in a graph with holes in the node indices couldn't we just detect the corner nodes in the edge loop and not add the edge?

mtreinish (commented):
A few quick questions inline. Also do we want to add a directed graph version of this generator (networkx lets you do this, for example: `nx.hexagonal_lattice_graph(10, 10, create_using=nx.DiGraph)`)?

src/generators.rs
```diff
@@ -985,6 +985,101 @@ pub fn directed_grid_graph(
...
+///   with tempfile.TemporaryDirectory() as tmpdirname:
+///       tmp_path = os.path.join(tmpdirname, 'dag.png')
+///       dot.write_png(tmp_path)
+///       image = Image.open(tmp_path)
+///       os.remove(tmp_path)
+///   image
+///
+#[pyfunction(multigraph = true)]
+#[text_signature = "(/, rows=None, cols=None, multigraph=True)"]
+pub fn hexagonal_graph(
```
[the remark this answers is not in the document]
nahumsa: Not really. It can be changed to `hexagonal_latice_graph`.

src/generators.rs
nahumsa: Sure. I will do an implementation without removing nodes.

nahumsa:
> A few quick questions inline. Also do we want to add a directed graph version of this generator (networkx lets you do this, for example: `nx.hexagonal_lattice_graph(10, 10, create_using=nx.DiGraph)`)?

I guess so. I will do an implementation for directed hexagonal graph.

src/generators.rs
nahumsa: Actually deleting the nodes is required for the implementation. Thus, we have two options:

- Documenting that we are deleting nodes and which nodes will be missing;
- Reindexing all nodes.

I will take a look at what will be the overhead of reindexing all nodes.

nahumsa: I ran a benchmark comparing reindexing nodes:

| Branches        | Time           | 
| ------------- |:-------------:| 
| [without reindexing](https://github.com/nahumsa/retworkx/tree/Expand_Generators)      | 0.115 |
| [with reindexing](https://github.com/nahumsa/retworkx/tree/hex-reindexing)     |   1.043  | 

Reindexing this graph adds a significant overhead, therefore I will choose to document which nodes that we are deleting.

nahumsa:
I added the directed version of the hexagonal lattice graph: `directed_hexagonal_lattice_graph`, but I'm not sure if the name is too big.

tests/generators/test_hexagonal.py
```diff
@@ -0,0 +1,57 @@
...
+        graph = retworkx.generators.directed_hexagonal_lattice_graph(2, 2)
+        self.assertEqual(len(graph), 16)
+        self.assertEqual(len(graph.edges()), 19)
+
+    def test_directed_hexagonal_graph_3_2(self):
+        graph = retworkx.generators.directed_hexagonal_lattice_graph(3, 2)
+        self.assertEqual(len(graph), 22)
+        self.assertEqual(len(graph.edges()), 27)
+
+    def test_hexagonal_graph_2_4(self):
```
mtreinish, line 29:
```suggestion
    def test_directed_hexagonal_graph_2_4(self):
```

nahumsa: Thanks! I was about to do that now. :)

tests/generators/test_hexagonal.py
```diff
@@ -0,0 +1,57 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import unittest
+import retworkx
+
+
+class TestHexagonalLatticeGraph(unittest.TestCase):
```
mtreinish, line 17: Could you add some tests that use `bidirectional=True` for `directed_hexagonal_lattice_graph`?

tests/generators/test_hexagonal.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def test_directed_hexagonal_graph_0_0(self):
+        graph = retworkx.generators.directed_hexagonal_lattice_graph(0, 0)
+        self.assertEqual(len(graph), 0)
+        self.assertEqual(len(graph.edges()), 0)
+
+    def test_hexagonal_graph_2_2(self):
+        graph = retworkx.generators.hexagonal_lattice_graph(2, 2)
+        self.assertEqual(len(graph), 16)
+        self.assertEqual(len(graph.edges()), 19)
```
mtreinish, line 42: It'd be nice to verify that this is actually a hexagonal lattice so we know moving forward that if we make changes that we're not breaking this. Right now all we know is that there are 16 nodes and 19 edges in the graph, not how they're connected.

mtreinish (commented):
The function code LGTM, just a couple comments on the tests. It would be good if we had the tests validating the output edge list is actually a hexagonal lattice. I think once the tests are updated this is ready to go.

tests/generators/test_hexagonal.py
nahumsa: Sure.

tests/generators/test_hexagonal.py
nahumsa: Sure, I'll add that.

mtreinish (approved):
LGTM! Thanks for updating the tests.

PR #278, opened 2021-03-18
Add dijkstra all_pairs functions

### LINKED ISSUE #271: Add functions for all_pairs_dijkstra_path and all_pairs_dijkstra_path_length

Implement `all_pairs_dijkstra_path` and `all_pairs_dijkstra_path_length`

The equivalent functions in networkx are: https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.shortest_paths.weighted.all_pairs_dijkstra_path.html and https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.shortest_paths.weighted.all_pairs_dijkstra_path_length.html

These functions are basically a for loop over all nodes in the `dijkstra_shortest_paths` and `dijkstra_shortest_path_lengths` functions. Ideally we should implement these for loops in parallel using rayon but the complication will be in getting the edge weights since we will need a python callback function. I'm thinking we can just graph all the edge weights once up front to a `Vec` and then use a parallel for loop with rayon calling dijkstra's on each iteration and for weights just do a lookup.

mtreinish:
This commit adds 2 new api functions all_pairs_dijkstra_shortest_paths()
and all_pairs_dijkstra_path_lengths(). These two functions are similar
to the existing dijkstra functions but they differ in that it will
return dictionaries over all nodes in the graph instead of just for a
single user provided node. Additionally these functions are multihreaded
and the level of parallelism can be controlled by the appropriate
environment variables.

Fixes #271


src/lib.rs
```diff
@@ -1875,6 +1875,343 @@ fn digraph_dijkstra_shortest_path_lengths(
...
+    for (index, entries) in out_map {
+        let inner_dict = PyDict::new(py);
+        for (map_index, val) in entries {
+            if map_index == index {
+                continue;
+            }
+            inner_dict.set_item(map_index, val)?;
+        }
+        out_dict.set_item(index, inner_dict)?;
+    }
```
nahumsa, line 1957: Since the dict creation is similar for all instances, in order to not write redundant code, wouldn't be better to abstract this part of the code to be reusable in other instances?

nahumsa (approved):

mtreinish:
I ran some benchmarks last night on this:

```python3
import time

import retworkx as rx
import networkx as nx

rx_graph = rx.directed_gnp_random_graph(5000, .9)
nx_graph = nx.DiGraph(list(rx_graph.edge_list()))

start = time.time()
rx.digraph_all_pairs_dijkstra_shortest_paths(rx_graph, lambda _: 1)
stop = time.time()

nx_start = time.time()
# Need to cast as a dict because networkx returns an iterator
dict(nx.all_pairs_dijkstra_path(nx_graph))
nx_stop = time.time()
print("networkx: %s" % str(nx_stop - nx_start))
```

| Library| Time (sec) |
| --- | --- |
| retworkx | 93.04283165931702 |
| networkx | 14554.81346654892 |

src/lib.rs
mtreinish: This is a good point, and I think that we actually really want to go even one step further and create some custom return types that implement the mapping protocol, that way we don't need the overhead of dict creation. I will do that in a separate PR because it will be a larger scope than this as it will modify the existing dijkstra functions as well.

mtreinish: I've done this here: https://github.com/Qiskit/retworkx/pull/287 if this PR merges first, I'll update that to add a new custom return type that wraps those. If that merges first I'll update it here and use new classes here that wrap those.

nahumsa: Awesome! I'll take a look on #287.

src/lib.rs
```diff
@@ -1956,6 +1956,15 @@ pub fn digraph_all_pairs_dijkstra_path_lengths(
     graph: &digraph::PyDiGraph,
     edge_cost_fn: PyObject,
 ) -> PyResult<PyObject> {
+    if graph.node_count() == 0 {
+        return Ok(PyDict::new(py).into());
```
nahumsa, line 1960: Is this done just to return an empty dict if is an empty graph?

mtreinish: Yep, exactly. Just fast path returning an empty dict for empty graphs.

src/lib.rs
mtreinish: Ok, done in: https://github.com/Qiskit/retworkx/pull/278/commits/4c0478ac942231be54d7d0fbc47abe50693a2c6b and https://github.com/Qiskit/retworkx/pull/278/commits/59145f2935d4c048c59b0c795ef12e5b83e6dc10

This has a very noticeable difference in performance running the benchmark script I did before it goes from ~93 seconds to ~42 seconds locally.

retworkx/__init__.py
```diff
@@ -398,6 +398,74 @@ def _graph_dijkstra_shortest_path(
...
+    This function will generate the shortest path from all nodes int the graph
+    using Dijkstra's algorithm. This function is multithreaded and will run
+    launch a thread pool with threads equal to the number of CPUs by default.
+    You can tune the number of threads with the ``RAYON_NUM_THREADS``
+    environment variable. For example, setting ``RAYON_NUM_THREADS=4`` would
+    limit the thread pool to 4 threads.
+
+    :param graph: The input graph to use. Can either be a
+        :class:`~retworkx.PyGraph` or :class:`~retworkx.PyDiGraph`
+    :param edge_cost_fn: An optional weight function for an edge. It will
```
nahumsa, line 414: Is `edge_cost_fn` really optional?

retworkx/__init__.py
```diff
@@ -398,6 +398,74 @@ def _graph_dijkstra_shortest_path(
...
+        accept a single argument, the edge's weight object and will return a
+        float which will be used to represent the weight/cost of the edge
+
+    :return: Dictionary of paths. The keys are destination node indices and
+        the dict values are lists of node indices making the path.
```
nahumsa, line 419: Would it be important to point out that is a read-only dict?

src/lib.rs
```diff
@@ -2022,6 +2022,309 @@ fn digraph_dijkstra_shortest_path_lengths(
...
 }
 
+fn _all_pairs_dijkstra_path_lengths<Ty: EdgeType + Sync>(
+    py: Python,
+    graph: &StableGraph<PyObject, PyObject, Ty>,
+    edge_cost_fn: PyObject,
+) -> PyResult<AllPairsPathLengthMapping> {
+    if graph.node_count() == 0 {
+        return Ok(AllPairsPathLengthMapping {
+            path_lengths: HashMap::new(),
```
nahumsa, line 2032: Wouldn't `AllPairsPathLenghtMapping::new()` do the same thing?

nahumsa (commented):
Great implementation overall. Using `AllPairsPathLenghtMapping` and `AllPairsPathMapping` makes it a clean implementation. Just had a few questions regarding the docs.

retworkx/__init__.py
mtreinish: Yep, we should definitely call that out I'll do this now.

retworkx/__init__.py
mtreinish: Good catch, yeah it's required here it's optional in some of the other functions (where I copied the docstring from)

src/lib.rs
mtreinish: It would, This was less typing :) and also consistent with how we build the other returns.

nahumsa: That makes sense, thanks for clarifying it

retworkx/__init__.py
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/278/commits/e9ee3657eab1cc02bcfa7cb5829e32ce2a23d1e3

retworkx/__init__.py
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/278/commits/e9ee3657eab1cc02bcfa7cb5829e32ce2a23d1e3

retworkx/__init__.py
```diff
@@ -398,6 +398,91 @@ def _graph_dijkstra_shortest_path(
...
+    launch a thread pool with threads equal to the number of CPUs by default.
+    You can tune the number of threads with the ``RAYON_NUM_THREADS``
+    environment variable. For example, setting ``RAYON_NUM_THREADS=4`` would
+    limit the thread pool to 4 threads.
+
+    :param graph: The input graph to use. Can either be a
+        :class:`~retworkx.PyGraph` or :class:`~retworkx.PyDiGraph`
+    :param edge_cost_fn: An optional weight function for an edge. It will
+        accept a single argument, the edge's weight object and will return a
+        float which will be used to represent the weight/cost of the edge
```
nahumsa, line 459:
```suggestion
    :param edge_cost_fn: A callable object that acts as a weight function for
        an edge. It will accept a single positional argument, the edge's weight
        object and will return a float which will be used to represent the
        weight/cost of the edge
```

nahumsa (approved):

PR #281, opened 2021-03-20
Add networkx converter function

mtreinish:
A commonly asked for piece of functionality of existing networkx users
is an easy way to convert networkx objects into retworkx objects. While
in the past we've avoided this because there is a strong desire to
**never** add a dependency on networkx to retworkx; implementing a
converter from networkx could be built by just duck typing and never
needing to import networkx. This commit implements that function so that
if a user of retworkx is taking in networkx graphs they have a simple
function to get a retworkx graph object from that. We still will never
add networkx as a dependency, which means going from retworkx graphs to
networkx graphs will be an exercise for users. But this commit also adds
a code example on how to do this to the retworkx for networkx users
guide to hopefully make that simpler.


nahumsa (approved):
LGTM. 
I was thinking about a converter because networkx have a considerably more robust drawing tools, for instance using matplotlib. Wouldn't be a good feature to have a matplotlib drawing interface implemented for retworkx?

mtreinish:
> LGTM.
> I was thinking about a converter because networkx have a considerably more robust drawing tools, for instance using matplotlib. Wouldn't be a good feature to have a matplotlib drawing interface implemented for retworkx?

Sure, that's something I've been debating too. Although if we did add a matplotlib drawer we would need to make sure matplotlib was an optional extra (and not required to install retworkx). Do you mind opening an issue for it?

nahumsa:
> > LGTM.
> > I was thinking about a converter because networkx have a considerably more robust drawing tools, for instance using matplotlib. Wouldn't be a good feature to have a matplotlib drawing interface implemented for retworkx?
> 
> Sure, that's something I've been debating too. Although if we did add a matplotlib drawer we would need to make sure matplotlib was an optional extra (and not required to install retworkx). Do you mind opening an issue for it?

Sure, I've done that in issue #298.

retworkx/__init__.py
```diff
@@ -593,3 +593,39 @@ def _digraph_core_number(graph):
...
+    if graph.is_directed():
+        if graph.is_multigraph():
+            new_graph = PyDiGraph()
+        else:
+            new_graph = PyDiGraph(multigraph=False)
+    else:
+        if graph.is_multigraph():
+            new_graph = PyGraph()
+        else:
+            new_graph = PyGraph(multigraph=False)
```
IvanIsCoding, line 624: This can be simplified
```suggestion
    if graph.is_directed():
        new_graph = PyDiGraph(multigraph=graph.is_multigraph())
    else:
        new_graph = PyGraph(multigraph=graph.is_multigraph())
```

IvanIsCoding (commented):
LGTM, there's still some code that can be simplified

IvanIsCoding:
> Is there a rustworkx to networkx converter? The reason I ask is that I ultimately have to import a network for use in PyTorch. There is a function in PyTorch to import from networkx. On the documentation, I only notice a networkx to rustworkx converter, but none for the other way around. Has this been implemented?

I think in general we prefer to do computations in rustworkx, hence the NetworkX -> rustworkx path is more frequent.

For your own use case rustworkx <-> PyTorch would be a better conversion. If you open an issue in their repository and link it here we can try to implement the conversion without going through NetworkX, which probably skips some overhead

PR #282, opened 2021-03-21
Add MST algorithms to retworkx

### LINKED ISSUE #194: Add a minimum spanning tree function

Add a minimum spanning tree function to retworkx's algorithm library. Something like networkx's `minimum_spanning_tree` https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.tree.mst.minimum_spanning_tree.html

IvanIsCoding:
Closes #194 

Add two new functions: 
* `retworkx.minimum_spanning_edges` to calculate the edges in the MST using Kruskal's algorithm
* `retworkx.minimum_spanning_tree` that returns a new graph with only the MST edges

Also, this PR makes `PyGraph` clonable In Rust (and `PyDigraph` as well for consistency). This makes keeping the same node indexes in `retworkx.minimum_spanning_tree` much easier. Let me know if there's any concerns about adding the clone feature


src/lib.rs
```diff
@@ -2729,6 +2731,68 @@ pub fn is_maximal_matching(
...
+///
+///     Note: The new graph will keep the same node indexes, but edge indexes might differ.
+/// :rtype: PyGraph
+#[pyfunction]
+#[text_signature = "(graph,)"]
+pub fn minimum_spanning_tree(
+    py: Python,
+    graph: &graph::PyGraph,
+) -> PyResult<graph::PyGraph> {
+    let mut spanning_tree = (*graph).clone();
```
Chriscrosser3310, line 2786: Is there any benefit of cloning the graph and clearing the edges over creating an empty graph and copying nodes from the original graph?

IvanIsCoding: Performance wise? No.

For keeping node indexes? It's a lot easier to code. `PyGraph` uses `StableGraph` as its base to store the graph, so there's no guarantee the node indexes are sequential. So we might have indexes `(0, 2, 4, 5)` in the old graph and when we insert it into the new graph they'll become sequential `(0, 1, 2, 3)` losing the mapping. Also, we cannot instantiate a graph with custom node indices because the variables from `petgraph` are private.

We can preserve this by looping `0 .. graph.node_bound()` and adding nodes to the new graph. Then we remove the indexes that are not in the old graph. And then we set the data to be the same as before. I found that code harder to implement and maintain

Chriscrosser3310: Thanks! That makes sense, and I think the trade-off is reasonable here.

Chriscrosser3310:
Great work! The clone mechanics might need to be updated as discussed in #289, but besides that, I don't have any more comments.

src/lib.rs
```diff
@@ -2729,6 +2731,68 @@ pub fn is_maximal_matching(
...
+    graph: &graph::PyGraph,
+) -> WeightedEdgeList {
+    let mut subgraphs = UnionFind::<usize>::new(graph.graph.node_bound());
+    let mut edge_list: Vec<EdgeReference<PyObject>> =
+        graph.graph.edge_references().collect();
+
+    edge_list.sort_by(|a, b| {
+        let weight_a = a.weight().as_ref(py);
+        let weight_b = b.weight().as_ref(py);
+        weight_a.compare(weight_b).unwrap()
```
mtreinish, line 2755: If you change the return type to `PyResult<WeightedEdgeList>` can't you just do:

```suggestion
        weight_a.compare(weight_b)?
```

here? We don't want to `unwrap()` a potential error, especially from python. We need to preserve those Python errors as they're exceptions which need to be returned to the caller. Especially here, where you're relying on python to do the object comparison the user needs to know that they're using objects which couldn't be compared.

src/lib.rs
```diff
@@ -2729,6 +2731,68 @@ pub fn is_maximal_matching(
     })
 }
 
+/// Find the edges in the minimum spanning tree or forest of a graph
+/// using Kruskal's algorithm.
+///
+/// :param PyGraph graph: Undirected weighted graph with comparable weights
```
mtreinish, line 2737: I would probably expand this a bit as it's a unique constraint for a retworkx function, something like: "python comparisons (ie `>` or `<`)  are used to compare the weights of edges. Ensure that the objects on the input graph are types that can be compared in this way."

mtreinish (commented):
Thanks for pushing this up, it's a great start! I left some comments inline. The other thing I'm a bit hesitant about still is the use of python for weight comparisons, mostly because it's different from other weight usage in retworkx. But, I also am wondering how the performance differs vs the callback function approach we use for the other functions. Did you do any comparisons against having a callback?

src/lib.rs
IvanIsCoding: We cannot use `weight_a.compare(weight_b)?` because the call is in the closure, and unfortunately the closure can't return `Result<cmp::Ordering>`, it must return `cmp::Ordering`. I will try find an alternative

src/lib.rs
```diff
@@ -2729,6 +2731,68 @@ pub fn is_maximal_matching(
...
+        let weight_a = a.weight().as_ref(py);
+        let weight_b = b.weight().as_ref(py);
+        weight_a.compare(weight_b).unwrap()
+    });
+
+    let mut answer: Vec<(usize, usize, PyObject)> = Vec::new();
+    for edge in edge_list.iter() {
+        let u = edge.source().index();
+        let v = edge.target().index();
+        let w = edge.weight().clone();
```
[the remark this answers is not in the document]
IvanIsCoding: Done

IvanIsCoding:
Alright, with regards to performance, I did some benchmarks with code similar to those:

```python
from random import randint

import time
import retworkx as rx
import networkx as nx

N = 1000

all_edges = []

for i in range(N):
	for j in range(i+1, N):
		all_edges.append((i, j, randint(0, 2**31 - 1)))

graph = rx.PyGraph()
graph.extend_from_weighted_edge_list(all_edges)

start = time.time()
mst_edges = rx.minimum_spanning_tree(graph)
stop = time.time()

G = nx.Graph()
G.add_weighted_edges_from(all_edges)

nx_start = time.time()
nx_mst_edges = nx.minimum_spanning_tree(G)
nx_stop = time.time()

print("networkx: %s" % str(nx_stop - nx_start))
```

And got:
```
retworkx: 1.1417829990386963
networkx: 1.724998950958252
```

We're winning but not by a large margin. I will try to benchmark against a code that uses weight callables

IvanIsCoding:
Ok, I switched to an implementation using function calls and conversion to floats. The calls to the functions are longer: `mst_edges = rx.minimum_spanning_tree(graph, weight_fn=lambda x: x)`. But the results are also better:

```
retworkx: 0.21052837371826172
networkx: 1.727597713470459
```

I came to the conclusion that sorting in Rust with Python callbacks cannot be much faster than a default sort in Python. Our only advantage over a pure Python implementation was the Union Find, but it wasn't yielding good margins because it's sorting that dominates the time

src/lib.rs
```diff
@@ -3069,6 +3071,122 @@ pub fn digraph_core_number(
...
+        let weight_a = a.0;
+        let weight_b = b.0;
+        if weight_a == weight_b {
+            Ordering::Equal
+        } else if weight_a < weight_b {
+            Ordering::Less
+        } else if weight_a > weight_b {
+            Ordering::Greater
+        } else if weight_a.is_nan() && weight_b.is_nan() {
+            Ordering::Equal
```
mtreinish, line 3124: Per IEEE 754 I don't think this is actually true. This is why rust's f64 doesn't implement the `Ord` trait.

src/lib.rs
```diff
@@ -3069,6 +3071,122 @@ pub fn digraph_core_number(
...
+        } else if weight_a.is_nan() {
+            Ordering::Greater
+        } else {
+            Ordering::Less
+        }
```
mtreinish, line 3129: If we're willing to say `NaN` isn't allowed I think we can just use `a.partial_cmp(b).unwrap());` here. This is actually what the rust `sort_by` docs point out. I think this is a fair compromise and we probably can just return a python exception in the edge references loop above if `weight_callable` returns a nan (we also should have a test for that).

mtreinish (commented):
This is looking good. Just some inline comments. The biggest thing is the float comparisons in the sort closure. I think we should just not allow NaN and use `partial_cmp` which makes that closure much simpler.

IvanIsCoding:
Ok, there is an advantage in using parallel sorting:
```
retworkx: 0.06562638282775879
networkx: 1.6518995761871338
```
Most of the type is spent in converting Python and Rust objects now. The algorithm is no longer the bottleneck.

src/lib.rs
```diff
@@ -3069,6 +3071,122 @@ pub fn digraph_core_number(
...
+
+    let mut edge_list: Vec<(f64, EdgeReference<PyObject>)> = Vec::new();
+    for edge in graph.edge_references() {
+        edge_list.push((
+            weight_callable(py, &weight_fn, &edge.weight(), default_weight)?,
+            edge,
+        ));
+    }
+
+    edge_list.sort_unstable_by(|a, b| {
```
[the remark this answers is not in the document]
IvanIsCoding: There is an advantage, it halves the time. I firmly believe most of the time is spent converting objects now, that sort is fast

src/lib.rs
IvanIsCoding: Done (with test added).

Also, I opted to use `a.partial_cmp(b).unwrap_or(Orderding::Less)` because it provides a slight performance boost (Rust inlines code better if there's no `panic!()`).

IvanIsCoding:
Ok all comment should have been addressed  now

src/lib.rs
```diff
@@ -3069,6 +3071,116 @@ pub fn digraph_core_number(
...
+///         minimum_spanning_edges(graph, weight_fn: float)
+///
+///     to cast the edge object as a float as the weight.
+/// :param float default_weight: If ``weight_fn`` isn't specified this optional
+///     float value will be used for the weight/cost of each edge.
+///
+/// :returns: The :math:`N - |c|` edges of the Minimum Spanning Tree (or Forest, if :math:`|c| > 1`)
+///     where :math:`N` is the number of nodes and :math:`|c|` is the number of connected components of the graph
+/// :rtype: WeightedEdgeList
+#[pyfunction(weight_fn = "None", default_weight = "1.0")]
```
mtreinish, line 3096: I don't think you actually need `None` for `weight_fn` here. It is an implied default for a `Option<T>` in pyfunction iirc.

mtreinish (approved):

PR #287, opened 2021-03-22
Add custom returns types for path dicts and path length dicts

mtreinish:
This commit adds 2 new custom return types PathMapping and
PathLengthMapping. These are used for functions that return either a
path dictionary or path length dictionary. These return types implement
the python mapping protocol and should be usable inplace of the previous
dict replacement.


mtreinish:
Just for comparisons sake I ran this test:

```python3
import time

import retworkx as rx

rx_graph = rx.generators.directed_path_graph(50000)

start = time.time()
rx.digraph_dijkstra_shortest_paths(rx_graph, 0)
stop = time.time()
```

for a worst case example and compared this PR against master:

| Branch | Run time (sec) |
| -- | --- |
| Master | 42.72968530654907 |
| #287 | 4.248757839202881 |

mtreinish:
I think this still needs a bit of work, an object with the mapping protocol set is not equivalent to a dict. There is no iterator support and common methods like keys, values, etc don't exist. I'll have to add those onto the custom types explicitly so this is actually a drop in replacement.

mtreinish:
Ok, I added the missing methods from `dict` to the custom return types here so they should be drop in replacements now.

tests/test_custom_return_types.py
```diff
@@ -276,3 +276,251 @@ def test_hash_invalid_type(self):
...
+class TestPathMapping(unittest.TestCase):
+
+    def setUp(self):
+        self.dag = retworkx.PyDAG()
+        node_a = self.dag.add_node('a')
+        self.dag.add_child(node_a, 'b', "Edgy")
+
+    def test__eq__match(self):
+        self.assertTrue(
+            retworkx.dijkstra_shortest_paths(self.dag, 0) == {1: [0, 1]})
```
lcapelluto, line 290: The existing dijkstra tests would also exercise this now, for larger graphs, correct?

src/iterators.rs
```diff
@@ -550,3 +556,538 @@ impl PyGCProtocol for WeightedEdgeList {
...
+                nodes: data.clone(),
+            }),
+            None => Err(PyIndexError::new_err("No node found for index")),
+        }
+    }
+}
+
+#[pyproto]
+impl PySequenceProtocol for PathMapping {
+    fn __len__(&self) -> PyResult<usize> {
```
lcapelluto, line 742: it's interesting that this has to be implemented multiple times depending on the protocol! Am I reading that right? (this line and 727) This means these could theoretically be implemented differently depending on the context?

releasenotes/notes/custom-path-returns-e1d5fc1e02809179.yaml
```diff
@@ -0,0 +1,37 @@
...
+    * :func:`retworkx.k_shortest_path_lengths`
+    * :func:`retworkx.digraph_dijkstra_shortest_path_lengths`
+    * :func:`retworkx.graph_dijkstra_shortest_path_lengths`
+    * :func:`retworkx.dijkstra_shortest_path_lengths`
+
+    no longer are returning a ``dict`` and instead are returning
+    :class:`retworkx.PathLengthMapping` objects. This new return type
+    is much faster to build and it implements the python mapping protocol in a
+    read-only fashion and should not be noticeable unless explicit type
+    checking or mutating the result were done.
```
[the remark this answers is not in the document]
mtreinish: It is, this is technically is an API change because the return type changes from a dict to a `PathMapping` or a `PathLengthMapping`, while it should be backwards compatible there are still edge cases that could require manual intervention on upgrading.

tests/test_custom_return_types.py
mtreinish: Yeah, I just added it here to have an explicit test of the python comparison api for `__eq__`

src/iterators.rs
mtreinish: Yeah, it's weird. But looking at the Python C API docs these are just the rust methods wrapping 2 different functions in the C protocols api:

https://docs.python.org/3/c-api/sequence.html#c.PySequence_Length
https://docs.python.org/3/c-api/mapping.html#c.PyMapping_Length

I'm not sure what circumstances one would be called over the other tbh, I was just implementing the required methods for the protocol `impl`.

tests/test_custom_return_types.py
```diff
@@ -278,6 +278,254 @@ def test_hash_invalid_type(self):
...
+        self.assertTrue(
+            retworkx.dijkstra_shortest_paths(self.dag, 0) != ['a', None])
+
+    def test__gt__not_implemented(self):
+        with self.assertRaises(NotImplementedError):
+            retworkx.dijkstra_shortest_paths(self.dag, 0) > {1: [0, 2]}
+
+    def test_deepcopy(self):
+        paths = retworkx.dijkstra_shortest_paths(self.dag, 0)
+        paths_copy = copy.deepcopy(paths)
```
nahumsa, line 344: Should the `PathMapping` return a shallow copy as well, using `.copy()` just like a dict does?

nahumsa (approved):

tests/test_custom_return_types.py
mtreinish: I hadn't thought of that (and honestly completely missed the `.copy()`method in the dict api). Thinking about it a bit I'm not sure it makes sense here though. Since `PathMapping` is read only a copy would be identical and couldn't diverge. The one way that would be different is if it returned a `dict` copy so you could change the object after copying. But that would be the same as just running `dict(path_mapping)`. So while the mapping types here are read-only I don't think we need `.copy()`.

PR #289, opened 2021-03-23
Add complement function

### LINKED ISSUE #285: Add complement function

Add an equivalent function to `networkx.complement()` to retworkx.

https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.operators.unary.complement.html

https://github.com/networkx/networkx/blob/4da254c095f8bd13610577974b749499e66f345b/networkx/algorithms/operators/unary.py#L7-L31

This is needed for converting qiskit-optimization to use retworkx internally:

https://github.com/Qiskit/qiskit-optimization/blob/03a4710fd31b583d4c2f4b6c0af98922ea045d63/qiskit_optimization/applications/clique.py#L55

IvanIsCoding:
Closes #285 

Adds a new `retworkx.complement()` function for calculating the complement of `PyGraph` and `PyDiGraph`


georgios-ts:
LGTM but i am a bit hesitant to `#[derive(Clone)]`
An artificial example:

```python
import retworkx as rx
from qiskit import QuantumCircuit

qc = QuantumCircuit(1)

g = rx.PyGraph()
g.add_nodes_from([qc])

cg = rx.complement(g)

# later modifiy circuit
qc.x(0)

print(cg.get_node_data(0))
---
Output:
     ┌───┐
q_0: ┤ X ├
     └───┘
```

I would expect to find the original empty circuit in the complement graph but maybe i am wrong.

IvanIsCoding:
> LGTM but i am a bit hesitant to `#[derive(Clone)]`
> An artificial example:
> 
> ```python
> import retworkx as rx
> from qiskit import QuantumCircuit
> 
> qc = QuantumCircuit(1)
> 
> g = rx.PyGraph()
> g.add_nodes_from([qc])
> 
> cg = rx.complement(g)
> 
> # later modifiy circuit
> qc.x(0)
> 
> print(cg.get_node_data(0))
> ---
> Output:
>      ┌───┐
> q_0: ┤ X ├
>      └───┘
> ```
> 
> I would expect to find the original empty circuit in the complement graph but maybe i am wrong.

That's because we're cloning a Python reference. Both graphs reference the same mutable object, so update in one affects the other.

I'd see three options here:

1. Keep the same references as it is
2. Create a deep copy of the referenced data
3. Just delete the data in the clone, as `networkx` does

Any preferences?

georgios-ts:
I would go with 3 but after some thought 1 seems fine too. I would leave to you and @mtreinish

mtreinish:
> I would go with 3 but after some thought 1 seems fine too. I would leave up to you and @mtreinish

For other functions where we return a subgraph (like `PyGraph.subgraph()`) we've been going with approach 1. But when we do we make it clear in the docstring that the weights from the original graph are copied by reference (and also add a unittest to assert that behavior). I think 3 is also a valid approach but we would also need to make it clear in the dcostring that behavior.

IvanIsCoding:
> > I would go with 3 but after some thought 1 seems fine too. I would leave up to you and @mtreinish
> 
> For other functions where we return a subgraph (like `PyGraph.subgraph()`) we've been going with approach 1. But when we do we make it clear in the docstring that the weights from the original graph are copied by reference (and also add a unittest to assert that behavior). I think 3 is also a valid approach but we would also need to make it clear in the dcostring that behavior.

In this case it's slightly different, the new complement edges have no data associated to them. But the data for node labels are a shared reference.

I'll check the behavior of `PyGraph.subgraph()`, and see if we copy node labels by reference there as well

mtreinish:
> > > I would go with 3 but after some thought 1 seems fine too. I would leave up to you and @mtreinish
> > 
> > 
> > For other functions where we return a subgraph (like `PyGraph.subgraph()`) we've been going with approach 1. But when we do we make it clear in the docstring that the weights from the original graph are copied by reference (and also add a unittest to assert that behavior). I think 3 is also a valid approach but we would also need to make it clear in the dcostring that behavior.
> 
> In this case it's slightly different, the new complement edges have no data associated to them. But the data for node labels are a shared reference.
> 
> I'll check the behavior of `PyGraph.subgraph()`, and see if we copy node labels by reference there as well

So far everywhere in retworkx where we return a node's payload as part of a container (which includes newly created graphs containing nodes from a graph passed in)  we copy by reference. This is mostly because that's all we really have access to both `clone()` and `clone_ref(py)` just increment the refcount and create a new pyobject with the same reference to the python heap (which is how all shared ownership types in rust work when `clone()` is called). Deepcopying isn't something natively exposed over the python c api either (it's also horrible .performance-wise because by default it pickles an object and the unpickles it to make the copy). Subgraph was just the best (and closest to this case) example of this (mostly because it has the docstring), but there are other examples, like `PyDigraph.to_undirected()` which makes an undirected graph from a digraph.

IvanIsCoding:
We copy by reference in both `PyDigraph.to_undirected()` and `PyGraph.subgraph()`, and we'll probably continue copying by reference in the future as pointed out, because its the behaviour of `.clone()` and `.clone_ref(py)`.

So the new `retworkx.complement()` addition follows the pattern

src/lib.rs
```diff
@@ -3069,6 +3069,86 @@ pub fn digraph_core_number(
...
+/// Compute the complement of a graph.
+/// :param graph: The graph to be used.
+///     :class:`~retworkx.PyGraph`
+
+///     :returns: The complement of the graph.
+///     :rtype: :class:`~retworkx.PyGraph`
+///
+/// .. note::
+///     Paralell edges and self-loops are never created,
+///     even if multigraph is set to True
```
mtreinish, line 3081:
```suggestion
/// Compute the complement of a graph.
///
/// :param PyGraph graph: The graph to be used.
///
/// :returns: The complement of the graph.
/// :rtype: PyGraph
///
/// .. note::
///
///     Parallel edges and self-loops are never created,
///     even if the :attr:`~retworkx.PyGraph.multigraph`
///     attribute is set to ``True``
```

retworkx/__init__.py
```diff
@@ -593,3 +593,30 @@ def _digraph_core_number(graph):
...
+    :param graph: The graph to be used, can be either a
+        :class:`~retworkx.PyGraph` or :class:`~retworkx.PyDiGraph`.
+
+    :returns: The complement of the graph.
+    :rtype: :class:`~retworkx.PyGraph` or :class:`~retworkx.PyDiGraph`
+
+    .. note::
+        Paralell edges and self-loops are never created,
+        even if multigraph is set to True
+    """
```
mtreinish, line 611:
```suggestion
    """Compute the complement of a graph.

    :param graph: The graph to be used, can be either a
        :class:`~retworkx.PyGraph` or :class:`~retworkx.PyDiGraph`.

    :returns: The complement of the graph.
    :rtype: :class:`~retworkx.PyGraph` or :class:`~retworkx.PyDiGraph`

    .. note::
        Parallel edges and self-loops are never created,
        even if the ``multigraph`` is set to ``True``
    """
```

src/lib.rs
```diff
@@ -3069,6 +3069,86 @@ pub fn digraph_core_number(
...
+/// Compute the complement of a graph.
+/// :param graph: The graph to be used.
+///     :class:`~retworkx.PyDiGraph`.
+
+///     :returns: The complement of the graph.
+///     :rtype: :class:`~retworkx.PyDiGraph`
+///
+/// .. note::
+///     Paralell edges and self-loops are never created,
+///     even if multigraph is set to True
```
mtreinish, line 3123:
```suggestion
/// Compute the complement of a graph.
///
/// :param PyDiGraph graph: The graph to be used.
///
/// :returns: The complement of the graph.
/// :rtype: :class:`~retworkx.PyDiGraph`
///
/// .. note::
///
///     Parallel edges and self-loops are never created,
///     even if the :attr:`~retworkx.PyDiGraph.multigraph`
///     attribute is set to ``True``
```

docs/source/api.rst
```diff
@@ -111,6 +111,7 @@ type functions in the algorithms API but can be run with a
 .. autosummary::
    :toctree: stubs
 
+   retworkx.complement
```
mtreinish, line 114: You also need to add the per type variants to the specific graph type methods general algorithms toctree above

mtreinish (commented):
This LGTM, the code looks great, thanks for tackling this so quickly. Just some minor docs formatting issues inline, but once those are fixed I think it's ready to go.

mtreinish:
FWIW, if you're having trouble getting `tox -edocs` to run locally the CI docs builds upload the built html output as an artifact, so you can download a zip file of the built docs. For example from the most recent CI run: https://github.com/Qiskit/retworkx/suites/2395802405/artifacts/50943773 is the artifact link from the [most recent CI run](https://github.com/Qiskit/retworkx/pull/289/checks?check_run_id=2242814574)

docs/source/api.rst
IvanIsCoding: Done

mtreinish (approved):
LGTM, thanks for updating the docs.

PR #290, opened 2021-03-25
Build universal2 wheels for macOS

messense:
https://cibuildwheel.readthedocs.io/en/stable/faq/#universal2

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -68,7 +68,7 @@ jobs:
           TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
           TWINE_USERNAME: retworkx-ci
   build-mac-arm-wheels:
-    name: Build wheels on macos for arm
+    name: Build wheels on macos for arm and universal2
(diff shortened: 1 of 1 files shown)


messense:
```bash
❯ lipo -info retworkx/retworkx.cpython-39-darwin.so
Architectures in the fat file: retworkx/retworkx.cpython-39-darwin.so are: x86_64 arm64
```

Verified that the architectures in the wheel is ok.

mtreinish (approved):
Thanks for doing this! It LGTM. I'm a bit surprised it works since we're explicitly setting the target to `aarch64-apple-darwin`, but I had mac users test it on different OS versions and architectures test the universal2 wheel and it seems to work as advertised.

messense:
> I'm a bit surprised it works since we're explicitly setting the target to `aarch64-apple-darwin`

setuptools-rust ignores that when building universal2 wheel: https://github.com/PyO3/setuptools-rust/pull/115/files#diff-91756cbf6b189507f711e1506b2ec255fc67de97478081e702ffdd50b19fc8afR85-R87

PR #291, opened 2021-03-25
Parallelize floyd_warshall_numpy with Rayon

### LINKED ISSUE #142: Add parallel numpy floyd warshall algorithm implementation

We have 2 implementations of the floyd warshall algorithm, one using ndarray/numpy and the other using dictionaries and looping over the graph objects.Right now they're both serially executed and as the graphs grow hit some scaling limits. We should look at adding parallel implementations of these algorithms, especially for the ndarray/numpy based one using:

https://en.wikipedia.org/wiki/Parallel_all-pairs_shortest_path_algorithm#Floyd_algorithm

since after we get the initial adjacency matrix there is no python involvement so we can easily parallelize the function.

IvanIsCoding:
Closes #142

* Rewrite `floyd_warshall_numpy` using `ndarray::parallel` iterators from Rayon to improve performance
* Refactor test case for negative cycles that was implementation dependent


tests/test_floyd_warshall.py
```diff
@@ -111,12 +111,15 @@ def test_weighted_numpy_negative_cycle(self):
             (3, 0, -1),
         ])
         dist = retworkx.graph_floyd_warshall_numpy(graph, lambda x: x)
-        expected = numpy.array(
-            [[-6, -7, -8, -13],
-             [-7, -8, -9, -14],
-             [-8, -9, -10, -15],
-             [-13, -14, -15, -20]], dtype=numpy.float64)
-        self.assertTrue(numpy.array_equal(dist, expected))
```
IvanIsCoding, line 119: Brief explanation why this test is implementation dependent and shouldn't be there:

1. If there's a negative cycle, the shortest path is not well defined. You can loop forever in the cycle and go as negative as you want making the path as short as you want.

So, take for example if we changed the nested inner loop we have for each `k`:
```
for i in 0 .. N {
  for j in 0 .. N {
  /* minimize m */
  }
}
```

For one where we swap `i` and `j`:
```
for j in 0 .. N {
  for i in 0 .. N {
  /* minimize m */
  }
}
```
Both ways of writing the loops visit all pairs `mat[i][j]`. For graphs without negative cycles, swapping `i` and `j` doesn't change correctness.

But for graphs with negative cycles, we would visit those negative cycles in a different order if we swap `i` and `j`. So the matrix results that would be slightly different. Even though the matrix results are meaningless because there's no shortest path!

Hence the reason to remove the matrix comparison

tests/test_floyd_warshall.py
```diff
@@ -111,12 +111,15 @@ def test_weighted_numpy_negative_cycle(self):
...
+        floyd_graph = retworkx.PyGraph()
+        floyd_graph.add_nodes_from(list(range(4)))
+        floyd_graph.add_edges_from(
+            [(i, j, dist[i][j]) for i in range(4) for j in range(4)]
+        )
+        second_dist = retworkx.graph_floyd_warshall_numpy(
+            floyd_graph, lambda x: x
+        )
+        self.assertTrue(numpy.all(second_dist < dist))
```
IvanIsCoding, line 122: Instead of comparing exact matrices, we compare if there's a negative cycle. If there's a negative cycle and I run Floyd-Warshall's algorithm again, it will give me shorter paths than before

IvanIsCoding:
I still need to run formal benchmarks, but I compared multiple versions with informal benchmarks and opted to use the one that only parallelizes the first nested loop.

For `N = 5000`, the version that is currently in `master` with no parallelization ran on 95s on my i5 machine with 8 cores.

For the version in the PR, the same test ran in 44s. It uses `.into_par_iter()`, which should have a small overhead because of how Rayon handles parallel iterators.

And for a version with both nested loops parallelized, the runtime was of 65s. It was faster than the code `master`, but slower than the version in the PR. Adding another `.into_par_iter()` caused more overhead.

retworkx/__init__.py
```diff
@@ -207,32 +207,40 @@ def _graph_all_simple_paths(graph, from_, to, min_depth=None, cutoff=None):
...
     :param float default_weight: If ``weight_fn`` is not used this can be
         optionally used to specify a default weight to use for all edges.
 
-     :return: The adjacency matrix for the input dag as a numpy array
-     :rtype: numpy.ndarray
+    :returns: A matrix of shortest path distances between nodes. If there is no
+        path between two nodes then the corresponding matrix entry will be
+        ``np.inf``.
+    :rtype: numpy.ndarray
     """
```
IvanIsCoding, line 244: I think we copied `adjacency_matrix` and forgot to change it! Profiting that we are changing it to fix it

IvanIsCoding:
Here's a benchmark with this code on some random graphs:

```python
from random import randint

import time
import retworkx as rx
import random

branch = "parallel-floyd" # change for each branch

def get_random_graph(N, p):
	random.seed(10**9 + 13)

	edge_list = [
		(i, j, random.randint(0, 2**20 - 1)) 
                for i in range(N) 
                for j in range(i+1, N) 
                if random.random() <= p
	]

	graph = rx.PyGraph()

	graph.extend_from_weighted_edge_list(edge_list)

	return graph

for N in [5, 25, 125,  625, 1000, 2000, 3000, 4000, 5000]:

	graph = get_random_graph(N, 0.5)
	start = time.time()
	floyd_result = rx.graph_floyd_warshall_numpy(graph, weight_fn=lambda x: x)
	stop = time.time()

	print("N = {} retworkx at {}: {}".format(N, branch, str(stop - start)))

```

On master, the results are:
```
N = 5 retworkx at master: 0.10829281806945801
N = 25 retworkx at master: 0.00010085105895996094
N = 125 retworkx at master: 0.0062580108642578125
N = 625 retworkx at master: 0.21328234672546387
N = 1000 retworkx at master: 0.9113898277282715
N = 2000 retworkx at master: 6.730209589004517
N = 3000 retworkx at master: 21.81312608718872
N = 4000 retworkx at master: 49.6114866733551
N = 5000 retworkx at master: 99.33520412445068
```

With this PR, the results are (considering 4 cores with 8 threads):
```
N = 5 retworkx at parallel-floyd: 0.16567063331604004
N = 25 retworkx at parallel-floyd: 0.010886430740356445
N = 125 retworkx at parallel-floyd: 0.0028476715087890625
N = 625 retworkx at parallel-floyd: 0.07497501373291016
N = 1000 retworkx at parallel-floyd: 0.2848215103149414
N = 2000 retworkx at parallel-floyd: 2.6836845874786377
N = 3000 retworkx at parallel-floyd: 9.107330083847046
N = 4000 retworkx at parallel-floyd: 21.59263300895691
N = 5000 retworkx at parallel-floyd: 42.03385925292969
```

I also got similar results for `digraph_floyd_warshall_numpy`.

However, for `floyd_warshall_numpy` there was no difference in performance... And that was because of a bug I suppose! `floyd_warshall_numpy` doesn't call Floyd Warshall, it calls `adjacency_matrix`.

tests/test_floyd_warshall.py
georgios-ts: Could you just check for negative values in the diagonal of `dist`?

mtreinish:
Can you split https://github.com/Qiskit/retworkx/pull/291/commits/5ac80fefc72c5421533be1f2eab4f52da21a6502 out into a separate PR? It's an independent fix from this PR and having a standalone commit for that fix (with it's own release note and test) will be good so we can potential backport it for a 0.8.1 release if needed.

mtreinish:
I ran a similar script:
```python3
import time
import retworkx as rx

branch = "serial-floyd" # change for each branch
sizes = [5, 25, 125,  625, 1000, 2000, 3000, 4000, 5000]
for N in sizes:
    graph = rx.undirected_gnp_random_graph(N, 0.5)
    rx.graph_floyd_warshall_numpy(graph, weight_fn=lambda _: 1.0)
    
for N in sizes:
    graph = rx.undirected_gnp_random_graph(N, 0.5)
    start = time.time()
    floyd_result = rx.graph_floyd_warshall_numpy(graph, weight_fn=lambda _: 1.0)
    stop = time.time()
    print("N = {} retworkx at {}: {}".format(N, branch, str(stop - start)))
```
and got consistent results (with 8 cores, 16 threads):
| branch | 5 nodes | 25 nodes | 125 nodes | 625 nodes | 1000 nodes | 2000 nodes | 3000 nodes | 4000 nodes | 5000 nodes |
| --- | --- | --- | ---| --- | --- | --- | --- | --- | --- |
| master | 2.0266e-05 | 3.0279e-05 | 0.0016 | 0.1619 | 0.6518 | 6.863 | 23.429 | 54.456 | 104.8159 |
| #291 | 0.00029 | 0.0007 | 0.00517 | 0.06161 | 0.1945 | 1.599 | 5.4470 | 10.5573 | 19.4377 |

Looking at these numbers I wonder if it makes sense to have an adjustable parallelization threshold (like what we do for `distance_matrix()`: https://github.com/Qiskit/retworkx/blob/master/src/lib.rs#L1381-L1393) since it looks like there is an overhead for building the threadpool that is larger than the gains from parallelization for smaller graphs.

tests/test_floyd_warshall.py
IvanIsCoding: Indeed, I switched to that. I think it's a reasonable check as we instantiate the diagonals to 0 at the start.

IvanIsCoding:
> I ran a similar script:
> 
> ```python
> import time
> import retworkx as rx
> 
> branch = "serial-floyd" # change for each branch
> sizes = [5, 25, 125,  625, 1000, 2000, 3000, 4000, 5000]
> for N in sizes:
>     graph = rx.undirected_gnp_random_graph(N, 0.5)
>     rx.graph_floyd_warshall_numpy(graph, weight_fn=lambda _: 1.0)
>     
> for N in sizes:
>     graph = rx.undirected_gnp_random_graph(N, 0.5)
>     start = time.time()
>     floyd_result = rx.graph_floyd_warshall_numpy(graph, weight_fn=lambda _: 1.0)
>     stop = time.time()
>     print("N = {} retworkx at {}: {}".format(N, branch, str(stop - start)))
> ```
> 
> and got consistent results (with 8 cores, 16 threads):
> 
> branch	5 nodes	25 nodes	125 nodes	625 nodes	1000 nodes	2000 nodes	3000 nodes	4000 nodes	5000 nodes
> master	2.0266e-05	3.0279e-05	0.0016	0.1619	0.6518	6.863	23.429	54.456	104.8159
> #291	0.00029	0.0007	0.00517	0.06161	0.1945	1.599	5.4470	10.5573	19.4377
> Looking at these numbers I wonder if it makes sense to have an adjustable parallelization threshold (like what we do for `distance_matrix()`: https://github.com/Qiskit/retworkx/blob/master/src/lib.rs#L1381-L1393) since it looks like there is an overhead for building the threadpool that is larger than the gains from parallelization for smaller graphs.

I added the `parallel_threshold` parameter to Floyd-Warshall, also defaulting to 300 nodes. I also added a test class and rewrote the test to make sure execute both the parallel and sequential version in CI.

mtreinish (approved):
LGTM, thanks for doing this.

PR #293, opened 2021-03-25
Add Heavy hex code to generators

nahumsa:
Added implementation of [heavy hex](https://journals.aps.org/prx/abstract/10.1103/PhysRevX.10.011022), the input is the code distance, d. In the implementation I chose to add nodes referring to each type of qubit used in the Heavy Hex Code (data, syndrome, and flag) which _may cause the node order to be confusing_, but that was the best way that I figure out how to implement. Now I will explain the way that I implemented the code, which could be better for people without experience with the heavy hex code.

In the heavy hex code, there are `d ** 2` data qubits, `(d + 1) * (d - 1) / 2` syndrome qubits, and `d * (d - 1)` flag qubits. In order to implement I follow 3 steps:

- 1) Each flag qubit is conneceted with 2 data qubits forming a chain on the graph, the chain has size `d`. 
- 2) Connect syndrome qubits that are connected to data qubits.
- 3) Connect flag and syndrome qubits.

In steps (ii) and (iii), there are two patterns depending on the eve (odd) columns of syndrome qubits, which are treated accordingly using an `if` statement. It was also required to use `.chunk()` for `Vectors` in order to represent them in a column.

Related to #150


src/generators.rs
```diff
@@ -983,6 +983,139 @@ pub fn directed_grid_graph(
...
+        if i % 2 == 0 {
+            graph.add_edge(nodes_data[i * d], syndrome_chunk[0], py.None());
+            graph.add_edge(
+                syndrome_chunk[0],
+                nodes_data[(i + 1) * d],
+                py.None(),
+            );
+        } else if i % 2 == 1 {
+            graph.add_edge(
+                nodes_data[(i + 1) * d - 1],
```
ecpeterson, line 1065: Same comment as in #313 : I'd prefer this to be written `i * d + (d - 1)`.

src/generators.rs
```diff
@@ -983,6 +983,139 @@ pub fn directed_grid_graph(
...
+                syndrome_chunk[syndrome_chunk.len() - 1],
+                py.None(),
+            );
+            graph.add_edge(
+                syndrome_chunk[syndrome_chunk.len() - 1],
+                nodes_data[(i + 2) * d - 1],
```
ecpeterson, line 1071: Same comment as above.

src/generators.rs
```diff
@@ -983,6 +983,139 @@ pub fn directed_grid_graph(
     })
 }
 
+/// Generate an undirected heavy hex graph.
```
ecpeterson, line 986: A useful reference here is Fig 2 of https://arxiv.org/pdf/1907.09528.pdf , and here's some ASCII art:

```
... D-S-D   D ...
    |   |   |
...-F   F-S-F ...
    |   |   |
... D   D   D ...
    |   |   |
... F-S-F   F-...
    |   |   |
    .........
    |   |   |
... D   D   D ...
    |   |   |
...-F   F-S-F ...
    |   |   |
... D   D   D ...
    |   |   |
... F-S-F   F-...
    |   |   |
    .........
    |   |   |
... D   D   D ...
    |   |   |
...-F   F-S-F ...
    |   |   |
... D   D   D ...
    |   |   |
... F-S-F   F-...
    |   |   |
... D   D-S-D ...
```

ecpeterson (approved):
Same as in #313 : no comment on the Rust, quantum stuff all looks good, and I left a few notes on clarifying intent.

ecpeterson:
One other note for onlookers, not really germane to the PR: it's easier to understand these lattices as perfect square grids with holes poked in them, rather than assembling the pieces from scratch.

src/generators.rs
```diff
@@ -1072,6 +1072,173 @@ pub fn directed_binomial_tree_graph(
     })
 }
 
+/// Generate an undirected heavy hex graph. Fig. 2 of
+/// https://arxiv.org/abs/1907.09528
+/// An ASCII diagram of the graph is given by:
+///
+/// .. code-block:: console
+///     ... D-S-D   D ...
```
mtreinish, line 1080: This is a formatting error (and why ci is failing) because of the missing newline sphinx is interpretting the entire ascii art block as the type string for the codeblock. To fix this you can do:

```suggestion
/// .. code-block:: text
///
///     ... D-S-D   D ...
```

nahumsa: Thanks!

mtreinish (commented):
LGTM, thanks for doing this. The only question I have before merging is do you think it makes sense to add a directed version too? If so I think that would be useful to have.

ecpeterson:
Good catch, and up to y'all. It's ambiguous what directionality is preferred by the circuit embodiments of the stabilizer checks, so I don't think it'd be punting to skip this — but I also think that "flag qubits are targets of edges" is the maxim that's compatible with device engineering.

nahumsa:
I'll implement the directed version of the heavy hex code in order to be used on the `couplingMap` class.

src/generators.rs
```diff
@@ -1072,6 +1072,393 @@ pub fn directed_binomial_tree_graph(
...
+    }
+
+    // connect flag and syndromes
+    for (i, syndrome_chunk) in nodes_syndrome.chunks((d + 1) / 2).enumerate() {
+        if i % 2 == 0 {
+            for (j, syndrome) in syndrome_chunk.iter().enumerate() {
+                if j != 0 {
+                    graph.add_edge(
+                        nodes_flag[i * (d - 1) + 2 * (j - 1) + 1],
+                        *syndrome,
```
mtreinish, line 1401: Same here assuming I'm reading the figures in the paper correctly (the color change is annoying)

```suggestion
                        *syndrome,
                        nodes_flag[i * (d - 1) + 2 * (j - 1) + 1],
```

src/generators.rs
```diff
@@ -1072,6 +1072,393 @@ pub fn directed_binomial_tree_graph(
...
+                    );
+                    graph.add_edge(
+                        *syndrome,
+                        nodes_flag[(i + 1) * (d - 1) + 2 * (j - 1) + 1],
+                        py.None(),
+                    );
+                    if bidirectional {
+                        graph.add_edge(
+                            *syndrome,
+                            nodes_flag[i * (d - 1) + 2 * (j - 1) + 1],
```
mtreinish, line 1412:
```suggestion
                            nodes_flag[i * (d - 1) + 2 * (j - 1) + 1],
                            *syndrome,
```

src/generators.rs
```diff
@@ -1072,6 +1072,393 @@ pub fn directed_binomial_tree_graph(
...
+                        );
+                    }
+                }
+            }
+        } else if i % 2 == 1 {
+            for (j, syndrome) in syndrome_chunk.iter().enumerate() {
+                if j != syndrome_chunk.len() - 1 {
+                    graph.add_edge(
+                        nodes_flag[i * (d - 1) + 2 * j],
+                        *syndrome,
```
mtreinish, line 1428:
```suggestion
                        *syndrome,
                        nodes_flag[i * (d - 1) + 2 * j],
```

src/generators.rs
```diff
@@ -1072,6 +1072,393 @@ pub fn directed_binomial_tree_graph(
...
+                    );
+                    graph.add_edge(
+                        *syndrome,
+                        nodes_flag[(i + 1) * (d - 1) + 2 * j],
+                        py.None(),
+                    );
+                    if bidirectional {
+                        graph.add_edge(
+                            *syndrome,
+                            nodes_flag[i * (d - 1) + 2 * j],
```
mtreinish, line 1439:
```suggestion
                            nodes_flag[i * (d - 1) + 2 * j],
                            *syndrome,
```

mtreinish (commented):
LGTM, thanks for adding the directed variant. I do wonder if we can de-duplicate the implementation since the code is mostly identical but we can do that in a follow up (as it also applies to most of the generators).

Just one inline comment on the edge direction based on @ecpeterson's comment: https://github.com/Qiskit/retworkx/pull/293#issuecomment-892036650 and the figures in the paper about the hardware implementation. Once that's fixed I think this is good to go

src/generators.rs
```diff
@@ -1072,6 +1072,393 @@ pub fn directed_binomial_tree_graph(
...
+
+    // connect data and syndromes
+    for (i, syndrome_chunk) in nodes_syndrome.chunks((d + 1) / 2).enumerate() {
+        if i % 2 == 0 {
+            graph.add_edge(nodes_data[i * d], syndrome_chunk[0], py.None());
+            graph.add_edge(
+                syndrome_chunk[0],
+                nodes_data[(i + 1) * d],
+                py.None(),
+            );
```
mtreinish, line 1359:
```suggestion
            graph.add_edge(
                nodes_data[(i + 1) * d],
                syndrome_chunk[0],
                py.None(),
            );
```

src/generators.rs
```diff
@@ -1072,6 +1072,393 @@ pub fn directed_binomial_tree_graph(
...
+            if bidirectional {
+                graph.add_edge(syndrome_chunk[0], nodes_data[i * d], py.None());
+                graph.add_edge(
+                    nodes_data[(i + 1) * d],
+                    syndrome_chunk[0],
+                    py.None(),
+                );
```
mtreinish, line 1366:
```suggestion
                graph.add_edge(
                    syndrome_chunk[0],
                    nodes_data[(i + 1) * d],
                    py.None(),
                );
```

src/generators.rs
```diff
@@ -1072,6 +1072,393 @@ pub fn directed_binomial_tree_graph(
...
+            }
+        } else if i % 2 == 1 {
+            graph.add_edge(
+                nodes_data[i * d + (d - 1)],
+                syndrome_chunk[syndrome_chunk.len() - 1],
+                py.None(),
+            );
+            graph.add_edge(
+                syndrome_chunk[syndrome_chunk.len() - 1],
+                nodes_data[i * d + (2 * d - 1)],
```
mtreinish, line 1376:
```suggestion
                nodes_data[i * d + (2 * d - 1)],
                syndrome_chunk[syndrome_chunk.len() - 1],
```

src/generators.rs
```diff
@@ -1072,6 +1072,393 @@ pub fn directed_binomial_tree_graph(
...
+            );
+            if bidirectional {
+                graph.add_edge(
+                    syndrome_chunk[syndrome_chunk.len() - 1],
+                    nodes_data[i * d + (d - 1)],
+                    py.None(),
+                );
+                graph.add_edge(
+                    nodes_data[i * d + (2 * d - 1)],
+                    syndrome_chunk[syndrome_chunk.len() - 1],
```
mtreinish, line 1387:
```suggestion
                    syndrome_chunk[syndrome_chunk.len() - 1],
                    nodes_data[i * d + (2 * d - 1)],
```

mtreinish (commented):
Looking at figure 10 more closely for #313 I think the direction for the data-> syndrome edges:

Screenshot_2021-08-17_13-24-02
Screenshot_2021-08-17_12-58-35

src/generators.rs
```diff
@@ -1072,6 +1072,393 @@ pub fn directed_binomial_tree_graph(
...
+    }
+
+    Ok(graph::PyGraph {
+        graph,
+        node_removed: false,
+        multigraph,
+    })
+}
+
+/// Generate an directed heavy hex graph. Fig. 2 of
```
ecpeterson, line 1243:
```suggestion
/// Generate a directed heavy hex graph. Fig. 2 of
```

src/generators.rs
```diff
@@ -1072,6 +1072,393 @@ pub fn directed_binomial_tree_graph(
...
+///
+///   with tempfile.TemporaryDirectory() as tmpdirname:
+///       tmp_path = os.path.join(tmpdirname, 'dag.png')
+///       dot.write_png(tmp_path)
+///       image = Image.open(tmp_path)
+///       os.remove(tmp_path)
+///   image
+///
+#[pyfunction(bidirectional = false, multigraph = true)]
+#[pyo3(text_signature = "(d, /, bidirecational=False, multigraph=True)")]
```
ecpeterson, line 1314:
```suggestion
#[pyo3(text_signature = "(d, /, bidirectional=False, multigraph=True)")]
```

src/generators.rs
ecpeterson: I'm like ~95% confident that this suggested change is right. The circuit described in the paper would prefer the other natural direction on this small set of edges, but if natural source/targets are required to alternate in hardware, then having to simulate flipping these edges through Hadamards is preferable to flipping the alternating pattern through the bulk.

But there is a 5% chance that I'm missing something.

ecpeterson (approved):
With @mtreinish's amendments, this LGTM.

mtreinish (approved):

mtreinish:
LGTM, I just went ahead and applied the suggestions to get this in for the release. Thanks for doing this @nahumsa and sticking with it for so long

nahumsa:
Thanks both @mtreinish and @ecpeterson for the review.
Sorry, I didn't update the code quickly.

PR #296, opened 2021-03-27
Cleanup test organization

### LINKED ISSUE #183: Cleanup test organization

Currently the Python unittests for retworkx lack a cohesive organizational structure. As pointed out in #144 there is a mix of some tests splitting across `tests/` and `tests/graph` for tests using `PyDiGraph`/`PyDAG` and `PyGraph`, while others just put all the tests in the same file under `tests/` normally using 2 test classes. We really should make these consistent, so either split all tests into subdirectories by the class being tested or have a single module per function and test both classes in that module with a separate test classes. Which pattern we adopt doesn't matter so much as long as its consistent for all tests.

nahumsa:
Closes #183.

Creates a cohesive structure for tests by splitting class tests into folders, all `PyGraph` class tests are in the `tests/graph` folder and all `PyDiGraph/PyDAG` tests are in the `tests/DiGraph` folder. As before, all `retworkx.generators` tests are in the `tests/generators` folder.

There are two exceptions:
- `test_dispatch.py` is in the `tests` folder because it tests both `PyGraph` and `PyDigraph` and moving this test to separate folders would lead to duplicated code.
- `test_random.py` is in the `tests` folder because it is similar to generators and generates random graphs, however they are not in the `generators` module, so I was unsure if it is needed to create a `random`  folder to put this kind of tests.


mtreinish (commented):
This LGTM, thanks for doing this! The only nitpick I have is instead of `tests/DiGraph` could we name that directory `tests/digraph`?

nahumsa:
> This LGTM, thanks for doing this! The only nitpick I have is instead of `tests/DiGraph` could we name that directory `tests/digraph`?

Sure.

mtreinish (approved):
Awesome, thanks for the quick update. This LGTM, I especially appreciate you catching things like `test_unique_neighbors_on_graphs` which was put in the wrong file (even though the test_neighbors files were already split). Heh, if I had done this I would have completely missed that and just left it in `digraph/test_neighbors`.

Before I merge this PR though we should give a heads up to @georgios-ts @IvanIsCoding @Chriscrosser3310 and @jlapeyre (which hopefully this comment will be sufficient for) who all have open PRs as this will require them to rebase and move the tests into the new consistent directory structure.

PR #299, opened 2021-04-01
Add binomial tree graph generator

nahumsa:
Related to #150 

Add binomial tree graph for a given order.


src/generators.rs
```diff
@@ -985,6 +987,105 @@ pub fn directed_grid_graph(
...
+    py: Python,
+    order: Option<usize>,
+    weights: Option<Vec<PyObject>>,
+    multigraph: bool,
+) -> PyResult<graph::PyGraph> {
+    let mut graph = StableUnGraph::<PyObject, PyObject>::default();
+
+    if order.is_none() {
+        return Err(PyIndexError::new_err("order should be specified"));
+    }
```
mtreinish, line 1038: Is there a reason to make this optional if it's always required? Couldn't we change the order argument to be: `order: usize,` above and just remove this?

nahumsa: Sure. I was thinking to use weights to initialize as well, but was having some issues with rust `Option`, but not making optional would work.

nahumsa:
I ran some benchmarks comparing to `networkx`:
```python
import time
import retworkx as rx
import networkx as nx 


libraries = ["retworkx", "networkx"] # change for each branch
sizes = [5, 10, 20]

start = 0
stop = 0

for lib in libraries:
    
    for N in sizes:
        if lib == "retworkx":
            start = time.time()
            graph = rx.generators.binomial_tree_graph(N)
            stop = time.time()
        
        if lib == "networkx":
            start = time.time()
            graph = nx.generators.binomial_tree(N)
            stop = time.time()

        print(f"N = {N} {lib}: {str(stop - start)}")
```

| Library        |    N=5      | N=10  | N=20 |
| ------------- |:-------------:| -------------:| -----:|
| retworkx      | 1.5e-05s | 0.000129s | 0.07646989822387695s |
| networkx     |  0.0198s | 0.0075s | 5.8s |

src/generators.rs
```diff
@@ -985,6 +987,101 @@ pub fn directed_grid_graph(
...
+pub fn binomial_tree_graph(
+    py: Python,
+    order: usize,
+    weights: Option<Vec<PyObject>>,
+    multigraph: bool,
+) -> PyResult<graph::PyGraph> {
+    let mut graph = StableUnGraph::<PyObject, PyObject>::default();
+
+    // try_into is used to convert from usize to u32
+    let num_nodes = usize::pow(2, order.try_into().unwrap());
```
mtreinish, line 1037: If a `u32` is required for `usize::pow` why not just make the input to order `u32` instead of `usize`? If you do then a type error will get raised if something outside the bounds is passed to the function. With it like this the user will see a panic exception

nahumsa: Sure. When I was doing the benchmarking I stumbled upon the panic exception and was not able to figure out why it happened. 
Thanks for pointing that out!

src/generators.rs
```diff
@@ -985,6 +986,204 @@ pub fn directed_grid_graph(
...
+///
+///   with tempfile.TemporaryDirectory() as tmpdirname:
+///       tmp_path = os.path.join(tmpdirname, 'dag.png')
+///       dot.write_png(tmp_path)
+///       image = Image.open(tmp_path)
+///       os.remove(tmp_path)
+///   image
+///
+#[pyfunction(multigraph = true)]
+#[text_signature = "(/, order=None, weights=None, multigraph=True)"]
```
mtreinish, line 1026: Unless I'm misreading the code `order` is a required positional argument:

```suggestion
#[text_signature = "(order=None, /, weights=None, multigraph=True)"]
```

src/generators.rs
```diff
@@ -985,6 +986,204 @@ pub fn directed_grid_graph(
...
+) -> PyResult<graph::PyGraph> {
+    let mut graph = StableUnGraph::<PyObject, PyObject>::default();
+
+    let num_nodes = usize::pow(2, order);
+
+    let nodes: Vec<NodeIndex> = match weights {
+        Some(weights) => {
+            let mut node_list: Vec<NodeIndex> = Vec::new();
+
+            for weight in weights {
```
mtreinish, line 1041: What happens if `weights.len() != num_nodes`?

tests/generators/test_binomial_tree.py
```diff
@@ -0,0 +1,53 @@
...
+
+class TestBinomailTreeGraph(unittest.TestCase):
+
+    def test_binomial_tree_graph(self):
+        for n in range(10):
+            graph = retworkx.generators.binomial_tree_graph(n)
+            self.assertEqual(len(graph), 2**n)
+            self.assertEqual(len(graph.edges()), 2**n - 1)
+
+    def test_binomial_tree_graph_weights(self):
```
mtreinish, line 26: Can you add a negative test case where `len(weights) != 2^order` ?

tests/generators/test_binomial_tree.py
```diff
@@ -0,0 +1,53 @@
...
+import retworkx
+
+
+class TestBinomailTreeGraph(unittest.TestCase):
+
+    def test_binomial_tree_graph(self):
+        for n in range(10):
+            graph = retworkx.generators.binomial_tree_graph(n)
+            self.assertEqual(len(graph), 2**n)
+            self.assertEqual(len(graph.edges()), 2**n - 1)
```
mtreinish, line 24: Can we add some assertions here on the edge list. It would be good to assert that the output graph is actually a binomial tree instead of just a graph with `2**n` nodes and `2**n - 1` edges

mtreinish, line 24:
```suggestion
            with self.subTest(n=n):
                graph = retworkx.generators.binomial_tree_graph(n)
                self.assertEqual(len(graph), 2**n)
                self.assertEqual(len(graph.edges()), 2**n - 1)
```

tests/generators/test_binomial_tree.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_binomial_tree_no_order(self):
+        with self.assertRaises(TypeError):
+            retworkx.generators.binomial_tree_graph(weights=list(range(4)))
+
+    def test_directed_binomial_tree_graph(self):
+        for n in range(10):
+            graph = retworkx.generators.directed_binomial_tree_graph(n)
+            self.assertEqual(len(graph), 2**n)
+            self.assertEqual(len(graph.edges()), 2**n - 1)
```
mtreinish, line 41:
```suggestion
            with self.subTest(n=n):
                graph = retworkx.generators.directed_binomial_tree_graph(n)
                self.assertEqual(len(graph), 2**n)
                self.assertEqual(len(graph.edges()), 2**n - 1)
```

tests/generators/test_binomial_tree.py
nahumsa: Sure

src/generators.rs
nahumsa: I'm adding extra None nodes if the number of weights is less than 2**order, and raise an error if `weights.len() > num_node`.

tests/generators/test_binomial_tree.py
nahumsa: Added assertions to check if the edges match what is expected for a binomial tree graph

tests/generators/test_binomial_tree.py
nahumsa: I added `test_binomial_tree_graph_weight_less_nodes` for the case when `weights < 2^order` and `test_binomial_tree_graph_weights_greater_nodes` when `weights > 2^order`

src/generators.rs
```diff
@@ -983,6 +984,246 @@ pub fn directed_grid_graph(
...
+///
+///   with tempfile.TemporaryDirectory() as tmpdirname:
+///       tmp_path = os.path.join(tmpdirname, 'dag.png')
+///       dot.write_png(tmp_path)
+///       image = Image.open(tmp_path)
+///       os.remove(tmp_path)
+///   image
+///
+#[pyfunction(multigraph = true)]
+#[text_signature = "(order=None, /, weights=None, multigraph=True)"]
```
mtreinish, line 1025: I think this should be

```suggestion
#[text_signature = "(order, /, weights=None, multigraph=True)"]
```

`order` is a required positional argument

mtreinish (commented):
Looks good, just a couple of quick inline comments. I think this is almost ready.

src/generators.rs
```diff
@@ -983,6 +984,246 @@ pub fn directed_grid_graph(
...
+            {
+                graph.add_edge(
+                    nodes[target_index + n],
+                    nodes[source_index + n],
+                    py.None(),
+                );
+            }
+        }
+
+        if graph.find_edge(nodes[0], nodes[n]) == None {
```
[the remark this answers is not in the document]
nahumsa: Sure, this looks better.

src/generators.rs
```diff
@@ -983,6 +984,246 @@ pub fn directed_grid_graph(
...
+///   image
+///
+#[pyfunction(bidirectional = "false")]
+#[text_signature = "(/, order=None, weights=None, bidirectional=False)"]
+pub fn directed_binomial_tree_graph(
+    py: Python,
+    order: u32,
+    weights: Option<Vec<PyObject>>,
+    bidirectional: bool,
+) -> PyResult<digraph::PyDiGraph> {
```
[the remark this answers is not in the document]
nahumsa: Not really, I just did this because all other directed generators have this pattern. Should we add `multigraph` on all of them?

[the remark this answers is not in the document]
mtreinish: Ah, I didn't realize it was missing as an option from all the digraph generators. I think then it's fine to exclude it here for now, but we should push a follow up PR and add a multigraph option to all the digraph generators.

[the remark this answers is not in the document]
nahumsa: Sure, I'll raise an issue and submit a PR for that

src/generators.rs
```diff
@@ -983,6 +984,246 @@ pub fn directed_grid_graph(
...
+            .edge_references()
+            .map(|e| (e.source(), e.target()))
+            .collect();
+
+        for (source, target) in edges {
+            let source_index = source.index();
+            let target_index = target.index();
+
+            if graph.find_edge(nodes[source_index + n], nodes[target_index + n])
+                == None
```
mtreinish, line 1185:
```suggestion
            if graph.find_edge(nodes[source_index + n], nodes[target_index + n])
                .is_none()
```

src/generators.rs
```diff
@@ -983,6 +984,246 @@ pub fn directed_grid_graph(
...
+                    nodes[source_index + n],
+                    nodes[target_index + n],
+                    py.None(),
+                );
+            }
+
+            if bidirectional
+                && graph
+                    .find_edge(nodes[target_index + n], nodes[source_index + n])
+                    == None
```
mtreinish, line 1197:
```suggestion
                    .is_none()
```

src/generators.rs
```diff
@@ -983,6 +984,246 @@ pub fn directed_grid_graph(
...
+                    py.None(),
+                );
+            }
+        }
+
+        if graph.find_edge(nodes[0], nodes[n]).is_none() {
+            graph.add_edge(nodes[0], nodes[n], py.None());
+        }
+
+        if bidirectional && graph.find_edge(nodes[n], nodes[0]) == None {
```
mtreinish, line 1211:
```suggestion
        if bidirectional && graph.find_edge(nodes[n], nodes[0]).is_none() {
```

mtreinish (approved):
LGTM, I found a couple more places to use `is_none()` but other than that I think this is ready.

PR #302, opened 2021-04-05
Fix floyd_warshall_numpy dispatcher

IvanIsCoding:
Fix bug found during #291 benchmarking. 

`floyd_warshall_numpy` was dispatching to the wrong function and calling `adjacency_matrix` instead of `graph_floyd_warshall_numpy` and `digraph_floyd_warshall_numpy`.


tests/releasenotes/notes/floyd-warshall-fix-507ddeb9b792c89f.yaml
```diff
@@ -0,0 +1,6 @@
+---
```
mtreinish, line 1: This file is in the wrong directory, it should be `releasenotes/notes/` but this is in in `tests/releasenotes/notes` which means it won't show up in the release notes in the documentation.

mtreinish (commented):
Thanks for splitting this out, it looks good ot me, but I think you ran `reno` from `tests/` and the release note file ended up in the wrong directory.

tests/releasenotes/notes/floyd-warshall-fix-507ddeb9b792c89f.yaml
IvanIsCoding: Nice catch! Apparently I did the same thing for another release note too. It should be fixed now

mtreinish (approved):

PR #304, opened 2021-04-06
Add matplotlib based drawer

### LINKED ISSUE #298: Add a matplotlib drawing tool for Retworkx

Networkx has a robust function to draw graphs using [matplotlib](https://matplotlib.org/)

https://networkx.org/documentation/stable//reference/generated/networkx.drawing.nx_pylab.draw.html#networkx.drawing.nx_pylab.draw

We should add an equivalent tool for retworkx, but this function should be an optional extra.

mtreinish:
This commit adds a a new python visualization module which contains a
matplotlib drawer for retworkx graphs. This first iteration of the
visualization function is porting networkx's equivalent functionality.
For now the api is basically identical to networkx's functionality
except it takes a retworkx graph object instead of a networkx graph
object. However, in follow-on commits to this branch the API will be
adapted to be more retworkx native (i.e. using callbacks to generate
labels from nodes and edges, etc).

Fixes #298

TODO:

- [x] Update API to be flatter and more retworkx-like
    - [x] Reduce number of public facing APIs (only need a single draw() entrypoint)
    - [x] remove `retworkx` from function names
    - [x] Remove label functions and rely on a callback for a node's payload
    - [x] Return a `Figure` object
- [x] Split out layout functionality into a separate PR (integrated here for first draft testing) #305 
- [x] Add docs to build
- [x] Rewrite docs in standard sphinx style and update for new api
- [x] Add tests
- [x] Add release notes


mtreinish:
Just to show the basics here, running:
```python
import retworkx
from retworkx import visualization

g = retworkx.directed_gnp_random_graph(10, .9)
fig = plt.figure()
ax = fig.gca()
visualization.draw(g, ax=ax)
fig.savefig('test.png', dpi=600)
```

yields:

test

There is still a lot of work to do here, but the basics work for now.

mtreinish:
I've split out the layout functionality into #305

mtreinish:
I think this is ready for review now. But this is on hold until #305 (which an early draft of is included inline for now) merges because for the `mpl_draw()` function to work we need a layout function available. Ideally the default will be #306 but #305 will probably be ready first.

tests/visualization/test_mpl.py
```diff
@@ -0,0 +1,194 @@
...
+
+import retworkx
+from retworkx.visualization import mpl_draw
+
+try:
+    import matplotlib as mpl
+    import matplotlib.pyplot as plt
+
+    mpl.use("PS")
+    plt.rcParams["text.usetex"] = False
```
nahumsa, line 27: Why it is needed to disable usetex?

mtreinish: It's mostly for reproducibility in the tests, matplotlib's usetex mode relies on external latex installations to run: https://matplotlib.org/stable/tutorials/text/usetex.html and even if present can cause different behavior on different platforms or latex distributions. So just simplicity it makes sense to disable it here so things run consistently between different systems (although I don't think it should come up in any of the tests). 

That being said this line was just borrowed from networkx's equivalent drawer test module: https://github.com/networkx/networkx/blob/ead0e65bda59862e329f2e6f1da47919c6b07ca9/networkx/drawing/tests/test_pylab.py#L9

retworkx/visualization/matplotlib.py
```diff
@@ -0,0 +1,1153 @@
...
+        edge_color = "k"
+
+    # set edge positions
+    edge_pos = np.asarray([(pos[e[0]], pos[e[1]]) for e in edge_list])
+
+    # Check if edge_color is an array of floats and map to edge_cmap.
+    # This is the only case handled differently from matplotlib
+    if (
+        np.iterable(edge_color)
+        and (len(edge_color) == len(edge_pos))
```
nahumsa, line 657: Could it be possible for `edge_color` to be a dict? 
My reasoning is that we could have a default color and just change some of the nodes that we want to change using a dictionary with `edge_pos` as key values. Is that reasonable?

mtreinish: Well `edge_pos` isn't user facing (it's just internal array of node positions for the nodes of an edge), but I think I see what your getting at here. The only way we can do different colors for a subset of edges now is multiple calls to `mpl_draw` with a subset of nodes and edges (the tests show examples of this). In networkx this kind of thing is handled normally by calling `draw_edges()` separately, but for this PR I opted for a single function for simplicity. This however causes some tension for more custom things like different colors for specific edges or nodes. I think there are 3 ways to address this:

1. Make the internal `draw_nodes`, `draw_edges`, etc functions part of the user facing api like networkx does.
2. Add support for dicts/mappings as an input to `edge_colors`, `node_colors`, `node_shape` (and maybe `node_size`) so that these properties can be set per node and per edge.
3. Add callback function support for `edge_colors`, `node_colors`, etc so that when we evaluate colors we can just call the user provided function to get a color string (or whatever other property.

None of these options are mutually exclusive so we can do any combination of them. How do you think we should proceed here?

nahumsa: The most natural solution, at least for me, is the first one. However, exposing those functions may lead to some trouble. For the long term, I think the second solution can be better. What you think about doing both 1 and 2?

nahumsa (approved):
LGTM. Overall it looks great, and the tests are excellent. 
The only comment is to make the drawer function more customizable, but this could be a future improvement.

retworkx/visualization/matplotlib.py
mtreinish: I was on the fence about doing 1, but I agree it seems like it makes the most sense for user expectations. I think it'll be good as a followup (either before or after release). Same for 2, we can do that in a follow up. I'm going to merge this now (with your review) I'll open an issue about this after merging.

Issue #280, opened 2021-03-20
Add spring_layout/fruchterman_reingold_layout function to retworkx


mtreinish:
### What is the expected enhancement?

Networkx has a `spring_layout` function which returns a layout for nodes in the graph (which can be used for drawing):

https://networkx.org/documentation/stable/reference/generated/networkx.drawing.layout.spring_layout.html

and the code for this networkx function is:

https://github.com/networkx/networkx/blob/4da254c095f8bd13610577974b749499e66f345b/networkx/drawing/layout.py#L345-L494

We should add an equivalent function to retworkx. (it might also be good to add simpler layout functions first)

Chriscrosser3310:
Has anyone been working on this already? If not I'll do it

georgios-ts:
In fact i have started working on it but we can coordinate or add different layouts

mtreinish:
Just a heads up I wrote a `random_layout` function for #304 (since to get that working I needed a layout function and that looked like the simplest). I will eventually split that out into a separate PR (it's technically independent of the mpl drawer), I just pushed it up all together to show the PoC for the drawer. One thing we might want to share between all the layout implementations is the `Pos2DLayout` custom return type in that, since it should be a lot more efficient to create than building a python dict.

I think once we have a spring layout function that should be the default we use in the matplotlib drawer.

georgios-ts:
Although there is still work to do in #306 , i'm wondering if it makes sense to implement some improvements in fruchterman reingold layout algorithm as described [here](http://yifanhu.net/PUB/graph_draw_small.pdf). Specifically, we can use:

- Quad trees to efficiently (O(n logn) instead of O(n^2)) approximate repulsive forces
- A multilevel approach with a coarsening scheme.

Both are necessary for fast and nice layouts in large graphs. So, my question is if it's worth spending the time to implement it in retworkx or rely on dedicated tools like GraphViz (this algorithm is in `sfdp` layout)

mtreinish:
My first response is that I would probably say for more complex algorithms we probably should just rely on graphviz. The drawer/layouts built into retworkx are more about convenience (so you don't need to install an external tool), I don't think we should expect to match the performance or ability of graphviz with a built-in drawer or layout function. We still will always have the `to_dot` method to generate a dot file that we can pass to graphviz (a follow up to #304 I plan is to add a python function that wraps `pydot` as an optional requirement to make it a bit easier to draw).

[...]

PR #305, opened 2021-04-07
Add random_layout function and Pos2DMapping

mtreinish:
This commit adds a new function `random_layout()` which is used to
generate a random layout for a graph that can be used in visualization.
This is necessary for building a matplotlib drawer (issue #298 and a
first draft of the implementation #304). To make the function more
efficient it also adds a new custom return type `Pos2DMapping` which is
used to build an immutable readonly dict compatible result container for
the output type from this function.

Related to #280

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -1,11 +1,11 @@
 .. _retworkx:
 
-======================
+######################
 Retworkx API Reference
-======================
+######################
 
 Graph Classes
--------------
+=============
 
 .. autosummary::
    :toctree: stubs
@@ -15,7 +15,7 @@ Graph Classes
     retworkx.PyDAG
 
 Generators
-----------
+==========
 
 .. autosummary::
    :toctree: stubs
@@ -32,7 +32,7 @@ Generators
     retworkx.generators.directed_grid_graph
 
 Random Circuit Functions
-------------------------
+========================
 
 .. autosummary::
    :toctree: stubs
@@ -43,10 +43,10 @@ Random Circuit Functions
     retworkx.undirected_gnm_random_graph
 
 Algorithm Functions
--------------------
+===================
 
 Specific Graph Type Methods
-'''''''''''''''''''''''''''
+---------------------------
 
 .. autosummary::
    :toctree: stubs
@@ -101,13 +101,11 @@ Specific Graph Type Methods
    retworkx.digraph_core_number
    retworkx.graph_complement
    retworkx.digraph_complement
-   retworkx.graph_random_layout
-   retworkx.digraph_random_layout
 
 .. _universal-functions:
 
 Universal Functions
-'''''''''''''''''''
+-------------------
 
 These functions are algorithm functions that wrap per graph object
 type functions in the algorithms API but can be run with a
@@ -133,8 +131,18 @@ type functions in the algorithms API but can be run with a
    retworkx.core_number
    retworkx.random_layout
 
+Layout Functions
+================
+
+.. autosummary::
+   :toctree: stubs
+
+   retworkx.random_layout
+   retworkx.graph_random_layout
+   retworkx.digraph_random_layout
+
 Exceptions
-----------
+==========
 
 .. autosummary::
    :toctree: stubs
@@ -147,8 +155,8 @@ Exceptions
    retworkx.NoPathFound
    retworkx.NullGraph
 
-Return Iterator Types
----------------------
+Custom Return Types
+===================
 
 .. autosummary::
    :toctree: stubs
@@ -157,3 +165,4 @@ Return Iterator Types
    retworkx.NodeIndices
    retworkx.EdgeList
    retworkx.WeightedEdgeList
+   retworkx.Pos2DMapping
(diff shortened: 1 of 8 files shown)


mtreinish:
There is a potential merge conflict here with #287 since the `Pos2DMapping` type borrows a lot of the implementation from that PR. We probably should merge #287 first and then rebase this on top.

src/lib.rs
```diff
@@ -3263,6 +3265,73 @@ fn digraph_complement(
...
+}
+
+/// Generate a random layout
+///
+/// :param PyGraph graph: The graph to generate the layout for
+/// :param tuple center: An optional center position. This is a 2 tuple of two
+///     ``float`` values for the center position
+/// :param int seed: An optional seed to set for the random number generator.
+///
+/// :returns: The complement of the graph.
```
Chriscrosser3310, line 3304:
```suggestion
/// :returns: The random layout of the graph.
```

src/lib.rs
```diff
@@ -3263,6 +3265,73 @@ fn digraph_complement(
...
+}
+
+/// Generate a random layout
+///
+/// :param PyDiGraph graph: The graph to generate the layout for
+/// :param tuple center: An optional center position. This is a 2 tuple of two
+///     ``float`` values for the center position
+/// :param int seed: An optional seed to set for the random number generator.
+///
+/// :returns: The complement of the graph.
```
Chriscrosser3310, line 3323:
```suggestion
/// :returns: The random layout of the graph.
```

retworkx/__init__.py
```diff
@@ -624,3 +624,28 @@ def _digraph_complement(graph):
...
+@functools.singledispatch
+def random_layout(graph, center=None, seed=None):
+    """Generate a random layout
+
+    :param PyGraph graph: The graph to generate the layout for
+    :param tuple center: An optional center position. This is a 2 tuple of two
+        ``float`` values for the center position
+    :param int seed: An optional seed to set for the random number generator.
+
+    :returns: The complement of the graph.
```
Chriscrosser3310, line 638:
```suggestion
    :returns: The random layout of the graph.
```

releasenotes/notes/add-random-layout-c1c2751be971e5d0.yaml
```diff
@@ -0,0 +1,16 @@
+---
+features:
+  - |
+    Added a new layou function, :func:`retworkx.random_layout` (and it's
```
nahumsa, line 4:
```suggestion
Added a new layout function, :func:`retworkx.random_layout` (and it's
```

nahumsa (approved):
LGTM.

src/lib.rs
```diff
@@ -3263,6 +3265,73 @@ fn digraph_complement(
     Ok(complement_graph)
 }
 
+fn _random_layout<Ty: EdgeType>(
+    graph: &StableGraph<PyObject, PyObject, Ty>,
+    center: Option<(f64, f64)>,
```
georgios-ts, line 3270: Maybe?
```suggestion
    center: Option<[f64; 2]>,
```
not important but i'm commenting on this to keep things consistent since i use `[f64; 2]` on my spring-layout pr.

docs/source/api.rst
```diff
@@ -154,3 +165,4 @@ Return Iterator Types
    retworkx.NodeIndices
    retworkx.EdgeList
    retworkx.WeightedEdgeList
+   retworkx.Pos2DMapping
```
georgios-ts, line 168: with `retworkx.*_layout(graph).keys()` a user will see `Pos2DMappingKeys`. similar with`.values()`. Should we have docs for these types?

mtreinish: I'm not sure if it matters too much, we can add it to the documentation it doesn't hurt. I just left it out since there really isn't any documentation for those types and we don't actually export them in the retworkx module. I basically modeled these types on how python's dict works it 'returns a `dict_keys`, `dict_values`, and `dict_items` class for `.keys()`, `.values()`, and `.items()`. I couldn't find any documentation on those classes except in the C-API: https://docs.python.org/3/c-api/dict.html?highlight=dict_keys#c.PyDict_Keys

src/lib.rs
mtreinish: Yeah I agree it'll be good to be consistent. The only reason I went with a tuple here was I wasn't certain if pyo3 handled the conversion from `(0., 0.5)` in python to `[f64; 2]` in rust. Since it's a fixed 2 element array for center the python side interface seems more pythonic as a tuple instead of a list. But I think in the rust side using `[f64; 2]` makes more sense since it has a contiguous memory layout. Let me test this locally and see what we can do here. At the worst case we can have the `#[pyfunction]` functions convert from a tuple to an array.

mtreinish: I tested this locally (and the unit tests cover it too), it looks like pyo3 will do the conversion from `(0.0, 0.5)` to an `[f64; 2]` in rust. So I've made this change in: https://github.com/Qiskit/retworkx/pull/305/commits/82fe3f1afc8edb6000beb40df100e54036422f43

georgios-ts (approved):

PR #306, opened 2021-04-07
Add spring layout

### LINKED ISSUE #280: Add spring_layout/fruchterman_reingold_layout function to retworkx

georgios-ts:
Resolves #280 .
The algorithm is taken from [here](http://yifanhu.net/PUB/graph_draw_small.pdf) (section 3).
Left to do:
- [x] Add more tests.
- [x] Add weights in the edges.


georgios-ts:
Some examples using this new function together with `networkx.draw` functionality [here](https://gist.github.com/georgios-ts/b0f487b0ab79a6d30240303bf2ddf8cb#file-layout-ipynb)

mtreinish:
I just merged #305 which has some merge conflicts here, while you're updating it would be good to change the return types to a `Pos2DMapping` instead of a hashmap too.

src/layout.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    if fixed.is_empty() {
+        if let Some(scale) = scale {
+            rescale(
+                &mut pos,
+                scale,
+                graph.node_indices().map(|n| n.index()).collect(),
+            );
+        }
+
```
Chriscrosser3310, line 288: To match the behavior of networkx we need to set scale == 1.0 instead of not calling it, because rescale always moves the mean of points to (0, 0).

src/layout.rs
```diff
@@ -0,0 +1,295 @@
...
+        step = cs.update_step(energy);
+        if converged {
+            break;
+        }
+    }
+
+    if fixed.is_empty() {
+        if let Some(scale) = scale {
+            rescale(
+                &mut pos,
```
Chriscrosser3310, line 283: To match the behavior of networkx we need to set scale == 1.0 instead of not calling it when `scale` is `None`, because rescale always moves the mean of points to (0, 0).

tests/digraph/test_spring_layout.py
```diff
@@ -0,0 +1,64 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import unittest
+
+import retworkx
+
+
+class TestSpringLayout(unittest.TestCase):
+
```
Chriscrosser3310, line 19: It might be better to add tests which examine the accuracy of output positions, such as comparing to outputs generated by networkx’s spring layout function

src/layout.rs
georgios-ts: i match this behavior by setting `scale=1.0` by default in `spring_layout` function in `retworkx/__init__.py` . i was thinking that it makes sense to explicitly define `scale=None` if you don't want any rescaling of the output.

tests/digraph/test_spring_layout.py
georgios-ts: Well, i'm not sure. This function and networkx's spring layout function don't give identical results but rather "equivalent" since:
- there are some hardcoded values that may not match
- there are some parameters that networkx doesn't expose (e.g repulsive force exponent, cooling scheme)
- we do a kind of numeric integration that's quite sensitive in the order you process nodes, "when" you apply forces etc.

retworkx/__init__.py
```diff
@@ -740,6 +740,78 @@ def _graph_random_layout(graph, center=None, seed=None):
...
+    as repelling objects, sometimes called an anti-gravity force.
+    Simulation continues until the positions are close to an equilibrium.
+
+    :param graph: Graph to be used. Can either be a
+        :class:`~retworkx.PyGraph` or :class:`~retworkx.PyDiGraph`.
+    :param dict (default=None) pos:
+        Initial node positions as a dictionary with node ids as keys and values
+        as a coordinate list. If None, then use random initial positions.
+    :param set (default=None) fixed: Nodes to keep fixed at initial position.
+        Error raised if fixed specified and pos not.
```
mtreinish, line 762:
```suggestion
    :param set fixed: Nodes to keep fixed at initial position.
        Error raised if fixed specified and ``pos`` is not. (``default=None``) 
```

retworkx/__init__.py
```diff
@@ -740,6 +740,78 @@ def _graph_random_layout(graph, center=None, seed=None):
...
+    :param float (default=None) k:
+        Optimal distance between nodes. If None the distance is set to
+        1/sqrt(n) where n is the number of nodes.  Increase this value
+        to move nodes farther apart.
```
mtreinish, line 766:
```suggestion
    :param float  k:
        Optimal distance between nodes. If ``None`` the distance is set to
        :math:`\frac{1}{\sqrt{n}}` where :math:`n` is the number of nodes.  Increase this value
        to move nodes farther apart. (``default=None``)
```

retworkx/__init__.py
```diff
@@ -740,6 +740,78 @@ def _graph_random_layout(graph, center=None, seed=None):
...
+    :param int (default=2) p:
+        Repulsive force exponent.
```
mtreinish, line 768:
```suggestion
    :param int p:
        Repulsive force exponent. (``default=2``)
```

retworkx/__init__.py
```diff
@@ -740,6 +740,78 @@ def _graph_random_layout(graph, center=None, seed=None):
...
+    :param bool (default=True) adaptive_cooling:
+        Use an adaptive cooling scheme. If set to False,
+        a linear cooling scheme is used.
```
mtreinish, line 771:
```suggestion
    :param bool adaptive_cooling:
        Use an adaptive cooling scheme. If set to ``False``,
        a linear cooling scheme is used. (``default=True``)
```

retworkx/__init__.py
```diff
@@ -740,6 +740,78 @@ def _graph_random_layout(graph, center=None, seed=None):
...
+    :param int (default=50) niter:
+        Maximum number of iterations.
```
mtreinish, line 773:
```suggestion
    :param int niter:
        Maximum number of iterations. (``default=50``)
```

src/lib.rs
```diff
@@ -3466,6 +3467,251 @@ fn digraph_complement(
...
+#[allow(clippy::too_many_arguments)]
+pub fn digraph_spring_layout(
+    py: Python,
+    graph: &digraph::PyDiGraph,
+    pos: Option<HashMap<usize, layout::Point>>,
+    fixed: Option<HashSet<usize>>,
+    k: Option<f64>,
+    p: Option<i32>,
+    adaptive_cooling: Option<bool>,
+    niter: Option<usize>,
```
mtreinish, line 3689: What do you think about renaming this `max_iter` or `num_iter`? When I first saw this I though it was [niter](https://en.wikipedia.org/wiki/Niter) and was very confused.

retworkx/__init__.py
```diff
@@ -740,6 +740,78 @@ def _graph_random_layout(graph, center=None, seed=None):
...
+    :param float (default = 1e-6) tol:
+        Threshold for relative error in node position changes.
+        The iteration stops if the error is below this threshold.
```
mtreinish, line 776:
```suggestion
    :param float tol:
        Threshold for relative error in node position changes.
        The iteration stops if the error is below this threshold.
        (``default = 1e-6``) 
```

retworkx/__init__.py
```diff
@@ -740,6 +740,78 @@ def _graph_random_layout(graph, center=None, seed=None):
...
+    :param weight_fn: An optional weight function for an edge. It will accept
+        a single argument, the edge's weight object and will return a float
+        which will be used to represent the weight of the edge.
+    :param float (default=1) default_weight: If ``weight_fn`` isn't specified
+        this optional float value will be used for the weight/cost of each edge
+    :param float or None (default=1) scale: Scale factor for positions.
+        Not used unless fixed is None. If scale is None, no rescaling is
+        performed.
```
mtreinish, line 784:
```suggestion
    :param float|None scale: Scale factor for positions.
        Not used unless fixed is None. If scale is ``None``, no re-scaling is
        performed. (``default=1.0``)
```

retworkx/__init__.py
```diff
@@ -740,6 +740,78 @@ def _graph_random_layout(graph, center=None, seed=None):
...
+    :param list (default=None) center: Coordinate pair around which to center
+        the layout. Not used unless fixed is None.
```
mtreinish, line 786:
```suggestion
    :param list center: Coordinate pair around which to center
        the layout. Not used unless fixed is ``None``. (``default=None``)
```

src/lib.rs
```diff
@@ -3466,6 +3467,251 @@ fn digraph_complement(
...
+                     niter=50, tol=1e-6, weight_fn=None, default_weight=1, scale=1,
+                     center=None, seed=None, /)"]
+#[allow(clippy::too_many_arguments)]
+pub fn digraph_spring_layout(
+    py: Python,
+    graph: &digraph::PyDiGraph,
+    pos: Option<HashMap<usize, layout::Point>>,
+    fixed: Option<HashSet<usize>>,
+    k: Option<f64>,
+    p: Option<i32>,
```
mtreinish, line 3687: I realize these are straight from the paper, but I do wonder if we can think of more descriptive argument names in case someone hasn't read the paper. Not really a blocker, but something to consider.

mtreinish (commented):
Overall the functional component of the code LGTM, I highlighted some docs syntax issues in the docstring for the universal function. But, they'll need to be updated in all the docstrings. One thing to point out is you can download the built and rendered docs from the CI docs jobs, they upload an artifact zip file of the rendered html that you can view locally and test formatting (assuming `tox -edocs` doesn't work locally for some reason).

There are also some comments inline about some of the parameter names which I think could be renamed for clarity.

src/lib.rs
georgios-ts: `k` is also used by networkx, so it might be better to leave it as it is. 
what about `repulsive_exponent` for `p`?

src/lib.rs
georgios-ts: haha sure, i renamed to `num_iter`

retworkx/__init__.py
georgios-ts: for some reason, this line can't pass the doc tests.

src/lib.rs
mtreinish: Sure that works for me

retworkx/__init__.py
mtreinish: Ah, that was my bad I forgot this docstring wasn't a raw string, so it's trying to escape `\f`, just replace `\frac` here with `\\frac` so it parses correctly (or alternatively you could start the docstring with `r"""` and leave it as `\frac`).

retworkx/__init__.py
```diff
@@ -740,6 +740,136 @@ def _graph_random_layout(graph, center=None, seed=None):
...
+    :param graph: Graph to be used. Can either be a
+        :class:`~retworkx.PyGraph` or :class:`~retworkx.PyDiGraph`.
+    :param dict pos:
+        Initial node positions as a dictionary with node ids as keys and values
+        as a coordinate list. If ``None``, then use random initial positions. (``default=None``)
+    :param set fixed: Nodes to keep fixed at initial position.
+        Error raised if fixed specified and ``pos`` is not. (``default=None``)
+    :param float  k:
+        Optimal distance between nodes. If ``None`` the distance is set to
+        :math:`\frac{1}{\sqrt{n}}` where :math:`n` is the number of nodes.  Increase this value
```
mtreinish, line 776:
```suggestion
        :math:`\\frac{1}{\sqrt{n}}` where :math:`n` is the number of nodes.  Increase this value
```

retworkx/__init__.py
georgios-ts: oh thanks, that was subtle.

src/layout.rs
```diff
@@ -0,0 +1,295 @@
...
+}
+
+#[inline]
+fn max(x: Nt, y: Nt) -> Nt {
+    if x > y {
+        x
+    } else {
+        y
+    }
+}
```
mtreinish, line 41: Is there an advantage to defining a custom inline function over just calling `x.max(y)` where we need it?

src/lib.rs
```diff
@@ -3466,6 +3467,255 @@ fn digraph_complement(
...
+    let fixed = fixed.unwrap_or_default();
+    let k = k.unwrap_or(1.0 / (graph.node_count() as f64).sqrt());
+    let f_a = layout::AttractiveForce::new(k);
+    let f_r = layout::RepulsiveForce::new(k, repulsive_exponent.unwrap_or(2));
+
+    let num_iter = num_iter.unwrap_or(50);
+    let tol = tol.unwrap_or(1e-6);
+    let step = 0.1;
+
+    let mut weights: HashMap<(usize, usize), f64> = HashMap::default();
```
mtreinish, line 3518:
```suggestion
    let mut weights: HashMap<(usize, usize), f64> = HashMap::with_capacity(graph.graph.edge_count());
```

Just to avoid multiple memory allocations.

src/lib.rs
```diff
@@ -3466,6 +3467,255 @@ fn digraph_complement(
...
+    graph.edge_references().for_each(|e| {
+        let w = weight_callable(py, &weight_fn, &e.weight(), default_weight)
+            .unwrap();
+        let source = e.source().index();
+        let target = e.target().index();
+
+        weights.insert((source, target), w);
+        weights.insert((target, source), w);
+    });
```
mtreinish, line 3527:
```suggestion
    for e in graph.edge_references() {
        let w = weight_callable(py, &weight_fn, &e.weight(), default_weight)?;
        let source = e.source().index();
        let target = e.target().index();

        weights.insert((source, target), w);
        weights.insert((target, source), w);
    });
```

So we don't lose an inner exception raised by the python code in the callback function.

mtreinish (commented):
LGTM, I just did another pass through the code and found a couple of things. But this is getting close.

src/layout.rs
```diff
@@ -0,0 +1,295 @@
...
+
+use hashbrown::{HashMap, HashSet};
+
+use pyo3::prelude::*;
+
+use petgraph::graph::NodeIndex;
+use petgraph::prelude::*;
+use petgraph::EdgeType;
+
+type Nt = f64;
```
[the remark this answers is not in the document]
georgios-ts: well, not really. my initial thought was that this could be an "adjustable" numerical precision. do you feel that i should explicitly use f64 in the code?

src/layout.rs
georgios-ts: no, i removed it.

src/layout.rs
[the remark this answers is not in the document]
mtreinish: It was more as I was reading the module I kind of skipped past the type definition at first and saw everything taking `Nt` where I would have expected `f64` and didn't know what the reason for aliasing it was. It's not a big deal I personally probably would have stuck with `f64` everywhere. But it's just code churn at this point to find replace `Nt` with `f64` so it's probably not worth updating.

mtreinish (commented):
One last thing and then I think this ready to merge could you add a test case that tests on a graph and digraph with holes in the node and edge indices. Just to complement the empty graph test. From my reading of the code it should work fine, but it'd be good to assert that in the tests. After that I think this is ready.

mtreinish (approved):

src/layout.rs
Chriscrosser3310: ah yeah, thanks for clarifying! I thought it might be confusing for the user that `scale = 1.0` and `scale = None` have different behaviors, but yeah it's not that big a problem and yours makes complete sense too. I think maybe it's just better to not use `Option` and set `scale = 1.0` in `retworkx/__init__.py`, so passing `None` wouldn't be allowed

PR #309, opened 2021-04-16
Add write_edge_list methods

mtreinish:
In #113 we added a constructor method to the graph classes to create a
graph object from a edge list file. However there was not a native way
to generate such a file from retworkx. This commit adds that function
with a new method, write_edge_list(), which can be used to write an edge
list file in the same format that can be read by read edge list.


georgios-ts:
LGTM, my only comment is that the documentation of `BufWriter` recommends calling `flush` before dropping.

mtreinish:
> LGTM, my only comment is that the documentation of `BufWriter` recommends calling `flush` before dropping.

Good call, I'll update that now

georgios-ts (approved):

PR #310, opened 2021-04-23
Add four simple layouts

Chriscrosser3310:
Related to #280. Four simplest layouts were added: bipartite, circular, shell, and spiral. A few things to note:

1. The rescale and recenter functions are duplicated with the ones in #306, we only need to keep one of each.
2. The layout functions return the new type Pos2DMapping added in #305.
3. The tests are performed as follows: generate a position dictionary from a layout function in [networkx](https://networkx.org/), and compare it to that generated by retworkx with the same layout function and inputs. If for all nodes, the difference between positions of the same nodes is smaller than some threshold (I chose 1E-6), the tests pass.


src/layouts.rs
```diff
@@ -0,0 +1,284 @@
...
+            let r = step * theta;
+            theta += chord / r;
+            pos.push([theta.cos() * r, theta.sin() * r]);
+        }
+    } else {
+        let mut angle: f64 = 0.0;
+        let mut dist = 0.0;
+        let step = 1.0;
+        let radius = node_num as f64;
+        while dist < radius {
```
georgios-ts, line 221: is this equivalent?
```suggestion
        for _ in 0..node_num {
```

src/layouts.rs
```diff
@@ -0,0 +1,284 @@
...
+
+fn recenter(pos: &mut Vec<Point>, center: Option<Point>) {
+    let num_pos = pos.len();
+    if num_pos == 0 {
+        return;
+    }
+    let dim = pos[0].len();
+    if let Some(center) = center {
+        for p in pos.iter_mut().take(num_pos) {
+            for (d, c) in center.iter().enumerate().take(dim) {
```
georgios-ts, line 250: not really needed.
```suggestion
        for p in pos.iter_mut() {
            for (d, c) in center.iter().enumerate() {
```

georgios-ts (commented):
Great work! 
My only comment is that you should consider updating all layout functions to handle graphs with holes in node indices.

src/layouts.rs
Chriscrosser3310: yes I guess they're used to be cases with stepsize < 1, but it's not really relevant here and the latter might be better

src/layout.rs
```diff
@@ -284,3 +286,215 @@ where
...
+}
+
+pub fn shell_layout<Ty: EdgeType>(
+    graph: &StableGraph<PyObject, PyObject, Ty>,
+    nlist: Option<Vec<Vec<usize>>>,
+    rotate: Option<f64>,
+    scale: Option<f64>,
+    center: Option<Point>,
+) -> Pos2DMapping {
+    let node_num = graph.node_count();
```
georgios-ts, line 410:
```suggestion
    let node_num = graph.node_bound();
```

src/layout.rs
```diff
@@ -284,3 +286,215 @@ where
...
+    let mut pos: Vec<Point> = vec![[0.0, 0.0]; node_num];
+    let pi = std::f64::consts::PI;
+
+    let shell_list: Vec<Vec<usize>> = match nlist {
+        Some(nlist) => nlist,
+        None => vec![(0..node_num).collect()],
```
georgios-ts, line 416:
```suggestion
        None => vec![graph.node_indices().map(|n| n.index()).collect()],
```

src/layout.rs
```diff
@@ -284,3 +286,215 @@ where
...
+        radius += radius_bump;
+        first_theta += rot_angle;
+    }
+
+    if let Some(center) = center {
+        recenter(&mut pos, center);
+    }
+
+    Pos2DMapping {
+        pos_map: graph.node_indices().map(|n| n.index()).zip(pos).collect(),
```
georgios-ts, line 451:
```suggestion
        pos_map: graph
            .node_indices()
            .map(|n| {
                let n = n.index();
                (n, pos[n])
            })
            .collect(),
```

src/layout.rs
```diff
@@ -20,6 +20,8 @@ use petgraph::graph::NodeIndex;
 use petgraph::prelude::*;
 use petgraph::EdgeType;
```
georgios-ts, line 22:
```suggestion
use petgraph::visit::NodeIndexable;

```

georgios-ts (commented):
This
```python
n = 5
graph = rx.generators.mesh_graph(n)
graph.remove_node(2)
rx.shell_layout(graph, nlist=[graph.node_indexes()])
```
will raise a panick exception.
I suggested some changes to fix this but feel free to proceed in your own way.

src/layout.rs
Chriscrosser3310: thanks! I certainly missed this part. I was trying to fix pos in the size of node_num, but it seems this is the easiest way to fix this, I'll just apply your suggestion

src/layout.rs
```diff
@@ -284,3 +287,221 @@ where
...
+            pos.push([dist * angle.cos(), dist * angle.sin()]);
+            dist += step;
+            angle += ros;
+        }
+
+        if let Some(scale) = scale {
+            rescale(&mut pos, scale, (0..node_num).collect());
+        }
+    }
+
```
georgios-ts, line 499:
```suggestion
    }
    
    if let Some(scale) = scale {
        rescale(&mut pos, scale, (0..node_num).collect());
    }

```

georgios-ts (approved):
Great! One comment left and then it should be ready to go

src/layout.rs
Chriscrosser3310: Nice catch, I forgot there's an `else if` -- but the problem of moving it out is that if the graph only has 1 node, `lim` in `rescale` will be `0.0` and the point will become `[NaN, NaN]`. I guess I can modify `rescale` so it only does something when `lim > 0`.

src/lib.rs
```diff
@@ -4091,6 +4091,217 @@ pub fn digraph_random_layout(
...
+#[pyfunction]
+#[text_signature = "(graph, /, scale=1, center=None, resolution=0.35, 
+                     equidistant=False)"]
+pub fn digraph_spiral_layout(
+    graph: &digraph::PyDiGraph,
+    scale: Option<f64>,
+    center: Option<layout::Point>,
+    resolution: Option<f64>,
+    equidistant: Option<bool>,
+) -> Pos2DMapping {
```
mtreinish, line 4301: Personally I would do something like:

```suggestion
#[pyfunction(scale="1.0", resolution="0.35", equidistant="false")]
#[text_signature = "(graph, /, scale=1, center=None, resolution=0.35, 
                     equidistant=False)"]
pub fn digraph_spiral_layout(
    graph: &digraph::PyDiGraph,
    scale: f64,
    center: Option<layout::Point>,
    resolution: f64,
    equidistant: bool,
) -> Pos2DMapping {
```

so we set the default in the macro generated python c api ffi function. But this also works fine

mtreinish (approved):
Overall this LGTM, great work! I have 2 nits inline but not really worth blocking over. I might just update the test one myself real quickly so we can merge this (because getting good error messages from the tests are useful).

PR #312, opened 2021-04-26
Add substitute_node_with_subgraph method to PyDiGraph

mtreinish:
This commit adds a new method substitute_node_with_subgraph that takes
in a node id and a PyDiGraph object and then procedes to replace the
specified node with the subgraph.

TODO:

- [x] Add tests
- [x] Add release notes
- [x] Verify functionality for use with qiskit-terra's `substitute_node_with_dag`


mtreinish:
Ok, I've updated this and I think it's ready for review now. There is a matching terra PR which shows this in use: https://github.com/Qiskit/qiskit-terra/pull/6302

nahumsa:
I think that the node index ordering of the substituted graph needs to be explicit in the documentation.

For instance, I have the `original_graph` being
orig_graph

and the `other_graph` being
other_graph

using `substitute_node_with_subgraph` on node `2`, we get
result

All new nodes start at the max value of the 'original_graph` node index, but this is not explicit in the documentation. Or maybe it could be useful to reindexing all nodes or even giving the choice for the user to choose the index of the added nodes.

Here is the code to reproduce what I said:
```python
import retworkx
from retworkx.visualization import mpl_draw
original_graph = retworkx.generators.directed_path_graph(5)
other_graph = retworkx.generators.directed_path_graph(2)
original_graph.substitute_node_with_subgraph(2, other_graph, lambda _, __, ___: 0)
mpl_draw(original_graph, with_labels=True)
```

mtreinish:
This is unavoidable because we can't preserve the indices between the 2 graphs as there might be overlap and we're essentially just calling `add_node()` for each node in the graph. But this is why the return from the function is a dict of old node indices in `other_graph` to it's new index in `original_graph` so you can run this and then use that dictionary  I will try to improve the documentation around this to make this clear.

For the second half I'm actually not actually sure how node index 2 is getting added there. I agree it should be 7, let me dig into that with your example locally and see if I can figure out what's going on

nahumsa:
> This is unavoidable because we can't preserve the indices between the 2 graphs as there might be overlap and we're essentially just calling `add_node()` for each node in the graph. But this is why the return from the function is a custom mapping class of old node indices in `other_graph` to it's new index in `original_graph` so you can run this and then use that dictionary I will try to improve the documentation around this to make this clear.

Sure, that makes sense. I forgot about the return, sorry about that.

releasenotes/notes/add-subgraph-replace-ec54792df1d641fb.yaml
```diff
@@ -0,0 +1,33 @@
...
+      mpl_draw(other_graph)
+
+    finally replace a node in the original graph with the second graph:
+
+    .. jupyter-execute::
+
+      def edge_map_fn(_source, _target, _weight):
+          return 0
+
+      original_graph.substitute_node_with_subgraph(2, other_graph, edge_map_fn)
```
nahumsa, line 32: Maybe it would be good to show on this example the `NodeMap` output

tests/digraph/test_substitute_node_with_subgraph.py
```diff
@@ -0,0 +1,150 @@
...
+        self.assertEqual({1: 5, 2: 6}, res)
+        expected = [
+            (5, 6, "migrated"),
+            (6, 4, None),
+            (6, 3, None),
+            (5, 2, None),
+            (5, 1, None),
+        ]
+        self.assertEqual(expected, graph.weighted_edge_list())
+
```
nahumsa, line 128: Maybe could add a case with where the substituted node have an inward and outward edge.

releasenotes/notes/add-subgraph-replace-ec54792df1d641fb.yaml
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/312/commits/681fcabb121593911950e257cabb96491173a6e2

tests/digraph/test_substitute_node_with_subgraph.py
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/312/commits/cca111b90121a94639911df788297429efa988b8

tests/digraph/test_substitute_node_with_subgraph.py
```diff
@@ -0,0 +1,185 @@
...
+    def test_none_mapping(self):
+        in_graph = retworkx.PyDiGraph()
+        in_graph.add_node(0)
+        in_graph.add_child(0, 1, "edge")
+        res = self.graph.substitute_node_with_subgraph(
+            2, in_graph, lambda _, __, ___: None
+        )
+        self.assertEqual([(0, 1), (3, 4), (5, 6)], self.graph.edge_list())
+        self.assertEqual(res, {0: 5, 1: 6})
```
nahumsa, line 82: Testing this locally. I got an issue similar to #347 which sometimes the output is `{0: 5, 1: 6}` and `{1: 6, 0: 5}`. Asserting equal to the dictionary doesn't actually catch this difference.
Does order matter for this case?

mtreinish: I don't think it matters too much here, because it's a mapping of nodes in the input subgraph to the equivalent node after they're added to the graph. But I will add it to the docstring of the function that it's unordered to make it clear (and so there are no surprises).

mtreinish: I added the documentation for this in https://github.com/Qiskit/retworkx/pull/312/commits/80d12164edbeee9c12917125be9b5d6573de1db8

nahumsa (approved):

src/digraph.rs
```diff
@@ -2217,6 +2217,161 @@ impl PyDiGraph {
...
+    /// :rtype: NodeMap
+    ///
+    /// .. note::
+    ///
+    ///    The return type is a :class:`retworkx.NodeMap` which is an unordered
+    ///    type. So it does not provide a deterministic ordering between objects
+    ///    when iterated over (although the same object will have a consistent
+    ///    order when iterated over multiple times).
+    ///
+    #[text_signature = "(self, node, other, edge_map_fn, /, node_filter=None, edge_weight_map=None)"]
```
1ucian0, line 2250: I think `edge_map_fn` should be optional.

releasenotes/notes/add-subgraph-replace-ec54792df1d641fb.yaml
```diff
@@ -0,0 +1,34 @@
...
+    :class:`~retwork.PyDiGraph` object. For example, first creating a new
+    graph:
+
+    .. jupyter-execute::
+
+      import retworkx
+      from retworkx.visualization import mpl_draw
+
+      original_graph = retworkx.generators.directed_path_graph(5)
+      mpl_draw(original_graph)
```
1ucian0, line 16: Because the feature refers to node labels, I would include the labels in this drawing.

```suggestion
      mpl_draw(original_graph, with_labels=True)
```

releasenotes/notes/add-subgraph-replace-ec54792df1d641fb.yaml
```diff
@@ -0,0 +1,34 @@
...
+    finally replace a node in the original graph with the second graph:
+
+    .. jupyter-execute::
+
+      def edge_map_fn(_source, _target, _weight):
+          return 0
+
+      node_mapping = original_graph.substitute_node_with_subgraph(2, other_graph, edge_map_fn)
+      print(node_mapping)
+      mpl_draw(original_graph)
```
1ucian0, line 34: [Similarly](https://github.com/Qiskit/retworkx/pull/312/files#r657937039) here

```suggestion
      mpl_draw(original_graph, with_labels=True)
```

src/digraph.rs
mtreinish: This is required right now because we without it we don't know what to do with edges to or from `node`. If we switched it to optional what do you think the default behavior should it was not set? The only option I could think of is to just drop all the edges into or out of `node` if this wasn't set, but if we did this it wouldn't really feel like it was substituting the node with a graph.

1ucian0: Is this a typo?
```suggestion
    #[text_signature = "(self, node, other, edge_map_fn, node_filter=None, edge_weight_map=None)"]
```

mtreinish: It's actually intentional, the `/` indicates the end of positional only arguments (it's part of the cPython interface and is actually a valid python syntax for >=3.8). See the docs for the rust python interface here: https://pyo3.rs/v0.9.2/function.html#making-the-function-signature-available-to-python which explains this.

1ucian0 (approved):

PR #313, opened 2021-04-29
Add Heavy square code to generators

nahumsa:
Related to #150.

Added implementation of [heavy hex](https://journals.aps.org/prx/abstract/10.1103/PhysRevX.10.011022), the input is the code distance, d. In the implementation I chose to add nodes referring to each type of qubit used in the Heavy Hex Code (data, syndrome, and flag) which _may cause the node order to be confusing_, but that was the best way that I figure out how to implement.


src/generators.rs
```diff
@@ -983,6 +983,139 @@ pub fn directed_grid_graph(
...
+            graph.add_edge(nodes_data[i * d + j], *flag, py.None());
+            graph.add_edge(*flag, nodes_data[i * d + j + 1], py.None());
+        }
+    }
+
+    // connect data and syndromes
+    for (i, syndrome_chunk) in nodes_syndrome.chunks(d).enumerate() {
+        if i % 2 == 0 {
+            graph.add_edge(
+                nodes_data[(i + 1) * d - 1],
```
ecpeterson, line 1058: This is correct, but it's written in a way I find less legible. This intends the last (i.e., `d-1`st) qubit in the `i`th column, which I'd prefer to describe as `i * d + (d - 1)`.

src/generators.rs
```diff
@@ -983,6 +983,139 @@ pub fn directed_grid_graph(
...
+                syndrome_chunk[syndrome_chunk.len() - 1],
+                py.None(),
+            );
+            graph.add_edge(
+                syndrome_chunk[syndrome_chunk.len() - 1],
+                nodes_data[(i + 2) * d - 1],
```
ecpeterson, line 1064: Same complaint here. I think the others are safe.

src/generators.rs
```diff
@@ -983,6 +983,139 @@ pub fn directed_grid_graph(
     })
 }
 
+/// Generate an undirected heavy square graph.
```
ecpeterson, line 986: I think it would benefit the reader a lot to inclue (1) a reference to the paper (e.g., Fig 6 of https://arxiv.org/pdf/1907.09528.pdf ) and (2) an ASCII diagram. Here's my attempt at artistry:

```
...       S   ...
   \     / \
... D   D   D ...
    |   |   |
... F-S-F-S-F-...
    |   |   |
... D   D   D ...
    |   |   |
... F-S-F-S-F-...
    |   |   |
    .........
    |   |   |
... D   D   D ...
     \ /     \
...   S       ...
```

ecpeterson (approved):
Rust is not my native language, so I have no comments about style, but the QEC-related stuff is all correct. I did leave some comments about legibility of intent.

ecpeterson:
Oh, one other comment: there is a variant of the heavy square lattice (see Fig 10.b, right-hand side) whose edge case handling (lines 1054-1075 in the present PR) is a bit different. No comment on whether it's worth including as an option, but it is definitely worth mentioning in the docstring.

src/generators.rs
ecpeterson: There is a variant of the heavy square lattice (see Fig 10.b, right-hand side) whose edge case handling (lines 1054-1075 in the present PR) is a bit different. No comment on whether it's worth including as an option, but it is definitely worth mentioning in the docstring.

src/generators.rs
```diff
@@ -983,6 +983,139 @@ pub fn directed_grid_graph(
...
+///
```
ecpeterson, line 987: There is a variant of the heavy square lattice (see Fig 10.b, right-hand side) whose edge case handling (lines 1054-1075 in the present PR) is a bit different. No comment on whether it's worth including as an option, but it is definitely worth mentioning in the docstring.

src/generators.rs
nahumsa: Thanks for pointing this out. I'll change it to be more legible.

src/generators.rs
nahumsa: Sure, that makes sense. I'll include both.

src/generators.rs
nahumsa: Do you have any suggestion how this can be mentioned in the docstring?

ecpeterson: The only important thing is to bring to the reader's attention that there is another variant and to disambiguate which you mean. So, something like:

```
NOTE: This function generates the four-frequency variant of the heavy square code. See Fig 10.b left of /1907.09528 , and compare Fig 10.b right.
```

nahumsa: The note was added on [`60a150b`](https://github.com/Qiskit/retworkx/pull/313/commits/60a150b0292713220e21a5d00d00935b736fdab9).

mtreinish (commented):
LGTM, thanks for doing this. Same as on heavy hex the only question I have before merging is do you think it makes sense to add a directed version too? If so I think that would be useful to have.

src/generators.rs
```diff
@@ -1072,6 +1072,371 @@ pub fn directed_binomial_tree_graph(
...
+    }
+
+    // connect flag and syndromes
+    for (i, syndrome_chunk) in nodes_syndrome.chunks(d).enumerate() {
+        if i % 2 == 0 {
+            for (j, syndrome) in syndrome_chunk.iter().enumerate() {
+                if j != syndrome_chunk.len() - 1 {
+                    graph.add_edge(
+                        nodes_flag[i * (d - 1) + j],
+                        *syndrome,
```
mtreinish, line 1379:
```suggestion
                        *syndrome,
                        nodes_flag[i * (d - 1) + j],
```

src/generators.rs
```diff
@@ -1072,6 +1072,371 @@ pub fn directed_binomial_tree_graph(
...
+                    );
+                    graph.add_edge(
+                        *syndrome,
+                        nodes_flag[(i + 1) * (d - 1) + j],
+                        py.None(),
+                    );
+                    if bidirectional {
+                        graph.add_edge(
+                            *syndrome,
+                            nodes_flag[i * (d - 1) + j],
```
mtreinish, line 1390:
```suggestion
                            nodes_flag[i * (d - 1) + j],
                            *syndrome,
```

src/generators.rs
```diff
@@ -1072,6 +1072,371 @@ pub fn directed_binomial_tree_graph(
...
+                        );
+                    }
+                }
+            }
+        } else if i % 2 == 1 {
+            for (j, syndrome) in syndrome_chunk.iter().enumerate() {
+                if j != 0 {
+                    graph.add_edge(
+                        nodes_flag[i * (d - 1) + j - 1],
+                        *syndrome,
```
mtreinish, line 1406:
```suggestion
                        *syndrome,
                        nodes_flag[i * (d - 1) + j - 1],
```

src/generators.rs
```diff
@@ -1072,6 +1072,371 @@ pub fn directed_binomial_tree_graph(
...
+                    );
+                    graph.add_edge(
+                        *syndrome,
+                        nodes_flag[(i + 1) * (d - 1) + j - 1],
+                        py.None(),
+                    );
+                    if bidirectional {
+                        graph.add_edge(
+                            *syndrome,
+                            nodes_flag[i * (d - 1) + j - 1],
```
mtreinish, line 1417:
```suggestion
                            nodes_flag[i * (d - 1) + j - 1],
                            *syndrome,
```

src/generators.rs
```diff
@@ -1072,6 +1072,371 @@ pub fn directed_binomial_tree_graph(
...
+    for (i, syndrome_chunk) in nodes_syndrome.chunks(d).enumerate() {
+        if i % 2 == 0 {
+            graph.add_edge(
+                nodes_data[i * d + (d - 1)],
+                syndrome_chunk[syndrome_chunk.len() - 1],
+                py.None(),
+            );
+            graph.add_edge(
+                syndrome_chunk[syndrome_chunk.len() - 1],
+                nodes_data[i * d + (2 * d - 1)],
```
mtreinish, line 1339:
```suggestion
                nodes_data[i * d + (2 * d - 1)],
                syndrome_chunk[syndrome_chunk.len() - 1],
```

src/generators.rs
```diff
@@ -1072,6 +1072,371 @@ pub fn directed_binomial_tree_graph(
...
+            );
+            if bidirectional {
+                graph.add_edge(
+                    syndrome_chunk[syndrome_chunk.len() - 1],
+                    nodes_data[i * d + (d - 1)],
+                    py.None(),
+                );
+                graph.add_edge(
+                    nodes_data[i * d + (2 * d - 1)],
+                    syndrome_chunk[syndrome_chunk.len() - 1],
```
mtreinish, line 1350:
```suggestion
                    syndrome_chunk[syndrome_chunk.len() - 1],
                    nodes_data[i * d + (2 * d - 1)],
```

src/generators.rs
```diff
@@ -1072,6 +1072,371 @@ pub fn directed_binomial_tree_graph(
...
+                    py.None(),
+                );
+            }
+        } else if i % 2 == 1 {
+            graph.add_edge(nodes_data[i * d], syndrome_chunk[0], py.None());
+            graph.add_edge(
+                syndrome_chunk[0],
+                nodes_data[(i + 1) * d],
```
mtreinish, line 1358:
```suggestion
                nodes_data[(i + 1) * d],
                syndrome_chunk[0],
```

src/generators.rs
```diff
@@ -1072,6 +1072,371 @@ pub fn directed_binomial_tree_graph(
...
+                py.None(),
+            );
+            if bidirectional {
+                graph.add_edge(syndrome_chunk[0], nodes_data[i * d], py.None());
+                graph.add_edge(
+                    nodes_data[(i + 1) * d],
+                    syndrome_chunk[0],
```
mtreinish, line 1365:
```suggestion
                    syndrome_chunk[0],
                    nodes_data[(i + 1) * d],
```

mtreinish (commented):
Looking at the figure 10 in the paper I think this has the same direction issues as #313 

Screenshot_2021-08-17_12-58-13
Screenshot_2021-08-17_12-58-35

Other than that I think this is good to go.

mtreinish (approved):

mtreinish:
LGTM, just as on #293 I took care of applying the last suggestions and fixing the tests to include this in the release. Thanks for adding these generators @nahumsa both this and heavy hex will be really nice to integrate with qiiskit-terra's `CouplingMap` after 0.10.0 goes out so we can have an a qiskit api to generate heavy hex and heavy square coupling maps.

PR #316, opened 2021-05-07
Format code with black

IvanIsCoding:
Format Python code with [black](https://github.com/psf/black) to align code formatting with [recent changes to other Qiskit repositories](https://github.com/Qiskit/qiskit-terra/pull/6361).

This formatting changes a massive amount of test files, so I will highlight the files that actually need to be reviewed:
*  `tox.ini` for adding `tox -eblack` and making `flake8` agree with `black`
* `pyproject.toml` for adding `black` settings
* `CONTRIBUTING.md` with instructions to run `black` for new contributors
* `.github/workflows/main.yml` for checking `black` codestyle in CI

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,6 @@
 [build-system]
 requires = ["setuptools", "wheel", "setuptools-rust"]
+
+[tool.black]
+line-length = 100
+target-version = ['py36', 'py37', 'py38', 'py39']
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,5 +2,5 @@
 requires = ["setuptools", "wheel", "setuptools-rust"]
 
 [tool.black]
-line-length = 100
+line-length = 80
 target-version = ['py36', 'py37', 'py38', 'py39']
--- a/tests/digraph/test_pred_succ.py
+++ b/tests/digraph/test_pred_succ.py
@@ -150,9 +150,21 @@ class TestBfsSuccessors(unittest.TestCase):
         node_j = dag.add_child(node_i, 9, {})
         dag.add_child(node_j, 10, {})
         res = {n: sorted(s) for n, s in retworkx.bfs_successors(dag, node_b)}
-        expected = {1: [2], 2: [3, 7], 3: [4], 4: [5], 5: [6], 7: [8], 8: [9], 9: [10]}
+        expected = {
+            1: [2],
+            2: [3, 7],
+            3: [4],
+            4: [5],
+            5: [6],
+            7: [8],
+            8: [9],
+            9: [10],
+        }
         self.assertEqual(expected, res)
-        self.assertEqual([(7, [8]), (8, [9]), (9, [10])], retworkx.bfs_successors(dag, node_h))
+        self.assertEqual(
+            [(7, [8]), (8, [9]), (9, [10])],
+            retworkx.bfs_successors(dag, node_h),
+        )
 
     def test_bfs_successors_sequence(self):
         dag = retworkx.PyDAG()
--- a/tests/digraph/test_pred_succ.py
+++ b/tests/digraph/test_pred_succ.py
@@ -18,85 +18,123 @@ import retworkx
 class TestPredecessors(unittest.TestCase):
     def test_single_predecessor(self):
         dag = retworkx.PyDAG()
-        node_a = dag.add_node('a')
-        dag.add_child(node_a, 'b', {'a': 1})
-        node_c = dag.add_child(node_a, 'c', {'a': 2})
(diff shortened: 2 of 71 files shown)


tox.ini
```diff
@@ -22,10 +22,12 @@ commands =
 [testenv:lint]
 basepython = python3
 deps =
+  black
   flake8
   setuptools-rust
 whitelist_externals=cargo
 commands =
+  black --check {posargs} '../retworkx' '../tests'
```
mtreinish, line 30: Just to make it a bit more consistent with `cargo fmt` (so the diff shows up)

```suggestion
  black --check --diff {posargs} '../retworkx' '../tests'
```

pyproject.toml
```diff
@@ -1,2 +1,6 @@
 [build-system]
 requires = ["setuptools", "wheel", "setuptools-rust"]
+
+[tool.black]
+line-length = 100
```
mtreinish, line 5: I don't feel too strongly about this, but I have the rustfmt settings at 80 characters because of my personal preference (I do most of my dev in 80 character wide terminals). If we want to go for 100 characters here we probably should update the rustfmt config in a follow up PR.

tests/digraph/test_pred_succ.py
```diff
@@ -18,85 +18,123 @@
...
+                        {"numeral": 4},
+                        {"numeral": 3},
+                        {"numeral": 2},
+                        {"numeral": 1},
+                        {"numeral": 0},
+                    ],
+                )
+            ],
+            res,
+        )
```
mtreinish, line 137: Wow this is pretty ugly.

tox.ini
```diff
@@ -39,11 +41,20 @@ changedir = {toxinidir}/docs
 commands =
   sphinx-build -W -b html source build/html {posargs}
 
+[testenv:black]
```
mtreinish, line 44: Can we share the venv here between black and lint by adding a line:
```
envdir = .tox/lint
```

to both job definitions? Right now we're building 2 venvs to just run black while these 2 jobs basically run the same thing (just save a little install time)

tox.ini
```diff
@@ -22,10 +22,12 @@ commands =
 [testenv:lint]
 basepython = python3
 deps =
+  black
```
mtreinish, line 25: We should pin the black version here, new releases of black can change the behavior which will cause CI failures and also a potentially a ping pong issue with local environments if there is a mismatch between versions. So we want to make sure everyone is running the same version of black.

```suggestion
  black==21.5b0
```

tox.ini
```diff
@@ -39,11 +41,20 @@ changedir = {toxinidir}/docs
 commands =
   sphinx-build -W -b html source build/html {posargs}
 
+[testenv:black]
+basepython = python3
+deps =
+  black
```
mtreinish, line 47:
```suggestion
  black==21.5b0
```

mtreinish (commented):
Overall this LGTM. Personally I've resisted using black here because I personally really don't like some of it's style choices it makes. But, I think we've already (for quite some time) reached the point where retworkx is no longer solely my domain. With that in mind I think using black to handle the code formatting makes sense since it simplifies the code formatting for contribution in the same way rustfmt does (also the style is a closer match to rustfmt).

 I just left a couple of comments inline. Of those comments most are just idle thoughts or questions for the future (mostly around line length) the only required changes are in the tox.ini.
Aside from the inline comments though, you will also need to manually update the lint github actions job. We don't use tox for that job because I wanted to run `cargo` directly for the well formatted compiler error message and we already build twice, so doing it again for tox seemed unnecessary. I guess you could update the job either to just explicitly run `black --check --diff` or change the lint jobs to be `tox --skip-pkg-install -elint` either would work (the later might be better to be more consistent between local and ci, especially with a pinned black version).

IvanIsCoding:
I have addressed some comments and:
* `black` version is now fixed
* `black` is executed in CI. I opted to add a new command because I found it simpler
* Line lenght is now 80

With regards to line lenght: I agree with the idea of using the same line lenght for `cargo fmt` and `black`, consistency is a great idea!

It is easier though to set `black` to use line lenght of 80 than to set `cargo fmt` to use line length of 100. We have way more Rust code than Python code, so I tried to be mindful and minimize the diff. Also, most of our Rust code is core functions while most of our Python code is test cases. So let's keep Rust as it is 😃

tests/digraph/test_pred_succ.py
IvanIsCoding: Perhaps the beauty is that we didn't have to format it ourselves!

mtreinish (approved):

PR #317, opened 2021-05-08
Add `is_subgraph_isomorphic` function

georgios-ts:
Adds a function to determine if two graphs are induced subgraph isomorphic. 
This new feature is simple to implement since it's the same code with `is_isomorphic` function except in feasibility rules we check `>=` relation instead of `==`.

### DIFF
--- a/retworkx/__init__.py
+++ b/retworkx/__init__.py
@@ -621,6 +621,69 @@ def _graph_is_isomorphic_node_match(first, second, matcher, id_order=True):
 
 
 @functools.singledispatch
+def is_subgraph_isomorphic(
+    first, second, node_matcher=None, edge_matcher=None, id_order=True
+):
+    """Determine if 2 graphs are subgraph isomorphic
+
+    This checks if 2 graphs are subgraph isomorphic both structurally and also
+    comparing the node and edge data using the provided matcher functions.
+    The matcher functions take in 2 data objects and will compare them. A
+    simple example that checks if they're just equal would be::
+
+            graph_a = retworkx.PyGraph()
+            graph_b = retworkx.PyGraph()
+            retworkx.is_subgraph_isomorphic(graph_a, graph_b,
+                                            lambda x, y: x == y)
+
+    .. note::
+
+        For better performance on large graphs, consider setting
+        `id_order=False`.
+
+    :param first: The first graph to compare. Can either be a
+        :class:`~retworkx.PyGraph` or :class:`~retworkx.PyDiGraph`.
+    :param second: The second graph to compare. Can either be a
+        :class:`~retworkx.PyGraph` or :class:`~retworkx.PyDiGraph`.
+        It should be the same type as the first graph.
+    :param callable node_matcher: A python callable object that takes 2
+        positional one for each node data object. If the return of this
+        function evaluates to True then the nodes passed to it are viewed
+        as matching.
+    :param callable edge_matcher: A python callable object that takes 2
+        positional one for each edge data object. If the return of this
+        function evaluates to True then the edges passed to it are viewed
+        as matching.
+    :param bool id_order: If set to ``False`` this function will use a
+        heuristic matching order based on [VF2]_ paper. Otherwise it will
+        default to matching the nodes in order specified by their ids.
+
+    :returns: ``True`` if there is a subgraph of `first` isomorphic to `second`
+        , ``False`` if there is not.
+    :rtype: bool
+    """
+    raise TypeError("Invalid Input Type %s for graph" % type(first))
+
+
+@is_subgraph_isomorphic.register(PyDiGraph)
+def _digraph_is_subgraph_isomorphic(
+    first, second, node_matcher=None, edge_matcher=None, id_order=True
+):
+    return digraph_is_subgraph_isomorphic(
+        first, second, node_matcher, edge_matcher, id_order
+    )
+
+
+@is_subgraph_isomorphic.register(PyGraph)
+def _graph_is_subgraph_isomorphic(
+    first, second, node_matcher=None, edge_matcher=None, id_order=True
+):
+    return graph_is_subgraph_isomorphic(
+        first, second, node_matcher, edge_matcher, id_order
+    )
+
+
+@functools.singledispatch
 def transitivity(graph):
     """Compute the transitivity of a graph.
 
--- a/src/isomorphism.rs
+++ b/src/isomorphism.rs
@@ -346,65 +346,6 @@ where
     new_graph
 }
 
-/// [Graph] Return `true` if the graphs `g0` and `g1` are isomorphic.
-///
-/// Using the VF2 algorithm, examining both syntactic and semantic
-/// graph isomorphism (graph structure and matching node and edge weights).
-///
-/// The graphs should not be multigraphs.
-pub fn is_isomorphic<Ty, F, G>(
-    py: Python,
-    g0: &StablePyGraph<Ty>,
-    g1: &StablePyGraph<Ty>,
-    mut node_match: Option<F>,
-    mut edge_match: Option<G>,
-    id_order: bool,
-) -> PyResult<bool>
-where
-    Ty: EdgeType,
-    F: FnMut(&PyObject, &PyObject) -> PyResult<bool>,
-    G: FnMut(&PyObject, &PyObject) -> PyResult<bool>,
-{
-    let mut inner_temp_g0: StablePyGraph<Ty>;
-    let mut inner_temp_g1: StablePyGraph<Ty>;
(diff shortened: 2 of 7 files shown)


src/isomorphism.rs
```diff
@@ -346,13 +346,49 @@ where
...
+    #[inline]
+    fn enabled(&self) -> bool {
+        self.is_some()
+    }
+    #[inline]
+    fn eq(&mut self, a: &T, b: &T) -> PyResult<bool> {
+        let res = (self.as_mut().unwrap())(a, b)?;
+        Ok(res)
+    }
+}
```
mtreinish, line 380: Is `NoSemanticMatch` actually being used anywhere? I know it pre-existed prior to this PR (and was originally just from the fork of petgraph's module) and you're just moving it around. But maybe we should just delete it if it's not needed.

src/lib.rs
```diff
@@ -426,6 +426,142 @@ fn graph_is_isomorphic(
...
+///     function evaluates to True then the edges passed to it are vieded
+///     as matching.
+/// :param bool id_order: If set to ``False`` this function will use a
+///     heuristic matching order based on [VF2]_ paper. Otherwise it will
+///     default to matching the nodes in order specified by their ids.
+///
+/// :returns: ``True`` if there is a subgraph of `first` isomorphic to `second`,
+///     ``False`` if there is not.
+/// :rtype: bool
+#[pyfunction(id_order = "true")]
```
mtreinish, line 462: Since this is a new function we do have the flexability to change the default here and enable the vf2++ heuristic by default. I'll defer to what you think is best, either consistency with `is_isomorphic` or having the generally faster heuristic used by default.

src/isomorphism.rs
georgios-ts: no, we can safely delete it.

IvanIsCoding (commented):
This is great, `is_subgraph_isomorphic` reuses almost all `is_isomorphic` code which in my opinion is an indicator of good design!

Just a minor debate: should we use a macro and pass `$op`, or should we go with a more standard approach and use an [Ordering](https://doc.rust-lang.org/std/cmp/enum.Ordering.html) argument?

I am biased towards passing an argument, it feels closer to how we write our Python code.

src/isomorphism.rs
```diff
@@ -346,13 +346,36 @@ where
...
+
+macro_rules! impl_isomorphic {
+($is_isomorphic:ident, $try_match:ident, $op:tt) => {
+
+/// [Graph] Return `true` if the graphs `g0` and `g1` are (sub) graph isomorphic.
```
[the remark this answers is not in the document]
georgios-ts: Well, i see your point. I used a macro since it was fun, fast (required less coding) and it's not straight forward to use an `Ordering` here, since in `is_isomorphic` we would check:
```rust
a.cmp(&b) == Equal
```
but in `is_subgraph_isomorphic`:
```rust
a.cmp(&b) == Equal or Greater
```

> Is it idiomatic to use Rust macros like that?

Honestly, i don't know but i have seen similar things in petgraph.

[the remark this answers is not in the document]
IvanIsCoding: > Well, i see your point. I used a macro since it was fun, fast (required less coding) and it's not straight forward to use an `Ordering` here, since in `is_isomorphic` we would check:
> 
> ```rust
> a.cmp(&b) == Equal
> ```
> 
> but in `is_subgraph_isomorphic`:
> 
> ```rust
> a.cmp(&b) == Equal or Greater
> ```
> 
> > Is it idiomatic to use Rust macros like that?
> 
> Honestly, i don't know but i have seen similar things in petgraph.

You can overcome the `Equal or Greater` issue using [Ordering.then](https://doc.rust-lang.org/std/cmp/enum.Ordering.html#method.then):

```rust
a.cmp(&b).then(expected_ordering) == expected_ordering
```
If `a.cmp(&b)` happens to be `Orderign::Equal`, it will be converted to `expected_ordering`. Does this help making it easier to write the code without the macro?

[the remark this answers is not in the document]
georgios-ts: Yes it does. I will update it

retworkx/__init__.py
```diff
@@ -706,6 +706,69 @@ def _graph_is_isomorphic_node_match(first, second, matcher, id_order=True):
     return graph_is_isomorphic(first, second, matcher, id_order=id_order)
 
 
+@functools.singledispatch
+def is_subgraph_isomorphic(
+    first, second, node_matcher=None, edge_matcher=None, id_order=True
```
mtreinish, line 711:
```suggestion
    first, second, node_matcher=None, edge_matcher=None, id_order=False
```

retworkx/__init__.py
```diff
@@ -706,6 +706,69 @@ def _graph_is_isomorphic_node_match(first, second, matcher, id_order=True):
...
+
+            graph_a = retworkx.PyGraph()
+            graph_b = retworkx.PyGraph()
+            retworkx.is_subgraph_isomorphic(graph_a, graph_b,
+                                            lambda x, y: x == y)
+
+    .. note::
+
+        For better performance on large graphs, consider setting
+        `id_order=False`.
```
mtreinish, line 728:
```suggestion
```

retworkx/__init__.py
```diff
@@ -706,6 +706,69 @@ def _graph_is_isomorphic_node_match(first, second, matcher, id_order=True):
...
+    :returns: ``True`` if there is a subgraph of `first` isomorphic to `second`
+        , ``False`` if there is not.
+    :rtype: bool
+    """
+    raise TypeError("Invalid Input Type %s for graph" % type(first))
+
+
+@is_subgraph_isomorphic.register(PyDiGraph)
+def _digraph_is_subgraph_isomorphic(
+    first, second, node_matcher=None, edge_matcher=None, id_order=True
```
mtreinish, line 756:
```suggestion
    first, second, node_matcher=None, edge_matcher=None, id_order=False
```

retworkx/__init__.py
```diff
@@ -706,6 +706,69 @@ def _graph_is_isomorphic_node_match(first, second, matcher, id_order=True):
...
+):
+    return digraph_is_subgraph_isomorphic(
+        first, second, node_matcher, edge_matcher, id_order
+    )
+
+
+@is_subgraph_isomorphic.register(PyGraph)
+def _graph_is_subgraph_isomorphic(
+    first, second, node_matcher=None, edge_matcher=None, id_order=True
```
mtreinish, line 765:
```suggestion
    first, second, node_matcher=None, edge_matcher=None, id_order=False
```

retworkx/__init__.py
```diff
@@ -706,6 +706,69 @@ def _graph_is_isomorphic_node_match(first, second, matcher, id_order=True):
...
+    :param callable node_matcher: A python callable object that takes 2
+        positional one for each node data object. If the return of this
+        function evaluates to True then the nodes passed to it are viewed
+        as matching.
+    :param callable edge_matcher: A python callable object that takes 2
+        positional one for each edge data object. If the return of this
+        function evaluates to True then the edges passed to it are viewed
+        as matching.
+    :param bool id_order: If set to ``False`` this function will use a
+        heuristic matching order based on [VF2]_ paper. Otherwise it will
```
mtreinish, line 744: Is this footnote defined somewhere?

retworkx/__init__.py
```diff
@@ -706,6 +706,69 @@ def _graph_is_isomorphic_node_match(first, second, matcher, id_order=True):
...
+        It should be the same type as the first graph.
+    :param callable node_matcher: A python callable object that takes 2
+        positional one for each node data object. If the return of this
+        function evaluates to True then the nodes passed to it are viewed
+        as matching.
+    :param callable edge_matcher: A python callable object that takes 2
+        positional one for each edge data object. If the return of this
+        function evaluates to True then the edges passed to it are viewed
+        as matching.
+    :param bool id_order: If set to ``False`` this function will use a
```
mtreinish, line 743: It might be good to reverse this descriptions since this defaults to the vf2++ heuristic order unlike the other functions. (also in the the per type functions)

mtreinish (commented):
This lgtm, just some issues with the docstring for the universal function and the default for `id_order` python is different from the rust functions we probably want it to be consistent.

IvanIsCoding (approved):
LGTM. The diff is larger using `Ordering`, but ultimately I believe it will be easier to maintain.

retworkx/__init__.py
georgios-ts: it's defined in the docstring of `is_isomorphic`. do i have to redefine it here too?

mtreinish (approved):

PR #323, opened 2021-05-18
Add edge_indices methods to graph objects

mtreinish:
This commit adds a new method to the PyDiGraph (and PyDAG) and PyGraph
classes, node_indices, which returns a list of a edge indices for all
edges in the graph.


src/iterators.rs
```diff
@@ -1326,3 +1326,118 @@ impl PyIterProtocol for PathLengthMappingItems {
...
+/// that is integer indexed. If you want to use it as an iterator you
+/// can by wrapping it in an ``iter()`` that will yield the results in
+/// order.
+///
+/// For example::
+///
+///     import retworkx
+///
+///     graph = retworkx.generators.directed_path_graph(5)
+///     edges = retworkx.edge_indexes(0)
```
ajavadia, line 1344:
```suggestion
///     edges = retworkx.edge_indices(0)
```

ajavadia (commented):
A few questions::
- Are edge indices static identifiers for unique edges? What happens if I remove an edge, do others maintain their edge indices or get re-indexed? (I think this should be documented)

- Are edge indices assigned in the order the edges were created? If so then I think they are not properties of the graph, i.e. two identical graphs could have different edge indices if they were created in different orders.

ajavadia (approved):

PR #324, opened 2021-05-18
Add edge_index_map() method to graph classes

mtreinish:
This commit adds a new function edge_index_map() which will return a
mapping from edge indices to a tuple of the node indices and weight for
the edge.

TODO:

- [x] Add tests
- [x] Add release notes


releasenotes/notes/edge-index-map-cf07a035d02481a1.yaml
```diff
@@ -0,0 +1,14 @@
+---
+features:
+  - |
+    Added a new method, :meth:`~retworkx.PyDiGraph.edge_index_map`, to the
+    :class:`~retworkx.PyDiGraph` and :class:`~retworkx.PyGraph`
+    (:meth:`~retworkx.PyGraph.edge_index_map`) that will return a read-only
+    mapping of edge indices to a tuple of the form
+    ``(node index, node_index, weight/data payload)`` for every edge in the
```
ajavadia, line 8:
```suggestion
    ``(source_node_index, target_node_index, weight/data payload)`` for every edge in the
```

releasenotes/notes/edge-index-map-cf07a035d02481a1.yaml
```diff
@@ -0,0 +1,14 @@
...
+    graph object.
+  - |
+    Added a new custom return type :class:`~retworkx.EdgeIndexMap` which is
+    returned by :class:`retworkx.PyDiGraph.edge_index_map` and
```
ajavadia, line 12:
```suggestion
    returned by :meth:`retworkx.PyDiGraph.edge_index_map` and
```

releasenotes/notes/edge-index-map-cf07a035d02481a1.yaml
```diff
@@ -0,0 +1,14 @@
...
+    :class:`retworkx.PyGraph.edge_index_map`. It is equivalent to a read-only
```
ajavadia, line 13:
```suggestion
    :meth:`retworkx.PyGraph.edge_index_map`. It is equivalent to a read-only
```

releasenotes/notes/edge-index-map-cf07a035d02481a1.yaml
```diff
@@ -0,0 +1,14 @@
...
+    dict/mapping that represent a mapping of edge indices to the edge.
```
ajavadia, line 14: so is the correct way to identify an edge in retworkx to use the tuple (node1, node2, weight)? There's no Edge class or something?

mtreinish: Yeah more or less, the unique identifier for an edge is it's id/index. But other than that it's primarily represented by the tuple. There is not a dedicated class for an edge or node (mostly because we let the data payload for either be any python object).

mtreinish: Although I did intend for this sentence to be `edge tuple` instead of `edge` (I just made a typo).

ajavadia (approved):

PR #327, opened 2021-05-19
Add pydot drawer function

### LINKED ISSUE #321: Add pydot based drawer

In #304 we're adding a matplotlib based drawer to retworkx, which is standalone python code that just builds on the retworkx functionality and matplotlib is an optional dependency. Since retworkx has had a graphviz dot generator since #103 we should expand the built in visualizers to have `pydot_draw()` function in the new `retworkx.visualization` module (after #304 merges).
This would basically just be a prebuilt function version of the docstring we have on all the generator functions. Something like:

```python3
import os
import tempfile

import pydot 
from PIL import Image 

import retworkx.generators

def pydot_draw(graph):
    graph = retworkx.generators.directed_cycle_graph(5)
    dot_str = graph.to_dot(
        lambda node: dict(
            color='black', fillcolor='lightblue', style='filled'))
    dot = pydot.graph_from_dot_data(dot_str)[0]

    with tempfile.TemporaryDirectory() as tmpdirname:
        tmp_path = os.path.join(tmpdirname, 'dag.png')
        dot.write_png(tmp_path)
        image = Image.open(tmp_path)
        os.remove(tmp_path)
    return image
```

where obviously we either options for callback functions passed to `to_dot()` and different io options (like filename, file type, etc). Then we'll also need to add pydot and pillow as optional dependencies.

mtreinish:
This commit expands the retworkx.visualization module to include a new
pydot based drawer. This function wraps the pydot and pillow libraries
to generate graph visualizations leveraging graphviz. It internally uses
the to_dot() method of the retworkx graph types and then passes that dot
file to pydot to generate a visualization with graphviz. The result is
wrapped as a pillow image object.

Fixes #321

TODO:

- [x] Expand tests to have better coverage
- [x] Configure linux CI jobs to install graphviz so we can run the tests.
- [x] Add kwarg for pydot's `prog` kwarg to select between the different graphviz commands


retworkx/visualization/pydot.py
```diff
@@ -0,0 +1,99 @@
...
+    dot = pydot.graph_from_dot_data(dot_str)[0]
+    if image_type is None:
+        output_format = "png"
+    else:
+        output_format = image_type
+
+    if not filename:
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            tmp_path = os.path.join(tmpdirname, "dag.png")
+            dot.write(tmp_path, format=output_format)
```
georgios-ts, line 92: Might be good if we have an extra `prog` argument here to choose between different layouts implemented in graphviz, see ([1](https://github.com/pydot/pydot/blob/master/pydot.py#L1794), [2](https://github.com/pydot/pydot/blob/master/pydot.py#L1862-L1889))

mtreinish: Good call, I've added it to the TODO list

tests/visualization/test_pydot.py
```diff
@@ -0,0 +1,144 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+# Based on the equivalent tests for the networkx matplotlib drawer:
+# https://github.com/networkx/networkx/blob/ead0e65bda59862e329f2e6f1da47919c6b07ca9/networkx/drawing/tests/test_pylab.py
+
```
mtreinish, line 15:
```suggestion
```

retworkx/visualization/pydot.py
```diff
@@ -0,0 +1,116 @@
...
+        prog = "dot"
+    else:
+        prog = method
+
+    if not filename:
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            tmp_path = os.path.join(tmpdirname, "dag.png")
+            dot.write(tmp_path, format=output_format, prog=prog)
+            with Image.open(tmp_path) as temp_image:
+                image = temp_image.copy()
```
kdk, line 111: Is this a work around for https://github.com/Qiskit/qiskit-terra/issues/6005 ?

tests/visualization/test_pydot.py
```diff
@@ -0,0 +1,141 @@
...
+import unittest
+
+import retworkx
+from retworkx.visualization import pydot_draw
+
+try:
+    import pydot
+    import PIL
+
+    pydot.call_graphviz("dot", ["--version"], tempfile.gettempdir())
```
kdk, line 24: Should this also be in the `HAS_PYDOT` check of ` retworkx/visualization/pydot.py`? (In this case, the would the current code raise at runtime from `pydot.graph_from_dot_data`?)

kdk (commented):
Looks good, two small questions below. Should this have a release note?

retworkx/visualization/pydot.py
mtreinish: Yeah I think so, although I saw the fix first from this issue: https://github.com/Qiskit/qiskit-terra/issues/6256

tests/visualization/test_pydot.py
mtreinish: I can add it over there, that might not be a bad idea. Right now it will error when we call `pydot.write()` because that's where graphviz gets called. But the error isn't actually helpful, so it's probably more user friendly to detect graphviz is missing and raise a meaningful error.

mtreinish:
> Looks good, two small questions below. Should this have a release note?

Yeah it should have a release note, I'm pretty sure I wrote one I probably just forgot to commit it. I'll add it (or rewrite it if it's lost)

tests/visualization/test_pydot.py
kdk: :+1: if we can make the error message more meaningful. Realizing now though that this might have an impact on import performance, so worth weighing the best option..

mtreinish: Well I didn't include the visualization submodule in the default import path on purpose because of this so it's not critical. But I was also thinking of just doing it at runtime (since it depends on having pydot present).

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/327/commits/f346cb1a7d1ffa7f3654645cac1ce957dd07358b

kdk (approved):
Looks good, thanks for the updates.

PR #329, opened 2021-05-21
Add matplotlib to read the docs builds

mtreinish:
Since #304 merged building the retworkx documentation has required that
matplotlib be installed for generating the matplotlib visualizations in
the documentation of `mpl_draw()`. However that PR neglected to add it
to the read the docs conda environment configuration. This commit
corrects this oversight to hopefully fix the documentation builds on
read the docs.

PR #331, opened 2021-05-21
Add missing doc theme template file

mtreinish:
This commit adds a missing template file for the qiskit-sphinx-theme.
PR #329 previously attempted to fix the RTD docs builds and update the
theme formatting. While that worked fine for PR mode after it merged the
builds for the latest version of the docs did not work and fail with an
error about the missing template file. This will hopefully fix the
builds so we can see the rendered docs for the current state of the main
branch.


mtreinish:
It built for rtd successfully: https://retworkx--331.org.readthedocs.build/en/331/ although this by itself isn't an indication that this PR fixes the issue. The failures only seem to occur on builds from `main` for the latest version of the hosted docs.

PR #335, opened 2021-05-26
Add methods to get number of nodes and edges in the graph

mtreinish:
This commit adds two new methods num_nodes() and num_edges() to the
graph classes for returning the number of nodes and edges respectively
in the graph. While a user could do len(graph) to get the number of
nodes having an explicit method also is useful as it might not be
obvious that you can use len() on the graph.


mtreinish:
The one thing I wasn't sure here was whether we should make these attributes (so add a `#[getter]` macro to the functions) or leave them as methods that need to be called. Part of me thinks it makes more sense as an attribute, but that would be different from the equivalent in networkx. @t-imamichi do you have a preference?

t-imamichi (approved):
LGTM

PR #336, opened 2021-05-26
[WIP] Add frozen attribute to graph classes

mtreinish:
This commit adds a new attribute to a graph class, frozen, which when
set to True will make a graph object immutable. When it's set to True
all methods that mutate a graph object raise an exception. The one
exception is that all weight/data payload objects are stored and
returned by reference in the graph objects and if they're modified
in python there is no way to control that from retworkx.

TODO:

- [ ] Add tests
- [ ] Add release note

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -168,6 +168,7 @@ Exceptions
    retworkx.NoSuitableNeighbors
    retworkx.NoPathFound
    retworkx.NullGraph
+   retworkx.FrozenError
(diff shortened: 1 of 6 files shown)


mtreinish:
I'm going to close this for now, I think not being able to prevent mutability of returned data payloads/weights kind defeats the purpose of being able to make a frozen graph. We can maybe revisit this in the future if there is a use case for this.

PR #343, opened 2021-05-28
Add section to retworkx for networkx users guide on matrix converters

### LINKED ISSUE #342: Add `to_numpy_array` or `to_numpy_matrix`

It would be great if retworkx supports `to_numpy_array` or `to_numpy_matrix`.
It is used in a tutorial of qiskit-optimization.
https://github.com/Qiskit/qiskit-optimization/blob/main/docs/tutorials/06_examples_max_cut_and_tsp.ipynb
It is not urgent because we can implement a utility function in Python.

Related to https://github.com/Qiskit/qiskit-optimization/pull/150

mtreinish:
This commit adds documentation to the retworkx for networkx users guide
on the differences between networkx and retworkx when it comes to matrix
generation. Right now retworkx only has an `adjacency_matrix()` function
which is equivalent to networkx's `to_numpy_array()` and returns a numpy
array of the adjacency matrix. This is different from networkx's
`adjacency_matrix()` function which returns a scipy sparse array (and is
just an alias for networkx's `to_scipy_sparse_matrix()`). Documenting
this gap is important because it's easy to get confused by these API
differences.

This difference is fundamentally because we can effeciently create numpy
arrays in retworkx by just pass a pointer to numpy with a 2d array
that was constructed natively in rust (using rust-numpy) while other
matrix formats (such as scipy sparse matrices) don't expose a C api so
to create those objects we'd have to go through Python. Eventually we
might expand the API to include pure python functions to do additional
conversions (although I worry about increasing the external dependency
list) but numpy arrays will still be the default because of the inherent
performance benefits.

Fixes #342

### DIFF
--- a/docs/source/networkx.rst
+++ b/docs/source/networkx.rst
@@ -285,6 +285,27 @@ Graph Modifiers
 (note the retworkx version links to the :class:`~retworkx.PyDiGraph` version,
 but there are also equivalent :class:`~retworkx.PyGraph` methods available)
 
+Matrix Converter Functions
+--------------------------
+
+NetworkX has several functions for going back and forth between a NetworkX
+graph and matrices in other libraries. This includes ``to_numpy_matrix()``,
+``to_numpy_array()``, ``to_numpy_recarray()``, ``to_scipy_sparse_matrix()``,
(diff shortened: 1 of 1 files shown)


t-imamichi (approved):
LGTM. Thanks!

PR #344, opened 2021-05-28
Add visualization to retworkx for networkx users guide

### LINKED ISSUE #341: Rename `edge_list` arg with `edgelist` to be compatible with networkx

`mpl_draw` has an argument `edge_list` https://github.com/Qiskit/retworkx/blob/885f2de1588dc983591563f94b2ddcada443b3b2/retworkx/visualization/matplotlib.py#L101

The same argument has name `edgelist` in networkx
https://networkx.org/documentation/stable/reference/generated/networkx.drawing.nx_pylab.draw_networkx_edges.html

Related to https://github.com/Qiskit/qiskit-optimization/pull/150

mtreinish:
This commit adds a section to the retworkx for networkx users guide on
visualizations. This is important especially for the ``mpl_draw``
function since it's mostly identical to the ``networkx_drawer`` function
but there are some key differences to keep in mind.

This commit also makes small change to the graphviz drawer, the name of
the function is renamed from `pydot_draw` to `graphviz_draw` and the
Python module is renamed from `pydot` to `graphviz`. This was done to
be more descriptive as graphviz is the backend not pydot. It also makes
it easier to migrate away from pydot in the future if we needed to since
the key piece is using graphviz not a particular python interface to it.

Fixes #341

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ retworkx/*pyd
 *.stestr/
 *.ps
 *.png
+*.svg
+*.jpg
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -142,6 +142,8 @@ type functions in the algorithms API but can be run with a
    retworkx.random_layout
    retworkx.spring_layout
 
+.. _layout-functions:
+
 Layout Functions
 ================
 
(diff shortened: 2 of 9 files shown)


t-imamichi (approved):
LGTM

Issue #328, opened 2021-05-20
Documentation overhaul


mtreinish:
### What is the expected enhancement?

As we've been rapidly growing the functionality in retworkx the flat docs structure we were using before doesn't really scale. Before we push out the 0.9.0 we should work on reorganizing the documentation. The basic structure I had in mind was at the top level:

- README
- Graph Classes
- Algorithms
 -  Universal Functions and functions that are only applicably to single type
- Generators and Random Graph functions
- Visualization
  - Visualization Functions
  - Layout Functions
  - Examples (this can be in a follow up since we'd have to write all of these)
- Detailed API Docs
  - Per graph type algorithm/functions
  - custom return types
  - exceptions

But this isn't set in stone, that's just the idea I had in my head, feel free to propose alternative structures. The key here though is to make the docs organization a bit more clear.

As part of this we also should update the graph class docstrings to include explicit mention that node and edge indices are stable unique identifiers of a node or edge for the lifetime of the graph. This isn't really explicitly called out anywhere and it's very important.

IvanIsCoding:
I will work on this. I will start by splitting the Detailed API docs (`graph_*`, `digraph_*` and methods that are less important for the user let's say) from the Algorithms section.

PR #346, opened 2021-05-31
Major rework in documentation

IvanIsCoding:
Step towards #328. This PR changes the documentation by:

* Organizing algorithms by theme instead of splitting them into universal and type specific
* Creating two new sections for type specific algorithms in the docs, called Retworkx Advanced API
* Functions that have no `graph_*` or `digraph_*` equivalents are listed under Algorithms section instead of the type Advanced API section; no function is listed twice

This PR is opinionated, and is open to debate. Because the themes can be arbitrary, there could be other groupings. 

The proposed documentation groups every algorithm that did not belong to a theme under `Miscellaneous`. It also only groups algorithms under the Advanced Retworkx API if they are called by a dispatcher, otherwise they are listed under Algorithms.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -114,15 +114,10 @@ Connectivity and Cycles
    retworkx.cycle_basis
    retworkx.digraph_find_cycle
 
-.. _miscellaneous:
+.. _other-algorithms:
 
-Miscellaneous
--------------
-
-These functions are algorithm functions that wrap per graph object
-type functions in the algorithms API but can be run with a
-:class:`~retworkx.PyGraph`, :class:`~retworkx.PyDiGraph`, or
-:class:`~retworkx.PyDAG` object.
+Other Algorithm Functions
+-------------------------
 
 .. autosummary::
    :toctree: stubs
@@ -177,8 +172,7 @@ Layout Functions
 
    retworkx.random_layout
    retworkx.spring_layout
-   retworkx.graph_random_layout
-   retworkx.digraph_random_layout
+
 
 .. _converters:
 
@@ -190,14 +184,14 @@ Converters
 
    retworkx.networkx_converter
 
-.. _advanced-api-pydigraph:
+.. _api-functions-pydigraph:
 
-Advanced Retworkx API for PyDiGraph
+API functions for PyDigraph
 ====================================
 
 These functions are algorithm functions that are type specific for :class:`~retworkx.PyDiGraph` or
 :class:`~retworkx.PyDAG` objects. Universal functions from Retworkx API that work for both graph types 
-internally call the functions form the advanced API based on the data type.
+internally call the functions from the explicitly typed based on the data type.
 
 .. autosummary::
    :toctree: stubs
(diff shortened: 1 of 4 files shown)


releasenotes/notes/0.8.0/add-universal-functions-1e54351f1f7afa4b.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 features:
   - |
-    New :ref:`universal-functions` that can take in a :class:`~retworkx.PyGraph`
+    New `universal-functions` that can take in a :class:`~retworkx.PyGraph`
```
mtreinish, line 4: This will render kind of weirdly because it would basically be:

New _universal-functions_ that

so let's just drop the reference formatting and go with:

```suggestion
    New universal functions that can take in a :class:`~retworkx.PyGraph`
```

docs/source/api.rst
```diff
@@ -46,125 +167,87 @@ Random Circuit Functions
...
+Layout Functions
+================
+
+.. autosummary::
+   :toctree: stubs
+
+   retworkx.random_layout
+   retworkx.spring_layout
+   retworkx.graph_random_layout
+   retworkx.digraph_random_layout
```
mtreinish, line 181: Did you want to move these to the separate per type api functions?

docs/source/api.rst
```diff
@@ -14,6 +14,127 @@ Graph Classes
...
+   retworkx.strongly_connected_components
+   retworkx.number_weakly_connected_components
+   retworkx.weakly_connected_components
+   retworkx.is_weakly_connected
+   retworkx.cycle_basis
+   retworkx.digraph_find_cycle
+
+.. _miscellaneous:
+
+Miscellaneous
```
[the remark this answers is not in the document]
IvanIsCoding: Changed

docs/source/api.rst
IvanIsCoding: Nice catch, I did overlook those two! They're separate now

docs/source/api.rst
```diff
@@ -46,125 +167,87 @@ Random Circuit Functions
...
+==========
+
+.. autosummary::
+   :toctree: stubs
+
+   retworkx.networkx_converter
+
+.. _advanced-api-pydigraph:
+
+Advanced Retworkx API for PyDiGraph
```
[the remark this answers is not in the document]
IvanIsCoding: I added the type to the title, and it looks ok

IvanIsCoding:
> Overall this LGTM this is a way clearer and cleaner organizational structure for the api docs. I just have a few inline comments with bikeshedding suggestions about some of the naming and a couple of small cleanups. But, once we fix those and make a decision on the section names I think this is good to go.

Alright, the comments 

> Overall this LGTM this is a way clearer and cleaner organizational structure for the api docs. I just have a few inline comments with bikeshedding suggestions about some of the naming and a couple of small cleanups. But, once we fix those and make a decision on the section names I think this is good to go.

I adressed all comments, it should be better now. And I agree that Other Algorithms sounds better than Miscellaneous.

mtreinish (approved):

Issue #347, opened 2021-06-03
The result of `graph_greedy_color` is not deterministic


t-imamichi:
### What is the expected enhancement?

The result of `graph_greedy_color` is not deterministic as follows.
```python
import retworkx as rx
graph = rx.PyGraph()
graph.add_nodes_from(range(2))
coloring_dict = rx.graph_greedy_color(graph).items()
print(coloring_dict)
```

result:`dict_items([(0, 0), (1, 0)])` or `dict_items([(1, 0), (0, 0)])`.

This is perhaps due to hashmap order of rust.
It might be good to make the result deterministic or write a note in the documentation.

@ikkoham found this.

georgios-ts:
Indeed the order of iterating over a hashmap can be arbitary (https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.iter).
Is an order based on nodes indexes preferred?
We can easily do this by replacing this loop
https://github.com/Qiskit/retworkx/blob/010e0de6b8b475843df8aba511fc5731e386fa01/src/lib.rs#L929-L930
with
```rust
for index in graph.graph.node_indices() {
        let index = index.index();
        out_dict.set_item(index, colors[&index])?;
    }
```
but with a small performance  overhead.
Personally, i would leave it as it is and state this behavior in the documentation.

mtreinish:
@georgios-ts I wonder what the performance of using a `BTreeMap`  for `colors` would be. The rule of thumb is it's slower than hashmap for insertion and lookup/get but faster for iteration. It's probably worth benchmarking to see what the overhead is vs the current hash map, because a `BTreeMap` will give us a deterministic ordering too.

That being said from a performance perspective the other thing to consider is whether we want to add a custom mapping return type to avoid the copy overhead until access time. If we did this we should stick with a `HashMap` and just document the ordering limitation.

ikkoham2:
Personally, I expect the retworkx's result to be in a fixed order, because the Python dict is ordered. (This does not mean that I want them to be in order from 1..n). networkx also doesn't change the order.
I think it would be better if the results were fixed, since users may also be concerned about reproducibility of the results in their research.

I have checked the `BTreeMap` with the code
```
num_nodes = 10000
num_edges = num_nodes * 10

graph = rx.PyGraph()
graph.add_nodes_from(range(num_nodes))
graph.add_edges_from_no_data([tuple(np.random.randint(0, num_nodes, 2)) for _ in range(num_edges)])
[...]

PR #349, opened 2021-06-04
[Discussion] Use FnvHashMap

ikkoham:
Discussion for #347.
FNV hash is faster hash than Rust's default hash.
See https://crates.io/crates/fnv more details.

This can fix the order of the result.

Unfortunately, the results of my benchmarking were almost the same.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,6 +97,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37ab347416e802de484e4d03c7316c48f1ecb56574dfd4a46a80f173ce1de04d"
 
 [[package]]
+name = "fnv"
+version = "1.0.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
+
+[[package]]
 name = "getrandom"
 version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -525,6 +531,7 @@ name = "retworkx"
 version = "0.9.0"
 dependencies = [
  "fixedbitset",
+ "fnv",
  "hashbrown",
  "ndarray",
  "numpy",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ numpy = "0.13.1"
 rand = "0.8"
 rand_pcg = "0.3"
 rayon = "1.5"
+fnv = "1.0.7"
 
 [dependencies.pyo3]
 version = "0.13.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,6 +50,7 @@ use petgraph::visit::{
 };
 use petgraph::EdgeType;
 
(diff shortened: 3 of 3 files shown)


mtreinish:
For the hash speed issue we actually tackled this some time ago and switched all the hashmap usage to use hasbrown https://github.com/rust-lang/hashbrown/ everywhere which uses the ahash hasher: https://github.com/tkaitchuck/aHash by default.

I'm fine changing the hasher to one with a deterministic ordering, so fnv is fine if there isn't a performance regression (and https://github.com/tkaitchuck/aHash/blob/master/compare/readme.md indicates there might be a regression). But, I'd rather keep with the hashbrown hashmap type and just change the hasher for a stable ordering. This is mostly so we can keep the types the same and leverage it's native rayon support and other (also pyo3 natively supports hashbrown). 

That being said this order issue will be pretty much everywhere we return a dict or mapping in retworkx. Do you think this is something worth updating in other functions?

mtreinish:
Also did you try using a fixed seed for the hasher? Something like: 

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 14bb688..ea545dd 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ name = "retworkx"
 crate-type = ["cdylib"]
 
 [dependencies]
+ahash = "0.4.0"
 petgraph = "0.5.1"
 fixedbitset = "0.2.0"
 numpy = "0.13.1"
diff --git a/src/lib.rs b/src/lib.rs
index 2c6a998..3014688 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,6 +28,7 @@ mod union;
 use std::cmp::{Ordering, Reverse};
 use std::collections::{BTreeSet, BinaryHeap};
 
+use ahash::RandomState;
 use hashbrown::{HashMap, HashSet};
 
 use pyo3::create_exception;
@@ -898,7 +899,8 @@ fn graph_greedy_color(
     py: Python,
     graph: &graph::PyGraph,
 ) -> PyResult<PyObject> {
-    let mut colors: HashMap<usize, usize> = HashMap::new();
+    let stable_state = RandomState::with_seeds(42, 424242);
+    let mut colors: HashMap<usize, usize> = HashMap::with_hasher(stable_state);
     let mut node_vec: Vec<NodeIndex> = graph.graph.node_indices().collect();
     let mut sort_map: HashMap<NodeIndex, usize> =
         HashMap::with_capacity(graph.node_count());
```

ikkoham:
Thank you. 

I see. I was wondering why retworkx uses hashbrown, even though std uses it internally from Rust 1.36. (I forked retworkx yesterday and read the code) That is the reason.
I think your suggestion is the best. (When size is small, there is little difference in performance. `ahash` is probably faster and compatible with other crates.)

> That being said this order issue will be pretty much everywhere we return a dict or mapping in retworkx. Do you think this is something worth updating in other functions?

Unless we are in danger of HashDOS, we can fix it. In general, I think it's good to make test easier, and it's good for science to be able to reproduce experiments using simulators like Aer. But, I've recently seen retworkx, so I don't know what functions are nondeterministic.

Issue #350, opened 2021-06-04
Deduplicate custom return types with macros


mtreinish:
### What is the expected enhancement?

The custom return type definitions in `src/iterators.rs` are pyclass struct wrappers around a rust native container (like a `Vec` or `HashMap`) which we use to avoid the copy and overhead of going from rust->python for large return types. Most of the code there is copy and pasted boilerplate to fulfill the interface for exposing an Python Sequence or Python Mapping (and the auxiliary methods for a dict) and the only updates are the type information for the most part. This feel like the kind of thing that should be easy to simplify by just having a custom macro that expands the definition based on the input types and names.

mtreinish:
This was done in https://github.com/Qiskit/retworkx/pull/354

PR #354, opened 2021-06-08
Refactor custom return types

georgios-ts:
An approach for #350.

Right now in our custom return types we have some duplicate boilerplate code and the appropriate logic is inside `__richcmp__`, `__str__` and `__hash__`  functions. This PR moves the logic into separate traits and implements them for native rust types and PyObject. This allows us to introduce convenient macro rules to expand the definition of our custom iterator based on the input types and name. Hopefully this approach makes it easy and fast to introduce new return types (?).

That being said, I am a little bit concerned about the performance. For example, we need to acquire the python Gil even for methods that we don't actually use it (is there any chance that compiler will optimize this?). In any case, even if we decide that this approach makes sense, we should benchmark it (but how?).

There is also `PathMapping` that does not fit nicely into the macro rule since although it holds a `HashMap<usize, Vec<usize>>` the `PathMappingValues` are `Vec<NodeIndices>`.

### DIFF
--- a/src/iterators.rs
+++ b/src/iterators.rs
@@ -10,6 +10,33 @@
 // License for the specific language governing permissions and limitations
 // under the License.
 
+// There are two useful macros to quickly define a new custom return type:
+//
+// :`custom_vec_iter_impl` holds a `Vec<T>` and can be used as a
+//  read-only sequence/list. To use it, you should specify the name of the new type,
+//  the name of the vector that holds the data, the type `T` and a docstring.
+//
+//  e.g `custom_vec_iter_impl!(MyReadOnlyType, data, (usize, f64), "Docs");`
+//      defines a new type named `MyReadOnlyType` that holds a vector called `data`
+//      of values `(usize, f64)`.
+//
+// :`custom_hash_map_iter_impl` holds a `HashMap<K, V>` and can be used as
+//  a read-only mapping/dict. To use it, you should specify the name of the new type,
+//  the name of the hash map that holds the data, the type of the keys `K`,
+//  the type of the values `V` and a docstring.
+//
+//  e.g `custom_hash_map_iter_impl!(MyReadOnlyType, data, usize, f64, "Docs");`
+//      defines a new type named `MyReadOnlyType` that holds a mapping called `data`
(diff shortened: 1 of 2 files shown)


mtreinish:
> That being said, I am a little bit concerned about the performance. For example, we need to acquire the python Gil even for methods that we don't actually use it (is there any chance that compiler will optimize this?). In any case, even if we decide that this approach makes sense, we should benchmark it (but how?).

I wouldn't be too concerned about this, for the most part acquiring the gil from a function called by python is low overhead since we already have the handle from the calling python. I think it would only be a potential issue if we were calling these functions from a separate thread where we might have to wait for the gil, but that shouldn't matter in the python interface code. Also the functions which are using the gil now when there weren't before are things like `__str__` and `__hash__` which aren't super performance critical. It would be a bigger deal for things like `__getitem__` etc which get called a lot. The compiler is also probably smart enough to optimize away away code if it's never used in the generated function (but we'd have to look at the output from the compilation to confirm) there is a compiler flag to control this optimization: https://doc.rust-lang.org/rustc/codegen-options/index.html?highlight=dead-code#link-dead-code

As for benchmarking we can run some terra benchmarks that are good examples for how some of the custom return types are used in practice. For things like hash() and str() we'll probably need to come up with some synthetic benchmarks with large data sets in the custom container formats.

mtreinish (commented):
Overall this LGTM. But before I approve it could you add some docs to the top of the module on how to use the macros. It's more less obvious based on the examples there now. But having some documentation on how the macros are expected to be called would be really useful for when we need to add new custom return types.

georgios-ts:
@mtreinish done in c660181

mtreinish (approved):
LGTM, thanks for adding the docs

mtreinish:
This is super nice! I just rebased two pending commits with custom return types, it removes so much boiler plate: https://github.com/Qiskit/retworkx/pull/312/commits/4374d4ac3ff2c113aa0d0e7df1e43031504af7c0 and https://github.com/Qiskit/retworkx/pull/355/commits/a5473e1d6bf37c2c34aa81a491600c1f630d7ec3

PR #355, opened 2021-06-09
Add unweighted num shortest path functions

mtreinish:
This commit adds a new function num_shortest_paths_unweighted which is
used to get the number of parallel shortest paths from a source node to
all targets in the graph.

TODO:

- [x] Expand test coverage
- [x] Add release notes


georgios-ts:
I know it’s still WIP, but just to save you some time here, the updates in `out_map` do not match with the logic of this function https://github.com/yaleqc/qiskit_surface_codes/blob/master/topological_codes/fitters/graph_utils.py.

Also, not only relevant to this PR but what about using a vec of size `node.bound()` instead of HashMap to store the distances and the `out_map`? I know this might use more memory especially with many node removals (although if let’s say we only delete a constant ratio of all the nodes and the hashmap uses a constant factor more memory per entry than a vector does, the difference might be insignificant) but the tradeoff with performance is worth taking, e.g: 

```python
for n in [10000, 20000, 30000]:
    for p in [0.3, 0.5]:
        graph = rx.directed_gnp_random_graph(n, p, seed=42)

        start = time.time()
        res = rx.digraph_num_shortest_paths_unweighted(graph, 10)
        stop = time.time()

        print(stop - start)
----
with HashMap:
0.3144831657409668
0.5204041004180908
1.5746221542358398
2.211705207824707
5.68000602722168
11.261810064315796
---- 
with Vec:
0.11958813667297363
0.2027289867401123
0.45508575439453125
0.9924600124359131
2.9298949241638184
8.609977006912231

File: retworkx/__init__.py [hunk 1]
```diff
@@ -1245,3 +1245,27 @@ def _graph_spiral_layout(
...
+    :returns: A mapping of target node indices to the number of shortest paths
+        from ``source`` to that node
+    :rtype: NodesCountMapping
+    """
+    raise TypeError("Invalid input type %s for graph" % type(graph))
+
+
+@num_shortest_paths_unweighted.register(PyDiGraph)
+def _digraph_num_shortest_paths_unweighted(graph, source):
+    return digraph_num_shortest_paths(graph, source)
```
georgios-ts, line 1266:
```suggestion
    return digraph_num_shortest_paths_unweighted(graph, source)
```

retworkx/__init__.py
```diff
@@ -1245,3 +1245,27 @@ def _graph_spiral_layout(
...
+
+
+@num_shortest_paths_unweighted.register(PyGraph)
+def _graph_num_shortest_paths_unweighted(graph, source):
+    return graph_num_shortest_paths(graph, source)
```
georgios-ts, line 1271:
```suggestion
    return graph_num_shortest_paths_unweighted(graph, source)
```

mtreinish:
@georgios-ts Thanks, yeah I messed up the logic there. As for the data structure sure `Vec`s are almost always going to be faster if you can use them just because of memory locality, I went with a hashmap because of index accessing if there were holes in the indices because of removal. I'm not too concerned about the memory footprint it's not that much overhead to use size `node_bound()`, but it was more if you have node indices 0, 1, 3 in your graph what do you put in `out_vec[2]`? Writing this out the way I think we can handle this would be to just make it `Vec<Option<usize>>` and in the custom return just use the mapping protocol and return an index error if it's None.

georgios-ts:
I was thinking more to use a vector internally and then return a hashmap containing only valid node indices but yeah what you said works fine too

mtreinish:
> I was thinking more to use a vector internally, put whatever value in 'out_map[2]' (we are not going to use it anyway) and then return a hashmap containing only valid node indices but yeah what you said works fine too

I implemented this in https://github.com/Qiskit/retworkx/pull/355/commits/6ac18b808844cc8d965d2194b85e3b907bd23659 I liked it more than my idea (mostly because it was less work).

src/lib.rs
```diff
@@ -4442,6 +4443,82 @@ pub fn digraph_spiral_layout(
...
+            }
+        }
+    }
+    out_map[source] = None;
+    out_map
+        .into_iter()
+        .enumerate()
+        .filter(|i| i.1.is_some())
+        .map(|x| (x.0, x.1.unwrap()))
+        .collect()
```
georgios-ts, line 4481: Nodes that are not reachable from `source` will not be included in the final map. So either we state this clearly in the docs or we add them in the `out_map` with a zero value.

mtreinish: Sure, I'll update the documentation for the function to explicitly call this out. I think either is a valid approach but at least to me it makes a bit more sense to just not include it in the output map if there is no path from `source` to a node.

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/355/commits/30d5a47f21130ce8708029f5e9b8d2206f0026cd

src/lib.rs
```diff
@@ -4442,6 +4443,90 @@ pub fn digraph_spiral_layout(
     layout::spiral_layout(&graph.graph, scale, center, resolution, equidistant)
 }
 
+fn _num_shortest_paths_unweighted<Ty: EdgeType>(
+    graph: &StableGraph<PyObject, PyObject, Ty>,
+    source: usize,
+) -> PyResult<HashMap<usize, u128>> {
```
IvanIsCoding, line 4449: I am concerned about overflows here. The number of paths can grow very fast, it may not fit on a 128-bit integer. Can you try using [BigInt](https://docs.rs/num-bigint/0.4.0/num_bigint/struct.BigInt.html) and see how much it affects the performance? [PyO3 has supports it](https://github.com/PyO3/pyo3/blob/05e4bdf69cde2bc49865fbafe756a8507dd45f1d/src/num_bigint.rs#L39-L48).

IvanIsCoding (commented):
Commenting because I am a bit concerned about using 128-bit integers for a combinatorics. Counting functions can grow very fast, and if there is a lot of symmetry then they grow even faster.

We are adding this function to help some folks speed up their code, but I don't want to fall into the overflow benchmark trap. It is useless to have a speed up if the result is not correct because of an overflow

src/lib.rs
mtreinish: > Commenting because I am a bit concerned about using 128-bit integers for a combinatorics. Counting functions can grow very fast, and if there is a lot of symmetry then they grow even faster.
> 
> We are adding this function to help some folks speed up their code, but I don't want to fall into the overflow benchmark trap. It is useless to have a speed up if the result is not correct because of an overflow

Sure that's a good point, while the [max value of u128](https://doc.rust-lang.org/std/primitive.u128.html#associatedconstant.MAX) is: 340282366920938463463374607431768211455 which I honestly can't see being an issue for most people on this in practice; it could come up with large graphs. I did think that rust would panic on overflow for us, and it sort of does. By default only in debug mode addition panics on overflow, but in release mode (which is what we obviously ship in precompiled binaries on pypi) it wraps around on overflow which we definitely don't want (I mistakenly thought it always would panic). I'll benchmark both this current PR as it is, changing the addition to explicitly use a function call with overflow checking, and using `BigInt` and we can go from there.

src/lib.rs
```diff
@@ -4442,6 +4443,90 @@ pub fn digraph_spiral_layout(
...
+    let mut out_map: Vec<Option<u128>> = vec![None; graph.node_bound()];
+    let node_index = NodeIndex::new(source);
+    if graph.node_weight(node_index).is_none() {
+        return Err(PyIndexError::new_err(format!(
+            "No node found for index {}",
+            source
+        )));
+    }
+    let mut bfs = Bfs::new(&graph, node_index);
+    let mut distance: Vec<Option<u128>> = vec![None; graph.node_bound()];
```
IvanIsCoding, line 4459: `distance` does not need to be `u128`, `usize` is sufficient. The distance between two nodes is bounded by the number of nodes in a graph, which fits in a `usize` value.`out_map` is the only result that needs to be calculated as `u128/BigInt`.

src/lib.rs
mtreinish: I ran some quick benchmarks on this and it's 4-6x slower than using u128. But in that testing and benchmarking your point about the rapid growth of the count here was more obvious. The largest square grid graph I could run on without overflowing a u128 was 66x66. At a 66x66 graph the u128 took 0.0001586 sec and the BigUint version took 0.0004135 sec. So I'm comfortable with a 4x slowdown if we're that fast since anything slower would likely mean an invalid result or an overflow exception.

src/lib.rs
mtreinish: Done in https://github.com/Qiskit/retworkx/pull/355/commits/abbc61fddc881b8afe69f70af2794fde3742432e

src/lib.rs
```diff
@@ -4462,6 +4464,98 @@ pub fn digraph_spiral_layout(
...
+    source: usize,
+) -> PyResult<NodesCountMapping> {
+    Ok(NodesCountMapping {
+        map: _num_shortest_paths_unweighted(&graph.graph, source)?,
+    })
+}
+
+/// Get the number of unweighted shortest paths from a source node
+///
+/// :param PyDiGraph graph: The graph to find the number of shortest paths on
```
georgios-ts, line 4541:
```suggestion
/// :param PyGraph graph: The graph to find the number of shortest paths on
```

tests/digraph/test_num_shortest_path.py
```diff
@@ -0,0 +1,144 @@
...
+            20: 1,
+            21: 5,
+            22: 15,
+            23: 35,
+            24: 70,
+        }
+        for i in res:
+            print("(%s, %s)" % (i, res[i]))
+        print(res)
+        print(len(res))
```
georgios-ts, line 99:
```suggestion
```

georgios-ts (approved):
LGTM

src/lib.rs
jlapeyre: For a graph with `n` nodes, the number of shortest paths between two fixed nodes is bounded by the number of simple paths between these nodes, which is bounded by `n - 2`. So you will not get overflow using `u128`.

PR #357, opened 2021-06-13
Refactor floyd_warshall

### LINKED ISSUE #351: Refactor floyd_warshall to use new return types

After #278 was merged, retworkx has a new type [AllPairsPathLengthMapping](https://github.com/Qiskit/retworkx/blob/885f2de1588dc983591563f94b2ddcada443b3b2/src/iterators.rs#L1727) and [`all_pairs_dijkstra_path_lengths`](https://qiskit.org/documentation/retworkx/stubs/retworkx.all_pairs_dijkstra_path_lengths.html#retworkx.all_pairs_dijkstra_path_lengths) returns this type.

For consistency within retworkx, `floyd_warshall` should also return AllPairsPathLengthMapping. [`floyd_warshall`](https://qiskit.org/documentation/retworkx/stubs/retworkx.floyd_warshall.html#retworkx.floyd_warshall) and [`all_pairs_dijkstra_path_lengths`](https://qiskit.org/documentation/retworkx/stubs/retworkx.all_pairs_dijkstra_path_lengths.html#retworkx.all_pairs_dijkstra_path_lengths) calculate the same result (All Pairs Shortest Path) using different algorithms. Because they calculate the same result, their return types should also be the same.

Moreover, [`floyd_warshall`](https://qiskit.org/documentation/retworkx/stubs/retworkx.floyd_warshall.html#retworkx.floyd_warshall) could use of additional improvements such as:
* Supporting an `edge_cost_fn` argument such as in [`all_pairs_dijkstra_path_lengths`](https://qiskit.org/documentation/retworkx/stubs/retworkx.all_pairs_dijkstra_path_lengths.html#retworkx.all_pairs_dijkstra_path_lengths), currently it assumes all weights are 1
* Use Rayon parallelization for better performance as in [`floyd_warshall_numpy`](https://qiskit.org/documentation/retworkx/stubs/retworkx.floyd_warshall_numpy.html#retworkx.floyd_warshall_numpy) (#291 can be a guideline on how to accomplish this)

IvanIsCoding:
Closes #351:
* Uses new  `AllPairsPathLengthMapping` return type in `floyd_warshall`
* Support for PyGraph added
* Support for `weight_fn` added
* Parallelization addded

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1044,7 +1044,48 @@ fn graph_k_shortest_path_lengths(
     })
 }
 
-/// TODO
+/// Find all-pairs shortest path lengths using Floyd's algorithm
+///
+/// Floyd's algorithm is used for finding shortest paths in dense graphs
+/// or graphs with negative weights (where Dijkstra's algorithm fails).
+///
+/// This function is multithreaded and will launch a pool with threads equal
+/// to the number of CPUs by default if the number of nodes in the graph is
+/// above the value of ``parallel_threshold`` (it defaults to 300).
+/// You can tune the number of threads with the ``RAYON_NUM_THREADS``
+/// environment variable. For example, setting ``RAYON_NUM_THREADS=4`` would
+/// limit the thread pool to 4 threads if parallelization was enabled.
+///
+/// :param PyDiGraph graph: The directed graph to run Floyd's algorithm on
+/// :param weight_fn: A callable object (function, lambda, etc) which
+///     will be passed the edge object and expected to return a ``float``. This
+///     tells retworkx/rust how to extract a numerical weight as a ``float``
+///     for edge object. Some simple examples are::
+///
+///         graph_floyd_warshall_numpy(graph, weight_fn: lambda x: 1)
+///
+///     to return a weight of 1 for all edges. Also::
+///
+///         graph_floyd_warshall_numpy(graph, weight_fn: lambda x: float(x))
+///
+///     to cast the edge object as a float as the weight.
+/// :param as_undirected: If set to true each directed edge will be treated as
+///     bidirectional/undirected.
+/// :param int parallel_threshold: The number of nodes to execute
+///     the algorithm in parallel at. It defaults to 300, but this can
+///     be tuned
+///
+/// :return: A read-only dictionary of path lengths. The keys are source
+///     node indices and the values are dicts of the target node and the length
+///     of the shortest path to that node. For example::
+///
+///         {
+///             0: {0: 0.0, 1: 2.0, 2: 2.0},
+///             1: {1: 0.0, 2: 1.0},
+///             2: {0: 1.0, 2: 0.0},
+///         }
+///
+/// :rtype: AllPairsPathLengthMapping
 #[pyfunction]
 #[text_signature = "(graph, /, weight_fn=None, default_weight=1.0, parallel_threshold=300)"]
 fn digraph_floyd_warshall(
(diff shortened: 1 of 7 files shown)


IvanIsCoding:
I benchmarked with a code similar to #278. Performance is comparable to the All Pairs Dijkstar function.

```python
import time

import retworkx as rx
import networkx as nx

rx_graph = rx.directed_gnp_random_graph(5000, .9)

floyd_rx_start = time.time()
rx.digraph_floyd_warshall(rx_graph, lambda _: 1)
floyd_rx_stop = time.time()
print("Floyd-Warshall retworkx: %s" % str(floyd_rx_stop - floyd_rx_start))

start = time.time()
rx.digraph_all_pairs_dijkstra_shortest_paths(rx_graph, lambda _: 1)
stop = time.time()
print("Dijsktra All Pairs retworkx: %s" % str(stop - start))
```

Result:
```
Floyd-Warshall retworkx: 127.31177496910095
Dijsktra All Pairs retworkx: 131.72510266304016
```

src/lib.rs
```diff
@@ -1043,101 +1043,248 @@ fn graph_k_shortest_path_lengths(
...
 ///
-/// .. note::
+///         graph_floyd_warshall_numpy(graph, weight_fn: lambda x: 1)
 ///
-///     Paths that do not exist are simply not found in the return dictionary,
-///     rather than setting the distance to infinity, or -1.
+///     to return a weight of 1 for all edges. Also::
 ///
-/// .. note::
+///         graph_floyd_warshall_numpy(graph, weight_fn: lambda x: float(x))
```
mtreinish, line 1186:
```suggestion
///         graph_floyd_warshall_numpy(graph, weight_fn: lambda float)
```

src/lib.rs
```diff
@@ -1043,101 +1043,248 @@ fn graph_k_shortest_path_lengths(
             .collect(),
     })
 }
-/// Return the shortest path lengths between ever pair of nodes that has a
-/// path connecting them
+
+macro_rules! _floyd_warshall {
```
mtreinish, line 1047: Is there a reason you need to use a macro here? It feels like this could just be a common function

IvanIsCoding: I did it to avoid code duplication. I tried to make a function with similar same traits to `fn get_edge_iter_with_weights<G>` which accepts both directed and undirected graphs, and it did not work. No trait had `edge_count()` and `node_indices()`.

I think if I split the return statements from the code I can put the logic into functions, but that would still duplicate around 40 lines of code

georgios-ts: @IvanIsCoding In other functions we do something like:
```rust
fn _floyd_warshall<Ty: EdgeType>(
    py: Python,
    graph: &StableGraph<PyObject, PyObject, Ty>,
    ....
)
```

btw `petgraph::visit::IntoNodeIdentifiers` is the same as `node_indices`

mtreinish (approved):
LGTM, thanks for doing this and updating it to use a shared function instead of a macro. 

My only thought while reviewing it was that we probably could combine the implementation here with the ndarray based function and just adjust the output format based on which one we pick. In other words if the `_floyd_warhsall()` function you're adding here is as fast (or faster)  compared to the existing matrix based implementation is there a reason not to just have the numpy functions call `_floyd_warshall` and construct an output matrix from the hashmap? If there is a performance difference I think keeping them separate is fine. But either way that's something we can consider inf a follow up PR, I'm fine with the implementation here.

PR #359, opened 2021-06-14
Improve documentation of stable graph behaviour 

### LINKED ISSUE #356: Improve documentation around stable graph behavior

Currently the documentation for the graph classes mentions the node indices are stable for the lifetime of the graph, specifically:

```
Each node and edge (although rarely used for edges) is indexed by an integer id. These ids are stable for the lifetime of the graph object and on node or edge deletions you can have holes in the list of indices for the graph.
```

which is true, but doesn't describe the behavior sufficiently. We should also mention that node indices will be reused on additions after a removal. For example:

```python
import retworkx
graph = retworkx.PyDiGraph()
graph.add_nodes_from(list(range(5)))
graph.add_nodes_from(list(range(2)))
graph.remove_node(2)
res_manual = graph.add_parent(6, None, None)
print(res_manual)
```
will print `2`

We should have this documented so it doesn't catch people by surprise.

nahumsa:
fixes #356

The behaviour of node indices being reused after deletion is documented in both `PyGraph` and `PyDiGraph`.


src/digraph.rs
```diff
@@ -61,6 +61,7 @@ use super::{
 /// (although rarely used for edges) is indexed by an integer id. These ids
 /// are stable for the lifetime of the graph object and on node or edge
 /// deletions you can have holes in the list of indices for the graph.
+/// Node indices will be reused on additions after removal.
```
mtreinish, line 64: Can we add the example from the release note here instead? I think it would make more sense to have that example here. Also probably add a print of node indices after the removal so after `remove_node(2)` call print(graph.node_indexes())` to show that the list has the hole.

mtreinish (changes requested):
Thanks for getting this started. I think instead of putting the example in the release note I would just not bother with a release note and integrate the example into the docstring directly. That way it'll always be visible in the class docs and not hidden in a past release note.

Also can you update the PyDAG class docstring in `retworkx/__init__.py` too?

src/digraph.rs
nahumsa: Sure.

Updated the docstring of PyDAG on [`6eb08de`](https://github.com/Qiskit/retworkx/pull/359/commits/6eb08de6f39106b5e97ff72d5a45daab8a2be6c0)

Added the jupyter example to both PyGraph and PyDiGraph on [`2032410`](https://github.com/Qiskit/retworkx/pull/359/commits/20324107942ac9871b970e0ac0f859c96fec8e88)

mtreinish:
```suggestion
/// Node indices will be reused on additions after removal. For example:
```

src/graph.rs
```diff
@@ -54,6 +54,20 @@ use petgraph::visit::{
 /// (although rarely used for edges) is indexed by an integer id. These ids
 /// are stable for the lifetime of the graph object and on node or edge
 /// deletions you can have holes in the list of indices for the graph.
+/// Node indices will be reused on additions after removal.
```
mtreinish, line 57:
```suggestion
/// Node indices will be reused on additions after removal. For example:
```

retworkx/__init__.py
```diff
@@ -29,6 +29,20 @@ class PyDAG(PyDiGraph):
     (although rarely used for edges) is indexed by an integer id. These ids
     are stable for the lifetime of the graph object and on node or edge
     deletions you can have holes in the list of indices for the graph.
+    Node indices will be reused on additions after removal.
```
mtreinish, line 32:
```suggestion
    Node indices will be reused on additions after removal. For example:
```

mtreinish (approved):
Thanks for the quick update, just a quick inline suggestion but I think this is good to go otherwise

PR #361, opened 2021-06-16
Add collect_bicolor_runs

### LINKED ISSUE #265: Port collect 2q blocks function from qiskit-terra

In qiskit-terra there is a transpiler pass collect2q blocks which primarily consists of a graph traversal. Doing the actual graph traversal and search will be much faster in retworkx, so we should add a function that can be used in place of terra's python version. The terra function code is:

https://github.com/Qiskit/qiskit-terra/blob/b058856e2aee9c5819120dc9c95f1e2478c95f4e/qiskit/transpiler/passes/optimization/collect_2q_blocks.py#L55-L246

This is similar to collect_runs which already has a native retworkx version:

https://github.com/Qiskit/retworkx/blob/b1f973630cb352950fbdeabd4b4c60191a62a6e7/src/lib.rs#L1121-L1173

and will operate in a similar manner with a filter function argument except it will collect 2q blocks instead of 1q runs.

IvanIsCoding:
Closes #265 

This ports collect-2q-blocks to  Rust. Is ports the rewritten version from https://github.com/Qiskit/qiskit-terra/pull/6534 which is much simpler to maintain in Retworkx.

Update: There's a 1.25x speed up for using collect_bicolor_runs


IvanIsCoding:
I still need to benchmark the code more accurately, but this yield some gains. There is still a significant overhead converting the result to Python objects though. We have a list of lists, which is not the cheapest to generate. Ideally we would return a Python list of read-only Rust lists of `PyObject`.

@mtreinish do we have a type equivalent to read-only list of `PyObject` (equivalent to a Python tuple)? Do you think it is worth adding one for this?

src/lib.rs
```diff
@@ -1452,6 +1452,141 @@ fn collect_runs(
...
+        ($pending_list: expr, $block_id: expr, $color: expr) => {
+            if $color >= $pending_list.len() {
+                $pending_list.resize($color + 1, Vec::new());
+                $block_id.resize($color + 1, None);
+            }
+        };
+    }
+
+    for node in nodes {
+        if special_filter_node(&graph.graph[node])? {
```
georgios-ts, line 1501: why do we need this `special_filter_node` function? isn't `filter_node` enough?

src/lib.rs
```diff
@@ -1452,6 +1452,141 @@ fn collect_runs(
...
+                        && block_id[c0].unwrap_or_default()
+                            == block_id[c1].unwrap_or_default()
+                    {
+                        block_list[block_id[c0].unwrap_or_default()]
+                            .push(graph.graph[node].clone_ref(py));
+                    } else {
+                        let mut new_block: Vec<PyObject> = Vec::new();
+                        if !pending_list[c0].is_empty() {
+                            new_block.extend(pending_list[c0].iter().cloned());
+                            pending_list[c0].clear();
```
georgios-ts, line 1551: I think this is equivalent to 
```suggestion
                            new_block.append(&mut pending_list[c0]);
```
and we avoid a clone (it shouldn't be expensive but nevertheless)

src/lib.rs
IvanIsCoding: `special_filter_node` + topological sort  is the equivalent of [this](https://github.com/Qiskit/qiskit-terra/pull/6534/files#diff-4be18ab968355b22882ab1c3b52904b231c102a68a77d8ea3c5a2066f0511b14R41). That happens because [`topological_op_nodes`](https://github.com/Qiskit/qiskit-terra/blob/31a24c1a6fbbe62d7c1c568e405cde92f87fc0b7/qiskit/dagcircuit/dagcircuit.py#L945) filter only ops. To replicate the exact behavior I opted to add `speecial_filter_fn` which will receive the filter `nd.type == 'op'`

IvanIsCoding: With that said, I think it could be replaced only with `filter_fn` but performance will be impacted. The `special_filter_fn` is cheap to execute and prunes many loops in edges that call a more expensive `color_fn`

georgios-ts: i guess, in most dagcircuits `op` nodes will be way more than `in` + `out` nodes and it seems to me that altough `special_filter_fn` prunes some loops, we pay an overhead for calling it when most of the time it returns `true`.

IvanIsCoding: I opted to simplify this with `filter_fn` using ternary logic. If the result is `None`, then we simply ignore the node as it is not an operation. Otherwise, we get `True` or `False` and can handle the situation accordingly. This leads to a single Python call

src/lib.rs
```diff
@@ -1474,6 +1474,144 @@ fn collect_runs(
     Ok(out_list)
 }
 
+/// Collect runs that match a filter function
+///
+/// A bicolor run is a path of nodes connected by edges of up two colors,
```
georgios-ts, line 1479: I wonder if *path* is the correct term here since a collected block may have nodes with `out_degree > 1` and even be "disconnected" since if `filter_node` returns `None`, it doesn't "stop" the collection of a block.

src/lib.rs
```diff
@@ -1474,6 +1474,144 @@ fn collect_runs(
...
+/// :param PyDiGraph graph: The graph to find runs in
+/// :param filter_fn: The filter function to use for matching nodes. It takes
+///     in one argument, the node data payload/weight object, and will return a
+///     boolean whether the node matches the conditions or not.
+///     If it returns ``True``, it will continue the bicolor chain.
+///     If it returns ``False``, it will stop the bicolor chain.
+///     If it returns ``None`` it will skip that node.
+/// :param color_fn: The function that gives the color of the edge. It takes
+///     in one argument, the edge data payload/weight object, and will return
+///     an integer, the edge color. If the integer is negative, the edge is ignored.
```
georgios-ts, line 1492: Maybe using `None` to ignore the edge instead of a negative integer similar to `filter_fn`?

tests/digraph/test_collect_bicolor_runs.py
```diff
@@ -0,0 +1,313 @@
...
+            └─────────►│     cx      │◄────────┘
+            ┌──────────┤             ├─────────┐
+            │          │             │         │
+        q0  │          └─────────────┘         │  q1
+            │                                  │
+        ┌───▼─────────┐                 ┌──────▼──────┐
+        │             │                 │             │
+        │    q0       │                 │    q1       │
+        │             │                 │             │
+        └───┬─────────┘                 └──────┬──────┘
```
georgios-ts, line 66: It feels strange that you included this "intermediate" qubit nodes since they don't exist in a `DAGCircuit`

src/lib.rs
IvanIsCoding: Indeed, it is more consistent

src/lib.rs
IvanIsCoding: I went with list of group of nodes, and defined group as nodes connected by edges of exactly two colors.

tests/digraph/test_collect_bicolor_runs.py
IvanIsCoding: I forgot `DAGCircuit`, fortunately the code works on both test cases 😃. Updated with more accurate test cases without intermediate nodes

src/lib.rs
```diff
@@ -1785,6 +1785,139 @@ fn collect_runs(
...
+                    ensure_vector_has_index!(pending_list, block_id, c1);
+
+                    if block_id[c0].is_some()
+                        && block_id[c1].is_some()
+                        && block_id[c0] == block_id[c1]
+                    {
+                        block_list[block_id[c0].unwrap_or_default()]
+                            .push(graph.graph[node].clone_ref(py));
+                    } else {
+                        let mut new_block: Vec<PyObject> = Vec::new();
```
mtreinish, line 1889: Couldn't this be something like:

```suggestion
                        let mut new_block: Vec<PyObject> = Vec::with_capacity(pending_list[c0].len() + pending_list[c1].len() + 1);
```

to avoid the extra allocations?

mtreinish (approved):
Overall LGTM thanks for doing this. Just one small nit inline then I think this is good to go.

PR #339, opened 2021-05-27
Prepare 0.9.0 release

mtreinish:
This commit starts the preparation for the 0.9.0 release. It moves all
the release notes for the release into a separate folder (to distinguish
them from notes for future releases) and also reworks the content of the
release notes and updates the documentation for the release. This commit
should be the last one merged before releasing 0.9.0 to ensure we've
moved and updated all the release notes before the release.

Fixes #328

PR #364, opened 2021-06-16
Fix execution of graphivz tests in CI

mtreinish:
This commit fixes the tox job configuration so we actually install the
requirements for running the graphviz drawer tests. To run the graphviz
drawer tests you need graphviz installed in addition to the retworkx
extras requirements 'graphviz' (which contains pillow and pydot).
However, prior to #339 the extra was called 'pydot' not 'graphviz' and
this was never updated in the tox.ini. This has resulted in the graphivz
drawer tests all being skipped because we're not installing the python
requirements in the test environment. This commit fixes this oversight
and configures tox to install the correct extra so we can run the tests
(at least on linux where we install graphviz in CI).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -17,7 +17,7 @@ deps =
   stestr
 extras =
   mpl
-  pydot
+  graphviz
 passenv = RETWORKX_TEST_PRESERVE_IMAGES
 changedir = {toxinidir}/tests
 commands =


mtreinish:
The tests are running again now: https://github.com/Qiskit/retworkx/pull/364/checks?check_run_id=2838488448#step:7:1088

PR #366, opened 2021-06-17
Add a gitattributes file for linguist

mtreinish:
This commit adds a new gitattributes file to mark the documentation
and release note directories as containing documentation (and therefore
excluding them from language statistics on github). This is documented
here: https://github.com/github/linguist/blob/master/docs/overrides.md

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+docs/* linguist-documentation
+releasenotes/* linguist-documentation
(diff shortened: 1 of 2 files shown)


mtreinish:
This was part of my futile effort to get this to show up as a rust repo first and a python repo second since the addition of the visualization module it's flipped to be python first. But, even with this change the language breakdown it reports is:

```
50.51%  Python
49.23%  Rust
0.27%   Shell
```

while just the packaged code (what gets shipped via pypi and people actually use) has a breakdown of:

```
84.85%  Rust
14.69%  Python
```

I could remove the tests from the statistics to provide a better representation of what's included in the package but that doesn't seem correct either. I guess we can just take it as an indication of the level of testing we do.

PR #370, opened 2021-06-21
Add multigraph option to`PyDiGraph.to_undirected`

### LINKED ISSUE #369: Add option to `PyDiGraph.to_undirected()` to set multigraph false

Right now the `PyDiGraph.to_undirected()` method unconditionally sets the multigraph option to True (https://github.com/Qiskit/retworkx/blob/main/src/digraph.rs#L2317) it would be nice to have an option to control this behavior. This would be useful for directed graphs with bidirectional edges to condense that into a single edge in the undirected output from the method.

georgios-ts:
Closes #369.

Instead of relying to the default graph api that would keep the *last* edge data among the parallel edges we see, i added an optional weight_combo_fn` argument that combines all edge data. Do note here that if `weight_combo_fn` is omitted, the behavior is the same as the default graph api.

### DIFF
--- a/releasenotes/notes/multigraph-option-in-to-undirected-b56af4f26816be61.yaml
+++ b/releasenotes/notes/multigraph-option-in-to-undirected-b56af4f26816be61.yaml
@@ -1,0 +1,9 @@
+---
+upgrade:
+  - |
+    Two new kwargs, ``multigraph`` and ``weight_combo_fn``, was added to
+    :meth:`~retworkx.PyDiGraph.to_undirected`. This can be used to set
+    whether the generated :class:`~retworkx.PyGraph` object is a multi graph or not
+    (whether parallel edges can be added or not). By default this is set to ``True``.
+    If set to ``False`` an extra python callable object should be specified to combine
+    the weights/data of parallel edges.
--- a/releasenotes/notes/multigraph-option-in-to-undirected-b56af4f26816be61.yaml
+++ b/releasenotes/notes/multigraph-option-in-to-undirected-b56af4f26816be61.yaml
@@ -1,5 +1,5 @@
 ---
-upgrade:
+features:
   - |
     Two new kwargs, ``multigraph`` and ``weight_combo_fn``, was added to
     :meth:`~retworkx.PyDiGraph.to_undirected`. This can be used to set
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -2347,9 +2347,13 @@ impl PyDiGraph {
                 let exists = new_graph.find_edge(source, target);
                 match exists {
                     Some(index) => {
-                        let old_weight = new_graph.edge_weight_mut(index).unwrap();
-                        match combine(old_weight, edge.weight(), &weight_combo_fn)?
-                        {
+                        let old_weight =
+                            new_graph.edge_weight_mut(index).unwrap();
+                        match combine(
+                            old_weight,
+                            edge.weight(),
+                            &weight_combo_fn,
+                        )? {
                             Some(value) => {
                                 *old_weight = value;
                             }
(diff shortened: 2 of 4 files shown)


releasenotes/notes/multigraph-option-in-to-undirected-b56af4f26816be61.yaml
```diff
@@ -0,0 +1,9 @@
+---
+upgrade:
```
mtreinish, line 2:
```suggestion
features:
```

src/digraph.rs
```diff
@@ -2288,34 +2288,81 @@ impl PyDiGraph {
...
     /// graph. All edges in this graph will be created as undirected edges in
-    /// the new graph object.
+    /// the new graph object. For directed graphs with bidirectional edges, you
+    /// can set `multigraph=False` to condense them into a single edge and specify
+    /// a function to combine the weights/data of the edges.
     /// Do note that the node and edge weights/data payloads will be passed
     /// by reference to the new :class:`~retworkx.PyGraph` object.
     ///
+    /// :param bool multigraph: If set to `False` the output graph will not
+    ///     allow parallel edges. Default: `True`.
```
mtreinish, line 2298: We should explain a bit more here about what it does when it encounters a parallel edge (both with and without `weight_combo_fn` set). When I read this my first thought would be that if `multigraph=False` it raises when there are parallel edges because they're not allowed

src/digraph.rs
```diff
@@ -2288,34 +2288,81 @@ impl PyDiGraph {
...
+    ///     two edge weight/data object and return a new edge weight/data
+    ///     object that will be used when adding an edge between two nodes
+    ///     connected by multiple edges (of either direction) in the original
+    ///     directed graph.
     /// :returns: A new PyGraph object with an undirected edge for every
     ///     directed edge in this graph
     /// :rtype: PyGraph
-    #[text_signature = "(self)"]
-    pub fn to_undirected(&self, py: Python) -> crate::graph::PyGraph {
+    #[text_signature = "(self, multigraph, weight_combo_fn, /)"]
```
mtreinish, line 2307:
```suggestion
    #[text_signature = "(self, /, multigraph=True, weight_combo_fn=None)"]
```

src/digraph.rs
```diff
@@ -2288,34 +2288,81 @@ impl PyDiGraph {
...
+                        Some(value) => {
+                            *old_weight = value;
+                        }
+                        None => {
+                            *old_weight = weight;
+                        }
+                    }
+                } else {
+                    new_graph.add_edge(source, target, weight);
+                }
```
mtreinish, line 2358: I'm surprised you didn't do something like:

```suggestion
                match exists {
                    Some(index) => {
                        let old_weight = new_graph.edge_weight_mut(index).unwrap();
                        match combine(old_weight, edge.weight(), &weight_combo_fn)?
                        {
                            Some(value) => {
                                *old_weight = value;
                            }
                            None => {
                                *old_weight = weight;
                            }
                        }
                    },
                    None => new_graph.add_edge(source, target, weight),
                }
```

But using an if works fine too, there's no functional difference.

mtreinish (commented):
Overall LGTM, thanks for doing this so quickly. Just some small docs issues inline once those are fixed this should be good to go.

src/digraph.rs
georgios-ts:
```suggestion
    ///     allow parallel edges. Instead parallel edges will be condensed
    ///     into a single edge and their data will be combined using
    ///     `weight_combo_fn`. If `weight_combo_fn` is not provided, the data
    ///     of the edge with the largest index will be kept. Default: `True`.
```

mtreinish (approved):
LGTM, thanks for the quick update

PR #372, opened 2021-06-23
Add induced kwarg in `is_subgraph_isomorphic`

georgios-ts:
This commit adds a kwarg `induced` in `is_subgraph_isomorphic` to determine if we should check for a *node-induced* subgraph or *not* induced. Previously, only node-induced case was supported.
In not induced case, the changes are minimal: we shouldn't  check `R_new` pruning rule and check neighbors in `R_suc` and `R_pred` rules only in the second graph.

### DIFF
--- a/retworkx/__init__.py
+++ b/retworkx/__init__.py
@@ -744,14 +744,23 @@ def _graph_is_isomorphic_node_match(first, second, matcher, id_order=True):
 
 @functools.singledispatch
 def is_subgraph_isomorphic(
-    first, second, node_matcher=None, edge_matcher=None, id_order=False
+    first,
+    second,
+    node_matcher=None,
+    edge_matcher=None,
+    id_order=False,
+    induced=True,
 ):
     """Determine if 2 graphs are subgraph isomorphic
 
     This checks if 2 graphs are subgraph isomorphic both structurally and also
     comparing the node and edge data using the provided matcher functions.
-    The matcher functions take in 2 data objects and will compare them. A
-    simple example that checks if they're just equal would be::
+    The matcher functions take in 2 data objects and will compare them.
+    Since there is an ambiguity in the term 'subgraph', do note that we check
+    for an node-induced subgraph if argument `induced` is set to `True`. If it is
+    set to `False`, we check for a non induced subgraph, meaning the second graph
+    can have fewer edges than the subgraph of the first. A simple example that checks
+    if they're just equal would be::
 
             graph_a = retworkx.PyGraph()
             graph_b = retworkx.PyGraph()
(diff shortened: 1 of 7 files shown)


retworkx/__init__.py
```diff
@@ -744,14 +744,23 @@ def _graph_is_isomorphic_node_match(first, second, matcher, id_order=True):
...
 ):
     """Determine if 2 graphs are subgraph isomorphic
 
     This checks if 2 graphs are subgraph isomorphic both structurally and also
     comparing the node and edge data using the provided matcher functions.
-    The matcher functions take in 2 data objects and will compare them. A
-    simple example that checks if they're just equal would be::
+    The matcher functions take in 2 data objects and will compare them.
+    Since there is an ambiguity in the term 'subgraph', do note that we check
+    for an node-induced subgraph if argument `induced` is set to `True`. If it is
```
mtreinish, line 760: Maybe mention that `True` is the default setting here too

src/lib.rs
```diff
@@ -462,15 +464,16 @@ fn graph_is_isomorphic(
 /// :returns: ``True`` if there is a subgraph of `first` isomorphic to `second`,
 ///     ``False`` if there is not.
 /// :rtype: bool
```
mtreinish, line 466: Do you want to update the docstrings here (and for the graph function) to add the new parameter and the description of node-induced too?

mtreinish (commented):
Codewise this LGTM, it's a relatively simple change. Just a few minor docs issues then I think this is good to go.

mtreinish (approved):
LGTM< thanks for updating so quickly. There is one nit inline, but not worth blocking anything over.

tests/graph/test_subgraph_isomorphic.py
```diff
@@ -180,5 +180,27 @@ def test_non_induced_subgraph_isomorphic(self):
         for id_order in [False, True]:
             with self.subTest(id_order=id_order):
                 self.assertFalse(
-                    retworkx.is_subgraph_isomorphic(g_a, g_b, id_order=id_order)
+                    retworkx.is_subgraph_isomorphic(
+                        g_a, g_b, id_order=id_order, induced=True
+                    )
+                )
+            with self.subTest(id_order=id_order):
```
mtreinish, line 187: If adding an additional subtest here I would add an additional parameter here to make it clear that this is for induced=False too. Something like:

```suggestion
            with self.subTest(id_order=id_order, induced=False):
```
(and adding `induced=True` to the above subTest)

That being said it's not critical though, because this really only controls the error message and not much else. It's not selectable for isolating execution or anything like that.

georgios-ts: Good point, added in f53404b

PR #375, opened 2021-07-01
Stabilize Vf2++ matching order

georgios-ts:
This PR removes the unstable sort and the (unnecessary) use of a `HashSet` that would make the result of VF2++ heuristic matching order non-deterministic.

### DIFF
--- a/src/isomorphism.rs
+++ b/src/isomorphism.rs
@@ -15,10 +15,9 @@
 
 use fixedbitset::FixedBitSet;
 use std::cmp::Ordering;
-use std::iter::FromIterator;
 use std::marker;
 
-use hashbrown::{HashMap, HashSet};
+use hashbrown::HashMap;
 
 use super::NodesRemoved;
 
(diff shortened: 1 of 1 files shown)


mtreinish (approved):
LGTM

I'm curious is there any difference in performance form this? I expect the sort to be slower but the use of a vec for `next_level` to be faster, so I'm not sure what to expect.

IvanIsCoding (approved):
LGTM, it will make the output deterministic

> LGTM
> 
> I'm curious is there any difference in performance form this? I expect the sort to be slower but the use of a vec for `next_level` to be faster, so I'm not sure what to expect.

I wouldn't be surprised if the difference is negligible. This change only affects the matching order, no? If we benchmark the algorithm, I belive we will spend more time on `try_match`

1ucian0:
In case this has performance consequence, on my side, I would be okey if this is optional.

georgios-ts:
Running this benchmark, i don't really see any significant difference:
```python
import timeit
import random

import retworkx as rx


def permute(graph, seed=42):
    nodes = list(graph.node_indexes())

    rng = random.Random(seed)
    rng.shuffle(nodes)
    
    edges = [(nodes[u], nodes[v]) for (u, v) in graph.edge_list()]
    
    if isinstance(graph, rx.PyDiGraph):
        res = rx.PyDiGraph()
    else:
        res = rx.PyGraph()
        
    res.add_nodes_from(nodes)
    res.add_edges_from_no_data(edges)
    
    return res


def benchmark():
    n = 10000
    degrees = [10, 15, 50, 100]

    cases = []
    for deg in degrees:
        p = 2 * deg / (n - 1)

        g_a = rx.directed_gnp_random_graph(n, p, seed=42)
        g_b = permute(g_a, seed=4242)

        cases += [(g_a, g_a), (g_a, g_b)]
    
    return cases


def run(cases):
    for (g_a, g_b) in cases:
        res = rx.is_isomorphic(g_a, g_b,
                               id_order=False)
        assert res
        
cases = benchmark()
print(
    timeit.timeit("run(cases)", globals=globals(), number=5)
)
```
| Branch       | Run time (sec)          |
| ------------- |-------------| 
| master  | 20.348561446 |
| #375 | 20.359493619 |

PR #376, opened 2021-07-02
Return an iterator over all valid Vf2 mappings

georgios-ts:
Building on top of #368, this PR refactors Vf2 algorithm and implements a `vf2_mapping` function that returns a custom python class that acts as a python iterator that lazily evaluates all valid Vf2 mappings. The "trick" that allows the lazily evaluation is to take ownership of a copy of the graphs. We can do this (with minimal overhead) since we were already creating new graphs in case of `id_order=False` or `nodes_removed=True`.  At the same time, it adjusts `Vf2ppSorter` to handle graphs with non-contiguous node ids.

TODO
- [x] More tests


retworkx/__init__.py
```diff
@@ -750,6 +755,9 @@ def is_isomorphic(
     :param bool id_order: If set to ``False`` this function will use a
         heuristic matching order based on [VF2]_ paper. Otherwise it will
         default to matching the nodes in order specified by their ids.
+    :param int call_limit: An optional bound on the number of states that VF2
+        algorithm visits while searching for a solution. If it exceeds this limit,
+        the algorithm will stop and return ``False``. Default: ``None``.
```
mtreinish, line 760: Default `None` here means there is no limit right? It's just a bit ambiguous, I'd probably just drop it since it's covered by the optional part:

```suggestion
        the algorithm will stop and return ``False``.
```

retworkx/__init__.py
```diff
@@ -873,6 +892,9 @@ def is_subgraph_isomorphic(
     :param bool induced: If set to ``True`` this function will check the existence
         of a node-induced subgraph of first isomorphic to second graph.
         Default: ``True``.
+    :param int call_limit: An optional bound on the number of states that VF2
+        algorithm visits while searching for a solution. If it exceeds this limit,
+        the algorithm will stop and return ``False``. Default: ``None``.
```
mtreinish, line 897:
```suggestion
        the algorithm will stop and return ``False``.

File: src/isomorphism/mod.rs [hunk 3]
```diff
@@ -50,13 +50,17 @@ use pyo3::Python;
 /// :param bool id_order: If set to ``False`` this function will use a
 ///     heuristic matching order based on [VF2]_ paper. Otherwise it will
 ///     default to matching the nodes in order specified by their ids.
+/// :param int call_limit: An optional bound on the number of states that VF2 algorithm
+///     visits while searching for a solution. If it exceeds this limit, the algorithm
+///     will stop and return ``False``. Default: ``None``.
```
mtreinish, line 55:
```suggestion
///     will stop and return ``False``.
```

src/isomorphism/mod.rs
```diff
@@ -122,13 +113,17 @@ fn digraph_is_isomorphic(
 /// :param bool (default=True) id_order:  If set to true, the algorithm matches the
 ///     nodes in order specified by their ids. Otherwise, it uses a heuristic
 ///     matching order based in [VF2]_ paper.
+/// :param int call_limit: An optional bound on the number of states that VF2 algorithm
+///     visits while searching for a solution. If it exceeds this limit, the algorithm
+///     will stop and return ``False``. Default: ``None``.
```
mtreinish, line 118:
```suggestion
///     will stop and return ``False``.

File: src/isomorphism/mod.rs [hunk 5]
```diff
@@ -50,13 +50,17 @@ use pyo3::Python;
...
 ///     default to matching the nodes in order specified by their ids.
+/// :param int call_limit: An optional bound on the number of states that VF2 algorithm
+///     visits while searching for a solution. If it exceeds this limit, the algorithm
+///     will stop and return ``False``. Default: ``None``.
 ///
 /// :returns: ``True`` if the 2 graphs are isomorphic ``False`` if they are
 ///     not.
 /// :rtype: bool
-#[pyfunction(id_order = "true")]
+#[pyfunction(id_order = "true", call_limit = "None")]
```
mtreinish, line 60: You don't actually need this as it's implicit with the `Option<usize>` type, the pyo3 pyfunction macro will default it to `None` so we don't need to explicitly set the default for the argument (it's the reason `node_matcher` and `edge_matcher` aren't in the list here.

```suggestion
#[pyfunction(id_order = "true")]
```

src/isomorphism/mod.rs
```diff
@@ -201,13 +183,17 @@ fn graph_is_isomorphic(
 /// :param bool induced: If set to ``True`` this function will check the existence
 ///     of a node-induced subgraph of first isomorphic to second graph.
 ///     Default: ``True``.
+/// :param int call_limit: An optional bound on the number of states that VF2 algorithm
+///     visits while searching for a solution. If it exceeds this limit, the algorithm
+///     will stop and return ``False``. Default: ``None``.
```
mtreinish, line 188:
```suggestion
///     will stop and return ``False``.
```

src/isomorphism/mod.rs
```diff
@@ -122,13 +113,17 @@ fn digraph_is_isomorphic(
...
 ///     matching order based in [VF2]_ paper.
+/// :param int call_limit: An optional bound on the number of states that VF2 algorithm
+///     visits while searching for a solution. If it exceeds this limit, the algorithm
+///     will stop and return ``False``. Default: ``None``.
 ///
 /// :returns: ``True`` if the 2 graphs are isomorphic ``False`` if they are
 ///     not.
 /// :rtype: bool
-#[pyfunction(id_order = "true")]
+#[pyfunction(id_order = "true", call_limit = "None")]
```
mtreinish, line 123:
```suggestion
#[pyfunction(id_order = "true")]
```

src/isomorphism/mod.rs
```diff
@@ -201,13 +183,17 @@ fn graph_is_isomorphic(
...
 ///     Default: ``True``.
+/// :param int call_limit: An optional bound on the number of states that VF2 algorithm
+///     visits while searching for a solution. If it exceeds this limit, the algorithm
+///     will stop and return ``False``. Default: ``None``.
 ///
 /// :returns: ``True`` if there is a subgraph of `first` isomorphic to `second`,
 ///     ``False`` if there is not.
 /// :rtype: bool
-#[pyfunction(id_order = "false", induced = "true")]
+#[pyfunction(id_order = "false", induced = "true", call_limit = "None")]
```
mtreinish, line 193:
```suggestion
#[pyfunction(id_order = "false", induced = "true")]
```

src/isomorphism/mod.rs
```diff
@@ -281,13 +254,17 @@ fn digraph_is_subgraph_isomorphic(
 /// :param bool induced: If set to ``True`` this function will check the existence
 ///     of a node-induced subgraph of first isomorphic to second graph.
 ///     Default: ``True``.
+/// :param int call_limit: An optional bound on the number of states that VF2 algorithm
+///     visits while searching for a solution. If it exceeds this limit, the algorithm
+///     will stop and return ``False``. Default: ``None``.
```
mtreinish, line 259:
```suggestion
///     will stop and return ``False``.
```

src/isomorphism/mod.rs
```diff
@@ -281,13 +254,17 @@ fn digraph_is_subgraph_isomorphic(
...
 ///
 /// :returns: ``True`` if there is a subgraph of `first` isomorphic to `second`,
 ///     ``False`` if there is not.
 /// :rtype: bool
-#[pyfunction(id_order = "false", induced = "true")]
+#[pyfunction(id_order = "false", induced = "true", call_limit = "None")]
```
mtreinish, line 264:
```suggestion
#[pyfunction(id_order = "false", induced = "true")]
```

src/isomorphism/mod.rs
```diff
@@ -297,30 +274,182 @@ fn graph_is_subgraph_isomorphic(
...
+///     will stop. Default: ``None``.
+///
+/// :returns: An iterator over dicitonaries of node indices from ``first`` to node
+///     indices in ``second`` representing the mapping found.
+/// :rtype: Iterable[NodeMap]
+#[pyfunction(
+    id_order = "true",
+    subgraph = "false",
+    induced = "true",
+    call_limit = "None"
```
mtreinish, line 339:
```suggestion
```

src/isomorphism/mod.rs
```diff
@@ -297,30 +274,182 @@ fn graph_is_subgraph_isomorphic(
...
+///     heuristic matching order based on [VF2]_ paper. Otherwise it will
+///     default to matching the nodes in order specified by their ids.
+/// :param bool subgraph: If set to ``True`` the function will return the
+///     subgraph isomorphic found between the graphs.
+/// :param bool induced: If set to ``True`` this function will check the existence
+///     of a node-induced subgraph of first isomorphic to second graph.
+///     Default: ``True``.
+/// :param int call_limit: An optional bound on the number of states that VF2 algorithm
+///     visits while searching for a solution. If it exceeds this limit, the algorithm
+///     will stop. Default: ``None``.
```
mtreinish, line 330:
```suggestion
///     will stop.
```

src/isomorphism/mod.rs
```diff
@@ -297,30 +274,182 @@ fn graph_is_subgraph_isomorphic(
...
+///
+/// :returns: An iterator over dicitonaries of node indices from ``first`` to node
+///     indices in ``second`` representing the mapping found.
+/// :rtype: Iterable[NodeMap]
+#[pyfunction(
+    id_order = "true",
+    subgraph = "false",
+    induced = "true",
+    call_limit = "None"
```
mtreinish, line 421:
```suggestion
```

src/lib.rs
```diff
@@ -11,6 +11,7 @@
 // under the License.
 
 #![allow(clippy::float_cmp)]
+#![allow(clippy::too_many_arguments)]
```
mtreinish, line 14: Is this needed anymore?

src/isomorphism/vf2.rs
```diff
@@ -318,51 +335,20 @@ where
...
     }
     #[inline]
-    fn eq(&mut self, a: &T, b: &T) -> PyResult<bool> {
-        let res = (self.as_mut().unwrap())(a, b)?;
-        Ok(res)
+    fn eq(&mut self, py: Python, a: &PyObject, b: &PyObject) -> PyResult<bool> {
+        let res = self.as_mut().unwrap().call1(py, (a, b))?;
+        res.is_true(py)
     }
 }
```
mtreinish, line 353: It's not really related to this PR, but I'm wondering if this makes sense to just refactor to not be a trait? It seems we don't this to be generic over any type `T` that's a potential weight as we always are using a `PyObject`. This was originally forked from petgraph and they need the generic definition here because the weight in a graph type in petgraph can be any type. But that's something we can do in a followup, it's not worth changing here.

src/isomorphism/vf2.rs
```diff
@@ -318,51 +335,20 @@ where
...
     fn enabled(&self) -> bool {
         self.is_some()
     }
     #[inline]
-    fn eq(&mut self, a: &T, b: &T) -> PyResult<bool> {
-        let res = (self.as_mut().unwrap())(a, b)?;
-        Ok(res)
+    fn eq(&mut self, py: Python, a: &PyObject, b: &PyObject) -> PyResult<bool> {
+        let res = self.as_mut().unwrap().call1(py, (a, b))?;
+        res.is_true(py)
```
mtreinish, line 351: Nice! Definitely a lot easier to handle this in the sematic matcher than forcing callers to define a closure that does this. It also simplifies the types for the functions below since we only pass around a PyObject.

src/isomorphism/vf2.rs
```diff
@@ -765,76 +787,228 @@ where
...
+        impl PyGCProtocol for $name {
+            fn __traverse__(
+                &self,
+                visit: PyVisit,
+            ) -> Result<(), PyTraverseError> {
+                for j in 0..2 {
+                    for node in
+                        self.vf2.st[j].graph.node_indices().map(|node| {
+                            self.vf2.st[j].graph.node_weight(node).unwrap()
+                        })
```
mtreinish, line 982: Would it be easier to just iterate over[ `node_weights()`](https://docs.rs/petgraph/0.6.0/petgraph/stable_graph/struct.StableGraph.html#method.node_weights) here instead of mapping the indices? Something like:

```suggestion
                        self.vf2.st[j].graph.node_weights()
```

src/isomorphism/vf2.rs
```diff
@@ -765,76 +787,228 @@ where
...
+                    {
+                        visit.call(node)?;
+                    }
+                    for edge in
+                        self.vf2.st[j].graph.edge_indices().map(|edge| {
+                            self.vf2.st[j].graph.edge_weight(edge).unwrap()
+                        })
```
mtreinish, line 989: Same thing here we can't we just use [`edge_weights()`](https://docs.rs/petgraph/0.6.0/petgraph/stable_graph/struct.StableGraph.html#method.edge_weights)?

```suggestion
                        self.vf2.st[j].graph.edge_weights()
```

src/isomorphism/vf2.rs
```diff
@@ -10,74 +10,90 @@
 // License for the specific language governing permissions and limitations
 // under the License.
 
+#![allow(clippy::too_many_arguments)]
 // This module is a forked version of petgraph's isomorphism module @ 0.5.0.
 // It has then been modified to function with PyDiGraph inputs instead of Graph.
```
mtreinish, line 15: I would probably reword this comment now to something like

```suggestion
// This module was originally forked from petgraph's isomorphism module @ v0.5.0
// to handle PyDiGraph inputs instead of petgraph's generic Graph. However it has
// since diverged significantly from the original petgraph implementation.
```

since the petgraph lineage is still there but it's not longer just about handling `PyDiGraph`,
well specifically `StableGraph<PyObject, PyObject, Ty>` (the old petgraph isomorphism
only worked with `Graph`) and handling `PyResult` on the return for the semantic matcher
as we've added a lot of new functionality here since I originally forked it.

mtreinish (commented):
This looks great! I especially like a lot of the internal refactoring of the the core vf2 algorithm to be more explicitly defined (ie functions on structs functions instead of a bunch of nested closures) and the iterator built on the reordered copy of the graph is an awesome way to workaround the lifetime issue. I've got a few inline comments but most are on small mechanical things.

Besides those the only other thing I think would be good to add test coverage for some of the uncovered negative paths from the coverage report. https://coveralls.io/builds/42102310/source?filename=src%2Fisomorphism%2Fvf2.rs or leave a comment why it's unreachable like you did for the paths which you did for some of those paths. Assuming it's feasible for some of those it looked like it'd be kind of annoying to come up with a test case to cause those conditions.

src/lib.rs
georgios-ts: No it's not needed.
```suggestion
```

retworkx/__init__.py
georgios-ts: Yes, `None` means no limit.

src/isomorphism/vf2.rs
georgios-ts: yes, it's definitely easier. i just copy+pasted this from `graph.rs`.

src/isomorphism/vf2.rs
georgios-ts: Actually i like this even if we don't really use it but yeah, we can remove it in a followup PR

src/isomorphism/mod.rs
```diff
@@ -297,30 +275,172 @@ fn graph_is_subgraph_isomorphic(
...
+///         except StopIteration:
+///             pass
+///
+///
+/// :param PyDiGraph first: The first graph to find the mapping for
+/// :param PyDiGraph second: The second graph to find the mapping for
+/// :param node_matcher: An optional python callable object that takes 2
+///     positional arguments, one for each node data object in either graph.
+///     If the return of this function evaluates to True then the nodes
+///     passed to it are vieded as matching.
```
mtreinish, line 316:
```suggestion
///     passed to it are viewed as matching.
```

src/isomorphism/mod.rs
```diff
@@ -297,30 +275,172 @@ fn graph_is_subgraph_isomorphic(
...
+/// :param edge_matcher: A python callable object that takes 2 positional
+///     one for each edge data object. If the return of this
+///     function evaluates to True then the edges passed to it are vieded
```
mtreinish, line 319:
```suggestion
///     function evaluates to True then the edges passed to it are viewed
```

retworkx/__init__.py
```diff
@@ -1392,3 +1416,107 @@ def _digraph_num_shortest_paths_unweighted(graph, source):
...
+                mapping = next(vf2)
+            except StopIteration:
+                pass
+
+    :param first: The first graph to find the mapping for
+    :param second: The second graph to find the mapping for
+    :param node_matcher: An optional python callable object that takes 2
+        positional arguments, one for each node data object in either graph.
+        If the return of this function evaluates to True then the nodes
+        passed to it are vieded as matching.
```
mtreinish, line 1455:
```suggestion
        passed to it are viewed as matching.
```

retworkx/__init__.py
```diff
@@ -1392,3 +1416,107 @@ def _digraph_num_shortest_paths_unweighted(graph, source):
...
+    :param edge_matcher: A python callable object that takes 2 positional
+        one for each edge data object. If the return of this
+        function evaluates to True then the edges passed to it are vieded
```
mtreinish, line 1458:
```suggestion
        function evaluates to True then the edges passed to it are viewed
```

mtreinish (approved):
LGTM, thanks for making the quick updates and adding more tests to catch the different negative cases.

PR #377, opened 2021-07-04
Bump petgraph version to the latest release

mtreinish:
Petgraph 0.6.0 was just released, this commit bumps the version in the
cargo.toml to the new release. It also updates some of the petgraph
usage to take advantage of some new features.


IvanIsCoding (approved):
Very convenient upgrade, they improved their traits and most of the breaking changes do not impact us. Their bump in minimum Rust version will not impact us either, because PyO3 requires Rust 1.41+ anyway

Another enhancement from 0.6 that could help us reduce our code is `Added a generic graph complement implementation`. We could possibly use it to simplify ours implementation

PR #383, opened 2021-07-11
Add weight_fn kwarg to longest path functions

### LINKED ISSUE #381: Procedurally weight edges during longest path calculation

`dag_longest_path_length` presently does what it says on the tin: it computes the length of the longest chain of edges in the DAG. In applications, we often want to tweak this in two ways:

1. Rather than treating edges uniformly and tallying up their count, we might want to procedurally weight edges according to some user-specified scheme. It's possible to attach this data to the graph object itself, but it can be convenient to supply this to `dag_longest_path_length` itself in some non-persistent way, e.g., as a function.
2. Rather than adding edge weights and taking a maximum at each node to form the running tally, some other combinator might be called for.

As QISKit-motivated examples, we might like to estimate runtime infidelity of a circuit, where each gate incurs some cost (modeled as an edge weight in the `DAGCircuit`) and the errors propagate according to some compounding formula (encoded as a non-max combinator). Even just addressing 1. would be helpful to these ends.

mtreinish:
This commit adds a new kwarg, weight_fn, to the dag_longest_path()
function that is used to dynamical adjust the edge weight for finding
the longest path. Previously it assumed a constant weight of 1 for all
edges.

Fixes #381


tests/digraph/test_depth.py
```diff
@@ -69,3 +69,111 @@ def test_empty_graph(self):
...
+            [node_a, node_c, node_e],
+            retworkx.dag_longest_path(
+                dag, weight_fn=lambda _, __, weight: weight
+            ),
+        )
+
+    def test_degenerate_graph_with_weight(self):
+        dag = retworkx.PyDAG()
+        dag.add_node(0)
+        self.assertEqual([0], retworkx.dag_longest_path(dag, weight_fn=int))
```
georgios-ts, line 169: I'm a bit confused with `weight_fn=int` here. Isn't `weight_fn` supposed to take 3 arguments? I believe this would raise a `TypeError` if the graph was not trivial.

src/lib.rs
```diff
@@ -79,16 +94,28 @@ fn longest_path(graph: &digraph::PyDiGraph) -> PyResult<Vec<usize>> {
...
+                p_edge.weight(),
+            )?;
+            local_dist
+                .entry(p_node)
+                .and_modify(|x| *x = cmp::max(*x, weight))
+                .or_insert(weight);
+        }
+        for (p_node, weight) in local_dist.iter() {
+            let length = dist[p_node].0 + weight;
+            us.push((length, *p_node));
```
georgios-ts, line 118: Is it really necessary to store the "local" weights? Can't we just do something like:
```suggestion
        for p_edge in parents {
            let p_node = p_edge.source();
            let weight: usize = edge_weight_callable(
                p_node.index(),
                p_edge.target().index(),
                p_edge.weight(),
            )?;
            let length = dist[&p_node].0 + weight;
            us.push((length, p_node));
```
since after all we only keep the maximum over `us`.

tests/digraph/test_depth.py
mtreinish: Oops, yeah it's just sloppy copy and paste errors from me. You're correct this would raise a type error if the weight_fn was ever called. I'll fix this and add some negative tests to catch the type error case too.

src/lib.rs
mtreinish: My concern with doing this was in the parallel edge case won't `dist[&p_node].0` contain a sum of parallel edges from `p_node` to `node` while we want to have just the maximum weight.

mtreinish: Oh, yep I was misreading the code (jumping back and forth between python and rust too much), yeah this is fine we're not modifying things in dist until we find the max.

tests/digraph/test_depth.py
mtreinish: Done in https://github.com/Qiskit/retworkx/pull/383/commits/fb5f28ab687f789d8dab5841cdd66d84587ad581

georgios-ts: It seems that you fixed it for the function call below but not at this line.

mtreinish: sigh, not sure how I missed that, one sec I'll update it for real

mtreinish: ok, actually fixed in: https://github.com/Qiskit/retworkx/pull/383/commits/d5167ecc218d2c485a73c8ef288e170f124720e3

georgios-ts (approved):

PR #384, opened 2021-07-12
Bump pyo3, numpy, and bigint to the latest releases

mtreinish:
This commit bumps PyO3 to the latest release version 0.14 along with
bigint and rust-numpy all of which are dependent on the pyo3 release.
As part of this upgrade the code is updated to avoid newly deprecated
syntax and fix any breakages caused by the new versions.


mtreinish:
It looks like this is blocked on: https://github.com/PyO3/rust-numpy/issues/198 we'll have to wait for a numpy 0.14.1 I guess

PR #386, opened 2021-07-15
Deterministic results in custom return types with `IndexMap`

IvanIsCoding:
Related to the discussions in #347 about deterministic behaviour

Replaces HashMap with IndexMap from the [indexmap](https://github.com/bluss/indexmap) as the default hashmap in custom return types. IndexMap is inspired by Python's dict implementation, hence it presevers insertion order as long items are not removed. This leads to a deterministic behaviour, and removes the need to fix the seed in the hashmap to  reproduce the results.

TODO List:
- [x] ~~Benchmark code to compare the performance~~ 
- [x] ~~Make `all_pairs_dijkstra` parallel again~~ (that was easy, just had to enable `rayon` feature in indexmap) 
- [x] ~~Contribute the traits for `indexmap::IndexMap` to PyO3~~ (done in https://github.com/PyO3/pyo3/pull/1728)

## Comments

If this gets approved, we'd need to wait for a new PyO3 release. To make the conversions between PyDict and IndexMap work, I had to implement the traits on their crate.


mtreinish:
The pyo3 implementation will be pretty simple, I did the impl for hashbrown's HashMap: https://github.com/PyO3/pyo3/pull/1114 (although obviously check the current HEAD as I'm sure there have been updates, that's just a good starting point) it'll probably be more or less identical to that PR. If this does require a pyo3 change 2 options in the interim is to create a custom wrapper struct that contains a `IndexMap` and just implements the pyo3 traits locally, or use a wrapper function that takes an index map and returns a pydict.

IvanIsCoding:
I made progress on this, the trait is now in PyO3 so in the future we will be able to use it. I still have to consider if it is worth the effort to create temporary code while PyO3 doesn't release.

With regards to benchmark, I will have to redo them. Originally I was using the default hasher for IndexMap and it was slower than hashbrown's because hashbrown's ahash is super fast. I've switched to using hashbrown's default hasher with `indexmap::IndexMap<K, V, hashbrown::hash_map::DefaultHashBuilder>` and for example https://github.com/Qiskit/retworkx/issues/347#issuecomment-881135126 had a much smaller difference. I will post when I have more data.

mtreinish:
I ran the benchmarks from https://github.com/mtreinish/retworkx-bench and it showed an improvement on the time_graph_greed_coloring becnhmarks and a small regression on floyd warshall:

```
Benchmarks that have improved:

       before           after         ratio
     [9f24a937]       [21469ec9]
     <main>       <IvanIsCoding-deterministic-hashmap>
-        2.08±0μs      1.87±0.01μs     0.90  coloring.ColoringBenchmarks.time_graph_greedy_coloring(1, 100)
-     11.4±0.06ms      9.22±0.07ms     0.81  coloring.ColoringBenchmarks.time_graph_greedy_coloring(1, 1000000)
-        10.1±0μs      7.71±0.01μs     0.77  coloring.ColoringBenchmarks.time_graph_greedy_coloring(1, 1000)
-     91.2±0.04μs       67.3±0.1μs     0.74  coloring.ColoringBenchmarks.time_graph_greedy_coloring(1, 10000)
-       892±0.7μs        654±0.7μs     0.73  coloring.ColoringBenchmarks.time_graph_greedy_coloring(1, 100000)

Benchmarks that have stayed the same:

       before           after         ratio
     [9f24a937]       [21469ec9]
     <main>       <IvanIsCoding-deterministic-hashmap>
       63.7±0.2μs       70.0±0.5μs     1.10  floyd_warshall.FloydWarshall.time_floyd_warshall(100, 10)
          24.6±0s       26.9±0.03s     1.09  coloring.ColoringRoadMapFullUSA.time_graph_greedy_coloring
       6.32±0.02s       6.87±0.02s     1.09  coloring.TwoDimensionalDynamicSimulation.time_graph_greedy_coloring
       91.7±0.7μs       98.2±0.1μs     1.07  floyd_warshall.FloydWarshall.time_floyd_warshall(100, 1000)
        159±0.8ms          170±5ms     1.06  coloring.ColoringRoadMapNYC.time_graph_greedy_coloring
       28.8±0.2μs       29.8±0.2μs     1.04  floyd_warshall.FloydWarshall.time_floyd_warshall(10, 1000)
            1.94G            1.98G     1.02  coloring.TwoDimensionalDynamicSimulation.peakmem_graph_greedy_coloring
      1.17±0.01μs      1.20±0.01μs     1.02  coloring.ColoringBenchmarks.time_graph_greedy_coloring(1, 1)
       27.8±0.1ms       28.1±0.3ms     1.01  floyd_warshall.FloydWarshall.time_floyd_warshall(1000, 10)
         1.23±0μs      1.23±0.02μs     1.00  coloring.ColoringBenchmarks.time_graph_greedy_coloring(1, 10)
            71.5M            71.6M     1.00  floyd_warshall.FloydWarshall.peakmem_floyd_warshall(100, 1000)
            10.3G            10.3G     1.00  coloring.ColoringRoadMapFullUSA.peakmem_graph_greedy_coloring
            3.22G            3.22G     1.00  coloring.ColoringRoadMapAsia.peakmem_graph_greedy_coloring
            71.5M            71.4M     1.00  floyd_warshall.FloydWarshall.peakmem_floyd_warshall(100, 10)
            71.3M            71.3M     1.00  floyd_warshall.FloydWarshall.peakmem_floyd_warshall(10, 1000)
            71.5M            71.4M     1.00  floyd_warshall.FloydWarshall.peakmem_floyd_warshall(100, 100)
            71.5M            71.4M     1.00  floyd_warshall.FloydWarshall.peakmem_floyd_warshall(10, 10)
       27.2±0.4ms       27.1±0.3ms     1.00  floyd_warshall.FloydWarshall.time_floyd_warshall(1000, 100)
            71.5M            71.2M     1.00  floyd_warshall.FloydWarshall.peakmem_floyd_warshall(10, 100)
            72.3M              72M     1.00  floyd_warshall.FloydWarshall.peakmem_floyd_warshall(1000, 100)
            72.3M              72M     1.00  floyd_warshall.FloydWarshall.peakmem_floyd_warshall(1000, 10)
            72.4M            72.1M     1.00  floyd_warshall.FloydWarshall.peakmem_floyd_warshall(1000, 1000)
       27.1±0.1ms       26.9±0.4ms     0.99  floyd_warshall.FloydWarshall.time_floyd_warshall(1000, 1000)
             194M             192M     0.99  coloring.ColoringRoadMapNYC.peakmem_graph_greedy_coloring
            2.38G            2.34G     0.98  coloring.RandomGeometricGraph.peakmem_graph_greedy_coloring

Benchmarks that have got worse:

       before           after         ratio
     [9f24a937]       [21469ec9]
     <main>       <IvanIsCoding-deterministic-hashmap>
+     5.57±0.01μs       6.56±0.2μs     1.18  floyd_warshall.FloydWarshall.time_floyd_warshall(10, 10)
+     7.69±0.01μs       8.60±0.2μs     1.12  floyd_warshall.FloydWarshall.time_floyd_warshall(10, 100)
+      68.4±0.3μs       75.8±0.3μs     1.11  floyd_warshall.FloydWarshall.time_floyd_warshall(100, 100)
```

Everything else in the benchmark suite remained basically the same. So it looks like indexmap is a good solution for retaining insertion order without any performance penalty.

mtreinish:
So with #407 merging I think this is unblocked now (pyo3 0.14.2 include the indexmap support). But since we're close to 0.10.0 I think we should hold this after the release and wait we open 0.11.0 to make this change. Just so we have more time to live with the potential performance impact of switching to indexmap and make adjustments.

IvanIsCoding:
> So with #407 merging I think this is unblocked now (pyo3 0.14.2 include the indexmap support). But since we're close to 0.10.0 I think we should hold this after the release and wait we open 0.11.0 to make this change. Just so we have more time to live with the potential performance impact of switching to indexmap and make adjustments.

This is ready for review and no longer WIP. I will let you make the call for which release this goes in

src/dictmap.rs
```diff
@@ -0,0 +1,31 @@
+// Convenient alias to build IndexMap using a custom hasher.
+// For the moment, we use ahash which is the default hasher
+// for hashbrown::HashMap, another hashmap we use
+
+pub type DictMap<K, V> = indexmap::IndexMap<K, V, ahash::RandomState>;
+
+#[macro_export]
+macro_rules! _dictmap_new {
```
georgios-ts, line 8: A macro works fine here. In case we want to keep the common syntax `DictMap::new()` we could define a custom trait like 
```rust
pub trait Init {
    fn new() -> Self
    where
        Self: Sized;

    fn with_capacity(n: usize) -> Self
    where
        Self: Sized;
}

impl<K, V> Init for DictMap<K, V> {
    fn new() -> Self {
        indexmap::IndexMap::with_capacity_and_hasher(
            0,
            ahash::RandomState::default(),
        )
    }

    fn with_capacity(n: usize) -> Self {
        indexmap::IndexMap::with_capacity_and_hasher(
            n,
            ahash::RandomState::default(),
        )
    }
}
```

georgios-ts (commented):
Great work! 
There are still a few places, besides custom return types, where we return a `HashMap` and we could switch to `IndexMap`.  I have spotted `graph::adj`, `digraph::adj`, `digraph::adj_direction`, `graph::compose`, `digraph::compose`, `core_number` and `max_weight_matching`.

My guess is that we will not see a significant performance regression since most of the algorithms are still using a `HashMap` and they build an `IndexMap` only at the return phase.

IvanIsCoding:
> Great work!
> There are still a few places, besides custom return types, where we return a `HashMap` and we could switch to `IndexMap`. I have spotted `graph::adj`, `digraph::adj`, `digraph::adj_direction`, `graph::compose`, `digraph::compose`, `core_number` and `max_weight_matching`.
> 
> My guess is that we will not see a significant performance regression since most of the algorithms are still using a `HashMap` and they build an `IndexMap` only at the return phase.

Indeed we can replace those, I started replacing the one for custom types but because of #347 I think we can also extend the use to where we have return types like HashMap or PyDict. Performance will not be affected and we'd have a more Pythonic result

src/dictmap.rs
IvanIsCoding: I like this idea and I implemented it. One remarkable thing to say is that the trait must be in scope, otherwise we get:
```
    = help: items from traits can only be used if the trait is in scope
    = note: the following trait is implemented but not in scope; perhaps add a `use` for it:
            `use crate::dictmap::Init;`
```
To overcome that, I replaced the imports with `use crate::dictmap::*` to guarantee the trait would be in scope

IvanIsCoding:
I have incorporated the feedback, now more functions are deterministic and the diff is just replacing `HashMap` with `DictMap`. #434 has been merged I would like to get reviews and merge this PR

georgios-ts (commented):
Thanks for the updates.
I have left some inline comments that suggest reverting back to `HashMap` for some variables that do not affect the output result. The same goes for the internal changes in `vf2.rs` and `max_weight_matching.rs`, i.e we don't really need there the determinism that `IndexMap` offers and even if it's slightly slower we should avoid using it. 
 In the previous comment, i meant to replace the `HashSet` used in `max_weight_matching.rs`  https://github.com/Qiskit/retworkx/blob/main/src/matching/max_weight_matching.rs#L867 with an `IndexSet`.

Also, we probably need an `IndexMap` for the `path` variable in `dijkstra::dijkstra` (https://github.com/Qiskit/retworkx/blob/main/src/shortest_path/dijkstra.rs#L105) since we iterate over this in some places (e.g [here](https://github.com/Qiskit/retworkx/blob/main/src/shortest_path/mod.rs#L87) ) and might result in non-deterministic behavior.

IvanIsCoding:
> Thanks for the updates.
> I have left some inline comments that suggest reverting back to `HashMap` for some variables that do not affect the output result. The same goes for the internal changes in `vf2.rs` and `max_weight_matching.rs`, i.e we don't really need there the determinism that `IndexMap` offers and even if it's slightly slower we should avoid using it.
> In the previous comment, i meant to replace the `HashSet` used in `max_weight_matching.rs` https://github.com/Qiskit/retworkx/blob/main/src/matching/max_weight_matching.rs#L867 with an `IndexSet`.
> 
> Also, we probably need an `IndexMap` for the `path` variable in `dijkstra::dijkstra` (https://github.com/Qiskit/retworkx/blob/main/src/shortest_path/dijkstra.rs#L105) since we iterate over this in some places (e.g [here](https://github.com/Qiskit/retworkx/blob/main/src/shortest_path/mod.rs#L87) ) and might result in non-deterministic behavior.

I removed DictMap where it wasn't necessary. For `max_weight_matching.rs`, it turns out there is no change for including DictMap. Replacing HashSet with IndexSet would have no effects on the Python side, because sets do not guarantee order (one of the reasons why there is no conversion in https://github.com/PyO3/pyo3/pull/1728). Hence I left them out.

Also, `dijkstra::dijkstra` now uses DictMap for the [optional path variable](https://github.com/Qiskit/retworkx/blob/main/src/shortest_path/dijkstra.rs#L105).

georgios-ts (commented):
Great! Could you also revert the changes in `vf2.rs` and `max_weight_matching.rs`? I don't see a reason using an `IndexMap` there.

IvanIsCoding:
> Great! Could you also revert the changes in `vf2.rs` and `max_weight_matching.rs`? I don't see a reason using an `IndexMap` there.

Removed DictMap from`max_weight_matching.rs` and some internal `vf2.rs` functions. `vf2.rs` still needs to use one DictMap because of the custom return types so that's why there is one left

georgios-ts (approved):
LGTM, thanks for the updates

PR #388, opened 2021-07-19
Fix incorrect result with dijkstra path functions

### LINKED ISSUE #387: PyGraph dijkstra path functions return incorrect path in some cases
### Information

- **retworkx version**: 0.9 and `main` 
- **Python version**: 3.9
- **Rust version**: 1.53.0
- **Operating system**: Linux

Running this code:

```python
        graph = retworkx.PyGraph()
        graph.add_node(None)
        graph.extend_from_weighted_edge_list([
            (1, 2, 10),
            (2, 3, 10),
            (3, 4, 10),
            (4, 5, 10),
            (5, 6, 10),
            (2, 7, 1),
            (7, 5, 1)
        ])
        graph.remove_node(0)
        path = retworkx.graph_dijkstra_shortest_paths(
            self.graph, 5, weight_fn=float, target=3
        )
        print(path)
        expected = {3: [5, 7, 2, 3]}
        assert expected == path
        path = retworkx.graph_dijkstra_shortest_paths(
            self.graph, 3, weight_fn=float, target=5
        )
        print(path)
        expected = {5: [3, 2, 7, 5]}
        assert expected == path
```

fails. The returned paths look like they're not using the weight correctly as it's outputting a path of `[5, 4, 3]` which has a weighted distance of 20 vs the expected path `[5, 7, 2, 3]` which has a weighted distance of 12.

The interesting and unexpected thing is that the distance function is returning the expected correct result and the underlying rust code is the same function.

The expected path should take into account the weight and be returned correctly.

mtreinish:
This commit fixes an issue with the dijkstra path functions where they
were returning the incorrect result in the case of multiple paths
between nodes in the graph. Previously, in cases of parallel paths the
inner rust dijkstra function was not updating the path if it had a
shorter weight than the already found path. This could result in an
incorrect shortest path being returned, depsite the shortest length
being correct. This fixes this by making sure we always update the path
when we set the distance. Additionally, after fixing this several of the
dijkstra path tests began to fail because they were not returning the
correct result.

Fixes #387

### DIFF
--- a/releasenotes/notes/dijkstra-path-bugfix-2c3f8f16d40c47b9.yaml
+++ b/releasenotes/notes/dijkstra-path-bugfix-2c3f8f16d40c47b9.yaml
@@ -1,0 +1,16 @@
+---
+fixes:
+  - |
+    Fixed an issue with the Dijkstra path functions:
+
+      * :func:`retworkx.dijkstra_shortest_paths`
+      * :func:`retworkx.all_pairs_dijkstra_shortest_paths`
+      * :func:`retworkx.digraph_dijkstra_shortest_paths`
+      * :func:`retworkx.graph_dijkstra_shortest_paths`
+      * :func:`retworkx.digraph_all_pairs_dijkstra_shortest_paths`
+      * :func:`retworkx.graph_all_pairs_dijkstra_shortest_paths`
+
+    where the returned paths could be incorrect in cases where there were
+    multiple paths between nodes and an edge weight callback function was
+    specified.
+    Fixed `#387 <https://github.com/Qiskit/retworkx/issues/387>`__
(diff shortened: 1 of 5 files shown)


IvanIsCoding (approved):

IvanIsCoding:
At least we caught this bug! 

This fix will propagate to [CoupingMap](https://github.com/Qiskit/qiskit-terra/blob/10468e643472c7bae89e21d2c1c6846b476ae2ad/qiskit/transpiler/coupling.py#L198), which propagates to [BasicSwap](https://github.com/Qiskit/qiskit-terra/blob/10468e643472c7bae89e21d2c1c6846b476ae2ad/qiskit/transpiler/passes/routing/basic_swap.py#L81). Fortunately BasicSwap is not the default method routing method in the transpiler, so by default most users were not affected by this bug on Terra.

Issue #389, opened 2021-07-20
Add minimum steiner tree function


mtreinish:
### What is the expected enhancement?

Add a minimum steiner tree function equivalent to networkx's `steiner_tree()` function: https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.approximation.steinertree.steiner_tree.html#networkx.algorithms.approximation.steinertree.steiner_tree

The source can be found here: https://github.com/networkx/networkx/blob/80704838e60327f3e3a292f4fb4812039be843df/networkx/algorithms/approximation/steinertree.py

This will likely need to be done over 3 PRs, one adding `metric_closure()`, one for `edge_subgraph`, and the last combines the first 2 to make the `steiner_tree` function.

georgios-ts:
After reading the original [paper](http://aturing.umcs.maine.edu/~markov/SteinerTrees.pdf), it seems that some steps in the original algorithm are missing from NetworkX function (steps 4 and 5 to be specific). I tried to find an example that breaks the function but couldn't find any. Depending on how you choose the minimum spanning tree in step 3, steps 4 and 5 might be unnecessary.

Although you have almost done all the work in #392, i mention here for future reference that there is a similar algorithm with the same approximation ratio but better runtime complexity (https://people.mpi-inf.mpg.de/~mehlhorn/ftp/SteinerTrees.pdf).

Issue #300, opened 2021-04-03
Start to Organize Algorithms into Modules


mtreinish:
### What is the expected enhancement?

As retworkx continues to grow the number of algorithm functions we offer has grown substantially. Right now we have been putting standalone functions just in `lib.rs` except where the implementation was either forked from petgraph and modified (and done in a separate module to provide attribution) or when the implementation was large enough that it justified it's own file. But this has led the lib.rs file getting quite large and a bit disorderly.  To make it easier to find the code for the different functions we probably should start looking at splitting lib.rs into separate modules.

This is to start a discussion around what structure we should pick (ie function by type of algorithm, by graph type, by type of algorithm and graph type, etc) and start to organize things a bit into rust submodules (and also update the docs accordingly). Ideally I think that we still want to have the root `retworkx` python namespace still have all the functions present so lib.rs would probably just just contain the [pymodule](https://github.com/Qiskit/retworkx/blob/34a2cb01894ba1ab8438ca585ed0df3f280cce90/src/lib.rs#L3087-L3155) and be where we export the various pieces to python.

PR #390, opened 2021-07-20
Add metric closure function

mtreinish:
This commit adds a metric closure function to retworkx. The metric
closure of a graph is the complete graph in which each edge is weighted
by the shortest path distance between the nodes in the graph. This is
the first step towards implementing the minimum steiner graph function
because the first step for that is to compute the metric closure for the
graph.

Additionally, this function is entirely self contained to a new rust module
steiner_tree.rs (which we'll eventually add the steiner_tree() function
too) in the interest of showing the eventual organizational structure we
want for #300 (although having an algorithms namespace might make sense
for #300). In a separate PR addressing #300 we should start to
reorganize the rust code in lib.rs like this to make things easier to
find.

Partially implements #389


docs/source/api.rst
```diff
@@ -115,6 +115,16 @@ Connectivity and Cycles
...
+.. _approximations:
+
+Approximations and Heuristics
+-----------------------------
+
+.. autosummary::
+   :toctree: stubs
+
+   retworkx.metric_closure
+
```
georgios-ts, line 127: Although i like this new doc section, i'm not sure if this is the best place for `metric_closure`. i would suggest the other-algorithms section.

src/steiner_tree.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+    Ok(out_graph)
+}
+
+fn _metric_closure_edges(
+    py: Python,
+    graph: &graph::PyGraph,
+    weight_fn: PyObject,
+) -> PyResult<Vec<MetricClosureEdge>> {
+    let mut out_vec = Vec::with_capacity(graph.graph.edge_count());
```
georgios-ts, line 69: since the metric closure is a complete graph, the capacity of the output vec should be `n * (n - 1) / 2` where n is the number of nodes, right?

src/steiner_tree.rs
```diff
@@ -0,0 +1,106 @@
...
+        .copied()
+        .collect();
+    let mut nodes: HashSet<usize> =
+        graph.graph.node_indices().map(|x| x.index()).collect();
+    if nodes.difference(&path_keys).count() > 0 {
+        return Err(PyValueError::new_err(
+            "The input graph must be a connected graph. The metric closure is \
+            not defined for a graph with unconnected nodes",
+        ));
+    }
```
georgios-ts, line 90: This logic only captures disconnected graphs where it exists a node with no neighbors and the function panics for other cases. I think we should check instead if `paths[0].paths.keys()` contains all node indices. (of course `0` should be replaced with a node index that we are certain that it exists in the graph).

docs/source/api.rst
mtreinish: I just copied this from where networkx puts it, but I think that's only because the steiner tree function is there with it, I'll move it to the other algorithms section

src/steiner_tree.rs
mtreinish: Yeah, good catch. I'll update this now

docs/source/api.rst
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/390/commits/1641c8e4c3fc4207c308e9126a0c501e0f37eef0

src/steiner_tree.rs
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/390/commits/7b7a20d742cb3a837b6ce0d260cf22cbca9a1890

src/steiner_tree.rs
```diff
@@ -0,0 +1,106 @@
...
+        &graph.graph,
+        weight_fn,
+        Some(&mut distances),
+    )?
+    .paths;
+    let path_keys: HashSet<usize> = paths
+        .keys()
+        .filter(|x| !paths[x].paths.is_empty())
+        .copied()
+        .collect();
```
IvanIsCoding, line 82: Instead of iterating over the hashmap keys and accessing, I think we should do `.iter()` with `.filter_map()`. That way we avoid the lookup cost at the hashmap, we already have `paths[x].path` by iterating

src/steiner_tree.rs
mtreinish: Yep, good call. Fixed in: https://github.com/Qiskit/retworkx/pull/390/commits/de6d58c02166038345a221969a92ae3a2dc7408f

src/steiner_tree.rs
mtreinish: Yeah, that's a good idea. I tested something like:

```rust
let path_keys: HashSet<usize> = paths
    .iter()
    .filter_map(|(k, v)| if !v.paths.is_empty() { Some(k) } else { None })
    .copied()
    .collect();
```        
locally and it worked well. But I reworked the logic based on @georgios-ts's suggestion in https://github.com/Qiskit/retworkx/pull/390/commits/de6d58c02166038345a221969a92ae3a2dc7408f and didn't use this anymore.

src/steiner_tree.rs
```diff
@@ -0,0 +1,111 @@
...
+    Ok(out_graph)
+}
+
+fn _metric_closure_edges(
+    py: Python,
+    graph: &graph::PyGraph,
+    weight_fn: PyObject,
+) -> PyResult<Vec<MetricClosureEdge>> {
+    let node_count = graph.graph.node_count();
+    let mut out_vec = Vec::with_capacity(node_count * (node_count - 1) / 2);
```
IvanIsCoding, line 69: You cannot have `node_count - 1`, because it leads to a negative usize in the edge case. You can try `(node_count * node_count - node_count) / 2`, that never has an intermediate term that is zero

mtreinish: Oh, fun that is what's causing the CI failure on coverage runs because the overflow detection is only in debug mode, release mode removes the checks. We normally run tests in release mode because that's the default for setuptools-rust which means we're probably allocating a huge array in the `node_count == 0` case.

That being said I think we can fix this by just doing something like:

```suggestion
    if node_count == 0 {
        return Ok(Vec::new());
    }
    let mut out_vec = Vec::with_capacity(node_count * (node_count - 1) / 2);
```

to short circuit on an empty graph. That's what I did internally further down for: https://github.com/Qiskit/retworkx/pull/390/commits/de6d58c02166038345a221969a92ae3a2dc7408f but I should do that early and then we don't have to worry about it.

mtreinish: I did this in: https://github.com/Qiskit/retworkx/pull/390/commits/cade320de1454e64ab30a8e1fe074148cbb3ea97 which should avoid the overflow failure.

src/lib.rs
```diff
@@ -65,6 +67,7 @@ use crate::iterators::{
     NodesCountMapping, PathLengthMapping, PathMapping, Pos2DMapping,
     WeightedEdgeList,
 };
+use crate::steiner_tree::*;
```
IvanIsCoding, line 70:
```suggestion
use steiner_tree::*;
```

Module is declared on the same file, we don't need the `crate::` part

src/steiner_tree.rs
```diff
@@ -0,0 +1,115 @@
...
+        for v in &nodes {
+            let v_index = NodeIndex::new(*v);
+            out_vec.push(MetricClosureEdge {
+                source: node,
+                target: *v,
+                distance: distance[&v_index],
+                path: path_map[v].clone(),
+            });
+        }
+    }
```
IvanIsCoding, line 113: Maybe we can loop over a hashmap with key + value instead,
```rust
for (node, path) in &*paths {
    let path_map = path.paths;
}
```

mtreinish: Done in https://github.com/Qiskit/retworkx/pull/390/commits/3353b4b4cc3a7f859812145e25eb6290dc446ec4

IvanIsCoding (approved):

PR #391, opened 2021-07-20
Add edge_subgraph graph method

mtreinish:
This commit adds a new method, edge_subgraph(), to the PyGraph and
PyDiGraph classes for generating a edge induced subgraph from a graph
object.

Related to #389


src/digraph.rs
```diff
@@ -2426,6 +2426,65 @@ impl PyDiGraph {
...
+            edge_set.insert([source_index, target_index]);
+        }
+        let mut out_graph = self.clone();
+        for node in self
+            .graph
+            .node_indices()
+            .filter(|node| !nodes.contains(node))
+        {
+            out_graph.graph.remove_node(node);
+            out_graph.node_removed = true;
```
IvanIsCoding, line 2476: Calling `out_graph.node_removed = true` feels like breaking a unwritten rule, wouldn't it be better to just call `out_graph.remove_node()` on the PyDiGraph class?

mtreinish: I just did it this way because to avoid the double conversion of `NodeIndex` -> `usize` -> `NodeIndex`. But I can change this to be:
```suggestion
            out_graph.remove_node(node.index())?;
```
if you'd prefer. I don't think there will be any extra overhead from the double conversion (since a NodeIndex is basically just a thin struct around `usize` anyway (to provide a common trait and debug output for the various unsigned integer types)

IvanIsCoding: Double conversion is not ideal, and that would make it a PyResult. I think it is fine to manipulate `.node_removed` then

mtreinish: I'm actually not sure why `remove_node` returns a `PyResult`. Looking at the code for the method nothing returns an error. Petgraph's `remove_node` returns an option and `None` if the `NodeIndex` is invalid. But `PyGraph`'s `remove_node` is documented as just ignoring invalid node ids so we probably can remove the pyresult from the method.

georgios-ts (commented):
LGTM. Once the merge conflicts are resolved, it should be ready to go.

IvanIsCoding (approved):

PR #393, opened 2021-07-22
Add weighted longest path functions

mtreinish:
As a follow up to #383 what the underlying Qiskit user case for #381
would like is to use floats for the edge weighting. However for the
existing `dag_longest_path_length()` function our hands are tied with the
return type being an unsigned integer for backwards compatibility. To
workaround that limitation this commit adds a new function
`dag_weighted_longest_path_length()` which requires a weight function and
works with `f64` weights. To accomplish this, the inner `longest_path()` rust
function is made generic for any numeric type returned by the callback
function, which required leveraging the num-traits crate. The new
pyfunction is then just a thin wrapper that works with `f64` instead of
`usize`.

### DIFF
--- a/releasenotes/notes/add-dag_weighted_longest_path_length-5c4e9a7d19c97d4d.yaml
+++ b/releasenotes/notes/add-dag_weighted_longest_path_length-5c4e9a7d19c97d4d.yaml
@@ -1,0 +1,15 @@
+---
+features:
+  - |
+    Added a new algorithm function,
+    :func:`~retworkx.dag_weighted_longest_path_length`, to find the length
+    of the longest path in a:class:`~retworkx.PyDiGraph` object without any
+    cycles. This new function is basically equivalent to
+    :func:`~retworkx.dag_longest_path_length` except for two key differences,
+    the ``weight_fn`` parameter is required for
+    :func:`~retworkx.dag_weighted_longest_path_length` while it is optional
+    in :func:`~retworkx.dag_longest_path_length` and
+    :func:`~retworkx.dag_weighted_longest_path_length` works with ``float``
+    weights (the function returns a float and the ``weight_fn`` callback is
+    expected to return a ``float``) while
+    :func:`~retworkx.dag_longest_path_length` works with an unsigned ``int``.
--- a/releasenotes/notes/add-dag_weighted_longest_path_length-5c4e9a7d19c97d4d.yaml
+++ b/releasenotes/notes/add-dag_weighted_longest_path_length-5c4e9a7d19c97d4d.yaml
@@ -1,7 +1,7 @@
 ---
 features:
   - |
-    Added two new algorithm function,
+    Added two new algorithm functions,
     :func:`~retworkx.dag_weighted_longest_path_length` and
     :func:`~retworkx.dag_weighted_longest_path`, to find the longest path
     and the length of the longest path in a:class:`~retworkx.PyDiGraph` object
--- a/releasenotes/notes/add-dag_weighted_longest_path_length-5c4e9a7d19c97d4d.yaml
+++ b/releasenotes/notes/add-dag_weighted_longest_path_length-5c4e9a7d19c97d4d.yaml
@@ -1,15 +1,20 @@
 ---
 features:
   - |
-    Added a new algorithm function,
-    :func:`~retworkx.dag_weighted_longest_path_length`, to find the length
-    of the longest path in a:class:`~retworkx.PyDiGraph` object without any
-    cycles. This new function is basically equivalent to
+    Added two new algorithm function,
+    :func:`~retworkx.dag_weighted_longest_path_length` and
+    :func:`~retworkx.dag_weighted_longest_path`, to find the longest path
+    and the length of the longest path in a:class:`~retworkx.PyDiGraph` object
+    without any cycles. This new function is basically equivalent to
+    :func:`~retworkx.dag_longest_path` and
     :func:`~retworkx.dag_longest_path_length` except for two key differences,
     the ``weight_fn`` parameter is required for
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -195,7 +207,56 @@ fn dag_longest_path_length(
...
+#[pyo3(text_signature = "(graph, weight_fn, /)")]
+fn dag_weighted_longest_path_length(
+    py: Python,
+    graph: &digraph::PyDiGraph,
+    weight_fn: PyObject,
+) -> PyResult<f64> {
+    let edge_weight_callable =
+        |source: usize, target: usize, weight: &PyObject| -> PyResult<f64> {
+            let res = weight_fn.call1(py, (source, target, weight))?;
+            res.extract(py)
```
mtreinish, line 257: I probably should add a check here for `NaN` and raise because I don't think the `max_by()` call in `longest_path` will work with it (and I assume panic).

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/393/commits/45f3fb1261df51c1123454790cb8438bc70eaca7

georgios-ts (commented):
This looks great! I'm just wondering if we should add a `dag_weighted_longest_path` function or return the path and the length from one function.

mtreinish:
Yeah, that's a good idea, there wasn't an immediate use case for it but it's only a few extra lines so better to have it, I'll add it now.

releasenotes/notes/add-dag_weighted_longest_path_length-5c4e9a7d19c97d4d.yaml
```diff
@@ -1,15 +1,20 @@
 ---
 features:
   - |
-    Added a new algorithm function,
-    :func:`~retworkx.dag_weighted_longest_path_length`, to find the length
-    of the longest path in a:class:`~retworkx.PyDiGraph` object without any
-    cycles. This new function is basically equivalent to
+    Added two new algorithm function,
```
georgios-ts, line 4:
```suggestion
    Added two new algorithm functions,
```

src/lib.rs
```diff
@@ -214,6 +214,51 @@ fn dag_longest_path_length(
...
+/// This function differs from :func:`retworkx.dag_longest_path` in that
+/// this function requires a ``weight_fn`` parameter, and the ``weight_fn`` is
+/// expected to return a ``float`` not an ``int``.
+///
+/// :param PyDiGraph graph: The graph to find the longest path on. The input
+///     object must be a DAG without a cycle.
+/// :param weight_fn: A python callable that will be passed the 3
+///     positional arguments, the source node, the target node, and the edge
+///     weight for each edge as the function traverses the graph. It is expected
+///     to return an float weight for that edge. For example,
```
georgios-ts, line 228:
```suggestion
///     to return a float weight for that edge. For example,
```

src/lib.rs
```diff
@@ -214,6 +214,51 @@ fn dag_longest_path_length(
...
+///     ``dag_longest_path(graph, lambda: _, __, weight: weight)`` could be
+///     use to just use an float edge weight. It's also worth noting that this
```
georgios-ts, line 230:
```suggestion
///     used to just use a float edge weight. It's also worth noting that this
```

src/lib.rs
```diff
@@ -214,6 +214,51 @@ fn dag_longest_path_length(
...
+///     function traverses in topological order and only checks incoming edges to
+///     each node.
+///
+/// :returns: The node indices of the longest path on the DAG
+/// :rtype: NodeIndices
+///
+/// :raises Exception: If an unexpected error occurs or a path can't be found
+/// :raises DAGHasCycle: If the input PyDiGraph has a cyc
```
georgios-ts, line 238:
```suggestion
/// :raises DAGHasCycle: If the input PyDiGraph has a cycle
```

src/lib.rs
```diff
@@ -195,7 +200,106 @@ fn dag_longest_path_length(
...
+/// This function differs from :func:`retworkx.dag_longest_path_length` in that
+/// this function requires a ``weight_fn`` parameter, and the ``weight_fn`` is
+/// expected to return a ``float`` not an ``int``.
+///
+/// :param PyDiGraph graph: The graph to find the longest path on. The input
+///     object must be a DAG without a cycle.
+/// :param weight_fn: A python callable that will be passed the 3
+///     positional arguments, the source node, the target node, and the edge
+///     weight for each edge as the function traverses the graph. It is expected
+///     to return an float weight for that edge. For example,
```
georgios-ts, line 273:
```suggestion
///     to return a float weight for that edge. For example,
```

src/lib.rs
```diff
@@ -195,7 +200,106 @@ fn dag_longest_path_length(
...
+///     ``dag_longest_path(graph, lambda: _, __, weight: weight)`` could be
+///     use to just use an float edge weight. It's also worth noting that this
```
georgios-ts, line 275:
```suggestion
///     used to just use a float edge weight. It's also worth noting that this
```

src/lib.rs
```diff
@@ -195,7 +200,106 @@ fn dag_longest_path_length(
...
+///     function traverses in topological order and only checks incoming edges to
+///     each node.
+///
+/// :returns: The longest path length on the DAG
+/// :rtype: float
+///
+/// :raises Exception: If an unexpected error occurs or a path can't be found
+/// :raises DAGHasCycle: If the input PyDiGraph has a cyc
```
georgios-ts, line 283:
```suggestion
/// :raises DAGHasCycle: If the input PyDiGraph has a cycle
```

georgios-ts (approved):

PR #397, opened 2021-07-26
Hexagonal graph with consecutive node indices

### LINKED ISSUE #373: retworkx.generators.hexagonal_lattice_graph does not generate consecutive nodes
### Information

- **retworkx version**: main
- **Python version**: 3.8

While generating `hexagonal_lattice_graph` I noticed that the generated node indexes are not consecutive. This seems unexpected to me.

```python
import retworkx
G = retworkx.generators.hexagonal_lattice_graph(10, 10)
offset = 0
for i, j in enumerate(G.node_indexes()):
    if i+offset != j :
        offset += 1
        print(i, 'is missing')
```
```
21 is missing
219 is missing
```

Consecutive node indexes.

georgios-ts:
Closes #373.

This commit modifies `hexagonal_lattice_graph` and `directed_hexagonal_lattice_graph` to create graphs with consecutive node indices. This is possible with minor changes in the logic of the original source code, roughly we should shift the node indexes by one (or two in the last column) except the first column that we don't need a shift.

This is important if we ever use these functions to generate a `CouplingMap` in terra where the underlying graph assumes consecutive node indices.


mtreinish (approved):
Awesome, thanks for fixing this! Everything LGTM and I tested it locally for a bunch of inputs and it all looked good.

PR #398, opened 2021-07-27
Build non-x86 architecture wheels in QEMU emulation

mtreinish:
Since the introduction of tight travis quotas for open source projects
we have not been running travis CI on retworkx for ~2 months. Since we
can't rely on travis to reliably run our non-x86 testing and release
builds we need to come up with an alternative solution. However, travis
is the only avaiable CI provider with non-x86 nodes available for free
to open source projects (which is why we were using them, this quota
issue has been a long time coming). To workaround this commit tries
to leverage QEMU emulation in github actions for the release wheel
builds. In practice this is at least an order of magnitude slower than
running on a VM but the github actions job time limit is 6 hrs which
hopefully should be enough time.


mtreinish:
Ok, the output from the last build everything worked on it: https://github.com/Qiskit/retworkx/actions/runs/1079409435. The s390x build was >5 hrs so I'm a bit concerned we'll exceed the 6hr time limit at release time (the previous run hit the timeout). But we can always just manually retrigger it if we do have a timeout. This at least runs which is better than what travis was providing us.

PR #399, opened 2021-07-27
Use IndexMap for graph_greedy_color output

### LINKED ISSUE #347: The result of `graph_greedy_color` is not deterministic

mtreinish:
Right now functions that use or wrap a HashMap for their return type to
Python have a non-deterministic iteration order. This is because HashMap
is by definition an unordered type. However, the expectation for Python
users is that since Python 3.6 dictionaries are ordered as they preserve
insertion order. To fix this user expectation for graph_greedy_color()
this commit leverages IndexMap from the indexmap crate [1] which offers
similar performance but offers consistent order (and slightly faster
iteration, although for our use case it might not matter). This will
provide a consistent insertion order for the output dict from
graph_greedy_color().

Wider use of IndexMap for other places we return or wrap HashMaps to Python
is being done in #386 to get this advantage more broadly, but it depends on
a new PyO3 release to work. So in the meantime this just fixes the issue
for graph_greedy_color() where this a reported issue and need for the
deterministic iteration order (it's also doesn't require the PyO3 side
changeseasy as it returns a PyDict instead of relying on the pyfunction
macro to build to convert for us).

Fixes #347

Co-Authored-By: Ivan Carvalho <ivan.carvalho@alumni.ubc.ca>

[1] https://crates.io/crates/indexmap


IvanIsCoding (approved):
I agree. This is a quicker path to fix #347 for 0.10, #386 might take a while to merge and we can just pick the tiny part we want

Issue #352, opened 2021-06-06
Consider adding Python annotation support


IvanIsCoding:
### What is the expected enhancement?

Users are able to execute [mypy](https://github.com/python/mypy) to run a static type check in code that uses retworkx.

Python annotations are becoming more popular, and we should consider supporting them. Supporting annotations can help make retworkx more popular in codebases that have adopted mypy. Also, many popular libraries that do not support annotations yet have tickets claiming for them. 

#### Suggested Implementation

Create [`.pyi`](https://mypy.readthedocs.io/en/stable/stubs.html) stubs containing the type annotations for the existing code. See [typeshed](https://github.com/python/typeshed) for some inspiration on how this is done. Another useful inspiration is [orjson](https://github.com/ijl/orjson), they are another Python library that uses Rust/PyO3 and they also use the `.pyi` stub approach.

#### Challenges

* Duplication: if we use `.pyi` stubs, we will have to duplicate all the functions signatures in the stubs files; ideally PyO3 would generate all the stubs but https://github.com/PyO3/pyo3/issues/510 is far from being closed
* Packaging: making sure the stubs are installed and third-party code recognizes it when running mypy
* Testing: how do we test our stubs are correct and up to date?

#### Non-Goals

Use [mypy](https://github.com/python/mypy) internally within retworkx. The emphasis is: we want users to be able to run [mypy](https://github.com/python/mypy) when checking code that uses retworkx.

IvanIsCoding:
I had an idea that can help us avoid duplication and make type annotations manageable. Here is what I thought:

## Idea

The idea is to make `.pyi` stub files the source of `text_signature` in our Rust . The stub files look like:

```python
class PyGraph:
    
    def has_edge(self, node_a: int, node_b: int, /) -> bool: ...
```
In Rust, we would load the text signature using a `include_annotation!` macro to reuse the definition originally in `.pyi`:

```rust
#[pyo3(text_signature = include_annotation!("PyGraph.has_edge"))]
pub fn has_edge(&self, node_a: usize, node_b: usize) -> bool {
```

That way, we do not need two worry about conflicting text signatures and annotations. All annotations would "magically" come from `.pyi` files, once we update the stub file they propagate to the Rust code.

## Implementation

A Possible implementation for the idea is:

#### 1. Create a Python script that parses the `.pyi` stub and generates text signatures

[...]

PR #401, opened 2021-08-03
Add annotation stubs for `PyGraph` and `PyDiGraph`

IvanIsCoding:
Related to #352 

This PR adds type annotation stubs, it contains `.pyi` files with type annotations. I started by annotating the `PyGraph` and `PyDiGraph` files, as well as the custom types.

I tried to follow the [PEP 561](https://www.python.org/dev/peps/pep-0561/) convetion: we include `.pyi` files in the same directory as our Python code. Note that this is a first step in annotations: they are partial, we can add more as we go

We also test the annotations using [mypy.stubtest](https://mypy.readthedocs.io/en/stable/stubtest.html). This was the most reliable approach. For now, we ignore missing stubs and focus only on the functions imported from Rust code

Some items we need to handle because the PR has been open for long:
- [x] Add custom return types that were added after the PR was created


jakelishman:
I've no power on `retworkx` - can someone please assign me to this so I don't forget to check it?

tox.ini
```diff
@@ -71,6 +71,16 @@ deps =
...
 
+[testenv:stubs]
+basepython = python3
+envdir = .tox/stubs
+deps =
+  pytest
+  mypy==0.910
+  pytest-mypy-testing==0.0.8
+  dataclasses; python_version<'3.7'
+commands = python -m pytest '../stubs-tests'
```
jakelishman, line 82: It seems a bit odd that the tests aren't in the directory `tests`, but need to have their own completely separate directory.  Perhaps it would be nicer to restructure such that you have a structure
```text
src/
  | ...
tests/
  | unit/
     | digraph/
     | generators/
     | ...
  | stub/
     | pydigraph_test.py
     | pygraph_test.py
```
?

jakelishman (commented):
I'm generally in favour of libraries providing type hints if at all possible - it's hopefully not too much work, and lets downstream users make the choice over whether or not to use them.

I think there's a couple of problems with the typing here, though, some of which seem to stem from design decisions in `PyDiGraph` and `PyGraph`.  The edge data is inherently nullable, but the type hints here don't account for that.  This means that the types are inconsistent, and _all_ graph type hints should effectively be `PyGraph[S, Union[T, None]]`.  You don't want users to do that, so any `PyGraph` function that returns type `T` should instead return `Union[T, None]`.  Obviously this means that it'll be harder for users to write compliant code, but that's actually a function of the internal typing of `PyGraph`, and technically it's required whether or not the type hint says that.  (See also: why everyone hates nullable types!)

retworkx/custom_return_types.pyi
```diff
@@ -0,0 +1,237 @@
...
+    def __lt__(self, other) -> bool: ...
+    def __ne__(self, other) -> bool: ...
+    def __setstate__(self, state) -> None: ...
+
+class PathLengthMapping:
+    def __init__(self) -> None: ...
+    def items(self) -> Iterable[Tuple[int, float]]: ...
+    def keys(self) -> Iterable[int]: ...
+    def values(self) -> Iterable[float]: ...
+    def __contains__(self, other) -> int: ...
```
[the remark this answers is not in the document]
jakelishman: Also, I'd have thought that you get a lot of this just by inheriting from `Sequence[T]`?  Not 100% sure, though.

retworkx/custom_return_types.pyi
```diff
@@ -0,0 +1,237 @@
...
+
+# This file contains only type annotations for PyO3 functions and classes
+# For implementation details, see __init__.py and lib.rs
+
+from typing import Any, Generic, List, Iterable, TypeVar, Tuple
+from collections.abc import Sequence
+
+T = TypeVar("T")
+
+class NodeIndices(Sequence[int]):
```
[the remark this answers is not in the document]
jakelishman: Also: lol at the method being called `node_indexes` and the class being `NodeIndices`

retworkx/retworkx.pyi
```diff
@@ -0,0 +1,22 @@
...
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# This file contains only type annotations for PyO3 functions and classes
+# For implementation details, see __init__.py and lib.rs
+
+from .custom_return_types import *
+from .pygraph import PyGraph as PyGraph
+from .pydigraph import PyDiGraph as PyDiGraph
```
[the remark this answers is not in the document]
IvanIsCoding: We need to explicitly import `PyGraph` and `PyDiGraph` to ensure when someone imports retworkx as `from retworkx import *` they are included

tox.ini
IvanIsCoding: I agree it is weird, but I am trying to avoid tests using `unittest` with tests using `pytest`

tox.ini
```diff
@@ -71,6 +71,16 @@ deps =
...
 commands = black {posargs} '../retworkx' '../tests'
 
+[testenv:stubs]
+basepython = python3
+envdir = .tox/stubs
+deps =
+  pytest
+  mypy==0.910
+  pytest-mypy-testing==0.0.8
+  dataclasses; python_version<'3.7'
```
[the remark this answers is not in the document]
IvanIsCoding: This is more of a trick to make `pytest-mypy-testing` work for Python 3.6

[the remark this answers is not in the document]
jakelishman: I didn't see that on their github - all their project files indicate their library should work on Python 3.6 (`pyproject.toml`, their `tox` config, etc).  I'd raise it as a bug with them.

tox.ini
jakelishman: Having a structure like I suggested doesn't mean you have to mix tests for different runners - that's why I split the directories within the `tests` directory into two parts.  You can change the `tox`/`pytest` configurations to only search the correct subdirectories.

I don't feel strongly about this (I was only brought in to review the type hints) - it's a decision for you guys who actually maintain retworkx.

retworkx/retworkx.pyi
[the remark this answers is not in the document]
jakelishman: I guess I'm just confused why they're here and not in `__init__.pyi`, really.

retworkx/pydigraph.pyi
```diff
@@ -0,0 +1,176 @@
...
+        self,
+        /,
+        check_cycle: bool = ...,
+        multigraph: bool = ...,
+    ) -> None: ...
+    def add_child(self, parent: int, obj: S, edge: T, /) -> int: ...
+    def add_edge(self, parent: int, child: int, edge: T, /) -> int: ...
+    def add_edges_from(
+        self,
+        obj_list: List[Tuple[int, int, T]],
```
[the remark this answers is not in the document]
jakelishman: Also, I tried this locally with things like `add_nodes_from(range(10))` and everything went fine - that's also an example `retworkx` uses.

retworkx/custom_return_types.pyi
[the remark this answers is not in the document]
mtreinish: > Also: lol at the method being called `node_indexes` and the class being `NodeIndices`

https://github.com/Qiskit/retworkx/pull/541

IvanIsCoding:
> I'm generally in favour of libraries providing type hints if at all possible - it's hopefully not too much work, and lets downstream users make the choice over whether or not to use them.
> 
> I think there's a couple of problems with the typing here, though, some of which seem to stem from design decisions in `PyDiGraph` and `PyGraph`. The edge data is inherently nullable, but the type hints here don't account for that. This means that the types are inconsistent, and _all_ graph type hints should effectively be `PyGraph[S, Union[T, None]]`. You don't want users to do that, so any `PyGraph` function that returns type `T` should instead return `Union[T, None]`. Obviously this means that it'll be harder for users to write compliant code, but that's actually a function of the internal typing of `PyGraph`, and technically it's required whether or not the type hint says that. (See also: why everyone hates nullable types!)

I am back on this and I think the solution is [along the lines of restricting the methods you outlined](https://mypy.readthedocs.io/en/stable/more_types.html#restricted-methods-in-generic-classes).

So basically what we want is to never allows users to do this:
```python
graph: PyGraph[str, int] = PyGraph()
node_a: int = graph.add_node("A")
node_b: int = graph.add_node("B")
graph.add_edge_no_data(node_a, node_b) # mypy catches this error
```

But let users do this:
```python
graph: PyGraph[str, Optional[int]] = PyGraph()
node_a: int = graph.add_node("A")
node_b: int = graph.add_node("B")
graph.add_edge_no_data(node_a, node_b) # mypy should let this pass
```

That way we do not need to make users handle `None` unless they chose to

retworkx/pydigraph.pyi
```diff
@@ -0,0 +1,176 @@
...
+        node_map: Dict[int, Tuple[int, T]],
+        /,
+        node_map_func: Optional[Callable[[S], int]] = ...,
+        edge_map_func: Optional[Callable[[T], int]] = ...,
+    ) -> Dict[int, int]: ...
+    def copy(self) -> PyDiGraph[S, T]: ...
+    def edge_index_map(self) -> EdgeIndexMap: ...
+    def edge_indices(self) -> EdgeIndices: ...
+    def edge_list(self) -> EdgeList: ...
+    def edges(self) -> List[T]: ...
```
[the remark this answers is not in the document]
IvanIsCoding: We handle this now, thanks for catching it. I added a test to verify `graph.add_edges_from_no_data` is illegal unless we use `PyDiGraph[S, Optional[T]]`: https://github.com/Qiskit/retworkx/blob/63c87a6a9b7c94484992d1ad98f40dea080102ed/tests/stubs/restrict_operations_test.py

retworkx/pydigraph.pyi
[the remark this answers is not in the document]
IvanIsCoding: You are correct. PyO3 converts any `Sequence` from Python to `Vec` in Rust. But from Rust we can only convert `Vec` to a Python `List`, that is why I overlooked it earlier

retworkx/custom_return_types.pyi
```diff
@@ -0,0 +1,237 @@
...
+    def __init__(self) -> None: ...
+    def __eq__(self, other) -> bool: ...
+    def __ge__(self, other) -> bool: ...
```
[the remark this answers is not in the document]
IvanIsCoding: The tool I used thinks we have `__ge__` implemented because we have `__richcmp__` implemented, but in fact it just throws an error - so we just ignore it. For `__eq__` we use `Sequence[T]` now

retworkx/custom_return_types.pyi
```diff
@@ -0,0 +1,237 @@
...
+    def __getitem__(self, index: int) -> Tuple[T, List[T]]: ...
+    def __getstate__(self) -> Any: ...
+    def __gt__(self, other) -> bool: ...
+    def __hash__(self) -> int: ...
+    def __le__(self, other) -> bool: ...
+    def __len__(self) -> int: ...
+    def __lt__(self, other) -> bool: ...
+    def __ne__(self, other) -> bool: ...
+    def __setstate__(self, state) -> None: ...
```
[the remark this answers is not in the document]
IvanIsCoding: It is not mutable, but we are now using Covariant types

retworkx/custom_return_types.pyi
```diff
@@ -0,0 +1,237 @@
...
+    def __gt__(self, other) -> bool: ...
+    def __hash__(self) -> int: ...
+    def __iter__(self) -> Iterable[int]: ...
+    def __le__(self, other) -> bool: ...
+    def __len__(self) -> int: ...
+    def __lt__(self, other) -> bool: ...
+    def __ne__(self, other) -> bool: ...
+    def __setstate__(self, state) -> None: ...
+
+class WeightedEdgeList(Generic[T], Sequence[Tuple[int, int, T]]):
```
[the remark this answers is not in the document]
IvanIsCoding: Indeed they are covariant, I updated the code to reflect it

retworkx/custom_return_types.pyi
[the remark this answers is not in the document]
IvanIsCoding: It returns a bool now, thanks for catching it

IvanIsCoding:
I rewrote the custom return types annotations from zero. Because they are generated from the same Rust macro, I created an abstract class representing each macro and inherited from them. It shortens the annotations from that section considerably.

I also made the types covariant to handle some issues pointed earlier.

retworkx/custom_return_types.pyi
```diff
@@ -0,0 +1,62 @@
...
+# that they have been altered from the originals.
+
+# This file contains only type annotations for PyO3 functions and classes
+# For implementation details, see __init__.py and lib.rs
+
+from typing import Any, Generic, List, Iterable, Mapping, TypeVar, Tuple, overload
+from collections.abc import ABC, Sequence
+
+# Until PEP 673 is implemented in Python 3.11, we need to use this hack
+Self = TypeVar("Self", bound="RetworkxCustomVecIter")
```
jakelishman, line 16: I think you can get this from `typing_extensions` already if you'd rather, but that's no big deal.  Recent versions of `mypy` tend to support things as soon as they're specified in PEPs as well, I think?

tests/stubs/restrict_operations_test.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+@pytest.mark.mypy_testing
+def test_pygraph_add_edges_from_no_data_illegal() -> None:
+    graph: PyGraph[str, float] = PyGraph()
+    node_a: int = graph.add_node("A")
+    node_b: int = graph.add_node("B")
+
+    # fmt: off
+    graph.add_edges_from_no_data([(node_a, node_b)])  # E: Invalid self argument "PyGraph[str, float]" to attribute function "add_edges_from_no_data" with type "Callable[[PyGraph[S, Optional[T]], Sequence[Tuple[int, int]]], List[int]]"
```
jakelishman, line 26: This is really clever - you found a great solution to the typing problem!

jakelishman (commented):
I don't know enough about `retworkx` to comment much more than this, nor on the test failure, but the new form looks really good to me.  The solution you found about the restricted methods is great, and thanks for telling me about it too, since I didn't know it was possible in mypy.

The custom return type bits look right to me as well, assuming they line up with the actual Rust code.

I only had one question about the tests, but other than that, this is a logical approval from me.

tests/stubs/pydigraph_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+@pytest.mark.mypy_testing
+def test_pydigraph_simple() -> None:
+    graph: PyDiGraph[str, float] = PyDiGraph()
+    node_a: int = graph.add_node("A")
+    node_b: int = graph.add_node("B")
+    edge_ab: int = graph.add_edge(node_a, node_b, 3.5)
+    reveal_type(node_a)  # R: builtins.int
+    reveal_type(edge_ab)  # R: builtins.int
```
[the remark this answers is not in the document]
IvanIsCoding: They became a bit redundant because we test `reveal_type` anyway. I removed them

retworkx/custom_return_types.pyi
IvanIsCoding: Thanks for the heads up, I switched to using `typing_extensions`

retworkx/custom_return_types.pyi
```diff
@@ -0,0 +1,60 @@
...
+    def __init__(self) -> None: ...
+    def items(self) -> Iterable[Tuple[S, T_co]]: ...
+    def keys(self) -> Iterable[S]: ...
+    def values(self) -> Iterable[T_co]: ...
+    def __contains__(self, other: S) -> bool: ...
+    def __eq__(self, other: Mapping[S, T_co]) -> bool: ...
+    def __getitem__(self, index: S) -> T_co: ...
+    def __getstate__(self) -> Any: ...
+    def __hash__(self) -> int: ...
+    def __iter__(self) -> Iterable[S]: ...
```
jakelishman, line 42: I missed this first time - technically `__iter__` returns an `Iterator`, not an `Iterable` because it starts the iteration.  All the other `Iterable` hints are correct, though.

retworkx/custom_return_types.pyi
```diff
@@ -0,0 +1,60 @@
...
+
+class RetworkxCustomVecIter(Generic[T_co], Sequence[T_co], ABC):
+    def __init__(self) -> None: ...
+    def __eq__(self, other: Sequence[T_co]) -> bool: ...
+    @overload
+    def __getitem__(self, index: int) -> T_co: ...
+    @overload
+    def __getitem__(self: Self, index: slice) -> Self: ...
+    def __getstate__(self) -> Any: ...
+    def __hash__(self) -> int: ...
```
georgios-ts, line 27:
```suggestion
    def __hash__(self) -> int: ...
    def __str__(self) -> str: ...
```

retworkx/custom_return_types.pyi
```diff
@@ -0,0 +1,60 @@
...
+class RetworkxCustomHashMapIter(Generic[S, T_co], Mapping[S, T_co], ABC):
+    def __init__(self) -> None: ...
+    def items(self) -> Iterable[Tuple[S, T_co]]: ...
+    def keys(self) -> Iterable[S]: ...
+    def values(self) -> Iterable[T_co]: ...
+    def __contains__(self, other: S) -> bool: ...
+    def __eq__(self, other: Mapping[S, T_co]) -> bool: ...
+    def __getitem__(self, index: S) -> T_co: ...
+    def __getstate__(self) -> Any: ...
+    def __hash__(self) -> int: ...
```
georgios-ts, line 41:
```suggestion
    def __hash__(self) -> int: ...
    def __str__(self) -> str: ...
```

retworkx/custom_return_types.pyi
georgios-ts: Actually `__iter__` and `keys` return the same type in Rust so I'm thinking if we should use `Iterator` in all places.

retworkx/custom_return_types.pyi
```diff
@@ -0,0 +1,60 @@
...
+    def __eq__(self, other: Sequence[T_co]) -> bool: ...
+    @overload
+    def __getitem__(self, index: int) -> T_co: ...
+    @overload
+    def __getitem__(self: Self, index: slice) -> Self: ...
+    def __getstate__(self) -> Any: ...
+    def __hash__(self) -> int: ...
+    def __len__(self) -> int: ...
+    def __ne__(self, other: Sequence[T_co]) -> bool: ...
+    def __setstate__(self, state) -> None: ...
```
georgios-ts, line 30:
```suggestion
    def __setstate__(self, state: Sequence[T_co]) -> None: ...
```

retworkx/custom_return_types.pyi
jakelishman: If that's the case, it's possibly a bug - it's a break with the normal Python API.  If I use `collections.abc` to build up a mapping-like object from the minimal set of abstract methods, we get:
```python
In [1]: import collections.abc
   ...:
   ...: class A(collections.abc.Mapping):
   ...:     def __init__(self, d):
   ...:         self.d = d
   ...:     def __getitem__(self, k):
   ...:         return self.d[k]
   ...:     def __iter__(self):
   ...:         return iter(self.d)
   ...:     def __len__(self):
   ...:         return len(self.d)
   ...:
   ...: next(A({}).keys())
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-63ecf9188336> in <module>
     11         return len(self.d)
     12
---> 13 next(A({}).keys())

TypeError: 'KeysView' object is not an iterator
```

I would leave the others as `Iterable`, because if your Rust/Python conversion changes in the future, `Iterable` will still fulfill the requirements.  `Iterator` fulfills everything needed for `Iterable`, and I assume it's covariant in its argument, so it's ok for now, but you may as well be as restrictive as you can be when defining your output types.

retworkx/pydigraph.pyi
```diff
@@ -0,0 +1,155 @@
...
+    def compose(
+        self,
+        other: PyDiGraph[S, T],
+        node_map: Dict[int, Tuple[int, T]],
+        /,
+        node_map_func: Optional[Callable[[S], int]] = ...,
+        edge_map_func: Optional[Callable[[T], int]] = ...,
+    ) -> Dict[int, int]: ...
+    def copy(self) -> PyDiGraph[S, T]: ...
+    def edge_index_map(self) -> EdgeIndexMap: ...
```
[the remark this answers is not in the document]
IvanIsCoding: Thanks for catching this one!

retworkx/custom_return_types.pyi
IvanIsCoding: After almost half a year, some feedback: mypy agrees with Jake, I changed the code

src/digraph.rs
```diff
@@ -1066,7 +1066,7 @@ impl PyDiGraph {
     ///     ``(source, target, weight)`` where source and target are integer
     ///     node indices. If the node index is not present in the graph
     ///     nodes will be added (with a node weight of ``None``) to that index.
-    #[pyo3(text_signature = "(self, edge_lsit, /)")]
+    #[pyo3(text_signature = "(self, edge_list, /)")]
```
IvanIsCoding, line 1069: FWIW, this typo has been on our code base for 2 years. mypy caught it

tox.ini
```diff
@@ -56,12 +56,21 @@ deps =
   black~=22.0
 commands = black {posargs} '../rustworkx' '../tests' '../retworkx'
 
+[testenv:stubs]
+basepython = python3
+envdir = .tox/stubs
+deps =
+  mypy @ git+https://github.com/python/mypy.git
```
IvanIsCoding, line 63: We need to keep using the latest version until https://github.com/python/mypy/pull/14041 is out

IvanIsCoding:
Alright, this is ready for review after so long. `mypy.stubtest` checks that the stubs in this PR are identical to the `text_signature` from our Rust code and that the return types are valid.

Some things might be missing (e.g. all functions from `rustworkx.rustworkx`), but this is a good start to support annotations

retworkx/custom_return_types.pyi
jakelishman: Haha, I think Georgios was also correct at the time he wrote the comment. Maybe PyO3 changed their default conversions in the interim? Or mypy just got stricter about magic methods.

rustworkx/iterators.pyi
```diff
@@ -0,0 +1,144 @@
...
+    "CentralityMapping",
+    "BiconnectedComponents",
+    "ProductNodeMap",
+    "MultiplePathMapping",
+    "AllPairsMultiplePathMapping",
+]
+
+class RustworkxCustomVecIter(Generic[T_co], Sequence[T_co], ABC):
+    def __init__(self) -> None: ...
+    def __eq__(self, other: object) -> bool: ...
```
jakelishman, line 60: I doubt it matters at all since this is an input to a function that's a black box to Python-based type checkers (so no further operations will be performed on `other`), but it seemed marginally odd to me that this is typed as `object` rather than `Any`.

rustworkx/py.typed
```diff
@@ -0,0 +1 @@
+partial
```
jakelishman, line 1: What's actually missing, now?

tox.ini
jakelishman: I think this was released as of Mypy 1.0 a couple of weeks ago.

jakelishman (approved):
This looks good to merge to me!  Fantastic work, Ivan - I know I'm not one of the actual retworkx devs so it's a bit weird coming from me, but still!

I left a question or two inline (mostly for future reference), and maybe one place where's the possibility for _marginally_ better typing, but nothing that actually needs action on this PR.  I'd recommend to merge it now before it gets out-of-date again - it's been a great effort since the correct typing here is quite complex with the extra constraints on the `Self` type, and some of the additional co-/contra-/in- variant typing.

rustworkx/py.typed
IvanIsCoding: We are missing functions declared in Rust like `graph_floyd_warshall`` and `digraph_floyd_warshall` because I planned to do it after. Just landing this PR was hard, adding 100+ function declarations would make it worse

.github/workflows/main.yml
```diff
@@ -90,6 +90,31 @@ jobs:
...
+        with:
+          python-version: ${{ matrix.python-version }}
+          architecture: x64
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          target: x86_64-unknown-linux-gnu
+          profile: minimal
+          default: true
```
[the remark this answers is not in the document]
IvanIsCoding: You're right, but we're probably getting rid of this after #821 merges

rustworkx/iterators.pyi
IvanIsCoding: I tried to make it more specific but `mypy.stubgen` was complainining it wasn't the right one. I think `object` just avoids us comparing things to primitive types? It is still very broad

rustworkx/digraph.pyi
```diff
@@ -0,0 +1,174 @@
...
+        check_cycle: bool = ...,
+        multigraph: bool = ...,
+    ) -> None: ...
+    def add_child(self, parent: int, obj: S, edge: T, /) -> int: ...
+    def add_edge(self, parent: int, child: int, edge: T, /) -> int: ...
+    def add_edges_from(
+        self,
+        obj_list: Sequence[Tuple[int, int, T]],
+        /,
+    ) -> List[int]: ...
```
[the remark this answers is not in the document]
IvanIsCoding: PyO3 has some weird quirks with iterations (e.g. #614 that broke us) so sometimes I just accept the `Vec<_>` conversion and if it is not in our tests it is not "canonical"

rustworkx/digraph.pyi
```diff
@@ -0,0 +1,174 @@
...
+    def successor_indices(self, node: int, /) -> NodeIndices: ...
+    def successors(self, node: int, /) -> List[S]: ...
+    def to_dot(
+        self,
+        /,
+        node_attr: Optional[Callable[[S], Dict[str, str]]] = ...,
+        edge_attr: Optional[Callable[[T], Dict[str, str]]] = ...,
+        graph_attr: Optional[Dict[str, str]] = ...,
+        filename: Optional[str] = ...,
+    ) -> Optional[str]: ...
```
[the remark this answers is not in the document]
IvanIsCoding: I like this, I will try implementing it

rustworkx/iterators.pyi
jakelishman: Python 3's type system is "pure" OO in the sense that everything inherits from `object`, but the difference from `Any` is that `object` enforces the typing semantics that are associated with an object created by (e.g.) `object()` (effectively you can't do anything with it), while `Any` is the type that allows _any_ operation.

Now I think more about it, I think I can understand mypy would say that this (`__eq__`, particularly) has to be `object` not `Any`: you want to have the type system enforce that you narrow the type to the supported things before starting to access attributes on it.  Using `Any` doesn't do that.  So you and mypy are right here, and my comment was wrong.

PR #403, opened 2021-08-03
Add unweighted average shortest path length function

mtreinish:
This commit adds a new function unweighted_average_shortest_path_length.
This mirrors the default behvaior of networkx's
average_shortest_path_length() function without any weight function or
attribute specified. When we add support for weighted methods (using
dijkstra and floyd warshall) we can add a similar unified function like
networkx's.


mtreinish:
Before we merge this I probably should rework this to not use an intermediate matrix and just compute the sum over the iterator directly. Storing the shortest paths in a matrix just wastes a ton of memory, I was just being lazy when I wrote this since all the code was there to build the matrix already.

mtreinish:
https://github.com/Qiskit/retworkx/pull/403/commits/dfae66cceff3bd1948f40261711b96030d1f582c removed the matrix usage so this should be ready to go.

mtreinish:
Just out of curiosity not that it's really a fair comparison (especially as retworkx is multithreaded and I ran it locally with 64 threads) I ran this on a 500x500 grid graph to compare the performance of retworkx and networkx here with:

```python
import time
import retworkx as rx
import networkx as nx

rx_graph = rx.generators.grid_graph(500, 500)
nx_graph = nx.grid_graph((500, 500))
rx_start = time.time()
avg_rx = rx.unweighted_average_shortest_path_length(rx_graph)
rx_stop = time.time()
print(f"retworkx result: {avg_rx} took {rx_stop - rx_start}")
nx_start = time.time()
avg_nx = nx.average_shortest_path_length(nx_graph)
nx_stop = time.time()
print(f"networkx result: {avg_nx} took {nx_stop - nx_start}")
```

which returned:
```
retworkx result: 333.3333333333333 took 307.0236449241638
networkx result: 333.3333333333333 took 78192.08435726166
```

src/shortest_path/average_length.rs
```diff
@@ -0,0 +1,80 @@
...
+                            petgraph::Direction::Incoming,
+                        ) {
+                            next_level.insert(v);
+                        }
+                    }
+                } else {
+                    for v in graph.neighbors(node) {
+                        next_level.insert(v);
+                    }
+                }
```
georgios-ts, line 65: same thing, less code 
```suggestion
                for v in graph
                    .neighbors_directed(node, petgraph::Direction::Outgoing)
                {
                    next_level.insert(v);
                }
                if graph.is_directed() && as_undirected {
                    for v in graph
                        .neighbors_directed(node, petgraph::Direction::Incoming)
                    {
                        next_level.insert(v);
                    }
                }
```
```

**Review by georgios-ts — COMMENTED** [review]
Overall looks good but i think we still need to "fix" the output in case of:
- Disconnected graphs. Networkx raises an error in this case. Another possibility could be to return infinity.
- Graph with a single node. We return nan but i think zero is the right value.

File: src/lib.rs [hunk 2]
```diff
@@ -1876,37 +1876,11 @@ fn layers(
...
-    text_signature = "(graph, /, parallel_threshold=300, as_undirected=False)"
-)]
-pub fn digraph_distance_matrix(
-    py: Python,
+fn directed_distance_matrix(
     graph: &digraph::PyDiGraph,
     parallel_threshold: usize,
     as_undirected: bool,
-) -> PyResult<PyObject> {
+) -> Array2<f64> {
```
georgios-ts: This is an old commit but in my opinion it shows a general weakness of our source code: most of our functions are tightly connected to Python and it's not straight forward to re-use them inside Rust. As we expand retworkx's functionality i guess we will hit such cases more often.

src/shortest_path/average_length.rs
mtreinish: I was going to do it this way originally but didn't because I avoided it because we end up calling `insert` twice for each neighbor in an undirected graph because petgraph's `neighbors_directed() is equivalent to just `neighbors()` for undirected graphs. Not that it effects things in the output or would have a big runtime penalty it just felt better to avoid it. That being said we can probably tweak this logic to be simpler like what you have here and avoid the double loop.

src/lib.rs
mtreinish: Yeah, that's a good point and I agree with you. But it's kind of understandable as most of the code is designed with python in mind first and reuse in rust is secondary. I do hope having the new organizational structure in the repo will make us a bit better about writing things for reuse since it gives us a clear divide on the boundary between the python interface side and the internal rust functions.

FWIW, I did end up doing this as part of #405 as a separate change since it wasn't needed here anymore.

src/shortest_path/average_length.rs
mtreinish: Oh sorry, I misread your suggestion, this is correct and what I had in mind for the simpler logic that didn't duplicate. That's what I get for responding first thing in the morning :)

mtreinish:
> We already have a similar `is_weakly_connected` function. The difference is that `is_weakly_connected` takes a `PyDiGraph` as input. So my suggestion is to either adapt `is_weakly_connected` to work with `StableGraph` or adapt `is_weakly_connected` to use this new `is_connected` function. It's not that important but it seems bad having different code for two functions doing the same thing.

Yeah, I looked at that they're mostly the same (it is just a bfs though). But, the key difference besides the type difference, which is easy to fix/expand, was the use of `neighbors()`  vs `neighbors_undirected()`. I think for directed graphs in this case we want to respect edge direction while weekly connected definitely doesn't. My thinking was in a follow up we can combine the 2 so `weekly_connected_components()` calls the same inner function that does the bfs (and just make the undirected/directed a flag) and returns the set instead of a bool. But that seemed like a larger scope change than just this PR so I opted for this as the first step and we can do the deduplication in a separate PR.

georgios-ts:
@mtreinish Oh right, my bad. Let's proceed with this and we can simplify the code a bit in a follow-up PR. But as you said, we must respect the direction of the edges when `as_undirected=False`. So we should probably include `as_undirected` argument in `is_connected` function.

src/shortest_path/mod.rs
```diff
@@ -1089,3 +1091,117 @@ pub fn graph_distance_matrix(
...
+    }
+    if !is_connected(&graph.graph) {
+        return std::f64::INFINITY;
+    }
+    let sum = average_length::compute_distance_sum(
+        &graph.graph,
+        parallel_threshold,
+        true,
+    ) as f64;
+    sum / (n * (n - 1.0))
```
jlapeyre, line 1206: It looks a bit cleaner to me to have `let n = graph.node_count();`, i.e. `n` is an integer. Then something like `(n * (n - 1)) as f64`. That is, you still need only a single `as f64`, but you put it in the place where you really need it, rather than early, so that you need things like `n == 1.0` above.

mtreinish: Good idea, done in: https://github.com/Qiskit/retworkx/pull/403/commits/fbe1fa55db9401188a95bb93940eda5be0bbc500

src/shortest_path/mod.rs
```diff
@@ -1089,3 +1091,117 @@ pub fn graph_distance_matrix(
...
+    as_undirected: bool,
+) -> f64 {
+    let n = graph.node_count();
+    if n == 0 {
+        return std::f64::NAN;
+    }
+    if n == 1 {
+        return 0.0;
+    }
+    if !is_connected(&graph.graph) {
```
georgios-ts, line 1144: We need to rework the logic here for directed graphs since `is_connected` only checks if the first node is connected to all other nodes of the digraph (which does not really make any sense).

I see several possibilities depending on the definition we actually want for this function:
- Check `is_weakly_connected` if `as_undirected=True` and `is_strongly_connected` if `as_undirected=False`
- Check only `is_weakly_connected` as Networkx does and for disconnected pairs define the distance to be zero. This is not really consistent in my opinion.
- Check nothing and for disconnected pairs define the distance to be zero. This is the definition i found in [wikipedia](https://en.wikipedia.org/wiki/Average_path_length). A possible improvement here would be to divide by the number of connected pairs of nodes instead of `n`.

mtreinish: Nice catch! Yeah, this is clearly not correct. I also wasn't happy with the networkx weakly connected approach so I tried to do it differently, but you're right this doesn't make much sense. I kind of like the approach on Wikipedia since it means we only traverse the graph once, I'll give it a go and update it with this approach tomorrow.

mtreinish: I've reworked this in https://github.com/Qiskit/retworkx/pull/403/commits/44907620bc88ca3d48227b3cebccae0e4e1fb975 to use the wikipedia definition and added a test comparing a partially disconnected graph with the fully connected equivalent.

retworkx/__init__.py
```diff
@@ -176,6 +176,64 @@ def _graph_distance_matrix(graph, parallel_threshold=300, null_value=0.0):
...
+    :param int parallel_threshold: The number of nodes to calculate the
+        the distance matrix in parallel at. It defaults to 300, but this can
+        be tuned to any number of nodes.
+    :param bool as_undirected: If set to ``True`` the input directed graph
+        will be treated as if each edge was bidirectional/undirected while
+        finding the shortest paths. Default: ``False``.
+
+    :returns: The average shortest path length. If the graph is empty this
+        will return NaN, if there is a single node 0 will be returned and if
+        the graph is disconnected it will return infinity.
```
georgios-ts, line 212:
```suggestion
        will return NaN and if there is a single node 0 will be returned.
```

src/shortest_path/mod.rs
```diff
@@ -1089,3 +1090,113 @@ pub fn graph_distance_matrix(
...
+/// :param int parallel_threshold: The number of nodes to calculate the
+///     the distance matrix in parallel at. It defaults to 300, but this can
+///     be tuned to any number of nodes.
+/// :param bool as_undirected: If set to ``True`` the input directed graph
+///     will be treated as if each edge was bidirectional/undirected while
+///     finding the shortest paths. Default: ``False``.
+///
+/// :returns: The average shortest path length. If the graph is empty this
+///     will return NaN, if there is a single node 0 will be returned and if
+///     the graph is disconnected it will return infinity.
```
georgios-ts, line 1126:
```suggestion
///     will return NaN and if there is a single node 0 will be returned.
```

src/shortest_path/mod.rs
```diff
@@ -1089,3 +1090,113 @@ pub fn graph_distance_matrix(
...
+///
+/// :param PyGraph graph: The graph to compute the average shortest path length
+///     for
+/// :param int parallel_threshold: The number of nodes to calculate the
+///     the distance matrix in parallel at. It defaults to 300, but this can
+///     be tuned to any number of nodes.
+///
+/// :returns: The average shortest path length. If the graph is empty this
+///     will return NaN, if there is a single node 0 will be returned and if
+///     the graph is disconnected it will return infinity.
```
georgios-ts, line 1181:
```suggestion
///     will return NaN and if there is a single node 0 will be returned.
```

tests/digraph/test_avg_shortest_path.py
```diff
@@ -0,0 +1,81 @@
...
+        res = retworkx.unweighted_average_shortest_path_length(graph)
+        self.assertEqual(0.0, res)
+
+    def test_partially_connected_graph(self):
+        graph = retworkx.generators.directed_cycle_graph(32)
+        graph.add_nodes_from(list(range(32)))
+        res = retworkx.unweighted_average_shortest_path_length(graph)
+        s = 15872
+        n = 4032  # n*(n-1)
+        self.assertAlmostEqual(s / n, res, delta=1e-7)
```
georgios-ts, line 72: the name conflict in the variable and the inline comment is a bit confusing.
```suggestion
        den = 4032  # n*(n-1)
        self.assertAlmostEqual(s / den, res, delta=1e-7)
```

tests/digraph/test_avg_shortest_path.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def test_connected_cycle_graph(self):
+        # Show the difference between a fully connected example compared to
+        # a partially connected example in "test_partially_connected_graph":
+        graph = retworkx.generators.directed_cycle_graph(32)
+        res = retworkx.unweighted_average_shortest_path_length(graph)
+        s = 15872
+        n = 992  # n*(n-1)
+        self.assertAlmostEqual(s / n, res, delta=1e-7)
```
georgios-ts, line 81:
```suggestion
        den = 992  # n*(n-1)
        self.assertAlmostEqual(s / den, res, delta=1e-7)
```

tests/graph/test_avg_shortest_path.py
```diff
@@ -0,0 +1,81 @@
...
+        res = retworkx.unweighted_average_shortest_path_length(graph)
+        self.assertEqual(0.0, res)
+
+    def test_partially_connected_graph(self):
+        graph = retworkx.generators.cycle_graph(32)
+        graph.add_nodes_from(list(range(32)))
+        res = retworkx.unweighted_average_shortest_path_length(graph)
+        s = 8192
+        n = 4032  # n*(n-1)
+        self.assertAlmostEqual(s / n, res, delta=1e-7)
```
georgios-ts, line 72:
```suggestion
        den = 4032  # n*(n-1)
        self.assertAlmostEqual(s / den, res, delta=1e-7)
```

tests/graph/test_avg_shortest_path.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def test_connected_cycle_graph(self):
+        # Show the difference between a fully connected example compared to
+        # a partially connected example in "test_partially_connected_graph":
+        graph = retworkx.generators.cycle_graph(32)
+        res = retworkx.unweighted_average_shortest_path_length(graph)
+        s = 8192
+        n = 992  # n*(n-1)
+        self.assertAlmostEqual(s / n, res, delta=1e-7)
```
georgios-ts, line 81:
```suggestion
        den = 992  # n*(n-1)
        self.assertAlmostEqual(s / den, res, delta=1e-7)
```

georgios-ts (commented):
Nice, this looks better now. Once we have an extra test case with `as_undirected=True` for digraphs, it should be ready to go.

mtreinish:
I added tests for the digraph `as_undirected=True` path in: https://github.com/Qiskit/retworkx/pull/403/commits/5c1395a1ed3db00e37980b77ef0d65bb3b0d9b6a

georgios-ts (approved):
Thanks for the updates.

PR #404, opened 2021-08-04
Add a method to detect parallel edges

mtreinish:
This commit adds a new method has_parallel_edges() to the graph classes.
This method is used to detect if there are parallel edges in the graph.
While there is a multigraph attribute on the graph classes that only
dictates whether a graph allows parallel edges or not. A graph with
multigraph=True may still not have parallel edges. This method lets
users check this if they need to.


georgios-ts (approved):
LGTM

PR #405, opened 2021-08-06
Deduplicate distance matrix implementation

mtreinish:
This commit updates the graph_distance_matrix and
digraph_distance_matrix implementation to share a single inner function
that works for either an undirected or directed StableGraph and just
returns a 2D array. This both simplifies and deduplicates the code and
also gives other functions that need a distance matrix in the future a
function that can get a 2d ndarray object.


IvanIsCoding (approved):
This is a good idea, and maybe we should also copy it for `floyd_warshall_numpy` in a separate PR

PR #406, opened 2021-08-07
Deduplicate floyd_warshall_numpy implementation

IvanIsCoding:
Updates graph_floydwarshall_numpy and digraph_floydwarshall_numpy
to share a single inner function that works for both graph types and returns a 2D array. 

This deduplicates the code, similarly to #405


mtreinish (approved):
LGTM, thanks for doing this for floyd warshall too.

PR #409, opened 2021-08-12
Fix handling of null value in from_adjacency_matrix

### LINKED ISSUE #408: Make a graph from an adjacency matrix with negative elements
### Information

- **retworkx version**:0.9.0
- **Python version**:3.9.6
- **Operating system**:macOS Big Sur 11.4

When I tried to make a graph from an adjacency matrix with negative matrix elements, I noticed that edges corresponding to the negative elements are neglected.

When the adjacency matrix is as follows,
```python
adjacency_matrix = np.array([
    [1.0, -1.0, 2.0], 
    [0.0, 1.0, 1.0], 
    [0.0, 0.0, 1.0]
])
```
then the list of weighted edges,
```python
graph = retworkx.PyGraph.from_adjacency_matrix(adjacency_matrix)
list(graph.weighted_edge_list())
```
outputs,
```python
[(0, 0, 1.0), (0, 2, 2.0), (1, 1, 1.0), (1, 2, 1.0), (2, 2, 1.0)]
```
It means that there is no edge between node 0 and 1. 

On the other hand, if the adjacency matrix is 

```python
adjacency_matrix = np.array([
    [1.0, 1.0, 2.0], 
    [0.0, 1.0, 1.0], 
    [0.0, 0.0, 1.0]
])

graph = retworkx.PyGraph.from_adjacency_matrix(adjacency_matrix)
list(graph.weighted_edge_list())
```
then the list of weighted edges is like this:
```python
[(0, 0, 1.0), (0, 1, 1.0), (0, 2, 2.0), (1, 1, 1.0), (1, 2, 1.0), (2, 2, 1.0)]
```

Is the above behavior is a bug or intended?
If it is a bug, I expect that negative elements of a given adjacency matrix are not neglected.
For example, for the adjacency matrix
```python
adjacency_matrix = np.array([
    [1.0, -1.0, 2.0], 
    [0.0, 1.0, 1.0], 
    [0.0, 0.0, 1.0]
])
```
I want the following list of weighted edges:
```python
[(0, 0, 1.0), (0, 1, -1.0), (0, 2, 2.0), (1, 1, 1.0), (1, 2, 1.0), (2, 2, 1.0)]
```
Furthermore, it would be more convenient if we can create a graph from an adjacency matrix with complex matrix elements as follows.
```python
adjacency_matrix = np.array([
    [1.0, 1.0+1.0j, 2.0+2.0j], 
    [0.0, 2.0, 3.0+3.0j], 
    [0.0, 0.0, 3.0]
])
graph = retworkx.PyGraph.from_adjacency_matrix(adjacency_matrix)
list(graph.weighted_edge_list())
```
```python
[(0, 0, 1.0), (0, 1, 1.0+1.0j), (0, 2, 2.0+2.0), (1, 1, 2.0), (1, 2, 3.0+3.0j), (2, 2, 3.0)]
```

```python
adjacency_matrix = np.array([
    [1.0, -1.0, 2.0], 
    [0.0, 1.0, 1.0], 
    [0.0, 0.0, 1.0]
])
graph = retworkx.PyGraph.from_adjacency_matrix(adjacency_matrix)
list(graph.weighted_edge_list())
```

mtreinish:
This commit fixes an issue with the handling of null values in the
from_adjacency_matrix for PyDiGraph and PyGraph. Previously, the
function incorrectly treated any value <= 0 as a null value and didn't
add an edge for that. But the behavior should have been != instead of
less than or equal to. Additionally, the null value of 0.0 was hard
coded into the constructor methods which precluded using different
values. This commit fixes both issues by adding a new kwarg, null_value,
that lets users specify the null value they'd like to use. Additionally,
the behavior is changed to just use a not equals comparison to enable
using negative values (assuming null_value is zero).

Fixes #408

### DIFF
--- a/releasenotes/notes/adjacency_matrix_fixes-6e8c937398346c44.yaml
+++ b/releasenotes/notes/adjacency_matrix_fixes-6e8c937398346c44.yaml
@@ -4,7 +4,7 @@ features:
     Added a new kwarg, ``null_value`` to the
     :meth:`retworkx.PyDiGraph.from_adjacency_matrix` and
     :meth:`retworkx.PyGraph.from_adjacency_matrix` which is used to optionally
-    change the null value in the matrix treated as the absense of an edge. By
+    change the null value in the matrix treated as the absence of an edge. By
     default ``0.0`` is used. For example:
 
     .. jupyter-execute::
--- a/releasenotes/notes/adjacency_matrix_fixes-6e8c937398346c44.yaml
+++ b/releasenotes/notes/adjacency_matrix_fixes-6e8c937398346c44.yaml
@@ -1,0 +1,40 @@
+---
+features:
+  - |
+    Added a new kwarg, ``null_value`` to the
+    :meth:`retworkx.PyDiGraph.from_adjacency_matrix` and
+    :meth:`retworkx.PyGraph.from_adjacency_matrix` which is used to optionally
+    change the null value in the matrix treated as the absense of an edge. By
+    default ``0.0`` is used. For example:
+
+    .. jupyter-execute::
+
+      import numpy as np
+      import retworkx
+      from retworkx.visualization import mpl_draw
+
+      matrix = np.array([[np.nan, 1, 1], [1, np.nan, 1], [1, 1, 0]], dtype=np.float64)
+      graph = retworkx.PyDiGraph.from_adjacency_matrix(matrix, null_value=np.nan)
+      mpl_draw(graph, with_labels=True, edge_labels=str)
+
+fixes:
+  - |
+    Support for negative weights in the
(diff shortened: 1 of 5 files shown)


georgios-ts:
This is nice and we should probably add this new kwarg to `*_adjacency_matrix` (https://github.com/Qiskit/retworkx/blob/main/src/connectivity/mod.rs#L331) and `*_distance_matrix` (https://github.com/Qiskit/retworkx/blob/main/src/shortest_path/mod.rs#L1032) as well.

mtreinish:
Yeah, that's a good idea. I'll push up a follow up PR to do the matrix generation functions too. Want to keep this one more about the reported bug.

IvanIsCoding (approved):

PR #392, opened 2021-07-20
Add function to approximate minimum Steiner tree

mtreinish:
This commit adds a function to find an approximation of the minimum
Steiner tree using the algorithm described in "A fast algorithm for
Steiner tree" Kou, Markowsky & Berman (1981).
https://link.springer.com/article/10.1007/BF00288961
This algorithm produces a tree whose weight is within a (2 - (2 / t))
factor of the weight of the optimal Steiner tree where t is the number
of terminal nodes.
    
Closes #389

TODO:
- [x] Add docs
- [x] Expand test coverage
- [x] Add output graph edge de-duplication for parallel edges in a multigraph
- [x] Add release notes

PR #410, opened 2021-08-12
Add option to set null value for matrix generators

mtreinish:
This commit adds a new kwarg option to the *distance_matrix() and
*adjacency_matrix() functions, null_value, which is used to adjust the
float value that's used to represent the absence of an edge or a path in
the graph. This was previously hard coded as always 0.0 but this enables
users to specify an alternate value instead, such as NaN or infinity.

### DIFF
--- a/releasenotes/notes/add-null_value-matrix-generator-1225e7d553fe89a3.yaml
+++ b/releasenotes/notes/add-null_value-matrix-generator-1225e7d553fe89a3.yaml
@@ -1,0 +1,31 @@
+---
+features:
+  - |
+    The :func:`~retworkx.adjacency_matrix` function has a new kwarg
+    ``null_value`` which is used to adjust the value used in the output matrix
+    representing the absense of an edge. This can be set to any float value and
+    if not specified the default value of 0.0 is still used. For example:
+
+    .. jupyter-execute::
+
+      import numpy as np
+      import retworkx
+
+      graph = retworkx.generators.cycle_graph(25)
+      distance_matrix = retworkx.adjacency_matrix(graph, null_value=np.inf)
+      print(dsitance_matrix)
+
+  - |
+    The :func:`~retworkx.distance_matrix` function has a new kwarg
+    ``null_value`` which is used to adjust the value used in the output matrix
+    representing the absense a path. This can be set to any float value and
+    if not specified the default value of 0.0 is still used. For example:
+
+    .. jupyter-execute::
+
+      import numpy as np
+      import retworkx
+
+      graph = retworkx.generators.cycle_graph(25)
+      distance_matrix = retworkx.distance_matrix(graph, null_value=np.inf)
+      print(dsitance_matrix)
--- a/releasenotes/notes/add-null_value-matrix-generator-1225e7d553fe89a3.yaml
+++ b/releasenotes/notes/add-null_value-matrix-generator-1225e7d553fe89a3.yaml
@@ -3,7 +3,7 @@ features:
   - |
     The :func:`~retworkx.adjacency_matrix` function has a new kwarg
     ``null_value`` which is used to adjust the value used in the output matrix
-    representing the absense of an edge. This can be set to any float value and
+    representing the absence of an edge. This can be set to any float value and
     if not specified the default value of 0.0 is still used. For example:
 
     .. jupyter-execute::
@@ -13,12 +13,12 @@ features:
 
       graph = retworkx.generators.cycle_graph(25)
       distance_matrix = retworkx.adjacency_matrix(graph, null_value=np.inf)
-      print(dsitance_matrix)
+      print(distance_matrix)
 
   - |
     The :func:`~retworkx.distance_matrix` function has a new kwarg
     ``null_value`` which is used to adjust the value used in the output matrix
-    representing the absense a path. This can be set to any float value and
+    representing the absence of a path. This can be set to any float value and
     if not specified the default value of 0.0 is still used. For example:
 
     .. jupyter-execute::
@@ -28,4 +28,4 @@ features:
 
       graph = retworkx.generators.cycle_graph(25)
       distance_matrix = retworkx.distance_matrix(graph, null_value=np.inf)
-      print(dsitance_matrix)
+      print(distance_matrix)
--- a/retworkx/__init__.py
+++ b/retworkx/__init__.py
@@ -147,8 +147,8 @@ def distance_matrix(
         will be treat as if each edge was bidirectional/undirected in the
         output distance matrix.
     :param float null_value: An optional float that will treated as a null
-        value. For any node pair there isn't a path between the output matrix
-        will contain this value to indicate there isn't a path. By default this
+        value. This is the default value in the output matrix and it is used
+        to indicate the absence of an edge between 2 nodes. By default this
         is ``0.0``.
 
     :returns: The distance matrix
--- a/retworkx/__init__.py
+++ b/retworkx/__init__.py
@@ -124,7 +124,9 @@ class PyDAG(PyDiGraph):
 
 
 @functools.singledispatch
-def distance_matrix(graph, parallel_threshold=300):
+def distance_matrix(
+    graph, parallel_threshold=300, as_undirected=False, null_value=0.0
+):
     """Get the distance matrix for a graph
 
     This differs from functions like :func:`~retworkx.floyd_warshall_numpy` in
@@ -144,6 +146,9 @@ def distance_matrix(graph, parallel_threshold=300):
     :param bool as_undirected: If set to ``True`` the input directed graph
         will be treat as if each edge was bidirectional/undirected in the
         output distance matrix.
+    :param float null_value: An optional float that will treated as a null
+        value. If any element in the input matrix is this value it will be
+        treated as not an edge. By default this is ``0.0``.
 
     :returns: The distance matrix
     :rtype: numpy.ndarray
@@ -153,22 +158,25 @@ def distance_matrix(graph, parallel_threshold=300):
 
 @distance_matrix.register(PyDiGraph)
 def _digraph_distance_matrix(
-    graph, parallel_threshold=300, as_undirected=False
+    graph, parallel_threshold=300, as_undirected=False, null_value=0.0
 ):
     return digraph_distance_matrix(
         graph,
         parallel_threshold=parallel_threshold,
         as_undirected=as_undirected,
+        null_value=null_value,
     )
 
 
 @distance_matrix.register(PyGraph)
-def _graph_distance_matrix(graph, parallel_threshold=300):
-    return graph_distance_matrix(graph, parallel_threshold=parallel_threshold)
+def _graph_distance_matrix(graph, parallel_threshold=300, null_value=0.0):
+    return graph_distance_matrix(
+        graph, parallel_threshold=parallel_threshold, null_value=null_value
+    )
 
 
 @functools.singledispatch
-def adjacency_matrix(graph, weight_fn=None, default_weight=1.0):
+def adjacency_matrix(graph, weight_fn=None, default_weight=1.0, null_value=0.0):
     """Return the adjacency matrix for a graph object
 
     In the case where there are multiple edges between nodes the value in the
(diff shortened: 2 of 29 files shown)


releasenotes/notes/add-null_value-matrix-generator-1225e7d553fe89a3.yaml
```diff
@@ -0,0 +1,31 @@
...
+      import retworkx
+
+      graph = retworkx.generators.cycle_graph(25)
+      distance_matrix = retworkx.adjacency_matrix(graph, null_value=np.inf)
+      print(dsitance_matrix)
+
+  - |
+    The :func:`~retworkx.distance_matrix` function has a new kwarg
+    ``null_value`` which is used to adjust the value used in the output matrix
+    representing the absense a path. This can be set to any float value and
```
lcapelluto, line 21:
```suggestion
    representing the absence of a path. This can be set to any float value and
```

releasenotes/notes/add-null_value-matrix-generator-1225e7d553fe89a3.yaml
```diff
@@ -0,0 +1,31 @@
+---
+features:
+  - |
+    The :func:`~retworkx.adjacency_matrix` function has a new kwarg
+    ``null_value`` which is used to adjust the value used in the output matrix
+    representing the absense of an edge. This can be set to any float value and
```
lcapelluto, line 6:
```suggestion
    representing the absence of an edge. This can be set to any float value and
```

releasenotes/notes/add-null_value-matrix-generator-1225e7d553fe89a3.yaml
```diff
@@ -0,0 +1,31 @@
...
+    if not specified the default value of 0.0 is still used. For example:
```
lcapelluto, line 7: What happens with existing edges? Is it always `1.0`?

I image the adj matrix would either be bools (0/1) or weights; if weights then `inf` would make more sense as a default. And below, zero length existing path seems valid to me (especially when weights can be floats).

It seems this PR would be the first step towards changing the default. Is that part of the plan?

releasenotes/notes/add-null_value-matrix-generator-1225e7d553fe89a3.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+    .. jupyter-execute::
+
+      import numpy as np
+      import retworkx
+
+      graph = retworkx.generators.cycle_graph(25)
+      distance_matrix = retworkx.adjacency_matrix(graph, null_value=np.inf)
+      print(dsitance_matrix)
```
lcapelluto, line 16:
```suggestion
      distance_matrix = retworkx.adjacency_matrix(graph, null_value=np.inf)
      print(distance_matrix)
```

releasenotes/notes/add-null_value-matrix-generator-1225e7d553fe89a3.yaml
```diff
@@ -0,0 +1,31 @@
...
+    if not specified the default value of 0.0 is still used. For example:
+
+    .. jupyter-execute::
+
+      import numpy as np
+      import retworkx
+
+      graph = retworkx.generators.cycle_graph(25)
+      distance_matrix = retworkx.distance_matrix(graph, null_value=np.inf)
+      print(dsitance_matrix)
```
lcapelluto, line 31:
```suggestion
      print(distance_matrix)
```

retworkx/__init__.py
```diff
@@ -144,6 +146,9 @@ def distance_matrix(graph, parallel_threshold=300):
     :param bool as_undirected: If set to ``True`` the input directed graph
         will be treat as if each edge was bidirectional/undirected in the
         output distance matrix.
+    :param float null_value: An optional float that will treated as a null
+        value. If any element in the input matrix is this value it will be
+        treated as not an edge. By default this is ``0.0``.
```
lcapelluto, line 151: This does clarify one of my prev questions 👍

tests/digraph/test_dist_matrix.py
```diff
@@ -98,4 +98,48 @@ def test_digraph_distance_matrix_parallel_as_undirected(self):
...
+        graph.add_nodes_from(list(range(7)))
+        graph.add_edges_from_no_data(
+            [(0, 1), (0, 6), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6)]
+        )
+        dist = retworkx.distance_matrix(
+            graph, as_undirected=True, null_value=np.inf
+        )
+        expected = np.array(
+            [
+                [np.inf, 1.0, 2.0, 3.0, 3.0, 2.0, 1.0],
```
lcapelluto, line 114: Still feels like the distance from a node to itself should be `0.0`, especially as undirected -- at worst this would be 7

tests/digraph/test_adjacency_matrix.py
```diff
@@ -85,6 +85,20 @@ def test_multigraph_sum_cast_weight_func(self):
         self.assertIsInstance(res, np.ndarray)
         self.assertTrue(np.array_equal(np.array([[0.0, 7.5], [0.0, 0.0]]), res))
 
+    def test_multigraph_sum_cast_weight_func_non_zero_null(self):
+        graph = retworkx.PyDiGraph()
+        node_a = graph.add_node("a")
+        node_b = graph.add_node("b")
+        graph.add_edge(node_a, node_b, 7.0)
+        graph.add_edge(node_a, node_b, 0.5)
```
lcapelluto, line 93: Why does this update the edge to `prev_weight + new_weight`? No multi-edges? Why not update length?

src/connectivity/mod.rs
```diff
@@ -323,23 +323,39 @@ pub fn is_weakly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
     default_weight: f64,
+    null_value: f64,
 ) -> PyResult<PyObject> {
     let n = graph.node_count();
-    let mut matrix = Array2::<f64>::zeros((n, n));
+    let mut matrix = if null_value != 0.0 {
+        Array2::<f64>::from_elem((n, n), null_value)
+    } else {
+        Array2::<f64>::zeros((n, n))
+    };
```
lcapelluto, line 348: Is `zeros` much more efficient? Would this work?
```suggestion
    let mut matrix = Array2::<f64>::from_elem((n, n), null_value);
```

src/connectivity/mod.rs
```diff
@@ -323,23 +323,39 @@ pub fn is_weakly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
     for (i, j, weight) in get_edge_iter_with_weights(graph) {
         let edge_weight =
             weight_callable(py, &weight_fn, &weight, default_weight)?;
-        matrix[[i, j]] += edge_weight;
+        if matrix[[i, j]] == null_value
+            || (null_value.is_nan() && matrix[[i, j]].is_nan())
```
lcapelluto, line 353: On first read I thought "ha that's annoying of NaN" and then I thought "this is v smart" 😂

lcapelluto (changes requested):

src/connectivity/mod.rs
mtreinish: I don't know if it's more efficient but I think it should work. I'll go with it. I doubt initializing is more expensive than allocating an nxn array anyway

releasenotes/notes/add-null_value-matrix-generator-1225e7d553fe89a3.yaml
mtreinish: The adjacency matrix uses the edge weight for the value (and the sum of edge weights for parallel edges) as returned by the user specified callback weight function (or if that's not specified there is a default weight value https://qiskit.org/documentation/retworkx/stubs/retworkx.adjacency_matrix.html#retworkx.adjacency_matrix lists the existing options) for the elements in the matrix where there is an edge . Previously the default was to just use 0.0 for places where there was no edge the new option lets us use any float instead of 0.0 (including `inf` or `nan`) for elements where this no edge

mtreinish: I don't have any plan to change the default, more just to make it configurable for people that don't want to use 0.0 to indicate the lack of an edge. There is a potential conflict if the weight function returns 0.0 so letting people have the flexibility to pick makes sense. I'm adding the dual of this to `from_adjacency_matrix` in #409

src/connectivity/mod.rs
mtreinish: Hah, yeah I mostly found it annoying. But it's probably good that rust is stickler for avoiding undefined behavior and in the case of floats I think IEEE 754 does specifically say NaN doesn't equal NaN. It also makes sorting arrays of floats annoying too (which I had to do in #392) because of these gaps rust says f64 doesn't implement the `Ord` trait (only `PartialOrd`) which means we can't just call sort() on a Vec<f64>

tests/digraph/test_adjacency_matrix.py
mtreinish: This is adding a parallel edge between a and b, the test is specifically to test that for an adjacency matrix with parallel edges we sum the weights in the output matrix (which is the documented behavior).

tests/digraph/test_dist_matrix.py
mtreinish: That's a good point. When I was writing the PR I was annoyed at first that there were no `inf` in the output, but you're totally right we should be setting this to 0.0 not `inf` and only use `inf` for truly disconnected nodes in the graph.

mtreinish: I updated the logic in: https://github.com/Qiskit/retworkx/pull/410/commits/20594eca9fba2faec6d8f73e8f1fe48a68ebc1ee

retworkx/__init__.py
mtreinish: Oh, this is a copy and paste error from #409 there is no input matrix for this function, it generates a matrix :)

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/410/commits/ad32ec5fd08d00058b7a33e7771ab075730b33d8

src/connectivity/mod.rs
lcapelluto: ahh interesting implications

retworkx/__init__.py
```diff
@@ -147,8 +147,9 @@ def distance_matrix(
         will be treat as if each edge was bidirectional/undirected in the
         output distance matrix.
     :param float null_value: An optional float that will treated as a null
-        value. If any element in the input matrix is this value it will be
-        treated as not an edge. By default this is ``0.0``.
+        value. For any node pair there isn't a path between the output matrix
+        will contain this value to indicate there isn't a path. By default this
```
lcapelluto, line 151: Looks like a mistake, I'm guessing you meant to update this like the other ones:
```suggestion
        value. This is the default value in the output matrix and it is used
        to indicate the absence of an edge between 2 nodes. By default this
```

src/connectivity/mod.rs
```diff
@@ -366,24 +379,42 @@ pub fn digraph_adjacency_matrix(
...
     default_weight: f64,
+    null_value: f64,
 ) -> PyResult<PyObject> {
     let n = graph.node_count();
-    let mut matrix = Array2::<f64>::zeros((n, n));
+    let mut matrix = if null_value != 0.0 {
+        Array2::<f64>::from_elem((n, n), null_value)
+    } else {
+        Array2::<f64>::zeros((n, n))
+    };
```
lcapelluto, line 405: Update initialization here and in `distance_matrix.rs` as well?

lcapelluto (changes requested):

src/connectivity/mod.rs
mtreinish: Oh yeah good catch, I only updated it the one place you had the suggestion before. Updated the others here: https://github.com/Qiskit/retworkx/pull/410/commits/d58c5c3e27b8241bc21da72634c25a69aff96de2

lcapelluto (approved):

PR #412, opened 2021-08-16
Add from complex matrix function

### LINKED ISSUE #411: Add a method to make a graph from a complex matrix
Add a method like ```from_complex_adjacency_matrix``` , for makiing a graph from an adjacency matrix whose elements can be complex values.
I expect this method to work as follows:
```python
adjacency_matrix = np.array([
    [1.0, 1.0+1.0j, -2.0j], 
    [0.0, 2.0, 1.0+1.0j], 
    [0.0, 0.0, 3.0]
])

graph = retworkx.PyGraph.from_complex_adjacency_matrix(adjacency_matrix)
list(graph.weighted_edge_list())
```
output:
```python
[(0, 0, 1.0),
 (0, 1, (1+1j)),
 (0, 2, (-0-2j)),
 (1, 1, 2.0),
 (1, 2, (1+1j)),
 (2, 2, 3.0)]
```
When the input matrix is not symmetric, it may be convenient to make a graph from the elements of the upper triangular part of it, similar to ```from_adjacency_matrix```.

mtreinish:
This commit adds a new constructor method from_complex_adjacency_matrix
that will build a new graph from an adjacency matrix with complex
elements.

Fixes #411

TODO:

- [x] Undo dependency bumps to ndarray and num-complex (depends on PyO3/pyo3#1798 being released)
- [x] Add PyGraph version
- [x] Add tests
- [x] add release notes


IvanIsCoding (commented):
Update: I confirmed it is not possible, we have to stick with the current implementation

I agree that we should make #411 possible, but I would implement this differently.

If we make a generic function that can take arbitrary Python, it would have the same effect and avoid https://github.com/PyO3/pyo3/issues/1798. We could also support optional callables to make it even more flexible e.g:

```python
# without optional arguments
retworkx.PyGraph.from_custom_adjacency_matrix(adjacency_matrix) # complex numbers

# with optional arguments
retworkx.PyGraph.from_custom_adjacency_matrix(
    adjacency_matrix,
    weight_fn=lambda x: x.imag,
    null_value=0.0
) # example to do arbitrary conversions etc
```

src/digraph.rs
```diff
@@ -2183,6 +2195,79 @@ impl PyDiGraph {
...
+    ///
+    /// :returns: A new graph object generated from the adjacency matrix
+    /// :rtype: PyDiGraph
+    #[staticmethod]
+    #[pyo3(text_signature = "(matrix, /)")]
+    pub fn from_complex_adjacency_matrix<'p>(
+        py: Python<'p>,
+        matrix: PyReadonlyArray2<'p, Complex64>,
+        null_value: Option<Complex64>,
+    ) -> PyDiGraph {
```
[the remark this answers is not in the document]
mtreinish: I don't actually know if this will work as expected. Mostly because the `PyReadonlyArray2` is talking to the numpy C api to access the array in memory directly which will have the dtype set and we're basically accessing a raw c type instead of accessing it via python. While you can set an [`object_`](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.object_) as the dtype in a numpy array that's an explicit choice. I'll test this out locally though and confirm either way and report back here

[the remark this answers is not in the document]
IvanIsCoding: Indeed it looks like [because of  rust-numpy implementation](https://pyo3.github.io/rust-numpy/numpy/array/struct.PyArray.html#data-type-and-dimension), we won't be able to support `PyObject` or `numpy.object_` as they will not implement [numpy::TypeNum](https://pyo3.github.io/rust-numpy/numpy/types/trait.TypeNum.html)

src/digraph.rs
```diff
@@ -2183,6 +2195,79 @@ impl PyDiGraph {
...
+                }
+            });
+
+        PyDiGraph {
+            graph: out_graph,
+            cycle_state: algo::DfsSpace::default(),
+            check_cycle: false,
+            node_removed: false,
+            multigraph: true,
+        }
```
IvanIsCoding, line 2268: I think it is worth trying to avoid the `from_adjacency_matrix` code duplication. Maybe we can replace the logic with a generic function for [`numpy::TypeNum`](https://pyo3.github.io/rust-numpy/numpy/types/trait.TypeNum.html) and call that instead?

mtreinish: I was working in that direction before pushing up this WIP because I was blocked on the num-complex version issue. The thing I was getting most stuck on besides the num-complex versioning issues was the handling of the trait for `nan` handling. I agree we should do this as generically as possible instead of duplicating the code when we circle back to this after the next pyo3 release

IvanIsCoding: Does the [`num_traits::FloatCore`](https://docs.rs/num-traits/0.2.14/num_traits/float/trait.FloatCore.html) trait help you with handling `nan`? it is implemented for both float and complex types, and guarantees that `is_nan` will be present

georgios-ts: I believe `num_traits::FloatCore` is not implemented for complex.

We can avoid the code duplication either with a macro  (so we dont have to define any traits that a type must implement) or use a generic function over a type that implements the appropriate traits (`Copy`, `num_traits::Zero`, `std::cmp::PartialEq`, `numpy::Element`, `pyo3::ToPyObject`) plus an extra custom trait like:

```rust
trait IsNan {
    fn is_nan(self) -> bool;
}

/// https://doc.rust-lang.org/nightly/src/core/num/f64.rs.html#441
impl IsNan for f64 {
    #[inline]
    fn is_nan(self) -> bool {
        self != self
    }
}

/// https://docs.rs/num-complex/0.4.0/src/num_complex/lib.rs.html#572-574
impl IsNan for Complex64 {
    #[inline]
    fn is_nan(self) -> bool {
        self.re.is_nan() || self.im.is_nan()
    }
}
```

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/412/commits/c4460837e6d080d1c0dbd6191d519886fe96e66c

Cargo.toml
```diff
@@ -26,18 +26,19 @@ rand_pcg = "0.3"
...
-features = ["extension-module", "hashbrown", "num-bigint"]
+features = ["extension-module", "hashbrown", "num-bigint", "num-complex"]
 
 [dependencies.hashbrown]
 version = "0.11"
 features = ["rayon"]
 
 [dependencies.ndarray]
-version = "^0.13.1"
+version = "^0.15.3"
```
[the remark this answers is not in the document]
mtreinish: I had to manually rebase this but I corrected this in https://github.com/Qiskit/retworkx/pull/412/commits/fc75d042c380686c4981cc58dc82b8fd584bfcfe

georgios-ts (approved):

src/digraph.rs
```diff
@@ -2773,3 +2786,55 @@ impl PyGCProtocol for PyDiGraph {
...
 }
+
+fn _from_adjacency_matrix<'p, T>(
+    py: Python<'p>,
+    matrix: PyReadonlyArray2<'p, T>,
+    null_value: T,
+) -> PyDiGraph
+where
+    T: Copy
+        + num_traits::Zero
```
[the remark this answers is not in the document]
mtreinish: Good call removed in https://github.com/Qiskit/retworkx/pull/412/commits/55fb9b18a30603e49edbd9ffce585c8e2de132a6

src/digraph.rs
```diff
@@ -2761,3 +2774,50 @@ impl PyGCProtocol for PyDiGraph {
...
+    let mut out_graph = StableDiGraph::<PyObject, PyObject>::new();
+    let _node_indices: Vec<NodeIndex> = (0..shape[0])
+        .map(|node| out_graph.add_node(node.to_object(py)))
+        .collect();
+    array
+        .axis_iter(Axis(0))
+        .enumerate()
+        .for_each(|(index, row)| {
+            let source_index = NodeIndex::new(index);
+            for target_index in 0..row.len() {
```
georgios-ts, line 2797: One last comment before we merge this. It's not a blocker but while we are here, it might be better to iterate over the elements in `row` like:

```suggestion
            for (target_index, elem) in row.iter().enumerate() {
```
to avoid indexing `row[[target_index]]`.

mtreinish: Good idea, done in: https://github.com/Qiskit/retworkx/pull/412/commits/ada39eaa25fc5583e9506548e4495fc52e46bfb2

PR #417, opened 2021-08-18
Fix `number_weakly_connected_components` if holes in node indices.

georgios-ts:
`number_weakly_connected_components`  returns wrong answer if the graph has holes in the node indices. This commit modifies the source code of `petgraph::algo::connected_components` to handle this case.

### DIFF
--- a/src/connectivity/mod.rs
+++ b/src/connectivity/mod.rs
@@ -28,8 +28,7 @@ use pyo3::Python;
 use petgraph::algo;
 use petgraph::graph::NodeIndex;
 use petgraph::unionfind::UnionFind;
-use petgraph::visit::NodeCount;
-use petgraph::visit::{EdgeRef, IntoEdgeReferences, NodeIndexable};
+use petgraph::visit::{EdgeRef, IntoEdgeReferences, NodeCount, NodeIndexable};
 
 use ndarray::prelude::*;
 use numpy::IntoPyArray;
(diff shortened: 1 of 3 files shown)


src/connectivity/mod.rs
```diff
@@ -225,17 +227,41 @@ pub fn digraph_find_cycle(
...
+    for edge in graph.graph.edge_references() {
+        let (a, b) = (edge.source(), edge.target());
+
+        // union the two vertices of the edge
+        vertex_sets.union(graph.graph.to_index(a), graph.graph.to_index(b));
+    }
+    let mut labels = vertex_sets.into_labeling();
+    labels.sort_unstable();
+    labels.dedup();
+    labels.len()
```
IvanIsCoding, line 264: This works, but connecting the first node with the removed nodes is a bit hacky. Also, we can avoid the sort at the end.

A simpler algorithm is to subtract one from the number of components every time we do a union. If we initialize `weak_components` as the number of non-removed, we can tweak the existing code to get the answer:

```rust
for edge in graph.graph.edge_references() {
        let (u, v) = (edge.source(), edge.target());
        if vertex_sets.union(graph.graph.to_index(u), graph.graph.to_index(v)){
            weak_components -= 1;
        }
    }
```

The answer is `weak_components`

IvanIsCoding (commented):
Thanks for finding and fixing the bug! Commenting with a simpler implementation without sort

src/connectivity/mod.rs
georgios-ts: Thanks, this is much simpler! Done in 168f0d1

IvanIsCoding (approved):
LGTM!

Issue #108, opened 2020-07-30
Native rust documentation?


vlovich:
### What is the expected enhancement?
All the docs reference this as a replacement for the python networkx library. Is there a pure Rust interface available free of Python dependencies? If so some documentation about it would be helpful. If not, then this would be semi-related to #33 to add a native Rust interface.

mtreinish:
I'm not sure if there is a use case for retworkx without python. All the payloads for graphs in retworkx are python objects. You could probably leverage some of the code and functions purely from rust, but the you'd need to rely on [PyO3](https://github.com/PyO3/pyo3) to set for weights/payloads for nodes and edges in the graph. For example of what I mean, you could look at the random gnp functions that are used to generate graphs: https://github.com/Qiskit/retworkx/blob/master/src/lib.rs#L1183-L1241 That is rust code being used to generate the retworkx PyGraph objects which then gets returned to python. You could use retworkx in a manner like this but at the end of the day it would probably need python.

So I guess it depends on what you're looking to accomplish with a potential rust interface documentation. If you're just looking for a rust graph library most of what's being used inside retworkx today is based on [petgraph](https://github.com/petgraph/petgraph) and that's probably where I'd go for that. But if it's about writing a rust python extension that interacts with retworkx over a native interface than I agree there is a documentation gap. I actually hadn't thought of this before and it's something that shouldn't be hard to add docs and better support for since most of the functions and structs that exist today should be callable directly from rust.

The idea I had in mind when I opened #33 was to have a native defined stable interface for other python extensions to leverage retworkx (similar to numpy's C API). This use case probably already works for other rust python extensions (although I haven't tried this from an external project), but for that issue I was thinking of some applications with existing C/C++ (or cython) extensions that would probably want to interface with retworkx natively instead of going through python.

vlovich:
I was actually asking about a rust graph library. I googled a bunch but didn't come across petgraph so it looks like your SEO is better :). Might be worth calling out in the readme that you're wrapping petgraph. Might help other people like me trying to find that.

Glad the bug report helped you ideate documenting how to invoke from other Rust Python extensions 👍 .

mtreinish:
Having documentation on this is blocked on: https://github.com/PyO3/pyo3/issues/1444 once we have a good pattern for sharing graph objects between crates then we can document how to do it.

jlapeyre:
I also think it's a shame that retworkx only works with Python. I guess you are saying it's sort of because of
```rust
[...]

Issue #33, opened 2020-02-26
Add FFI


mtreinish:
To enable interfacing retworkx python objects with other languages directly instead of going through python for better performance we should add a foreign function interface and publish header files.

mtreinish:
https://github.com/eqrion/cbindgen/blob/master/docs.md has details on how to do this.

mtreinish:
Also: https://michael-f-bryan.github.io/rust-ffi-guide/ is useful

IvanIsCoding:
Apparently there is a Rust-to-Rust FF crate: https://github.com/rodrimati1992/abi_stable_crates/

I haven’t read the details yet but it could be interesting

PR #418, opened 2021-08-20
Stop implementing petgraph traits for `PyGraph` and `PyDiGraph`.

georgios-ts:
Stop implementing petgraph traits for `PyGraph` and `PyDiGraph`.

Previously, `PyGraph` and `PyDiGraph` implemented foreign traits defined in the petgraph library.
Most of them are useful in petgraph for a common interface for their few different graph structures.
But in retworkx some of them were never used. Others were redundant since the underlying `StableGraph`
graph structure already implements them and we can equivalently use this variable whenever it is needed.
Although it was nice implementing some of these traits, we did not make uniform use across retworkx
and it seemed more consistent to not have them at all. Likewise, `astar.rs`, `dijkstra.rs` `k_shortest_path.rs`
and `dot_utils` defined generic functions over petgraph traits and were modified to work with 
`StableGraph<PyObject, PyObject, Ty> where Ty: petgraph::EdgeType` instead. This is the easiest way
(and the pattern we have been using lately) to have the same function for directed and undirected `StableGraph`
graphs since after all, retworkx does not plan to use other graph structures from petgraph.

This is open to discussion but i'm mostly aiming here for code consistency.


src/centrality.rs
```diff
@@ -39,29 +31,16 @@ use petgraph::visit::{
 // P -- predecessors
 // S -- verts_sorted_by_distance,
 //      vertices in order of non-decreasing distance from s
-// Q -- Q
+// Q -- queue
```
jlapeyre, line 34: This list is meant to give correspondences between the identifier (variable name) used in the code and the mathematical symbol used in the expressions in the journal article. The symbol `Q` is the same in both instances.
There is a slight deviation in that I spell out "sigma" and "delta" rather that using the symbols, as the paper does. I suppose I should actually use a utf-8 character for sigma and delta.

georgios-ts: Yeah, I renamed the variable `Q` -> `queue` to avoid having `#[allow(non_snake_case)]` and this is reason i updated this list too. If you feel the code is more difficult to read, i can revert the changes.

jlapeyre: No, I think `queue` is better than `Q`. I can't think of a better name.

src/centrality.rs
```diff
@@ -78,10 +56,10 @@ where
     if graph.node_count() < parallel_threshold {
         node_indices
             .iter()
-            .map(|node_s| {
+            .map(|&node_s| {
                 (
-                    shortest_path_for_centrality(&graph, node_s),
-                    graph.to_index(*node_s),
+                    shortest_path_for_centrality(graph, node_s),
+                    graph.to_index(node_s),
```
mtreinish, line 62: we can probably change this to `node_s.index()` if we wanted and probably remove the visit traits. The use of `to_index()` before was mostly because of the trait implementations, we didn't know the type of the node id so we couldn't just convert to a usize, But, stablepygraph doesn't have this problem

mtreinish (commented):
Overall LGTM, this is great work. Having a defined type makes things a lot simpler, especially compared to petgraph's traits. I also think this will likely increase compilation speed a bit. Just, one comment inline, not a hard requirement though.

The other thing though is do we still need fixedbitset in the cargo.toml now? Since we're not using it in vf2 after #430 and the other thing using it (although not directly) was the adjacency matrix petgraph trait we're not implementing anymore.

src/centrality.rs
georgios-ts: Yeah but we would still need `petgraph::visit::NodeIndexable` for `graph.node_bound()`

georgios-ts:
We still need fixedbitset  since `visit_map` returns a `FixedBitSet` https://docs.rs/petgraph/0.6.0/petgraph/stable_graph/struct.StableGraph.html#impl-Visitable and we use it in `astar.rs` [L154](https://github.com/Qiskit/retworkx/blob/main/src/shortest_path/astar.rs#L154) and `diijkstra.rs` [L113](https://github.com/Qiskit/retworkx/blob/main/src/shortest_path/dijkstra.rs#L113)

mtreinish (approved):

jlapeyre:
As I said elsewhere https://github.com/Qiskit/retworkx/issues/108#issuecomment-919586084 I intentionally implemented betweenness centrality as python function wrappers around more generic implementations of the algorithms. My thinking was that layering the interfaces like this does not preclude providing a pure-rust API. I see now that there has been some discussion about this. But, @mtreinish said, IIUC, that it is not really possible to organize retworkx as a rust graph library with a separate Python layer. I don't see why this is so (for example betweenness centrality). But I also don't doubt that it's true, as I don't have a great grasp of the retworkx, petgraph, and rust.

This PR modifies betweenness centrality to operate on `StablePyGraph` only.  I just want to make sure we know (or I know) what we are potentially giving up in going this direction.

mtreinish:
> But, @mtreinish said, IIUC, that it is not really possible to organize retworkx as a rust graph library with a separate Python layer. I don't see why this is so (for example betweenness centrality). But I also don't doubt that it's true, as I don't have a great grasp of the retworkx, petgraph, and rust.
> 

Besides our graph types using pyobject as their weights. There are two or three other issues I see blocking here, the first and largest is our output is a c dynamic library and not a rust library, this makes using it as a dependency more difficult since it's not as simple as just adding retworkx, there is a pyo3 issue I referred to in #108 about having a pattern for sharing the generated python c api ffi between crates. This is actually probably the biggest blocker for reusing retworkx as a rust library. To unroll this we probably can have a feature flag for enabling/disabling the python extension and change the config to ship both a cdynlib and an rlib, but this would only make sense if we properly isolated all the python interface behind a feature flag which right now doesn't have a clean separation. The other option is split the rust functionality and the python interface code into separate crates then we make the python interface crate depend on the rust one. The issue with either approach is I'm not sure how cleanly we can separate the python interface from our algorithm code.

The other big place this is blocked are functions that work with weights are hard to make generic for python and rust. Something we've hit often is that to get a usable numeric weight, or a sorting key, or anything that's typed we have to call back to python which is has special requirements. Namely having a Gil handle and handling error/exception returns. This is why we have forks of upstream petgtaph function in places because the generic petgraph code wasn't compatible with how we had to extract weights from python.

None of these are necessarily insurmountable but it would require a potentially large amount work to enable it, which I'm not sure how much of a user demand there is. 

What I've actually been thinking about implementing for some time and think might be a more viable path for this sort of thing in the short term is working on #33. Which is basically expose a dedicated c api on our output c dynlib (like numpy's c api) and we avoid most of the problems I outlined above because we don't need to ship an rlib and a cdynlib for diff use cases and we can pass pyo3 ffi objects over our ffi. It's not a native rust api and it still requires python but it might be a simpler first step. But even if we implemented #33 it raises similar questions about whether we could expose different types instead of just pyobject over that c api.

Anyway I think you're raising a good point. I hadn't considered that this is making a potential future where we expose a native rust api much more difficult, and I agree we should maybe wait on this a bit and investigate whether we can and want to (assuming we can) expose a rust native api. The comments above are me mostly thinking out loud and notes for me (or someone else) on things to try when looking at this.

jlapeyre:
I can believe it would be a very big job, at least to separate the code into two APIs. I don't know how much more difficult it would be to develop thereafter. I would think twice about putting in the effort if it's not clear that it would be used as anything other than a python library. And actually think twice before going either direction (or other directions.)

I guess that currently a performant graph library for Python would have a very much higher demand than one for rust. I can imagine (but don't know how likely) that in 10 years the rust ecosystem might really want a performant graph library. Maybe if retworkx became more full featured, other bindings would be desirable as well. This is just speculation.

georgios-ts:
Thanks for raising these valid concerns. I've been thinking the same thing for a while now and i'm very much interested to take the project into this direction (functions defined over traits). Even retworkx itself, without ever exposing a native rust api, could benefit from this. https://github.com/Qiskit/retworkx/pull/403#discussion_r686638076

> The other big place this is blocked are functions that work with weights are hard to make generic for python and rust. Something we've hit often is that to get a usable numeric weight, or a sorting key, or anything that's typed we have to call back to python which is has special requirements. Namely having a Gil handle and handling error/exception returns. This is why we have forks of upstream petgtaph function in places because the generic petgraph code wasn't compatible with how we had to extract weights from python.

A workaround (haven't really checked if this pattern works everywhere but we certainly use something similar in a lot of places) might be to define
```rust
fn graph_algo<G, F, K, E>(
    graph: G,
    weight_fn: F
)
where
    G: petgraph traits, 
    K: ..,
    F: FnMut(G::EdgeRef or G::NodeRef) -> Result<K, E>,
{}
```
This can be used natively from rust (even if you wanted to return `K` from `weight_fn` you just have to wrap it in `Ok()`) and in python we can expose something like:
```rust
#[pyfunction]
fn py_graph_algo(
    py: Python,
    graph: PyGraph,
    weight_fn: PyObject,
) {
    graph_algo(&graph.graph, |w| -> PyResult<e.g f64, usize, bool> {
        let res = weight_fn.call1(py, (w,))?;
        res.extract(py)
    })
}
```

After this discussion, i still believe that the traits in `PyGraph` and `PyDiGraph` are redundant and we should remove them. This means that if i revert the changes made in generic functions, this PR can be merged without putting any obstacle in the future of the project. In follow-ups we can start working on defining more of our functions to work with petgraph traits.

mtreinish (approved):
This LGTM and is a good compromise in cleaning up unnecessary code and leaving things generic enough if we go down a path implementing a native rust api in the future. I think this just needs a rebase and then it'll be good to go.

PR #419, opened 2021-08-20
Add to_directed method to PyGraph

mtreinish:
This commit adds a new method, to_directed(), which is used to generate
a new directed graph object from a given PyGraph object. The output
PyDiGraph object has an bidirectional edge pair for each edge in the
original PyGraph object.


src/graph.rs
```diff
@@ -1080,6 +1082,46 @@ impl PyGraph {
...
+        let mut new_graph = StableDiGraph::<PyObject, PyObject>::with_capacity(
+            node_count,
+            2 * self.graph.edge_count(),
+        );
+        let mut node_map: HashMap<NodeIndex, NodeIndex> =
+            HashMap::with_capacity(node_count);
+        for node_index in self.graph.node_indices() {
+            let node = self.graph[node_index].clone_ref(py);
+            let new_index = new_graph.add_node(node);
+            node_map.insert(node_index, new_index);
```
IvanIsCoding, line 1107: Node indices might change and I think it is worth documenting it

IvanIsCoding (commented):
Implementation looks good to me, but I believe we need to document that node indices might change in case a node was removed previously. That behaviour can be counter-intuitive to users.

We could also support converting to directed graph with the gaps in the indices. But I don't think that is mandatory

src/graph.rs
mtreinish: Good call, I added a note to the docstring in: https://github.com/Qiskit/retworkx/pull/419/commits/706e07decdbb76abe8129ae9f038341b11f3f2ef (also for `to_undirected()`, which had the same thing)

IvanIsCoding (approved):

PR #427, opened 2021-08-29
Pin importlib-metadata in CI

mtreinish:
The recent release of importlib-metadata has broken an interface that
stevedore uses when looking for entrypoints (see:
https://github.com/python/importlib_metadata/issues/348 ). Several of
our test/ci dependecies use stevedore for their plugin interfaces
including stestr which is causing CI failures. To unblock CI this commit
pins the importlib metadata version in our constraints file while the
upstream issue is resolved.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,1 +1,2 @@
 decorator==4.4.2
+importlib-metadata==4.7.1


mtreinish:
The 4.8.0 release was yanked so this is no longer needed.

PR #428, opened 2021-08-29
Parallelize betweenness_centrality()

mtreinish:
This commit updates the internals of the betweenness_centrality()
function to leverage rayon parallel iterators to parallelize the loop
over all nodes in the graph which finds the shortest path for
centrality. 

TODO:

- [x] Benchmark function vs existing serial version
- [x] try to eliminate memory overhead
- [x] add parallel threshold kwarg


mtreinish:
@jlapeyre I'm not sure if you had any scripts or benchmarks you were using on #220 but if you did it would be good to test them with this as I just wrote the code and pushed this up and haven't had a chance to test it yet.

georgios-ts:
After skimming through the code, i think we can parallelize it a bit more if each thread computes and returns the `delta` vector that appears in `_accumulate_*` instead of returning `ShortestPathData`.  However this might need more changes in the code.

jlapeyre:
> I'm not sure if you had any scripts or benchmarks you were using on #220 

I didn't find a convenient workflow for scripts. I mean I had to install (in my venv) every time I wanted to run the script. This takes a minimum of three minutes, so not very interactive. But, I did use the following:

```python
import retworkx
import retworkx.generators

n = 50
g = retworkx.generators.grid_graph(n, n)
btw = retworkx.betweenness_centrality(g, normalized=False);
print(max(btw.values()))
```
This was to test correctness, speed, and overflow for various data types. The same test was used in an issue for [betweenness centrality](https://github.com/JuliaGraphs/LightGraphs.jl/blob/master/src/centrality/betweenness.jl) in the [Julia graph package](https://github.com/JuliaGraphs/LightGraphs.jl) referenced above. The result, `max(btw.values())` was nearly the same in the two implementations (I didn't count zeros, but the difference was maybe `10^(-14)`). The current retworkx version (without threads) was something like 30% faster than Julia for these tests. I  have not yet tried this on the branch for this PR.

mtreinish:
I just ran this and compared the serial to parallel versions. At a 50x50 grid the multi threaded version was comparable in performance but a bit slower (probably from thread pool overhead). But for a grid >=100x100 there is a noticeable improvement with this pr (on my desktop with 32 cores/64 threads a 100x100 took ~86.6 sec with this PR and ~104.1 sec with 0.10.1). However the intermediate storage uses a ton of memory especially for larger graphs. I hit the OOMKiller running a 200x200 grid (with 128GB of ram), so to make this approach viable I think that I need to do 2 things, add a configurable threshold for parallelization and try to eliminate (or significantly reduce as per @georgios-ts's suggestion) the intermediate storage when running in parallel.

mtreinish:
The most recent commit: https://github.com/mtreinish/retworkx/commit/a9a4d12d5c9deb14cf0f632e9da7e29a6b7b2e1a removes the intermediate storage and makes everything multithreaded and uses a lock around betweenness to get around the shared mutability. This actually improved performance pretty noticeably over the earlier version. I ran some scaling runs comparing this to the single threaded released version:

times

times-logy

(these two graphs are the same data with the second is just loglog)

I ran a linear sweep from 2x2 to a 200x200 grid graph on my laptop and my desktop. My laptop is more modest with a 6 core/12 thread cpu and 16gb of ram (and I think I started hitting thermal issues in the larger graph) and my desktop has a 32core/64 thread cpu and 128gb of ram.  The thing still left to decide is how many nodes to make the default parallel threshold (which is why I also put a loglog graph in there). Based on the loglog graph I'm thinking somewhere between 36 and 100 nodes is the point where the multithreaded version is faster.

I also have an idea to decrease the lock wait time a bit which should improve performance more. If I get that working I'll push up new graphs.

mtreinish:
The updated locking strategy I tried worked extremely well and I pushed it in: https://github.com/Qiskit/retworkx/pull/428/commits/d96408f1d54b57290e0d922cf7af7a166822de0f

The updated graphs with this new commit are (the 0.10.1 release data is unchanged but still in the graphs):

times

times-logy

mtreinish:
Ok, this is ready for review now. I added a `parallel_threshold` kwarg and went with a default of 50 nodes which in my scale testing seemed to be the point at which multithreaded was faster than a single thread.

georgios-ts (approved):
This looks great!

jlapeyre:
I'm sort of late to the party (!) But, I wonder if you can avoid some code duplication with something like the following. I'm sure this is not correct rust, but just to get the idea: `let iter_method = node_count.iter` or `let iter_method = node_count.par_iter` depending on the threshold count. Then you only need one code block with
```rust
        node_indices
            .iter_method()
            .map(|node_s| {
            ...
```

mtreinish:
> I'm sort of late to the party (!) But, I wonder if you can avoid some code duplication with something like the following. I'm sure this is not correct rust, but just to get the idea: `let iter_method = node_count.iter` or `let iter_method = node_count.par_iter` depending on the threshold count. Then you only need one code block with
> 
> ```rust
>         node_indices
>             .iter_method()
>             .map(|node_s| {
>             ...
> ```

I've tried something like this in the past for other functions with a parallel threshold like this and ran into a typing issue because the parallel iterator and the serial iterator were of different types so when you try to use the same variable for either it complains the types aren't the same. There might be a shared trait we can use a `dyn` type for, but I'll need to do some digging to see if there is a pattern that works. I agree it would be way nicer if we could just use a single iterator chain and switch between the serial and parallel (it would clean up the code in a bunch of places).

georgios-ts:
A macro could do the trick
```rust
macro_rules! do_work {
        ($iter_method:ident) => {
            node_indices
                .$iter_method()
                .map(|node_s| {
                ...
```
and then
```rust
if graph.node_count() < parallel_threshold {
        do_work!(iter)
    } else {
        do_work!(par_iter)
    }
```

jlapeyre:
Using a macro is not  bad. But of course, plain code is preferable if possible. Maybe refactoring the work into a function is the way to go. This might look very much like the macro above, but `do_work!` would be a function, not a macro.

jlapeyre:
After trying several things, I've discovered that rust is not Python (or Julia). The compiler insists on knowing the type of an argument (even by inference) at compile time. Looking online, it seems several  people accept that factoring the code is not possible (without a macro), although this particular problem is not uncommon. I found two other solutions that don't involve factoring the code into a function:
1. https://github.com/cuviper/rayon-cond
2. https://users.rust-lang.org/t/dynamically-using-or-not-using-rayon/18327/2

Issue #367, opened 2021-06-18
Stop relying on adjacency matrix in VF2State


mtreinish:
### What is the expected enhancement?

The `VF2State` struct in isomorphism.rs builds a fixedbitset adjacency matrix from petgraph for both graphs being checked. While this works well and is performant for smaller graphs, the memory requirements mean this doesn't work for large graphs. We should be able to get away without relying on an adjacency matrix for edge detection and instead just look up the presence of an edge via the normal graph api. However, there may be a runtime cost with this, so we should do some benchmarking to determine if there is and maybe set a threshold where we stop using an adjacency matrix for memory overhead concerns.

mtreinish:
Fixed by https://github.com/Qiskit/retworkx/pull/430

PR #430, opened 2021-08-30
Fix `is_subgraph_isomorphic` with parallel edges.

### LINKED ISSUE #429: Subgraph isomorphism fails with parallel edges
### Information

- **retworkx version**: main
- **Python version**: Any
- **Rust version**: Any
- **Operating system**: Any

`is_subgraph_isomorphic` returns `True` even in cases it shouldn't. This happens if the input "small" graph have parallel edges.

Handle correctly graphs with parallel edges.

```python
import retworkx as rx

first = rx.PyGraph()
first.extend_from_edge_list([
    (0, 1), (1, 2), (2, 3)
])
second = rx.PyGraph()
second.extend_from_edge_list([
    (0, 1), (0, 1)
])
rx.is_subgraph_isomorphic(first, second)
---
True
```

georgios-ts:
Closes #429.
The solution is to build an adjacency matrix with entries equal
to the number of edges connecting a pair of nodes instead of
0/1 values.

### DIFF
--- a/src/isomorphism/vf2.rs
+++ b/src/isomorphism/vf2.rs
@@ -38,18 +38,27 @@ use crate::iterators::NodeMap;
 
 type StablePyGraph<Ty> = StableGraph<PyObject, PyObject, Ty>;
 
-/// Returns the adjacency_matrix of a graph with `(i, j)` entry equal
-/// to number of edges from node `i` to node `j`.
-fn adjacency_matrix<Ty: EdgeType>(graph: &StablePyGraph<Ty>) -> Vec<usize> {
-    let n = graph.node_bound();
-    let mut matrix = vec![0; n * n];
+#[inline]
+fn sorted<N: std::cmp::PartialOrd>(x: &mut (N, N)) {
+    let (a, b) = x;
+    if b < a {
+        std::mem::swap(a, b)
+    }
+}
+
+/// Returns the adjacency matrix of a graph as a dictionary
+/// with `(i, j)` entry equal to number of edges from node `i` to node `j`.
+fn adjacency_matrix<Ty: EdgeType>(
+    graph: &StablePyGraph<Ty>,
+) -> HashMap<(NodeIndex, NodeIndex), usize> {
+    let mut matrix = HashMap::with_capacity(graph.edge_count());
     for edge in graph.edge_references() {
-        let i = edge.source().index() * n + edge.target().index();
-        matrix[i] += 1;
+        let mut item = (edge.source(), edge.target());
         if !graph.is_directed() {
-            let j = edge.source().index() + n * edge.target().index();
-            matrix[j] += 1;
+            sorted(&mut item);
         }
+        let entry = matrix.entry(item).or_insert(0);
+        *entry += 1;
     }
     matrix
 }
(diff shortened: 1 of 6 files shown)


src/isomorphism/vf2.rs
```diff
@@ -41,6 +38,48 @@ use crate::iterators::NodeMap;
 
 type StablePyGraph<Ty> = StableGraph<PyObject, PyObject, Ty>;
 
+/// Returns the adjacency_matrix of a graph with `(i, j)` entry equal
+/// to number of edges from node `i` to node `j`.
+fn adjacency_matrix<Ty: EdgeType>(graph: &StablePyGraph<Ty>) -> Vec<usize> {
```
mtreinish, line 43: I'm wondering if instead of using a matrix it would be better to fix #367 as part of this too and just query the graph object to get the edge count on demand. Not sure what the performance difference will be though, but it will at least scale better in the amount of memory required.

mtreinish:
When this is combined with #418 can we drop the fixedbitset dependency from the `Cargo.toml`?

src/isomorphism/vf2.rs
```diff
@@ -41,6 +38,48 @@ use crate::iterators::NodeMap;
...
+    let n = graph.node_bound();
+    let mut matrix = vec![0; n * n];
```
IvanIsCoding, line 45: Because of #367, I would say using a sparse matrix or hashmap would be more appropriate if we stick with the adjacency matrix approach

src/isomorphism/vf2.rs
georgios-ts: Yeah, I thought about it but in the interest of time, I decided to push this fix and in a follow up, once we also have a serious benchmark, we can check if there is a time performance difference using something like `HashMap`

georgios-ts:
This benchmark https://github.com/Qiskit/retworkx/pull/288#issuecomment-804448939 runs in `3.77 sec` with `Vec` while with `HashMap` in `2.64 sec`. So, I switched to `HashMap` in b0745b1 and we can close #367 with this PR.

I addressed also the further bugs reported https://github.com/Qiskit/retworkx/issues/429#issuecomment-910171179 in 538ba51

src/isomorphism/vf2.rs
```diff
@@ -41,6 +38,86 @@ use crate::iterators::NodeMap;
...
+        std::mem::swap(a, b)
+    }
+}
+
+/// Returns the adjacency matrix of a graph as a dictionary
+/// with `(i, j)` entry equal to number of edges from node `i` to node `j`.
+fn adjacency_matrix<Ty: EdgeType>(
+    graph: &StablePyGraph<Ty>,
+) -> HashMap<(NodeIndex, NodeIndex), usize> {
+    let mut matrix = HashMap::with_capacity(graph.edge_count());
```
mtreinish, line 79: Won't this potentially overallocate in the case of parallel edges?

src/isomorphism/vf2.rs
```diff
@@ -41,6 +38,86 @@ use crate::iterators::NodeMap;
...
+    matrix: &HashMap<(NodeIndex, NodeIndex), usize>,
+    a: NodeIndex,
+    b: NodeIndex,
+    val: usize,
+) -> bool {
+    let mut item = (a, b);
+    if !graph.is_directed() {
+        sorted(&mut item);
+    }
+    matrix.get(&item).cloned().unwrap_or(0) >= val
```
mtreinish, line 118: Do we need to actually clone here? Can't we just match on the get and use 0 if None, something like:
```suggestion
    match matrix.get(&item) {
        Some(count) => count >= val,
        None => 0 >= val, 
    } 
```

mtreinish (commented):
Overall LGTM, nice catch and fix. A couple inline comments. Also maybe we should add another fix note about the decrease in memory usage for #367?

mtreinish:
Also when this is ready I think between this and #423 that's probably big enough to start to prepare a 0.10.2 release. I'll create a stable branch and start staging that after this merges.

src/isomorphism/vf2.rs
georgios-ts: Hmm, yeah, it might over allocate but probably by not that much. It's quite difficult to obtain a more tight bound in the number of non-parallel edges. Are you suggesting to not pre-allocate at all?

src/isomorphism/vf2.rs
mtreinish: True, the overhead should be small, it just stood out as unnecessary. Thanks for updating it.

src/isomorphism/vf2.rs
mtreinish: Yeah, that's what I was getting at. But, thinking about it some more I agree it's probably not a bid deal we basically overallocate 3 usizes for each parallel edge in the graph which will probably hardly be noticeable. Especially compared to the extra allocation overhead for not pre-allocating.

src/isomorphism/vf2.rs
```diff
@@ -41,6 +38,86 @@ use crate::iterators::NodeMap;
...
+    matrix: &HashMap<(NodeIndex, NodeIndex), usize>,
+    a: NodeIndex,
+    b: NodeIndex,
+    val: usize,
+) -> bool {
+    let mut item = (a, b);
+    if !graph.is_directed() {
+        sorted(&mut item);
+    }
+    *matrix.get(&item).unwrap_or(&0) >= val
```
IvanIsCoding, line 118:
```suggestion
    edge_multiplicity(&graph, &matrix, a, b) >= val
```

I think the shorter way makes it more obvious

mtreinish (approved):
LGTM, thanks for fixing this.

PR #431, opened 2021-08-31
Add Stoer Wagner min cut algorithm.

georgios-ts:
This commits adds a new function `stoer_wagner_min_cut`
that computes a minimum cut for weighted undirected graphs.


src/connectivity/min_cut.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+
+fn stoer_wagner_phase<E>(
+    graph: &StableUnGraph<Vec<NodeIndex>, E>,
+) -> Option<((NodeIndex, NodeIndex), E)>
+where
+    E: Copy + Ord + Zero + AddAssign,
+{
+    let mut pq = PriorityQueue::<NodeIndex, E>::from(
```
mtreinish, line 83: Does it make sense to use ahash here like we do for hashmap and indexmap? https://docs.rs/priority-queue/1.1.1/priority_queue/struct.PriorityQueue.html#method.with_hasher

Cargo.toml
```diff
@@ -27,6 +27,7 @@ rayon = "1.5"
 num-traits = "0.2"
 num-bigint = "0.4"
 indexmap = "1.6.2"
+priority-queue = "1.1.1"
```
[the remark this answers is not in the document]
georgios-ts: I used priority_queue since i wanted to quickly change the priority of an element inside the queue (https://docs.rs/priority-queue/1.1.1/priority_queue/struct.PriorityQueue.html#method.change_priority_by) and `BinaryHeap` doesn't really offer this functionality. As for the general usage, i'm not 100% sure but i expect `BinaryHeap` to be faster (provided we only want to extract the max) as it's more lightweight: my understanding is that `PriorityQueue` is the same as a binary heap plus an extra `IndexMap` to store the position of an element in the heap.

src/connectivity/min_cut.rs
georgios-ts: Done in 2fc18e6

georgios-ts:
@mtreinish Yeah, it was a bit of extra work and i was forced to readjust the core logic of the function, but eventually it worked out. Since `retworkx-core` PR is now merged, I'm planning to move the generic version there with some extra docs.

georgios-ts:
@mtreinish I'm not sure what's the performance difference of safe/unsafe code here, we can do some becnhmarks if you wish. Personally I'm fine using unsafe blocks here and avoid unnecessary bounds checks since we have safety guarantees if we implement this right internally, i.e the apis are actually safe. In any case, if you feel really strong against using unsafe, i can replace the unchecked methods.

IvanIsCoding (commented):
I would like to push this PR to the 0.12 milestone. Mainly because it has the fork of petgraph's unsafe UnionFind and we keep reusing it on the codebase.

@mtreinish  implemented a custom UnionFind in #461 as well, so I think it is worth investing time to discuss the details and implement it right at first.

mtreinish:
I'm fine with that, I don't think this is critical for the 0.11.0 release nothing is blocked. It'd be nice to have (especially considering it was opened > 4 months ago. But let's defer it to early 0.12.0 so we can discuss your concerns then.

itoko:
Just my curiosity, why are we going to use DisjointSet data structure (with a slight extension) for implementing Stoer Wagner min cut algorithm? What are the advantages over another approach, e.g. copying the original graph and contracting it with storing the corresponding original nodes in each (super) node?

georgios-ts:
Performance. You avoid cloning the graph and it's faster (at least theoretically) to merge two super nodes using a DSU. But I haven't tested this (or I might have but it was a while back and can't remember). I'll benchmark the approach you suggested and if there isn't any significant performance differences I'll switch to it.

itoko:
That sounds good to me. I'm very interested in the benchmark result. I like DisjointSet/UnionFind (and want to have in some future). I know it's very fast in general, and current implementation would be already fast. However, I'm not sure this is a case where UnionFind can do the best job. I'm thinking the min-cut algorithm needs `union` function but may not need `find` function, so another approach may be comparable. I agree it's worth benchmarking.

georgios-ts:
I based b81ffa0 on NetworkX implementation [[1]](https://networkx.org/documentation/stable/_modules/networkx/algorithms/connectivity/stoerwagner.html#stoer_wagner) which doesn't use a DSU. From the limited benchmarks I did performance looks similar in most cases (especially if there're many nodes with low degree) with a single exception where DSU was significantly faster. I can't see a clear winner given the code complexity of DSU structure so I'll leave the reviewers to decide which one we pick.


For reference the following benchmark script was used with arbitrarily chosen graphs of medium sizes

```python
import time

import rustworkx
import networkx


benchmarks = [
    ("Ring of Cliques - 10 cliques", networkx.ring_of_cliques(10, 100)),
    ("Ring of Cliques - 100 cliques", networkx.ring_of_cliques(100, 10)),
    ("Mycielski Graph", networkx.mycielski_graph(10)),
    ("Margulis-Gabber-Galil graph", networkx.margulis_gabber_galil_graph(50)),
    ("Hexagonal graph", rustworkx.generators.hexagonal_lattice_graph(40, 40)),
    (
        "G_nm graph - 1_000 nodes",
        rustworkx.undirected_gnm_random_graph(1_000, 5_000, seed=2022),
    ),
    (
        "G_nm graph - 5_000 nodes",
        rustworkx.undirected_gnm_random_graph(5_000, 25_000, seed=2022),
    ),
]

for graph_name, graph in benchmarks:
    if isinstance(graph, networkx.Graph):
        graph = rustworkx.networkx_converter(graph)

    start = time.time()
    res = rustworkx.stoer_wagner_min_cut(graph)
    stop = time.time()

    run_time = stop - start

    print(
        f"{graph_name}: \n {run_time:.3} sec"
    )
```
resulting in

| Graphs | c83ad1f | b81ffa0 |
| ------- | ----------------- | ----------|
| Ring of Cliques - 10 cliques   | 1.33 sec     | 2.52 sec |
| Ring of Cliques - 100 cliques  | 0.22 sec     | 0.15 sec |
| Mycielski graph | 0.96 sec  | 6.59 sec |
| Margulis-Gabber-Galil graph | 2.59 sec | 2.36 sec |
| Hexagonal graph | 2.12 sec | 1.39 sec |
| G_nm graph - 1_000 nodes | 0.44 sec | 0.42 sec |
| G_nm graph - 5_000 nodes| 15.6 sec | 17.8 sec |

itoko:
Thank you for the quick benchmarking, @georgios-ts. I run the above code in my local environment and got the similar results.
I was curious about the surge in the Mycielski Graph case of b81ffa0 (networkx-like). I investigated it a bit and found out that is caused by adding parallel edges in the contraction of triangle subgraph structure. 
I examined the following changes to b81ffa0 
```
@@ -165,7 +165,13 @@ where
                 .map(|edge| (s, edge.target(), *edge.weight()))
                 .collect::<Vec<_>>();
             for (source, target, cost) in edges {
-                graph_with_super_nodes.add_edge(source, target, cost);
+                if let Some(edge_index) = graph_with_super_nodes.find_edge(source, target) {
+                    if let Some(edge_weight) = graph_with_super_nodes.edge_weight_mut(edge_index) {
+                        *edge_weight += cost;
+                    }
+                } else {
+                    graph_with_super_nodes.add_edge(source, target, cost);
+                }
```
and observed performance improvement as shown below.
```
Networkx-like implementation: (b81ffa0 + alpha)
Ring of Cliques - 10 cliques: 	 0.734 sec
Ring of Cliques - 100 cliques: 	 0.128 sec
Mycielski Graph: 		 0.512 sec
Margulis-Gabber-Galil graph: 	 1.49 sec
Hexagonal graph: 		 1.06 sec
G_nm graph - 1_000 nodes: 	 0.261 sec
G_nm graph - 5_000 nodes: 	 8.87 sec

DisjointSet implementation (c83ad1f):
Ring of Cliques - 10 cliques: 	 1.22 sec
Ring of Cliques - 100 cliques: 	 0.195 sec
Mycielski Graph: 		 0.814 sec
Margulis-Gabber-Galil graph: 	 2.27 sec
Hexagonal graph: 		 1.8 sec
G_nm graph - 1_000 nodes: 	 0.398 sec
G_nm graph - 5_000 nodes: 	 13.2 sec
```
I think now we can pick the networkx-like implementation with some confidence. What do you think?

georgios-ts:
Thanks for going deep into this. Good point! Parallel edges increase the number of updates in the priority queue and hence the slowdown. I applied your suggestion in 98c5f84.

rustworkx-core/src/connectivity/min_cut.rs
```diff
@@ -0,0 +1,197 @@
...
+pub fn stoer_wagner_min_cut<G, F, K, E>(graph: G, mut edge_cost: F) -> MinCut<K, G::NodeId, E>
+where
+    G: GraphProp<EdgeType = Undirected> + IntoEdges + IntoNodeIdentifiers,
+    G::NodeId: Hash + Eq,
+    F: FnMut(G::EdgeRef) -> Result<K, E>,
+    K: Copy + Ord + Zero + AddAssign,
+{
+    let mut graph_with_super_nodes = StableUnGraph::default();
+
+    let mut node_map = HashMap::new();
```
IvanIsCoding, line 130:
```suggestion
    let mut node_map = HashMap::with_capacity(graph.node_count());
```

rustworkx-core/src/connectivity/min_cut.rs
```diff
@@ -0,0 +1,197 @@
...
+    let mut rev_node_map = HashMap::new();
```
IvanIsCoding, line 131:
```suggestion
    let mut rev_node_map = HashMap::with_capacity(graph.node_count());
```

rustworkx-core/src/connectivity/min_cut.rs
```diff
@@ -0,0 +1,197 @@
...
+/// );
+/// ```
+pub fn stoer_wagner_min_cut<G, F, K, E>(graph: G, mut edge_cost: F) -> MinCut<K, G::NodeId, E>
+where
+    G: GraphProp<EdgeType = Undirected> + IntoEdges + IntoNodeIdentifiers,
+    G::NodeId: Hash + Eq,
+    F: FnMut(G::EdgeRef) -> Result<K, E>,
+    K: Copy + Ord + Zero + AddAssign,
+{
+    let mut graph_with_super_nodes = StableUnGraph::default();
```
IvanIsCoding, line 128:
```suggestion
    let mut graph_with_super_nodes = StableUnGraph::with_capacity(graph.node_count(), graph.edge_count());
```

IvanIsCoding (approved):
LGTM, I left minor comments but I must say I am happy with the implementation

rustworkx-core/src/connectivity/min_cut.rs
georgios-ts, line 130:
```suggestion
    let mut node_map = HashMap::with_capacity(graph.node_count());
```

rustworkx-core/src/connectivity/min_cut.rs
```diff
@@ -0,0 +1,197 @@
...
+/// let (min_cut, partition) = min_cut_res.unwrap().unwrap();
+/// assert_eq!(min_cut, 1);
+/// assert_eq!(
+///     HashSet::<NodeIndex>::from_iter(partition),
+///     HashSet::from_iter([e, f, g, h])
+/// );
+/// ```
+pub fn stoer_wagner_min_cut<G, F, K, E>(graph: G, mut edge_cost: F) -> MinCut<K, G::NodeId, E>
+where
+    G: GraphProp<EdgeType = Undirected> + IntoEdges + IntoNodeIdentifiers,
```
IvanIsCoding, line 123:
```suggestion
    G: GraphProp<EdgeType = Undirected> + IntoEdges + IntoNodeIdentifiers + NodeCount,
```

rustworkx-core/src/connectivity/min_cut.rs
```diff
@@ -0,0 +1,197 @@
...
+
+use hashbrown::HashMap;
+use num_traits::Zero;
+use std::{hash::Hash, ops::AddAssign};
+
+use priority_queue::PriorityQueue;
+
+use petgraph::{
+    stable_graph::StableUnGraph,
+    visit::{Bfs, EdgeRef, GraphProp, IntoEdges, IntoNodeIdentifiers},
```
IvanIsCoding, line 21:
```suggestion
    visit::{Bfs, EdgeRef, GraphProp, IntoEdges, IntoNodeIdentifiers, NodeCount},
```

IvanIsCoding (commented):
Forgot when I made the suggestion about the traits, it should fix the errors now

rustworkx-core/src/connectivity/min_cut.rs
```diff
@@ -0,0 +1,197 @@
...
+
+use hashbrown::HashMap;
+use num_traits::Zero;
+use std::{hash::Hash, ops::AddAssign};
+
+use priority_queue::PriorityQueue;
+
+use petgraph::{
+    stable_graph::StableUnGraph,
+    visit::{Bfs, EdgeRef, GraphProp, IntoEdges, IntoNodeIdentifiers, NodeCount},
```
IvanIsCoding, line 21:
```suggestion
    visit::{Bfs, EdgeCount, EdgeRef, GraphProp, IntoEdges, IntoNodeIdentifiers, NodeCount},
```

rustworkx-core/src/connectivity/min_cut.rs
```diff
@@ -0,0 +1,197 @@
...
+/// let (min_cut, partition) = min_cut_res.unwrap().unwrap();
+/// assert_eq!(min_cut, 1);
+/// assert_eq!(
+///     HashSet::<NodeIndex>::from_iter(partition),
+///     HashSet::from_iter([e, f, g, h])
+/// );
+/// ```
+pub fn stoer_wagner_min_cut<G, F, K, E>(graph: G, mut edge_cost: F) -> MinCut<K, G::NodeId, E>
+where
+    G: GraphProp<EdgeType = Undirected> + IntoEdges + IntoNodeIdentifiers + NodeCount,
```
IvanIsCoding, line 123:
```suggestion
    G: GraphProp<EdgeType = Undirected> + IntoEdges + IntoNodeIdentifiers + NodeCount + EdgeCount,
```

rustworkx-core/src/connectivity/min_cut.rs
```diff
@@ -0,0 +1,197 @@
...
+/// );
+/// ```
+pub fn stoer_wagner_min_cut<G, F, K, E>(graph: G, mut edge_cost: F) -> MinCut<K, G::NodeId, E>
+where
+    G: GraphProp<EdgeType = Undirected> + IntoEdges + IntoNodeIdentifiers + NodeCount + EdgeCount,
+    G::NodeId: Hash + Eq,
+    F: FnMut(G::EdgeRef) -> Result<K, E>,
+    K: Copy + Ord + Zero + AddAssign,
+{
+    let mut graph_with_super_nodes = StableUnGraph::with_capacity(graph.node_count(), graph.edge_count());
```
IvanIsCoding, line 128:
```suggestion
    let mut graph_with_super_nodes =
        StableUnGraph::with_capacity(graph.node_count(), graph.edge_count());
```

PR #437, opened 2021-09-06
Return adjacent nodes in both directions in `PyDiGraph.adj`

georgios-ts:
Documentation of `PyDiGraph.adj` states that the ouput dictionary
contains adjacent nodes in either direction. But previously, only
outbound neighbors were included. This has been fixed. At the same time,
this commit simplifies the source code of `PyDiGraph.adj_direction` and `PyGraph.adj`.


mtreinish (approved):
Good catch and good fix! This bug has probably been around since the earliest retworkx releases since these functions have been around since almost the beginning. Luckily I don't think that these functions are that widely used, so hopefully nobody will have been caught by this.

PR #439, opened 2021-09-07
Fixes `digraph_union` if `merge_edges` is set to true.

### LINKED ISSUE #432: `digraph_union` fails if `merge_edges` is set to true
### Information

- **retworkx version**: main
- **Python version**: Any
- **Rust version**: Any
- **Operating system**: Any

If `merge_edges` arg is set to `True`, `digraph_union` returns wrong output. This happens in many cases (e.g graphs with node holes, mismatch edge weights) and this means we need to rework the logic of the function.

Fix this.

An example where an edge is deleted when it shouldn't since edge weights don't match:
```python
import retworkx as rx

first = rx.PyDiGraph()
nodes = first.add_nodes_from([0, 1])
first.add_edges_from([
   (nodes[0], nodes[1], 'a')
])

second = rx.PyDiGraph()
nodes = second.add_nodes_from([0, 1])
second.add_edges_from([
    (nodes[0], nodes[1], 'b')
])

final = rx.digraph_union(first, second, 
                         merge_nodes=True,
                         merge_edges=True
)
print(final.weighted_edge_list())
---
WeightedEdgeList[(0, 1, a)]
```

An example with node index hole where an edge is *not* deleted but it should be removed.
```python
import retworkx as rx

first = rx.PyDiGraph()
nodes = first.add_nodes_from([0, 1])
first.add_edges_from([
   (nodes[0], nodes[1], 'a')
])

second = rx.PyDiGraph()
dummy = second.add_node('dummy')
nodes = second.add_nodes_from([0, 1])
second.add_edges_from([
    (nodes[0], nodes[1], 'a')
])
second.remove_node(dummy)

final = rx.digraph_union(first, second, 
                         merge_nodes=True,
                         merge_edges=True
)
print(final.weighted_edge_list())
---
WeightedEdgeList[(0, 1, a), (0, 1, a)]
```

georgios-ts:
Previously, `digraph_union` would falsely keep or delete
edges if `merge_edges` is set to true. This commit fixes
the logic of `digraph_union` to skip an edge from the second
graph if both its endpoints were merged to nodes from the
first graph and these nodes already share an edge with equal
weight data. At the same time, a new function `graph_union`
was added that returns the union of two `PyGraph`s.
Closes #432.

### DIFF
--- a/releasenotes/notes/bugfix-union-7da79789134a3028.yaml
+++ b/releasenotes/notes/bugfix-union-7da79789134a3028.yaml
@@ -1,7 +1,7 @@
 ---
 fixes:
   - |
-    Previously, :func:`~retworkx.digraph_union` would falsely keep or delete edges
+    Previously, :func:`~retworkx.digraph_union` would incorrectly keep or delete edges
     if argument `merge_edges` is set to true. This has been fixed and an edge from
     the second graph will be skipped if both its endpoints were merged to nodes from
     the first graph and these nodes already share an edge with equal weight data.
@@ -9,7 +9,9 @@ fixes:
 features:
   - |
     Adds a new function :func:`~retworkx.graph_union` that returns the union
-    of two :class:`~retworkx.PyGraph` objects.
+    of two :class:`~retworkx.PyGraph` objects. This is the equivalent to 
+    :func:`~retworkx.digraph_union` but for a :class:`~retworkx.PyGraph`
+    instead of for a :class:`~retworkx.PyDiGraph`.
     For example:
 
     .. jupyter-execute::
@@ -21,4 +23,3 @@ features:
       second = retworkx.generators.cycle_graph(3, weights=["node", "b_0", "b_1"])
       graph = retworkx.graph_union(first, second, merge_nodes=True)
       mpl_draw(graph)
-    
--- a/releasenotes/notes/bugfix-union-7da79789134a3028.yaml
+++ b/releasenotes/notes/bugfix-union-7da79789134a3028.yaml
@@ -1,0 +1,24 @@
+---
+fixes:
+  - |
+    Previously, :func:`~retworkx.digraph_union` would falsely keep or delete edges
+    if argument `merge_edges` is set to true. This has been fixed and an edge from
+    the second graph will be skipped if both its endpoints were merged to nodes from
+    the first graph and these nodes already share an edge with equal weight data.
+    Fixed `#432 <https://github.com/Qiskit/retworkx/issues/432>`__
+features:
+  - |
+    Adds a new function :func:`~retworkx.graph_union` that returns the union
+    of two :class:`~retworkx.PyGraph` objects.
+    For example:
+
+    .. jupyter-execute::
+
+      import retworkx
+      from retworkx.visualization import mpl_draw
+
+      first = retworkx.generators.path_graph(3, weights=["a_0", "node", "a_1"])
+      second = retworkx.generators.cycle_graph(3, weights=["node", "b_0", "b_1"])
+      graph = retworkx.graph_union(first, second, merge_nodes=True)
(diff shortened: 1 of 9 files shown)


src/union.rs
```diff
@@ -10,106 +10,167 @@
...
 
-    let compare_weights = |a: &PyAny, b: &PyAny| -> PyResult<bool> {
-        let res = a.compare(b)?;
-        Ok(res == Ordering::Equal)
-    };
-
-    for node in second.node_indices() {
-        let node_index = combined.add_node(second[node].clone_ref(py))?;
-        node_map.insert(node.index(), node_index);
+fn find_node_by_weight<Ty: EdgeType>(
```
mtreinish, line 24: Do we want to reuse this for: https://github.com/Qiskit/retworkx/blob/main/src/digraph.rs#L1400-L1419 (also just noticed it's not handling the python exception correctly)? Also should we expose a similar method for pygraph?

src/union.rs
```diff
@@ -10,106 +10,167 @@
...
-        let edge_index = combined.add_edge(
-            new_source,
-            new_target,
-            edge_weight.clone_ref(py),
-        )?;
+fn extract<T>(x: Entry<T>) -> T {
+    match x {
+        Entry::Merged(val) => val,
+        Entry::Added(val) => val,
+        Entry::None => panic!("called `Entry::extract()` on a `None` value"),
```
mtreinish, line 55: Instead if doing this I wonder if it makes more sense to have a custom exception class with a user facing error. I realize it shouldn't ever happen at runtime because we iterate over the second graph and populate all the nodes in node_map so this is a pathological edge case. I'm just worried because a user would see ```PanicException: called `Entry::extract()` on a `None` value``` if this happened right now and it's not really user actionable or something they could handle. Maybe all we need to do is change the error message to be something like:

```suggestion
        Entry::None => panic!("Unexpected internal error: called `Entry::extract()` on a `None` value. Please file an issue at https://github.com/Qiskit/retworkx/issues/new/choose with the details on how you encountered this."),
```

releasenotes/notes/bugfix-union-7da79789134a3028.yaml
```diff
@@ -0,0 +1,24 @@
+---
+fixes:
+  - |
+    Previously, :func:`~retworkx.digraph_union` would falsely keep or delete edges
+    if argument `merge_edges` is set to true. This has been fixed and an edge from
+    the second graph will be skipped if both its endpoints were merged to nodes from
+    the first graph and these nodes already share an edge with equal weight data.
+    Fixed `#432 <https://github.com/Qiskit/retworkx/issues/432>`__
+features:
```
mtreinish, line 9: I wonder if it's worth noting that `merge_nodes` and `merge_edges` for `digraph_union` are optional now and default False because previously they were both required I think.

mtreinish (commented):
Overall LGTM, definitely an improvement in the logic and does make the behavior of merge_edges more clear. Just  a few inline comments and suggestions.  The other thing is do you think it's worth adding a dispatch function `union` to `retworkx/__init__.py`?

src/union.rs
georgios-ts: Well, if we ever encounter a `None` value here it will be a "serious" bug and probably from petgraph side, i'm totally ok to just panic here with the improved error message.

releasenotes/notes/bugfix-union-7da79789134a3028.yaml
georgios-ts: Done in 088f27d

src/union.rs
georgios-ts: Done in 088f27d where i moved this in `lib.rs` so we can reuse it for `PyDiGraph` and `PyGraph` methods.

mtreinish (approved):

PR #447, opened 2021-09-13
Add support and CI config for Python 3.10

### LINKED ISSUE #415: Add python 3.10 support

Python 3.10rc1 was released recently (https://www.python.org/downloads/release/python-3100rc1/ ) which should be ABI stable so we can start to build and test against it in CI. We also will need to update our cibuildwheel version to the latest release https://github.com/pypa/cibuildwheel/releases/tag/v2.1.1 to include 3.10rc1. Although it might make sense to wait until the full 3.10 release on October 4th before bumping cibuildwheel and publishing 3.10 wheels.

mtreinish:
The release of Python 3.10 is coming soon on:[ 10/04/2021](https://www.python.org/dev/peps/pep-0619/#schedule). As of this
commit the second release candidate has been published which should be
stable to start testing. This commit adds the CI test jobs for all
platforms with Python 3.10 and updates the release wheel jobs to
including building python 3.10 wheels. In the interest of CI time this
commit also drops the win32 platform from the test matrix, as it
provides limited coverage over the win64 jobs and would be 5 (including
3.10) extra jobs.

Fixes #415

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,12 +47,11 @@ jobs:
     strategy:
       matrix:
         rust: [stable]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10-dev]
         platform: [
           { os: "macOS-latest", python-architecture: "x64", rust-target: "x86_64-apple-darwin" },
           { os: "ubuntu-latest", python-architecture: "x64", rust-target: "x86_64-unknown-linux-gnu" },
           { os: "windows-latest", python-architecture: "x64", rust-target: "x86_64-pc-windows-msvc" },
-          { os: "windows-latest", python-architecture: "x86", rust-target: "i686-pc-windows-msvc" },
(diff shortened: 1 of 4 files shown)


mtreinish:
Looks like this is blocked on building matplotlib on macOS. We can probably just disable the visualization tests to unblock this, or just wait on mpl to release on py3.10 before moving forward here

mtreinish:
Ok, skipping visualization tests on 3.10 works I've pushed up a throwaway commit https://github.com/Qiskit/retworkx/pull/447/commits/54e2c27780fa3575876d8aa830b865004351baff to verify we can build 3.10 wheels at release time. Assuming the wheels builds work, I'll revert that commit and merge this to add 3.10 ci.

mtreinish:
Fun, the py 3.10 wheel builds work fine, except with building 5 for s390x and ppc64le we don't have enough time to build them (each wheel takes ~5000 sec under qemu). I'll have to split the job into 2 halves to make it runnable in the timeout.

PR #449, opened 2021-09-14
Faster algorithm for `steiner_tree` problem.

georgios-ts:
This commits implements a faster algorithm that
achieves the same approximation ratio for the
`steiner_tree` problem based on
https://doi.org/10.1016/0020-0190(88)90066-X
There is a significant performance improvement since
we avoid computing all pairs shortest paths but rather
we solve only a single source shortest path problem.

### DIFF
--- a/src/steiner_tree.rs
+++ b/src/steiner_tree.rs
@@ -19,13 +19,15 @@ use pyo3::exceptions::PyValueError;
 use pyo3::prelude::*;
 use pyo3::Python;
 
-use petgraph::graph::{EdgeIndex, NodeIndex};
+use petgraph::stable_graph::{EdgeIndex, EdgeReference, NodeIndex};
 use petgraph::unionfind::UnionFind;
 use petgraph::visit::{EdgeRef, IntoEdgeReferences, NodeIndexable};
 
+use crate::dictmap::*;
 use crate::generators::pairwise;
 use crate::graph;
 use crate::shortest_path::all_pairs_dijkstra::all_pairs_dijkstra_shortest_paths;
+use crate::shortest_path::dijkstra;
 
 struct MetricClosureEdge {
     source: usize,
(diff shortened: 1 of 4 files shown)


georgios-ts:
Running the following benchmark
```python
import time
import random

import retworkx as rx


random.seed(42)
SIZES = [20, 30, 40]

for n in SIZES:
    graph = rx.generators.heavy_hex_graph(n + 1)

    nodes = list(graph.node_indexes())
    terminals = random.sample(nodes, len(nodes) // 4)

    start = time.time()
    tree = rx.steiner_tree(
        graph, terminals, weight_fn=lambda _: 1
    )
    stop = time.time()

    print(stop - start)
    print(
        ' Terminals:', len(terminals),
        ',Tree size:', len(tree),
    )
```

### main
```python
0.8138070106506348
 Terminals: 270 ,Tree size: 574
8.162646055221558
 Terminals: 592 ,Tree size: 1244
72.88786816596985
 Terminals: 1040 ,Tree size: 2153
```

### d583d26
```python
0.0024530887603759766
 Terminals: 270 ,Tree size: 574
0.004609107971191406
 Terminals: 592 ,Tree size: 1248
0.00787210464477539
 Terminals: 1040 ,Tree size: 2156
```

src/steiner_tree.rs
```diff
@@ -207,7 +310,7 @@ pub fn steiner_tree(
         .iter()
         .map(|x| x.iter())
         .flatten()
-        .copied()
+        .cloned()
```
mtreinish, line 313: Is there a reason this needs to be changed? I think `copied()` should still work and be more clear on the intent (I don't think for a `usize` there is much of a performance difference for `copy()` vs `clone()`.

src/steiner_tree.rs
```diff
@@ -121,6 +123,110 @@ fn _metric_closure_edges(
     Ok(out_vec)
 }
 
+fn fast_metric_edges(
```
mtreinish, line 126: Besides getting different results for the user api is there a reason we don't want to just replace `_metric_closure_edges` with this implementation? If so we should probably put that in a comment here.

mtreinish (commented):
Overall LGTM, thanks for improving the performance here. Just 2 small inline questions and then I think this is good to go.

src/steiner_tree.rs
georgios-ts: Oh, no, these functions compute different things. 
`_metric_closure_edges`  computes the shortest path between all pairs of nodes while `fast_metric_edges` computes the shortest path between all pairs `(s, t)` of the given `terminal_nodes` *provided* that:
- there is an edge `(u, v)` in the graph.
- node `s` is the closest node to  `u` among all `terminal_nodes`
- node `t`  is the closest node to `v` among all `terminal_nodes`

For example, if  all vertices are terminals, `fast_metric_edges` will return the original edges of the graph.

Probably, the name of this function was misleading. I'll add this explanation in a comment.

src/steiner_tree.rs
georgios-ts: Yeah, you're right. There is no reason for the change. I tried something that didn't work due to our MSRV requirement and apparently i did not fully reverted in 9754282.

mtreinish (approved):

Issue #448, opened 2021-09-13
Use rayon-cond to use threads conditionally without manual code duplication.


jlapeyre:
See this comment:

https://github.com/Qiskit/retworkx/pull/428#issuecomment-918391894

Sometimes one wants to choose whether to use multiple threads in a loop based on some runtime criteria. [rayon-cond](https://github.com/cuviper/rayon-cond) does the code duplication with a macro hidden behind a nice interface.

jlapeyre:
Unfortunately, rayon's `par_iter` iterates over items in a collection, but rayon-cond's `CondIterator` iterates over references to items. So code has to be changed a bit when modifying to use rayon-cond.

jlapeyre:
Finished examples of making this work are here https://github.com/Qiskit/rustworkx/pull/815 and reference therein.
It looks like at least some of the items in the opening comment above are still relevant.

PR #450, opened 2021-09-14
Use rayon-cond for using multiple threads conditionally

jlapeyre:
This PR uses rayon-cond to conditionally use rayon threads (if the job is big enough to warrant the overhead).
Before this PR, code blocks are duplicated manually for serial and parallel code paths.

Implements issue #448

 - [x] Use rayon-cond for betweenness centrality
 - [x] Use rayon-cond in the remaining places where code was duplicated by hand


jlapeyre:
I don't see an obvious way to use `CondIterator` here
https://github.com/Qiskit/retworkx/blob/747a4f6dad37e1e0495a7425e133dcc54b6cd6b7/src/shortest_path/floyd_warshall.rs#L132
since it doesn't seem to cover this functionality.

jlapeyre:
Perhaps adding a threshold in some other places might be beneficial. Such as,
https://github.com/Qiskit/retworkx/blob/747a4f6dad37e1e0495a7425e133dcc54b6cd6b7/src/shortest_path/all_pairs_dijkstra.rs#L171-L211

jlapeyre:
@mtreinish What do you think of merging this PR as it is? That is, it replaces previous uses of rayon with rayon-cond. But, it does not add any new uses of rayon/rayon-cond as suggested in https://github.com/Qiskit/retworkx/pull/450#issuecomment-919513145 . Those could be treated in a separate PR.

mtreinish:
> @mtreinish What do you think of merging this PR as it is? That is, it replaces previous uses of rayon with rayon-cond. But, it does not add any new uses of rayon/rayon-cond as suggested in [#450 (comment)](https://github.com/Qiskit/retworkx/pull/450#issuecomment-919513145) . Those could be treated in a separate PR.

Yeah I think this is fine as scoping for merging this PR. We can look at adding more uses in a follow-up PR, I think that makes sense because there will be some benchmarking needed to find the threshold values for running in multiple threads by default.

mtreinish (commented):
Did you want to update the betweenness centrality function to use rayon-cond too? It was moved to the retworkx-core crate as part of that split out: https://github.com/Qiskit/retworkx/blob/main/retworkx-core/src/centrality.rs#L106-L158

jlapeyre:
> want to update the betweenness centrality function to use rayon-cond 

Oh yeah. I missed that some how. I'll do that.

PR #454, opened 2021-09-25
Added undirected lollipop graph generator.

e-eight:

### DIFF
--- a/src/generators.rs
+++ b/src/generators.rs
@@ -2090,6 +2090,93 @@ pub fn directed_hexagonal_lattice_graph(
     }
 }
 
+/// Generate an undirected lollipop graph where a mesh graph is connected to a
+/// path
+///
+/// :param int num_mesh_nodes: The number of nodes to generate the mesh graph
+///     with. Node weights will be None if this is specified. If both
+///     ``num_mesh_nodes`` and ``mesh_weights`` are set this will be ignored and
+///     ``mesh_weights`` will be used.
+/// :param int num_path_nodes: The number of nodes to generate the path
+///     with. Node weights will be None if this is specified. If both
+///     ``num_path_nodes`` and ``path_weights`` are set this will be ignored and
+///     ``path_weights`` will be used.
+/// :param list mesh_weights: A list of node weights for the mesh graph. If both
+///     ``num_mesh_nodes`` and ``mesh_weights`` are set ``num_mesh_nodes`` will
+///     be ignored and ``mesh_weights`` will be used.
+/// :param list path_weights: A list of node weights for the path. If both
+///     ``num_path_nodes`` and ``path_weights`` are set ``num_path_nodes`` will
+///     be ignored and ``path_weights`` will be used.
+/// :param bool multigraph: When set to False the output
+///     :class:`~retworkx.PyGraph` object will not be not be a multigraph and
+///     won't  allow parallel edges to be added. Instead
+///     calls which would create a parallel edge will update the existing edge.
+///
+/// :returns: The generated mesh graph
+/// :rtype: PyGraph
+/// :raises IndexError: If neither ``num_mesh_nodes`` or ``mesh_weights`` are specified
+///
+/// .. jupyter-execute::
+///
+///   import retworkx.generators
+///   from retworkx.visualization import mpl_draw
+///
+///   graph = retworkx.generators.lollipop_graph(4, 2)
+///   mpl_draw(graph)
+///
+#[pyfunction(multigraph = true)]
+#[pyo3(
+    text_signature = "(/, num_mesh_nodes=None, num_path_nodes=None, mesh_weights=None, path_weights=None, multigraph=True)"
+)]
+pub fn lollipop_graph(
+    py: Python,
+    num_mesh_nodes: Option<usize>,
+    num_path_nodes: Option<usize>,
+    mesh_weights: Option<Vec<PyObject>>,
+    path_weights: Option<Vec<PyObject>>,
+    multigraph: bool,
+) -> PyResult<graph::PyGraph> {
+    if mesh_weights.is_none() && num_mesh_nodes.is_none() {
+        return Err(PyIndexError::new_err(
+            "num_mesh_nodes and mesh_weights list not specified",
+        ));
+    }
+    let meshlen: usize = match &mesh_weights {
+        Some(mesh_weights) => mesh_weights.len(),
+        None => num_mesh_nodes.unwrap(),
+    };
+    let mut graph = mesh_graph(py, num_mesh_nodes, mesh_weights, multigraph)?;
+
+    let path_nodes: Vec<usize> = match path_weights {
+        Some(path_weights) => {
+            let mut node_list: Vec<usize> = Vec::new();
+            for weight in path_weights {
+                let index = graph.add_node(weight)?;
+                node_list.push(index);
+            }
+            node_list
+        }
+        None => match num_path_nodes {
+            Some(num_path_nodes) => (0..num_path_nodes)
+                .map(|_| graph.add_node(py.None()).unwrap())
+                .collect(),
+            None => Vec::new(),
+        },
+    };
+
+    let pathlen = path_nodes.len();
+    if pathlen > 0 {
+        graph.add_edge(meshlen - 1, meshlen, py.None())?;
+        for (node_a, node_b) in pairwise(path_nodes) {
+            match node_a {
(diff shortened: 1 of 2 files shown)


src/generators.rs
```diff
@@ -2090,6 +2090,93 @@ pub fn directed_hexagonal_lattice_graph(
...
+    num_path_nodes: Option<usize>,
+    mesh_weights: Option<Vec<PyObject>>,
+    path_weights: Option<Vec<PyObject>>,
+    multigraph: bool,
+) -> PyResult<graph::PyGraph> {
+    if mesh_weights.is_none() && num_mesh_nodes.is_none() {
+        return Err(PyIndexError::new_err(
+            "num_mesh_nodes and mesh_weights list not specified",
+        ));
+    }
```
mtreinish, line 2143: Do we need a similar check for path weights and path nodes?

src/generators.rs
```diff
@@ -2090,6 +2090,93 @@ pub fn directed_hexagonal_lattice_graph(
...
+
+    let path_nodes: Vec<usize> = match path_weights {
+        Some(path_weights) => {
+            let mut node_list: Vec<usize> = Vec::new();
+            for weight in path_weights {
+                let index = graph.add_node(weight)?;
+                node_list.push(index);
+            }
+            node_list
+        }
```
mtreinish, line 2158: It'd be a bit more idiomatic to right it like this:

```suggestion
        Some(path_weights) => path_weights.into_iter().map(|node| graph.graph.add_node(node).index()),
```

(note I just did this in the web ui editor so there might be typos but it's the basic idea)

This also has the advantage of doing a single allocation instead of n which will be marginally faster (not that it matters too much here).

src/generators.rs
```diff
@@ -2090,6 +2090,93 @@ pub fn directed_hexagonal_lattice_graph(
...
+        None => match num_path_nodes {
+            Some(num_path_nodes) => (0..num_path_nodes)
+                .map(|_| graph.add_node(py.None()).unwrap())
+                .collect(),
+            None => Vec::new(),
+        },
+    };
```
mtreinish, line 2165: If you use `graph.graph.add_node()` for this match statement we can change the type of `path_nodes` to `Vec<NodeIndex>` and then avoid the double conversion with the edge loop below (it should be a no-op since NodeIndex is a simple wrapper, but it's just a little cleaner). This does require changing the `graph.add_edge` below to `graph.graph.add_edge` too. Also if you do this the `.index()` I used in my above suggestion should be removed.

src/generators.rs
```diff
@@ -2090,6 +2090,93 @@ pub fn directed_hexagonal_lattice_graph(
...
+                node_list.push(index);
+            }
+            node_list
+        }
+        None => match num_path_nodes {
+            Some(num_path_nodes) => (0..num_path_nodes)
+                .map(|_| graph.add_node(py.None()).unwrap())
+                .collect(),
+            None => Vec::new(),
+        },
```
mtreinish, line 2164: Instead of using a match here (which works fine) it'd probably be better to just `unwrap()` `num_path_nodes`. If you add the `if` condition I asked about in my comment above that would ensure that if this path is executed `num_path_nodes` is never `None` so unwrapping is safe. 

```suggestion
        None => (0..num_path_nodes.unwrap())
                .map(|_| graph.graph.add_node(py.None())
                .collect(),
            None => Vec::new(),
        },
```

That is unless your intent was to enable people to call: `lolipop_graph(4)` and get a 4 node mesh graph?

tests/generators/test_lollipop.py
```diff
@@ -0,0 +1,34 @@
...
+        self.assertEqual(len(graph), 20)
+        self.assertEqual(len(graph.edges()), 139)
+
+    def test_lollipop_graph_weights(self):
+        graph = retworkx.generators.lollipop_graph(
+            mesh_weights=list(range(17)), path_weights=list(range(17, 20))
+        )
+        self.assertEqual(len(graph), 20)
+        self.assertEqual(list(range(20)), graph.nodes())
+        self.assertEqual(len(graph.edges()), 139)
```
mtreinish, line 30: For these tests it would be good to assert the edge lists are consistent with what's expected for a lolipop graph. Right now these tests are just asserting the node and edge counts match up but the graph could be completely wrong. Obviously if you do this it would probably be good to reduce the size (nobody is gonna want to manually work with a 139 element edge list :)

src/generators.rs
```diff
@@ -2090,6 +2090,93 @@ pub fn directed_hexagonal_lattice_graph(
...
+///     be ignored and ``mesh_weights`` will be used.
+/// :param list path_weights: A list of node weights for the path. If both
+///     ``num_path_nodes`` and ``path_weights`` are set ``num_path_nodes`` will
+///     be ignored and ``path_weights`` will be used.
+/// :param bool multigraph: When set to False the output
+///     :class:`~retworkx.PyGraph` object will not be not be a multigraph and
+///     won't  allow parallel edges to be added. Instead
+///     calls which would create a parallel edge will update the existing edge.
+///
+/// :returns: The generated mesh graph
```
mtreinish, line 2115:
```suggestion
/// :returns: The generated lolipop graph
```

src/generators.rs
```diff
@@ -2090,6 +2090,93 @@ pub fn directed_hexagonal_lattice_graph(
...
+/// path
+///
+/// :param int num_mesh_nodes: The number of nodes to generate the mesh graph
+///     with. Node weights will be None if this is specified. If both
+///     ``num_mesh_nodes`` and ``mesh_weights`` are set this will be ignored and
+///     ``mesh_weights`` will be used.
+/// :param int num_path_nodes: The number of nodes to generate the path
+///     with. Node weights will be None if this is specified. If both
+///     ``num_path_nodes`` and ``path_weights`` are set this will be ignored and
+///     ``path_weights`` will be used.
```
mtreinish, line 2103: What happens if neither `num_path_nodes` or `path_weights` are specified?

mtreinish (commented):
Thanks for doing this, it's an excellent start! I left some suggestions and questions inline on some improvements I think we can make to the function. The other thing missing here are updating the api doc index here:

https://github.com/Qiskit/retworkx/blob/main/docs/source/api.rst#generators

to include your new function. This also needs a release note documenting the addition of a new feature for the 0.11.0 release (whenever we release it). The reasoning and process for doing a release note are documented in: https://github.com/Qiskit/retworkx/blob/main/CONTRIBUTING.md#release-notes

src/generators.rs
e-eight: According to networkx, if there are no path nodes/weights then the lollipop graph will simply be a mesh graph. So I think we do not need to check for that.

src/generators.rs
e-eight: Yeah, that was my intent, following networkx. But we need not exactly follow networkx exactly. We can make it a strictly non-mesh lollipop graph.

mtreinish: No that's fine but I think we can modify my earlier suggestion then after `let mut graph = mesh_graph(py, num_mesh_nodes, mesh_weights, multigraph)?;` we can just do something like:

```rust
if num_path_nodes.is_none() && path_weights.is_none() {
    return graph;
}
```

Then we can make this suggestion because the same holds true about num_path_nodes never being None, if the None branch of the outer match is reached.

e-eight: I was thinking about doing this, but was not sure if having multiple return-type statements is idiomatic or not (I think there's still some debate about this in C++ circles). I will go ahead and do this then.

tests/generators/test_lollipop.py
e-eight: Any suggestions how to do that? I was looking at the tests for the other graphs, but could not really figure out how to do that. From what I see it seems to me that the tests for the undirected graphs only check that the node and edge counts match up.

mtreinish: The other tests probably need to be updated (or predate the `*edge_list()` methods which might have been added after some of the generators) that's something we can fix in a new PR. The easiest way would be to do something like:
```python
graph = retworkx.generators.lolipop_graph(3, 3)
edge_list = graph.edge_list()
expected_edge_list = [
    (0, 1),
    (0, 2),
    (1, 2),
    (2, 3),
    (3, 4),
    (4, 5)
]
self.assertEqual(edge_list, expected_edge_list)
```
or use `weighted_edge_list()` if you want to validate the weights are set correctly too

PR #456, opened 2021-09-27
Add dfs-search

georgios-ts:
Widely used graph libraries like Boost Graph Library and graph-tool
provide the ability to insert callback functions at specified event
points for many common graph search algorithms. petgraph has a similar
concept in `petgraph::visit::depth_first_search` function. This commit
implements an iterative version of `depth_first_search`, that will be 
used in a follow-up in the pending PRs #444, #445. At the same time it
exposes this new functionality in Python by letting users subclassing
`retworkx.visit.DFSVisitor` and provide their own implementation for
the appropriate callback functions.

The benefit is less memory consumption since we avoid storing the results
but rather let the user take the desired action at specified points. For
example, if a user wants to process the nodes in dfs-order, we don't need
to create a new list with all the graph nodes in dfs-order but rather the user
can process a node on the fly. We can (probably) leverage this approach
in other algorithms as an alternative for our inability to provide "real" python
iterators.


mtreinish (commented):
I really like this idea. I was thinking about doing something similar for iterating over a dag in topological order (which terra does internally for a bunch of operations). But having this as a starting pattern lets us expose it for that and BFS which will give a lot of flexibility. I haven't reviewed the code in detail yet but the only other thing missing is some test coverage of the new functionality.

georgios-ts:
Yeah, test coverage is certainly missing. I was thinking that `depth_first_search` will be covered for the most part after i use it in #444, #445, but in any case i will go ahead and add some tests.

mtreinish:
I think this needs a rebase since #469 has merged and there is some overlap with that. I'd like to get this and #489 in for 0.11.0 which I'm hoping we can release in the first half of this month

retworkx-core/src/traversal/dfs_visit.rs
```diff
@@ -0,0 +1,276 @@
...
+// https://github.com/petgraph/petgraph/blob/0.6.0/src/visit/dfsvisit.rs
+
+use petgraph::visit::{
+    ControlFlow, EdgeRef, IntoEdges, Time, VisitMap, Visitable,
+};
+
+/// Return if the expression is a break value, execute the provided statement
+/// if it is a prune value.
+/// https://github.com/petgraph/petgraph/blob/0.6.0/src/visit/dfsvisit.rs#L27
+macro_rules! try_control {
```
mtreinish, line 24: I think this is duplicated with: https://github.com/Qiskit/retworkx/blob/main/retworkx-core/src/traversal/bfs_visit.rs#L19

retworkx-core/src/traversal/dfs_visit.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// An edge to an already visited node.
+    BackEdge(N, N, E),
+    /// A cross or forward edge.
+    ///
+    /// For an edge *(u, v)*, if the discover time of *v* is greater than *u*,
+    /// then it is a forward edge, else a cross edge.
+    CrossForwardEdge(N, N, E),
+    /// All edges from a node have been reported.
+    Finish(N, Time),
+}
```
mtreinish, line 58: It might be good to put a reference to the petgraph DFSEvent struct here too since this is mostly identical except for `E` generic.

mtreinish (approved):
Overall this LGTM, thanks for updating it and I'm sorry about the slow review. I think there are 2 minor things inline one around a duplicated macro definition and the other about a missing link to the upstream dfsevent struct in petgraph to show where things came from on it (since it'll have separate documentation). Other than that I think this is ready to go.

retworkx-core/src/traversal/dfs_visit.rs
georgios-ts: Done in f230431.

PR #459, opened 2021-10-09
Introduce retworkx-core crate

### LINKED ISSUE #108: Native rust documentation?

mtreinish:
This commit adds a new workspace crate to retworkx, retworkx-lib, which
contains a pure rust library built on top of petgraph for retworkx
algorithms and additional non-python specific data structures. Prior to
this we didn't really have a public rust API for retworkx as the main
output from the retworkx crate was a cdynlib for use with python. By
creating a new crate this gives us a place to enable rust reuse of the
library for functionality that is not specific to python. Any potential
functionality in retworkx-lib will need to be solely reusable in rust
without pyo3 or python.

The primary retworkx crate will still be primarily for the python
library. But for places where it makes sense to expose a rust api we can
use the retworkx-lib crate to export those for other rust usage.

Closes #108

TODO (~probably for 0.12.0~ Done)

- [x] Figure out why setuptools-rust doesn't like the new workspace
- [x] Add real API documentation (needed to close #108)
- [x] Add retworkx-lib README
- [x] Migrate additional generic functions still in base retworkx crate
- [x] Add description of retworkx-lib and what code goes in it to the contributing docs

### DIFF
--- a/retworkx-core/Cargo.toml
+++ b/retworkx-core/Cargo.toml
@@ -7,6 +7,7 @@ description = "Rust APIs used for retworkx algorithms"
 license = "Apache-2.0"
 readme = "README.md"
 repository = "https://github.com/Qiskit/retworkx"
+homepage = "https://github.com/Qiskit/retworkx/tree/main/retworkx-core"
 keywords = ["graph"]
 
 [dependencies]


mtreinish:
@jlapeyre this should introduce a standalone Rust library/API per our discussions in #418 and #108. This seemed like the cleanest way to do it, making a separate rust crate/package with the library API for the generic rust functionality built on top of petgraph. Then downstream rust users can just use `retworkx-lib` to get that rust api and the `retwork` crate remains solely the python pieces. I started by moving the existing generic functionality over to the lib crate here and we can expand it over time.

IvanIsCoding:
This is a great step forward! There are still some items we can untie from PyO3 (e.g. one of the Floyd-Warshalls), but just having `retworkx-lib` as is a big step.

I will take some time to review this because the expectations for Rust crates are different than those from Python libraries.

mtreinish:
> Thanks for the updates!
> 
> One last thing is that we could define a type in retworkx-lib like
> 
> ```rust
> use std::convert::Infallible;
> 
> /// A convenient type alias that by default assumes no error can happen.
> ///
> /// It can be used to avoid type annotations when the function you want
> /// to use needs a callback that returns [`Result`] but in your case no
> /// error can happen.
> pub type Result<T, E = Infallible> = core::result::Result<T, E>;
> ```
> 
> and use this as the return type in places like the shortest path algorithms and max weight matching (and in the docs). Users can also use this in order to avoid giving dummy type hints to the compiler.

I added this in: https://github.com/Qiskit/retworkx/pull/459/commits/2c0cb3ea5968e05e7ad1dc61bc8303cecffc8b04 but feel free to use tweak it if how I used it in the doc tests wasn't what you were thinking.

PR #469, opened 2021-10-20
Add bfs-search

georgios-ts:
This commit implements a `breadth_first_search` that provides
the ability to insert callback functions at specified event points.
Python users should subclass `retworkx.visit.BFSVisitor` and overwrite
the appropriate callback functions.

Similar to #456


retworkx-core/src/traversal/bfs_visit.rs
```diff
@@ -0,0 +1,254 @@
...
+    F: FnMut(BfsEvent<G::NodeId, &G::EdgeWeight>) -> C,
+    C: ControlFlow,
+{
+    let discovered = &mut graph.visit_map();
+    let finished = &mut graph.visit_map();
+
+    for start in starts {
+        try_control!(
+            bfs_visitor(graph, start, &mut visitor, discovered, finished),
+            unreachable!()
```
mtreinish, line 178: I think having a comment here would be nice to explain why this branch is unreachable. While it's pretty obvious from the context for these kind of things I find it better to be exlicit

retworkx/__init__.py
```diff
@@ -1730,3 +1733,65 @@ def _graph_union(
...
+                  if (color[v] = WHITE)
+                      color[v] = GRAY
+                  else                             (u,v) is a non - tree edge
+                      if (color[v] = GRAY)         (u,v) has a gray target
+                          ...
+                      else if (color[v] = BLACK)   (u,v) has a black target
+                          ...
+              end for
+              color[u] := BLACK                    finish vertex u
+          end while
```
mtreinish, line 1766: I think having this in the retworkx-core function's docstring would be nice too.

retworkx/__init__.py
```diff
@@ -1730,3 +1733,65 @@ def _graph_union(
...
+    If an exception is raised inside the callback function, the graph traversal
+    will be stopped immediately. You can exploit this to exit early by raising a
+    :class:`~retworkx.visit.StopSearch` exception. You can also prune part of the
+    search tree by raising :class:`~retworkx.visit.PruneSearch`.
+
+    In the following example we keep track of the tree edges:
+
+    .. note::
+
+        Graph can *not* be mutated while traversing.
```
mtreinish, line 1777: I assume you want this bold not italics:

```suggestion
        Graph can **not** be mutated while traversing.
```

retworkx/__init__.py
```diff
@@ -1730,3 +1733,65 @@ def _graph_union(
...
+
+    :param PyGraph graph: The graph to be used.
```
mtreinish, line 1779:
```suggestion
    :param graph: The graph to be used. This can be a :class:`~retworkx.PyGraph`
        or a :class:`~retworkx.PyDiGraph`
```

src/lib.rs
```diff
@@ -198,6 +199,10 @@ create_exception!(retworkx, NoSuitableNeighbors, PyException);
 create_exception!(retworkx, NullGraph, PyException);
 // No path was found between the specified nodes.
 create_exception!(retworkx, NoPathFound, PyException);
+// Prune part of the search tree while traversing a graph.
+import_exception!(retworkx.visit, PruneSearch);
+// Stop graph traversal.
+import_exception!(retworkx.visit, StopSearch);
```
mtreinish, line 205: Is there a difference between doing this in Python vs Rust? (TBH I didn't even know about this macro).

src/traversal/bfs_visit.rs
```diff
@@ -0,0 +1,67 @@
...
+        BfsEvent::Finish(u) => vis.finish_vertex.call1(py, (u.index(),)),
+    };
+
+    match res {
+        Err(e) => {
+            if e.is_instance::<PruneSearch>(py) {
+                Ok(Control::Prune)
+            } else {
+                Err(e)
+            }
```
mtreinish, line 63: So `StopSearch` isn't special here? I assumed  you added a custom class so it would be treated as a special exit condition  but this just raises. If this is the case do we need a special stop exception?

mtreinish (commented):
Overall this LGTM I'm looking forward to getting this and the dfs in retworkx. I just have a few questions and suggestions inline.

Also, sorry this has taken me so long to look at, I've been heads down on some qiskit-terra functionality and haven't had much time for retworkx reviews.

src/lib.rs
georgios-ts: No big difference, initially I did this in Rust but later thought it would be nicer to have a single Python file with these custom exceptions and the visitor base classes.

src/traversal/bfs_visit.rs
georgios-ts: Yeah, it's not really special but just defined for user convenience. It's treated as any other exception that may occur in python side, i.e exit early.

retworkx/__init__.py
georgios-ts: Done in aa27296

retworkx-core/src/traversal/bfs_visit.rs
georgios-ts: Done in aa27296

src/traversal/bfs_visit.rs
mtreinish: I'm wondering if we want to make it special, so if a user callback raises `StopSearch` it causes the function to return without an error kind of like how  `for i in iter` doesn't raise `StopIteration` when iter is exhausted. That was at least my expectation when I saw the 2 exception classes.

georgios-ts: Sure, that's a good point. My main thought was if we do not raise, a user might not know if the search terminated normally or exited early and a further action might depend on that (as in [#7211](https://github.com/Qiskit/qiskit-terra/pull/7211)) . But I guess users can define their own custom exception if they wish this kind of behavior.

georgios-ts: Done in 871d9eb.

retworkx-core/src/traversal/bfs_visit.rs
```diff
@@ -0,0 +1,282 @@
...
+                        );
+                    }
+                }
+            }
+
+            let first_finish = finished.visit(u);
+            debug_assert!(first_finish);
+            try_control!(
+                visitor(BfsEvent::Finish(u)),
+                panic!("Pruning on the `BfsEvent::Finish` is not supported!")
```
mtreinish, line 276: I get why you did this since a panic will raise to the user directly without needing to manually propogate an err return through the whole path. Also in the python case this gets exposed as a `PanicException` with the error message.

But I wonder if we want to have an error return associated with this. Especially for `retworkx-core` users it's a bit off-putting I think because to handle this they need to `catch_unwind` a call to `breadth_first_search()`. Especially as this is a user error right? It's a mistake for the user to request a `Prune` on a `Finish`  event.

mtreinish (commented):
Thanks for the fast updates, I think this is basically ready now. Before approving I just have one inline question/comment before I'm ready to approve. 

Also I was looking at coverage and I noticed a bunch of continue lines don't have coverage, for example: https://coveralls.io/builds/44296139/source?filename=retworkx-core%2Fsrc%2Ftraversal%2Fbfs_visit.rs#L247 is that an issue with the coverage collection or are we missing a path in the tests so that's not getting triggered.

retworkx-core/src/traversal/bfs_visit.rs
georgios-ts: Yeah, it's a user mistake and I'm in favor of using panic rather returning a `Result` since I consider this as a bug that should be fixed in their code rather giving them the chance to "recover". Also defining an error here would only complicate user valid code, i.e they might have to deal with return types like `Result<Result<_, _>, _>`.

georgios-ts:
> Also I was looking at coverage and I noticed a bunch of continue lines don't have coverage, for example: https://coveralls.io/builds/44296139/source?filename=retworkx-core%2Fsrc%2Ftraversal%2Fbfs_visit.rs#L247 is that an issue with the coverage collection or are we missing a path in the tests so that's not getting triggered.

Yeah, we miss some test paths, let me add some extra tests to cover these lines.

retworkx-core/src/traversal/bfs_visit.rs
mtreinish: Ok, that's fine, we have it documented as panicking in this case so that should be enough

mtreinish:
I think having a test case that just runs like normal when `PruneSearch` is raised in those edge cases make sense to me. We just make it explicit the test is expected to work fine and we're just testing pathological edge cases. I don't think we need to panic there too.

retworkx/__init__.py
```diff
@@ -1730,3 +1733,67 @@ def _graph_union(
...
+          end while
+
+    If an exception is raised inside the callback function, the graph traversal
+    will be stopped immediately. You can exploit this to exit early by raising a
+    :class:`~retworkx.visit.StopSearch` exception, in which case the search function
+    will return but without raising back the exception. You can also prune part of
+    the search tree by raising :class:`~retworkx.visit.PruneSearch`.
+
+    In the following example we keep track of the tree edges:
+
```
mtreinish, line 1775: The example here is missing

mtreinish (approved):
LGTM thanks for updating this. The dispatch function is just missing the example from it's docstring after that's added I think this is ready to merge.

retworkx/__init__.py
georgios-ts: Oops, good catch! Added in e6123aa.

PR #471, opened 2021-10-22
Added generator for barbell graphs

e-eight:

### DIFF
--- a/src/generators.rs
+++ b/src/generators.rs
@@ -2359,7 +2359,7 @@ pub fn barbell_graph(
     let right_mesh = left_mesh.clone();
     let meshlen = left_mesh.num_nodes();
 
-    if !num_path_nodes.is_none() || !path_weights.is_none() {
+    if num_path_nodes.is_some() || path_weights.is_some() {
         let path_nodes: Vec<NodeIndex> = match path_weights {
             Some(path_weights) => path_weights
                 .into_iter()
@@ -2399,8 +2399,8 @@ pub fn barbell_graph(
         py.None(),
     );
     for edge in right_mesh.graph.edge_references() {
-        let new_source = NodeIndex::new(lollipoplen + &edge.source().index());
-        let new_target = NodeIndex::new(lollipoplen + &edge.target().index());
+        let new_source = NodeIndex::new(lollipoplen + edge.source().index());
+        let new_target = NodeIndex::new(lollipoplen + edge.target().index());
         let weight = edge.weight();
         left_mesh
             .graph
--- a/src/generators.rs
+++ b/src/generators.rs
@@ -2362,8 +2362,8 @@ pub fn barbell_graph(
 
     let right_mesh = left_mesh.clone();
 
-    if num_path_nodes.is_some() {
-        let path_nodes: Vec<NodeIndex> = (0..num_path_nodes.unwrap())
+    if let Some(num_nodes) = num_path_nodes {
+        let path_nodes: Vec<NodeIndex> = (0..num_nodes)
             .map(|_| left_mesh.add_node(py.None()))
             .collect();
         left_mesh.add_edge(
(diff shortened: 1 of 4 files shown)


src/generators.rs
```diff
@@ -2216,6 +2216,111 @@ pub fn lollipop_graph(
...
+    py: Python,
+    num_mesh_nodes: Option<usize>,
+    num_path_nodes: Option<usize>,
+    mesh_weights: Option<Vec<PyObject>>,
+    path_weights: Option<Vec<PyObject>>,
+    multigraph: bool,
+) -> PyResult<graph::PyGraph> {
+    let mut left_mesh =
+        mesh_graph(py, num_mesh_nodes, mesh_weights, multigraph)?;
+    let right_mesh = left_mesh.clone();
```
IvanIsCoding, line 2271: This clone does not feel right, is it intended that the right mesh will always have the same weights as the left mesh?

src/generators.rs
```diff
@@ -2216,6 +2216,111 @@ pub fn lollipop_graph(
...
+                    None => continue,
+                };
+            }
+        }
+    }
+
+    let lollipoplen = left_mesh.num_nodes();
+    for node in right_mesh.graph.node_indices() {
+        let new_node = &right_mesh.graph[node];
+        left_mesh.graph.add_node(new_node.clone_ref(py));
```
IvanIsCoding, line 2306: Same comment about nodes in the right mesh having the same weight as in the left mesh

IvanIsCoding (commented):
The code generates a barbell graph and I have no comments about that, thanks for the contribution!

A minor concern is how to handle the weights in the two meshs: apparently the weights of the right mesh are just copies from the left? That does not match the behavior of other generators, the users should be able to provide weights for the right mesh as well. [Or we can do like networx, that avoids the problem  by not accepting weights](https://networkx.org/documentation/stable/_modules/networkx/generators/classic.html#barbell_graph)

e-eight:
The way I understand it, the right and left meshes are supposed to be identical, like how the weights on both ends of a barbell are same.

But yes forgoing the node weights option is a possibility.

Nov 11, 2021 16:23:35 Ivan Carvalho ***@***.***>:

IvanIsCoding:
I'm in favor of omitting weights or having separate arguments for the weights in the left and right mesh.

e-eight:
The way it is defined here in https://mathworld.wolfram.com/BarbellGraph.html, the left and right meshes should be identical. I think cloning the left mesh is the right way to produce the right mesh, but we can do away with the weights. 

If we want the left and right meshes to be different, we can just make a lollipop graph and then compose a mesh graph to that.

IvanIsCoding:
> The way it is defined here in https://mathworld.wolfram.com/BarbellGraph.html, the left and right meshes should be identical. I think cloning the left mesh is the right way to produce the right mesh, but we can do away with the weights.
> 
> If we want the left and right meshes to be different, we can just make a lollipop graph and then compose a mesh graph to that.

Wolfram's definition does not mention node weights, I agree that the meshes should be the same i.e. have the same structure. But not necessarily the same node weights. 

Having the same weights with shared Python references by default does not seems ideal, and it would be better not to have them.

e-eight:
Alright, sounds good. I will remove the weights. But I guess in that case this will be the only generator without the weights. Is that okay?

Nov 14, 2021 16:29:53 Ivan Carvalho ***@***.***>:

e-eight:
Sorry for the delay, it was a busy few weeks. But removed the weights options. So now it should behave in the same way as the barbell graph generator from networkx.

IvanIsCoding (approved):
LGTM, I think you need to rebase to solve a small merge conflict (we added a new generator recently). But apart from that, this is good to go.

e-eight:
I rebased it on the main branch. Not sure why it is still showing conflicting files.

PR #475, opened 2021-10-25
Add Left-Right Planarity test.

georgios-ts:
This commit implements a new function `is_planar` that
checks if an undirected graph can be drawn in a plane without any edge
intersections. The planarity check algorithm is based on the
Left-Right Planarity Test.

Reference: http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.217.9208


IvanIsCoding:
After #497, you can use the new generator to make some test cases. Table 2 of https://www.sciencedirect.com/science/article/pii/S0166218X08000371 has the crossing numbers of the Petersen graphs, zeros in the table are planar and the rest aren't

mtreinish:
@georgios-ts is there any update on this? It would be good to get this rebased as @enavarro51 is starting to look at adding a `planar_layout()` method and having some functionality around lr planarity in retworkx-core would be a good starting off point for that

georgios-ts:
@mtreinish This should be ready for review now.

retworkx-core/src/planar/lr_planar.rs
```diff
@@ -0,0 +1,699 @@
...
+        Interval {
+            inner: Some((low, high)),
+        }
+    }
+
+    fn is_empty(&self) -> bool {
+        self.inner.is_none()
+    }
+
+    fn unwrap(self) -> (T, T) {
```
enavarro51, line 99: Is there a reason this is not `&self`?

retworkx-core/src/planar/lr_planar.rs
```diff
@@ -0,0 +1,699 @@
...
+    Finish(N),
+    TreeEdge(N, N),
+    BackEdge(N, N),
+    FinishEdge(N, N),
+}
+
+// An error: graph is *not* planar.
+struct NonPlanar {}
+
+struct LRState<G: GraphBase>
```
enavarro51, line 213: Is there a reason for the use of `GraphBase` and `GraphProp` throughout instead of `StableGraph`? And is there a Python interface to `is_planar`?

retworkx-core/src/planar/lr_planar.rs
```diff
@@ -0,0 +1,699 @@
...
+            stack_emarker: HashMap::with_capacity(num_edges),
+            eref: HashMap::with_capacity(num_edges),
+            side: graph
+                .edge_references()
+                .map(|e| ((e.source(), e.target()), Sign::Plus))
+                .collect(),
+        }
+    }
+
+    fn lr_orientation_visitor(&mut self, event: DfsEvent<G::NodeId, &G::EdgeWeight>) {
```
enavarro51, line 270: It appears that you are following a different structure from `networkx`, using an event handler rather than a more procedural approach. Is there a reason for that?

tests/graph/test_planar.py
```diff
@@ -0,0 +1,283 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import unittest
+
+import itertools
+import retworkx
+
+
+class TestPlanarGraph(unittest.TestCase):
```
enavarro51, line 19: Is there a reason tests are duplicated in rust and python? Is that the standard approach for `retworkx-core` code?

enavarro51 (commented):
@georgios-ts. It was suggested by @mtreinish that I review this PR more as a process of learning. To start with some high-level questions
- `is_planar` is in `retworkx-core`. Is there a reason it's there instead of `retworkx`? It seems this would be a function accessed externally the same as `planar_layout` might be.
- The `networkx` approach creates the embedding at the same time the planarity test is being done. Is there a reason you only did the planarity test? As in, do you think the creation of the embedding should be separate from the planarity test, or incorporated within it?
- My current plan is to create a `PlanarEmbedding` struct and associated methods mirroring what is done in `networkx` and then when your PR is merged, to add the hooks to create the embedding within `lr_planar.rs` code. This would be followed by writing the `combinatorial_embedding_to_pos` function. Does this seem like a reasonable approach?
- Thanks.

retworkx-core/src/planar/lr_planar.rs
georgios-ts: Yeah, this function mirrors and calls [`Option::unwrap`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap) that consumes `self`.

retworkx-core/src/planar/lr_planar.rs
georgios-ts: Our goal in retworkx-core is to provide functionality that can work for any graph data structure in the external petgraph library that implement the required traits. See https://docs.rs/petgraph/0.6.0/petgraph/visit/index.html#graph-trait-implementations.

The python interface is here:
https://github.com/Qiskit/retworkx/blob/b6f4a9212afe909d943aaaceba2b5d6230c8b437/src/planar/mod.rs#L33-L37

retworkx-core/src/planar/lr_planar.rs
georgios-ts: It's mostly a personal preference. It's done since this PR makes use of [`depth_first_search`](https://github.com/Qiskit/retworkx/blob/main/retworkx-core/src/traversal/dfs_visit.rs#L159) that pre-existed and follows this approach.
In general I think it improves:
- Code readability. You develop/review the graph traversal and the custom logic inserted through the event handlers separately.
- Code reusability. We can re-use `lr_visit_ordered_dfs_tree` and implement a new `fn lr_embedding_visitor(&mut self, event: DfsEvent<G::NodeId, &G::EdgeWeight>)` event handler that will help in creating the combinatorial embedding.

tests/graph/test_planar.py
georgios-ts: It's not standard but it's something that we'd like to do in the future (see #587) since `retworkx-core` is a separate Rust crate and it deserves its own testing. Here I originally wrote the tests in rust since it's faster to built and run the rust tests and it saves some development time.

georgios-ts:
@enavarro51 Thanks for taking the time to look into this.

- It's in `retworkx-core` so we can expose it in our public rust API but there is also a python interface in `retworkx` that uses internally the implementation written in  `retworkx-core`.
- As you said, to create the embedding we need a `PlanarEmbedding` structure. This was additional work and I didn't want to overwhelm this PR since it was already quite long. But yeah, I do believe that the creation of the embedding should be separate from the public `is_planar` test. IMO, we should have two public functions: a test that returns true/false for planar/non-planar graphs and a method that creates a planar layout. All other function/structures (e.g  `PlanarEmbedding`) should be private.
- It sounds like a good plan. After implementing `PlanarEmbedding` you can leverage the [`lr_visit_ordered_dfs_tree`](https://github.com/Qiskit/retworkx/blob/b6f4a9212afe909d943aaaceba2b5d6230c8b437/retworkx-core/src/planar/lr_planar.rs#L583) function to implement Algorithm 6 in the linked paper (the NetworkX equivalent https://github.com/networkx/networkx/blob/main/networkx/algorithms/planarity.py#L650).

If you have any other questions feel free to reach out.

itoko (approved):
Thanks, @georgios-ts  (and good QAs with @enavarro51, the previous comments helped my review a lot.) This PR looks good to me. I mainly reviewed it from an algorithmic point of view especially comparing with the networkx implementation. It implements a planarity check algorithm using the visitor pattern in two depth-first searches nicely. I think that is making the code quite readable and also useful for another PRs such as #645. I've left a few minor comments inline but I think none of them would prevent from merging this after resolving conflicts.

retworkx-core/src/planar/lr_planar.rs
```diff
@@ -0,0 +1,699 @@
...
+                    // update lowpoints of parent edge.
+                    if let Some(e_par) = self.eparent.get(&v) {
+                        match self.lowpt[&ei].cmp(&self.lowpt[e_par]) {
+                            Ordering::Less => {
+                                self.lowpt_2
+                                    .insert(*e_par, self.lowpt[e_par].min(self.lowpt_2[&ei]));
+                                self.lowpt.insert(*e_par, self.lowpt[&ei]);
+                            }
+                            Ordering::Greater => {
+                                insert_or_min(&mut self.lowpt_2, *e_par, self.lowpt[&ei]);
```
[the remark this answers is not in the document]
georgios-ts: Yes it's equivalent but not necessarily faster since it evaluates the hash of `e_par` twice (ignoring any compiler optimization): 
1. retrieve the value `lowpt_2[e_par]`, and 
2. insert the updated value `lowpt_2.insert(*e_par, ..)`

But you've got a point: `e_par` is already contained in `lowpt_2` and hence the `or_insert` path in `insert_or_min` function can be omitted. I've done this in 0ef2844

IvanIsCoding (approved):
LGTM, I think having readable code for a long algorithm implementation like this one is a huge plus.

I am also confident on the implementation because of the tests. I think we cover a good amount of them, and further in #645 we have planar layout tests building on top of this so I believe this is good to merge

PR #484, opened 2021-11-08
Add colorbar to figure 1 example

mtreinish:
This commit adds a color bar to indicate the meaning of the color map
for the example betweenness centrality graph in Figure 1.


IvanIsCoding:
LGTM. I like the colorbar, I will tweak the proportion of the image in the paper and merge it. The new resolution is different, so the old proportions make the image look small

mtreinish:
While you're updating the formatting, maybe we should add to the caption that it's the normalized betweenness centrality too to explain what the visualization is. I realize it's there to just showcase we have visualization support but it's lacking context on what it is and some people commented to me they were wondering about it

IvanIsCoding:
> While you're updating the formatting, maybe we should add to the caption that it's the normalized betweenness centrality too to explain what the visualization is. I realize it's there to just showcase we have visualization support but it's lacking context on what it is and some people commented to me they were wondering about it

I agree, I will change the caption to be more descriptive. I am also thinking about adding more graphs to the image to focus more on examples than on that single graph itself.

mtreinish:
superseded by #485

PR #485, opened 2021-11-10
Improve Figure 1 and its description

IvanIsCoding:
Alternative to #484 to improve Figure 1

Instead of adding a color map, this commit adds more graphs to the example image. A better caption is also included, now specifying the types of graphs that are shown


paper.md
```diff
@@ -56,7 +56,7 @@ _retworkx_ operates on weights with callbacks. Callbacks are functions that take
 
 A defining characteristic of _retworkx_ graphs is that each node maps to a non-negative integer node index, and similarly, each edge maps to an edge index. Those indices uniquely determine nodes and edges in the graph. Moreover, the indices provide a clear separation between the underlying graph structure and the data associated with weights.
 
-![Visualization of a graph data structure made with the **`retworkx.visualization`** module.\label{fig:graphexample}](paper_img/example_graph.png){ width=40% height=40% }
+![A Petersen graph, a heavy-hexagon graph, and a binomial tree graph visualized with the **`retworkx.visualization`** module.\label{fig:graphexample}](paper_img/example_graph.png){ width=100% height==100% }
```
mtreinish, line 59: The middle graph is not actually a heavy hexagon graph, it's a hexagonal lattice. Also I still think we should have a color bar on it if we're using betwenness centrality to color the nodes and we should mention that it's colored by betwenness centrality.

IvanIsCoding: Thanks for catching the lattice part

IvanIsCoding:
I think coloring the graph with multiple colors was not the best choice, I simplified it now such that we will not need the color bar. I think the three examples are good to showcase our visualization module

mtreinish (approved):

PR #487, opened 2021-11-15
Add `contract_nodes` for graphs

### LINKED ISSUE #476: Add method for replacing block with node in graph

In https://github.com/Qiskit/qiskit-terra/pull/7185 terra is adding a method that replaces a block of graph nodes (as found with [`collect_bicolor_runs()`](https://qiskit.org/documentation/retworkx/stubs/retworkx.collect_bicolor_runs.html?highlight=collect_#retworkx.collect_bicolor_runs)) with a single node. The idea is that the block is consolidated to an equivalent single node. That PR adds a function to terra's dag class, `replace_block_with_op()` that iterates over all the predecessors and successors outside the block and adds equivalent edge to and from the replacement node. This should be easy to implement in retworkx natively as a method on `PyDiGraph` and `PyGraph` as it's primarily a function of the graph and doesn't involve any python interaction (except for the creation of the replacement node) so we should add a function for this to retworkx. The basic idea would be to add something like: `fn replace_block_with_node(&mut self, py: Python, block_nodes: Vec<usize>, node: usize)` to the `pymethods` of the graph classes which will do the copying of edges to and from the `block_nodes` to `node` and remove `block_nodes` from the graph.

kevinhartman:
Adds a new method `contract_nodes` to `PyDiGraph` and `PyGraph` to support replacing a set of nodes with an equivalent node.

Resolves #476


src/digraph.rs
```diff
@@ -2265,6 +2266,150 @@ impl PyDiGraph {
...
+            let mut visit_next = VecDeque::new();
+
+            // Start with successors of `nodes` that aren't in `nodes` itself.
+            for node in nodes {
+                for edge in graph.edges(*node) {
+                    if !nodes.contains(&edge.target()) {
+                        visit_next.push_back(edge.target());
+                    }
+                }
+            }
```
mtreinish, line 2307: You can replace this with a single iterator chain like:

```suggestion
            let mut visit_next: VecDeque<NodeIndex> = nodes
                .iter()
                .map(|n| graph.edges(*n))
                .flatten()
                .filter_map(|edge| {
                    let target_node = edge.target();
                    if !nodes.contains(&target_node) {
                        Some(target_node)
                    } else {
                        None
                    }
                })
                .collect();
```

This has the advantage of allocating at creation time so it should be slightly more efficient

src/digraph.rs
```diff
@@ -2265,6 +2266,150 @@ impl PyDiGraph {
...
+    ///     substitution would introduce cycle(s).
+    /// :raises ValueError: ``to_replace`` is empty or ``node``
+    ///     has existing connections.
+    #[pyo3(
+        text_signature = "(self, to_replace, node, /, force_check_cycle=None)"
+    )]
+    fn substitute_nodes_with_node(
+        &mut self,
+        py: Python,
+        to_replace: HashSet<usize>,
```
mtreinish, line 2289: I'm wondering if we want this to be `Vec<usize>` instead. Python users tend to use lists (it also enables users to pass any [sequence type](https://pyo3.rs/v0.15.0/conversions/tables.html#argument-types) not just a list) more than sets and `Vec<T>` iterates faster than a set. Since the first thing we do with `to_replace` is loop over it check each index is valid and convert it to a `HashSet<NodeIndex>` I think making the argument a vec will work better here

src/digraph.rs
```diff
@@ -2265,6 +2266,150 @@ impl PyDiGraph {
...
+        if (force_check_cycle == Some(true) || self.check_cycle)
+            && !can_contract(&self.graph, &indices_to_remove)
+        {
+            return Err(DAGWouldCycle::new_err(
+                "Substitution would create cycle(s)",
+            ));
+        }
+
+        let mut pred_to_weights = HashMap::new();
+        let mut succ_to_weights = HashMap::new();
```
mtreinish, line 2373: Wow, I'm surprised/impressed that rustc could implicitly tell this was a `HashMap<NodeIndex, Vec<PyObject>`

mtreinish (commented):
Thanks for starting this, it's a great start! realize this isn't ready yet I just had some quick inline comments on some things I caught from a quick scan.

src/digraph.rs
```diff
@@ -2265,6 +2266,150 @@ impl PyDiGraph {
...
+                    if nodes.contains(&edge.target()) {
+                        // we found a path back to `nodes`
+                        return false;
+                    }
+
+                    visit_next.push_back(edge.target());
+                }
+
+                visited.visit(decendant);
+            }
```
georgios-ts, line 2326: This works fine here, as an alternative we could use [`petgraph::visit::Dfs`](https://docs.rs/petgraph/0.6.0/petgraph/visit/struct.Dfs.html)
```suggestion
            let mut dfs = Dfs::from_parts(visit_next, visited);
            while let Some(node) = dfs.next(graph) {
                if nodes.contains(&node) {
                    // we found a path back to `nodes`
                    return false;
                }
            }
```

src/digraph.rs
```diff
@@ -2265,6 +2266,150 @@ impl PyDiGraph {
...
+        }
+
+        let node_index = NodeIndex::new(node);
+        assert_node_exists!(node_index);
+
+        if self.graph.neighbors_undirected(node_index).next().is_some() {
+            return Err(PyValueError::new_err(
+                "replacement node must not have connections",
+            ));
+        }
```
georgios-ts, line 2353: If `node` is an existing node, we should allow it to have edges and adjust the `can_contract` cycle check. Another option would be to take as input a weight `PyObject`, insert a new node in the graph with the given weight and return back the new index.

src/digraph.rs
```diff
@@ -2265,6 +2266,150 @@ impl PyDiGraph {
...
+
+        let mut indices_to_remove: HashSet<NodeIndex> =
+            HashSet::with_capacity(to_replace.capacity());
+
+        for node in to_replace {
+            let index = NodeIndex::new(node);
+            assert_node_exists!(index);
+            indices_to_remove.insert(index);
+        }
```
georgios-ts, line 2362: I don't feel it's necessary to raise an error here, we just ignore it if an index does not belong to the graph.
```suggestion
        let indices_to_remove: HashSet<NodeIndex> = to_replace
            .into_iter()
            .filter_map(|node| {
                let index = NodeIndex::new(node);
                if self.graph.contains_node(index) {
                    Some(index)
                } else {
                    None
                }
            })
            .collect();
```

src/digraph.rs
```diff
@@ -2265,6 +2266,150 @@ impl PyDiGraph {
...
+
+        if (force_check_cycle == Some(true) || self.check_cycle)
+            && !can_contract(&self.graph, &indices_to_remove)
+        {
+            return Err(DAGWouldCycle::new_err(
+                "Substitution would create cycle(s)",
+            ));
+        }
+
+        let mut pred_to_weights = HashMap::new();
```
georgios-ts, line 2372: It might be faster to use a `Vec` here with tuples `(pred, weight)`.

georgios-ts (commented):
Thanks for taking this! It already looks awesome!

src/digraph.rs
```diff
@@ -2265,6 +2266,150 @@ impl PyDiGraph {
...
+            indices_to_remove.insert(index);
+        }
+
+        if (force_check_cycle == Some(true) || self.check_cycle)
+            && !can_contract(&self.graph, &indices_to_remove)
+        {
+            return Err(DAGWouldCycle::new_err(
+                "Substitution would create cycle(s)",
+            ));
+        }
```
mtreinish, line 2370: I'm not sure this is quite right actually, shouldn't this be something like:

```suggestion
        if (force_check_cycle == Some(true) || (self.check_cycle && force_check_cycle != Some(false))
            && !can_contract(&self.graph, &indices_to_remove)
        {
            return Err(DAGWouldCycle::new_err(
                "Substitution would create cycle(s)",
            ));
        }
```

because if the user explicitly sets `force_check_cycle` to `false` we want that to override the value of `self.check_cycle`.

kevinhartman: Originally, I was using something like `check_cycle.unwrap_or(self.check_cycle)` to get exactly that behavior, but `_add_edge` will actually still find cycles when connecting in the new node, and that is connected to `self.check_cycle`, so it'd be possible to end up in a bad state with a modified DAG. The simplest workaround I figured would be to rename the `check_cycle` param to `force_check_cycle` so the user can choose to opt-in if cycle check is disabled at the class level, but cannot opt out if it's not.

src/digraph.rs
kevinhartman: Beautiful! TIL rustc has very similar collection APIs to Scala.

src/digraph.rs
mtreinish: Ah ok that makes sense, I didn't think of that. If we can't introduce a cycle if `self.cycle_check` is `true` regardless of this flag then I think we only need a `bool` here and not `Option<bool>`. Since we only are controlling whether we override this on the `self.cycle_check == false` branch. The `Option<bool>` is only needed if we wanted to override the both the true and false case.

src/digraph.rs
kevinhartman: This can potentially make the code a lot more concise too, so I'll make the switch.

src/digraph.rs
kevinhartman: Yes! It seems like if the usages aren't conflicting, type inference just works. I'm not sure if it's better to be explicit or not. In this case, I dropped the explicit type arguments because `rustfmt` was otherwise wrapping the lines in an ugly way. Let me know if you have a preference.

src/digraph.rs
kevinhartman: Would you also recommend then skipping check of empty `to_replace`? We could have that as a no-op, and just note the behavior in the docstring.

src/digraph.rs
kevinhartman: Wow, very cool!

src/digraph.rs
```diff
@@ -2265,6 +2266,150 @@ impl PyDiGraph {
...
+                    }
+
+                    visit_next.push_back(edge.target());
+                }
+
+                visited.visit(decendant);
+            }
+
+            true
+        }
```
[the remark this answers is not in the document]
kevinhartman: Sounds good. I was assuming closures would have overhead whereas perhaps function would not, but it looks like closures like this one are 0 overhead in Rust.

src/digraph.rs
mtreinish: It's fine to leave it out if the rust compiler can figure it out. I agree it's more concise and easy to read this way, I was just surprised that it could since the value is set inside the loop via an entry.

src/digraph.rs
mtreinish: I think we should probably leave this as a separate pre-check so that if a cycle were to be introduced we don't potentially add edges before we encounter the cycle. Doing the cycle checking explicitly up front means we never end up with a partial modified graph.

src/digraph.rs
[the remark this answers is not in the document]
mtreinish: Yeah a closure doesn't have any runtime overhead. It does have a compile time cost I think but that's a small cost especially compared to having slightly cleaner code

src/digraph.rs
```diff
@@ -2265,6 +2266,150 @@ impl PyDiGraph {
...
+        py: Python,
+        to_replace: HashSet<usize>,
+        node: usize,
+        force_check_cycle: Option<bool>,
+    ) -> PyResult<()> {
+        fn can_contract(
+            graph: &StablePyGraph<Directed>,
+            nodes: &HashSet<NodeIndex>,
+        ) -> bool {
+            let mut visited = graph.visit_map();
```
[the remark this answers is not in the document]
kevinhartman: It looks to me that `graph.visit_map()` returns a `FixedBitSet`, which stores bits across 32-bit blocks (e.g. with <=32 nodes, we'd have just a single element in the underlying `Vec<u32>`).

https://docs.rs/fixedbitset/0.2.0/src/fixedbitset/lib.rs.html#61

src/digraph.rs
```diff
@@ -2265,6 +2266,150 @@ impl PyDiGraph {
...
+            let mut visit_next = VecDeque::new();
```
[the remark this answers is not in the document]
kevinhartman: Done in 0500e13.

src/digraph.rs
kevinhartman: Okay by me!

src/digraph.rs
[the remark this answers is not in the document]
kevinhartman: Done in 0500e13.

src/digraph.rs
kevinhartman: I was able to make this even _slightly_ nicer by using `flat_map` instead of `map` followed by `flatten` :).

src/digraph.rs
georgios-ts: Yeah, I agree, cycle-checking should be done if requested as a pre-check before we modify the graph. But at the same time we should avoid doing cycle-checks again in `_add_edge` if for example `force_check_cycle == true && self.check_cycle == true`.

kevinhartman: I think what @georgios-ts is suggesting is that we add a `force_cycle_check` parameter to `_add_edge` as well so that we can skip the check if `substitute_nodes_with_node` is called with `force_cycle_check=false`. We would still do the check up front in `substitute_nodes_with_node` if the caller specified `force_cycle_check=true`.

In that case, I would also rename `force_cycle_check` back to `cycle_check`, since it then gives the user the power to turn off cycle checking for just this call, irrespective of the class setting.

src/digraph.rs
georgios-ts: Yeah but this might actually depend on how we address this comment https://github.com/Qiskit/retworkx/pull/487#discussion_r750309097, e.g if this function is expected to add a new node, it feels better to keep the empty-check

src/digraph.rs
IvanIsCoding: I'll agree with Georgios and say that taking an optional weight and returning a new node index seems like a better design choice. Creating an empty node with `add_node` and then calling this function is not as convenient.

src/digraph.rs
kevinhartman: Done in c8699ae.

src/digraph.rs
kevinhartman: Agreed. Done in c8699ae.

src/digraph.rs
kevinhartman: Hmm. I guess if `to_replace` is empty, this function behaves like `add_node`, assuming we keep the changes introduced by c8699ae. This seems reasonable to me, and allows us to skip checks. If we keep the empty check, then we also probably need to check that at least one node in `to_replace` exists, otherwise I'd think the method should also fail.

src/digraph.rs
kevinhartman: Done in 61cd6ef. With that change, I'm also including both endpoints in the tuple to make the code a bit simpler, though I'm not sure if this will add overhead or if the compiler is smart enough to do something like loop invariant code motion here.

src/digraph.rs
kevinhartman: Added behavior note in f734242. Happy to revise if you'd prefer this to work differently.

src/digraph.rs
kevinhartman: Done in 4d31566. I split `_add_edge` into two methods, moving the actual adding to `_add_edge_no_checks`. Then, we always call `_add_edge_no_checks` from `substitute_nodes_with_node` since checks are done up front, and we can avoid redoing them this way. I considered adding a parameter to `_add_edge` instead, but there's no function overloading or optional parameters in Rust, so I would have had to litter all ~18 callsites with `None`s.

src/digraph.rs
```diff
@@ -179,6 +179,24 @@ impl NodeCount for PyDiGraph {
 
 // Rust side only PyDiGraph methods
 impl PyDiGraph {
+    fn _add_edge_no_checks(
```
georgios-ts, line 182:
```suggestion
    fn add_edge_no_cycle_check(
```
since we're still doing a check if a graph is not a multigraph and in Rust the underscore usually means unused variable/function.

src/digraph.rs
```diff
@@ -179,6 +179,24 @@ impl NodeCount for PyDiGraph {
...
+        &mut self,
+        p_index: NodeIndex,
+        c_index: NodeIndex,
+        edge: PyObject,
+    ) -> PyResult<usize> {
```
georgios-ts, line 187:
```suggestion
    ) -> usize {
```
This function does not need to return a `Result` since it never actually fails.

src/digraph.rs
```diff
@@ -179,6 +179,24 @@ impl NodeCount for PyDiGraph {
...
+        if !self.multigraph {
+            let exists = self.graph.find_edge(p_index, c_index);
+            if let Some(index) = exists {
+                let edge_weight = self.graph.edge_weight_mut(index).unwrap();
+                *edge_weight = edge;
+                return Ok(index.index());
```
georgios-ts, line 193:
```suggestion
                return index.index();
```

src/digraph.rs
```diff
@@ -179,6 +179,24 @@ impl NodeCount for PyDiGraph {
...
+            }
+        }
+        let edge = self.graph.add_edge(p_index, c_index, edge);
+        Ok(edge.index())
```
georgios-ts, line 197:
```suggestion
        edge.index()
```

src/digraph.rs
```diff
@@ -2265,6 +2274,120 @@ impl PyDiGraph {
...
+    /// :param bool check_cycle: If set to ``True``, validates
+    ///     that the substitution will not introduce cycles before
+    ///     modifying the graph. If set to ``False``, validation is
+    ///     skipped. If not provided, inherits the value
+    ///     of ``check_cycle`` from this instance of
+    ///     :class:`retworkx.PyDiGraph`.
+    /// :raises DAGWouldCycle: The cycle check is enabled and the
+    ///     substitution would introduce cycle(s).
+    #[pyo3(
+        text_signature = "(self, to_replace, obj, /, force_check_cycle=None)"
```
georgios-ts, line 2293:
```suggestion
        text_signature = "(self, to_replace, obj, /, check_cycle=None)"
```

src/digraph.rs
```diff
@@ -2265,6 +2274,120 @@ impl PyDiGraph {
...
+
+        for index in indices_to_remove {
+            self.graph.remove_node(index);
+        }
+
+        for (start, end, weight) in edges {
+            self._add_edge_no_checks(start, end, weight)?;
+        }
+
+        Ok(())
```
georgios-ts, line 2388:
```suggestion
        Ok(node_index)
```
since we should return the index of the new node.

georgios-ts (commented):
Thanks for all the updates so far!

src/digraph.rs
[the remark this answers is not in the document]
georgios-ts: Yeah, you're right but still asymptotically the size will be roughly equal to the number of nodes. In any case, let's work with a `FixedBitSet`.

src/digraph.rs
kevinhartman: Fixed in 6b0369c.

src/digraph.rs
kevinhartman: Done in 31be2d0.

kevinhartman:
I discussed this offline with @mtreinish, but I realized that there's also a potential issue with non-multigraphs if there are edges out of `to_remove` to the same target nodes. Similar to `to_undirected`, I'll be adding an optional weight combiner function parameter to support this.

kevinhartman:
> there's also a potential issue with non-multigraphs if there are edges out of to_remove to the same target nodes.

This can also happen for incoming edges if multiple edges coming into `to_remove` originate from the same node. Will update to handle this as well.

src/digraph.rs
```diff
@@ -2265,6 +2274,159 @@ impl PyDiGraph {
...
+        // Remove nodes that will be replaced.
+        for index in indices_to_remove {
+            self.graph.remove_node(index);
+        }
+
+        // If `weight_combo_fn` was specified, merge edges according
+        // to that function, even if this is a multigraph. If unspecified,
+        // defer parallel edge handling to `add_edge_no_cycle_check`.
+        if let Some(merge_fn) = weight_combo_fn {
+            macro_rules! merge_parallel_edges {
```
georgios-ts, line 2396: Is there a reason for this to be a macro instead of a normal function? I'd actually go one step further and make it a generic function with signature like:
```rust
use std::hash::Hash;

fn merge_duplicates<K, V, F, E>(
    xs: Vec<(K, V)>,
    mut merge_fn: F,
) -> Result<Vec<(K, V)>, E>
where
    K: Hash + Eq,
    F: FnMut(&V, &V) -> Result<V, E>,
{
   ....
}
```

src/digraph.rs
```diff
@@ -2265,6 +2274,159 @@ impl PyDiGraph {
...
+                        match node_to_weight.remove(&node) {
+                            Some(existing_weight) => {
+                                let combined_weight = merge_fn
+                                    .call1(py, (weight, existing_weight))?;
+                                node_to_weight.insert(node, combined_weight);
+                            }
+                            None => {
+                                node_to_weight.insert(node, weight);
+                            }
+                        }
```
georgios-ts, line 2409: I think the following should be slightly faster without the need to first remove and then re-insert a node. I see in previous commits you did something similar with `and_modify/or_insert`, was there a reason for a change?

```suggestion
                        match node_to_weight.entry(node) {
                            Occupied(entry) => {
                                *entry.into_mut() = merge_fn
                                    .call1(py, (weight, entry.get()))?;
                            }
                            Vacant(entry) => {
                                entry.insert(weight);
                            }
                        }
```

georgios-ts (commented):
One extra comment is that it'd be nice for this function to be deterministic. Right now, invocations with the same input parameters may lead to a graph with different edge indices or in case of a not muligraph (with `weight_combo_fn = None`), the weight of a parallel edge that we finally pick may differ. As a solution, we could make `indices_to_remove` a `Vec` and remove duplicates like:
```rust
let mut indices_to_remove: HashSet<NodeIndex> =
            to_replace.iter().map(|&n| NodeIndex::new(n)).collect();
indices_to_remove.sort()
indices_to_remove.dedup()
```
and `node_to_weight` a [`DictMap`](https://github.com/Qiskit/retworkx/blob/main/retworkx-core/src/dictmap.rs) that has a consistent iteration order. I think this should remove the non-determinism.

But i don't feel very strongly about it, so i'd still be happy without any changes.

src/digraph.rs
```diff
@@ -2265,6 +2274,159 @@ impl PyDiGraph {
...
+                    // we found a path back to `nodes`
+                    return false;
+                }
+            }
+
+            true
+        };
+
+        let mut indices_to_remove: HashSet<NodeIndex> =
+            to_replace.iter().map(|&n| NodeIndex::new(n)).collect();
```
IvanIsCoding, line 2344:
```suggestion
        let mut indices_to_remove: indexmap::set::IndexSet<NodeIndex> =
            to_replace.iter().map(|&n| NodeIndex::new(n)).collect();
        indices_to_remove.sort()
```

Makes the code deterministic addressing https://github.com/Qiskit/retworkx/pull/487#pullrequestreview-811815333 without changing much of the code. `indexmap` is already a dependency we use extensively, might as well use it in this case we need determinism.

src/digraph.rs
kevinhartman: Ah yeah, something like this could be nice! I got rid of the `and_modify` stuff because it seemed like I'd then need to store the `PyObject` as a `PyResult<PyObject>` in the map, since we couldn't use the `?` inside the closure. Your suggestion doesn't have that problem :). I used `remove` because it moves the value out of the map, which meant I could avoid `.clone_ref(py)` on the value, but I _think_ what you have should work too since you've got a non-mutable ref of `entry`, it seems.

src/digraph.rs
kevinhartman: I'm not strongly attached to it being a macro, but I went with this over a closure to avoid having to specify types and do error propagation (I liked the idea of the `?` operator just working in the context of where the macro is expanded). Are you proposing that a generic function with this code might be valuable as a helper method at the class level?

src/digraph.rs
kevinhartman: That could work! Thoughts on using `collections::BTreeSet`? I've seen discussions of performance actually being better in some cases (better cache locality), and we can get an ordering without an explicit sort:

```rs
let mut indices_to_remove: BTreeSet<NodeIndex> =
  to_replace.into_iter().map(|n| NodeIndex::new(n)).collect();
```

I think we also need to fix `node_to_weight` as @georgios-ts mentioned as well to be deterministic. I am inclined to go for a `BTreeMap` in that case as well.

src/digraph.rs
georgios-ts: Yeah exactly. We can place it in `src/lib.rs`  and re-use it in the sister method for `PyGraph`.

kevinhartman: Done in 02f491a. Out of curiosity, is there a reason / best practice for making `F` be `FnMut`, rather than `Fn`?

src/digraph.rs
kevinhartman: Done in 3278664.

src/digraph.rs
kevinhartman: Done in a45fd8b. I went with `indexmap`, but omitted the sort. We should still have determinism from the caller's perspective, but IMO there's not much reason to ensure that different orderings of `to_remove` yield the same result.

src/digraph.rs
georgios-ts: It's just that `Fn` is a subtrait of `FnMut` so anything that implements `Fn` automatically implements `FnMut`. See https://doc.rust-lang.org/std/ops/trait.Fn.html

src/digraph.rs
```diff
@@ -2265,6 +2278,142 @@ impl PyDiGraph {
...
+                if nodes.contains(&node) {
+                    // we found a path back to `nodes`
+                    return false;
+                }
+            }
+
+            true
+        };
+
+        let mut indices_to_remove: IndexSet<NodeIndex> =
```
mtreinish, line 2347: I think we can make this a bit faster by using ahash as our hasher (which is what is the default in `hashbrown` and what we use for `DictMap` in the custom return types. It should be a bit faster then the default hasher for `IndexSet`. It is worth confirming this with a quick benchmark though.

```suggestion
        let can_contract = |nodes: &IndexSet<NodeIndex, ahash::RandomState>| {
            // Start with successors of `nodes` that aren't in `nodes` itself.
            let visit_next: Vec<NodeIndex> = nodes
                .iter()
                .flat_map(|n| self.graph.edges(*n))
                .filter_map(|edge| {
                    let target_node = edge.target();
                    if !nodes.contains(&target_node) {
                        Some(target_node)
                    } else {
                        None
                    }
                })
                .collect();

            // Now, if we can reach any of `nodes`, there exists a path from `nodes`
            // back to `nodes` of length > 1, meaning contraction is disallowed.
            let mut dfs = Dfs::from_parts(visit_next, self.graph.visit_map());
            while let Some(node) = dfs.next(&self.graph) {
                if nodes.contains(&node) {
                    // we found a path back to `nodes`
                    return false;
                }
            }

            true
        };

        let mut indices_to_remove: IndexSet<NodeIndex, ahash::RandomState> =
```

kevinhartman: Is it worth introducing a wrapper type that does this for sets like `DictMap`?

mtreinish: It doesn't hurt, but I'm not sure it's really necessary with only one user for right now. If we start using it more it'd be an easier call then.

We did it for `DictMap` because we were using `IndexMap` for all the custom returns to have a deterministic iteration order for everything with a custom return type (which was a lot of places).

.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
...
+        "-v",
+        "-s",
+        ".",
+        "-p",
+        "test_*.py"
+    ],
+    "python.testing.pytestEnabled": false,
+    "python.testing.unittestEnabled": true,
+    "python.testing.cwd": "/Users/kevin/Documents/Development/retworkx/tests"
+}
```
IvanIsCoding, line 12: This for sure shouldn't be in the PR, maybe we should add `.vscode/settings.json` to `.gitignore`

kevinhartman: Yikes. We could, though this is usually easy enough to avoid.

src/digraph.rs
```diff
@@ -2265,6 +2278,141 @@ impl PyDiGraph {
...
+            return Err(DAGWouldCycle::new_err(
+                "Contraction would create cycle(s)",
+            ));
+        }
+
+        // Create new node.
+        let node_index = self.graph.add_node(obj);
+
+        // Sanitize new node index from user input.
+        indices_to_remove.remove(&node_index);
```
mtreinish, line 2361: Is this coming up somewhere? At this point I would never expect this to actually do anything since `node_index` was just created it would be kind of weird to have user input that includes it.

src/graph.rs
```diff
@@ -1480,6 +1481,83 @@ impl PyGraph {
         Ok(out_dict.into())
     }
 
+    /// Substitute a set of nodes with a single new node.
+    ///
+    /// .. note::
+    ///     The new node is always made to be the target in any edge
+    ///     representation that includes it. For example, the new node
+    ///     will always appear as the target of its edges as returned by
+    ///     :meth:`~retworkx.PyGraph.edge_list`.
```
mtreinish, line 1490: `PyGraph` is undirected, what do you mean here? Just that the tuple `(u, v)` petgraph stores under the covers will always have `node` as `v` so when you call `edge_list()` which just returns those tuples you'll see this? If so, I'm not sure if it's worth documenting that, I'm mostly concerned about potentially having people depend on the behavior as it's mostly a implementation detail not something we want to commit to.

src/lib.rs
```diff
@@ -187,6 +191,28 @@ fn find_node_by_weight<Ty: EdgeType>(
...
 
+fn merge_duplicates<K, V, F, E>(
+    xs: Vec<(K, V)>,
+    mut merge_fn: F,
+) -> Result<Vec<(K, V)>, E>
+where
+    K: Hash + Eq,
+    F: FnMut(&V, &V) -> Result<V, E>,
+{
+    let mut kvs = DictMap::with_capacity(xs.len());
```
mtreinish, line 202: Is there a reason to use a `DictMap` here as opposed to using just a `HashMap`, like do we need to preserve order? Normally `HashMap` is a bit faster.

mtreinish (approved):
Overall this LGTM, just a few small questions/comments inline but nothing critical. Great work, thanks for doing this.

src/digraph.rs
mtreinish: Looking through the tests I understand now, you're guarding against someone passing in an invalid node that happens to end up being the new node. I don't think this will come up very often, but it should be low overhead as it's just a lookup so it's probably fine.

kevinhartman: Cool. Yeah that's right. We say that this method behaves like `add_node` when the user gives a list of nodes that is either completely empty or doesn't contain any nodes we know about. Adding this defensive removal ensures that is correct even for the case where the `nodes` parameter has an unknown node index that happens to become the index of the created node.

src/graph.rs
kevinhartman: > so when you call edge_list() which just returns those tuples you'll see this?

Yup. My thinking behind documenting it this way was because I would expect some users might be confused to find that the original ordering of edge tuples is not preserved (which is done to make our code cleaner (and possibly faster IIRC)). Perhaps I should change this note to warn that the original ordering is not preserved, but then not specify the implementation-specific order?

src/lib.rs
kevinhartman: I think it's not so much that we need to preserve order here, but more that we want a deterministic order based on the input `Vec` contents. This is because for our use cases, we take the result of this function and add edges to the graph following its iteration order. It seems like `DictMap` is an easy way to get this.

tests/graph/test_contract_nodes.py
```diff
@@ -15,6 +15,27 @@
 import retworkx
 
 
+class UE(tuple):
```
mtreinish, line 18: I'd probably name this `UndirectedEdge` just for clarity. When I saw `UE` at first I wasn't sure what that meant.

mtreinish (commented):
Just one small naming suggestion in the undirected tests and then I think this is good to merge.

tests/graph/test_contract_nodes.py
kevinhartman: Done in 44e0b79!

src/graph.rs
kevinhartman: Done in 1959cf8.

PR #489, opened 2021-11-22
Add dijkstra-search

georgios-ts:
This commit implements a `dijkstra_search` that traverses a graph
using Dijkstra algorithm and provides the ability to insert callback
functions at specified event points. Python users should subclass
`retworkx.visit.DijkstraVisitor` and overwrite the appropriate callback functions.


retworkx-core/src/shortest_path/dijkstra.rs
```diff
@@ -111,63 +108,31 @@ where
...
-                                    *new_vec = node_path;
-                                },
-                            );
-                        }
-                    }
+
+    dijkstra_search(graph, Some(start), edge_cost, |event| {
+        match event {
+            DijkstraEvent::Discover(node, score) => {
+                scores.insert(node, score);
```
IvanIsCoding, line 119: I would rather keep the code duplicated and straightforward for our Dijksta implementation. I do not think we get an advantage with the dijkstra_search rewrite, in fact I found that the code has more "magic" hiding the implementation details. Also, performance could be affected

retworkx-core/src/traversal/dijkstra_visit.rs
```diff
@@ -0,0 +1,324 @@
...
+    F: FnMut(G::EdgeRef) -> Result<K, E>,
+    K: Measure + Copy,
+    V: FnMut(DijkstraEvent<G::NodeId, &G::EdgeWeight, K>) -> C,
+    C: ControlFlow,
+{
+    if visited.is_visited(&start) {
+        return Ok(C::continuing());
+    }
+
+    let mut scores = HashMap::new();
```
IvanIsCoding, line 249: If we use `HashMap` instead of `DictMap`, we are changing the behavior fo the current implementation. For consitency and determinism, I think `DictMap` is a better choice

retworkx/__init__.py
```diff
@@ -1817,3 +1817,70 @@ def _digraph_bfs_search(graph, source, visitor):
...
     return graph_bfs_search(graph, source, visitor)
+
+
+@functools.singledispatch
+def dijkstra_search(graph, source, weight_fn, visitor):
+    """Dijkstra traversal of a graph.
+
+    The pseudo-code for the Dijkstra algorithm is listed below, with the annotated
+    event points, for which the given visitor object will be called with the
+    appropriate method.
```
IvanIsCoding, line 1828: Open to discussion, but having pseudocode that looks more like Python could improve this docstring

IvanIsCoding (commented):
Thanks for this, after #456 and #469 this feels natural and it will be a great addition.

One thing I would like to debate is not replacing our current implementation of Dijkstra, I understand it is shorter but I find it harder to maintain. It feels less straightforward with details hidden on another part of the code.

retworkx-core/src/shortest_path/dijkstra.rs
georgios-ts: Yeah, sure, I can revert the changes. I was just trying to avoid code duplication. However, if using `dijkstra_search` feels a bit unnatural and it's always more straightforward to rewrite Dijkstra's algorithm, I'm starting to doubt about the benefits of this series of PRs. My intention was to provide Python/Rust users or internal code the ability to customize the logic of common graph algorithms without the need to rewrite implementation details. Any performance benefits was a secondary concern. This means if you find this code less readable, which I totally get it, I have second thoughts about these PRs.

retworkx-core/src/traversal/dijkstra_visit.rs
georgios-ts: We do not iterate over `scores` or return it, so `DictMap`/`HashMap` has no difference here.

IvanIsCoding: `scores` is returned to Python on our top-level functions, so there is an implicit iteration when converting to a Python dictionary

retworkx-core/src/shortest_path/dijkstra.rs
IvanIsCoding: The PRs are useful. For end-users in Rust the code is convenient, re-implementing Dijkstra is a pain because they have to worry about priority queues, comparing elements, correctness, and other things. For Python users it has all the conveniences from before, and using the code from the PR is much faster than writting it in Python.

The thing is `retworkx` is a special case because we already have our Dijkstra code written and working. I commented more to debate: should we really change what is already working?

If performance is not affected, keeping the shorter version is fine for me.

retworkx-core/src/traversal/dijkstra_visit.rs
georgios-ts: Here we are in `traversal/dijkstra_visit.rs`. We return `scores` in `shortest_path/dijkstra.rs` where it's still a `DictMap`. https://github.com/Qiskit/retworkx/blob/dbbd9c9549920298c9e167cec37fac6f6fd79a12/retworkx-core/src/shortest_path/dijkstra.rs#L111

retworkx-core/src/shortest_path/dijkstra.rs
georgios-ts: @mtreinish Could you please run https://github.com/mtreinish/retworkx-comparison-benchmarks with this branch to see if there is a performance regression?

mtreinish: I've run the benchmark suite for shortest path, there is a noticeable regression with this PR:

single_source_shortest_path_2
single_source_shortest_path
all_pairs

What I'm not sure of though is how much `main` differs from `0.10.2` though, because as was discussed here the use of `DictMap` hasn't been released yet which is slightly slower than just hashbrown's hashmap.

mtreinish: I ran it with main too so we could compare. There is a regression with main on the single source benchmarks too. I think we might want to revisit our `DictMap` usage everywhere and only use it places where we think iteration order actually matters (I don't think dijkstra's output is one of those, at least for retworkx-core). But the regression with this PR is worse: 

single_source_shortest_path_2
single_source_shortest_path
all_pairs

georgios-ts: @mtreinish Thanks! Wow, didn't expect this huge performance regression! Reverted in 7e0e09f.

docs/source/api.rst
```diff
@@ -58,13 +58,15 @@ Traversal
    retworkx.dfs_edges
    retworkx.bfs_successors
    retworkx.bfs_search
+   retworkx.dijkstra_search
```
IvanIsCoding, line 61: One enhancement we could do on the documentation after #456 and #469 is having a subsection in the  Traversal for the visitors, including Dijkstra's. Their usage is quite different from the other methods, I think they deserve to have its own subsection

retworkx-core/src/traversal/dijkstra_visit.rs
IvanIsCoding: Indeed, we do not return scores here

src/traversal/mod.rs
```diff
@@ -385,3 +389,151 @@ pub fn graph_bfs_search(
...
+) -> PyResult<()> {
+    let starts: Vec<_> = match source {
+        Some(nx) => nx.into_iter().map(NodeIndex::new).collect(),
+        None => graph.graph.node_indices().collect(),
+    };
+
+    dijkstra_search(
+        &graph.graph,
+        starts,
+        |e| weight_callable(py, &weight_fn, e.weight(), 1.0),
```
IvanIsCoding, line 460: We need to raise an error if a negative weight appears as we did in #530

IvanIsCoding (changes requested):

src/traversal/mod.rs
georgios-ts: Done in 29fd43c.

IvanIsCoding (approved):
LGTM, fantastic addition to the 0.11 release

PR #490, opened 2021-11-22
Set max line length to 100 characters

mtreinish:
Previously we had been setting rustfmt and black to use an 80 character
line length. This was primarily because my personal dev workflow is
setup around 80 character-wide windows. However, as the project has a
growing number of contributors my personal archaic taste in dev workflow
should not be subjected upon everyone. This commit bumps the max line
length in rustfmt and black up to 100 which is a more common default
that will work better in most editors and not needlessly wrap lines
everywhere. Most of this diff is just the auto-formatting adjustments
now that there are an additional 20 characters available for them to
use.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: 3.8
-      - run: pip install -U flake8 black==21.5b0
+      - run: pip install -U flake8 black~=22.0
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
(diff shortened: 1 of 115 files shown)


IvanIsCoding:
I like this, I think it is reasonable to bump the line lenght to 100. Our Rust code is shorter and I don't think the readability is affected.

Also, we could also profit from this PR to update black to `21.11b1`

IvanIsCoding (approved):
LGTM

mtreinish:
Given that this will cause a merge conflict for everyone with an open PR, I'd like to merge more of our open reviews first. We can merge this right before we're ready to release 0.11.0 to minimize the pain this causes.

Issue #492, opened 2021-11-23
Revisit DictMap usage in some places


mtreinish:
### Information

- **retworkx version**: main @ https://github.com/Qiskit/retworkx/commit/9e853d42f4b35343adef081e06a90e7aa8ea40a9
- **Python version**: 3.9
- **Rust version**: 1.55.0
- **Operating system**: Linux

### What is the current behavior?

Since the use of `DictMap` was introduced in #386 there has been a regression caused in places that used to use a `HashMap` for the output type. This is probably because of the additional overhead in generating an indexmap vs a hashmap. For example, looking at #489 I benchmarked dijkstra from https://github.com/mtreinish/retworkx-comparison-benchmarks with 0.10.2, main, and #489 (which we can ignore for this issue):

![all_pairs](https://user-images.githubusercontent.com/2447371/143055845-ffb57409-90c8-4342-91b0-c8ee56b7d8ae.png)
![single_source_shortest_path_2](https://user-images.githubusercontent.com/2447371/143055846-52cb0dea-bac4-46e8-8059-92b0807af478.png)
![single_source_shortest_path](https://user-images.githubusercontent.com/2447371/143055848-a88f9ea8-4f9a-4ae6-82ae-3863d17cabe2.png)

As you can see there for the single source case there is a noticeable regression for `main` vs 0.10.2. Before the 0.11.0 release I think we should revisit #386 and determine if there is a performance regression where `DictMap` is used and if so whether we think a deterministic iteration order is worth that extra overhead.

IvanIsCoding:
Can you also include retworkx at the commits right after? https://github.com/Qiskit/retworkx/commit/747a4f6dad37e1e0495a7425e133dcc54b6cd6b7

We made some other changes to Dijkstra and we bumped PyO3, I just want to narrow it down to #386

mtreinish:
Sure, I reran the benchmarks and replaced the PR entry with 747a4f6 

![single_source_shortest_path_2](https://user-images.githubusercontent.com/2447371/143093556-eec33704-aa7d-4c93-859b-8567aeca6189.png)
![single_source_shortest_path](https://user-images.githubusercontent.com/2447371/143093559-7fdbb20f-56f3-4c78-b647-c4de883fdf94.png)
![all_pairs](https://user-images.githubusercontent.com/2447371/143093561-dd3486d9-0073-4ea6-86c1-93419eba21d3.png)

[...]

PR #493, opened 2021-11-24
Use `Vector` for intermediate computations in Dijkstra

IvanIsCoding:
Related to #492 

Replaces `DictMap` with `Vector` for intermediate computations in our Dijkstra's implementation.

* Instead of always using `DictMap` to store the scores, we use a `Vector` and create the `DictMap` the end. That way, users can get a nice deterministic output and we can keep computations fast

### DIFF
--- a/retworkx-core/src/distancemap.rs
+++ b/retworkx-core/src/distancemap.rs
@@ -28,6 +28,7 @@ use std::hash::Hash;
 use petgraph::graph::IndexType;
 
 use crate::dictmap::*;
+use hashbrown::HashMap;
 
 pub trait DistanceMap<K, V> {
     fn build(num_elements: usize) -> Self;
@@ -56,6 +57,23 @@ impl<K: Eq + Hash, V: Clone> DistanceMap<K, V> for DictMap<K, V> {
     #[inline]
     fn build(_num_elements: usize) -> Self {
         DictMap::<K, V>::default()
+    }
+
+    #[inline]
+    fn get_item(&self, pos: K) -> Option<&V> {
+        self.get(&pos)
+    }
+
+    #[inline]
+    fn put_item(&mut self, pos: K, val: V) {
+        self.insert(pos, val);
+    }
+}
+
+impl<K: Eq + Hash, V: Clone> DistanceMap<K, V> for HashMap<K, V> {
+    #[inline]
+    fn build(_num_elements: usize) -> Self {
+        HashMap::<K, V>::default()
     }
 
     #[inline]
--- a/retworkx-core/src/distancemap.rs
+++ b/retworkx-core/src/distancemap.rs
@@ -10,13 +10,6 @@
 // License for the specific language governing permissions and limitations
 // under the License.
 
-// This module was originally copied and forked from the upstream petgraph
-// repository, specifically:
-// https://github.com/petgraph/petgraph/blob/0.5.1/src/dijkstra.rs
-// this was necessary to modify the error handling to allow python callables
-// to be use for the input functions for edge_cost and return any exceptions
-// raised in Python instead of panicking
-
 //! This module contains the [`DistanceMap`] trait which is used in
 //! [`shortest_path`].
 //!
--- a/retworkx-core/src/distancemap.rs
+++ b/retworkx-core/src/distancemap.rs
@@ -11,7 +11,7 @@
 // under the License.
 
 //! This module contains the [`DistanceMap`] trait which is used in
-//! [`shortest_path`].
+//! [`shortest_path`](crate::shortest_path).
 //!
 //! The trait allows the shortest path functions to support multiple
 //! return types.
--- a/retworkx-core/src/distancemap.rs
+++ b/retworkx-core/src/distancemap.rs
@@ -27,11 +27,11 @@ use hashbrown::HashMap;
 pub trait DistanceMap<K, V> {
     /// Create mapping with support for items between 0 and `num_elements - 1`.
     fn build(num_elements: usize) -> Self;
-    
+
     /// Get the distance to the item at `pos`. If the distance does not exist,
     /// the function returns `None`.
     fn get_item(&self, pos: K) -> Option<&V>;
-    
+
     /// Insert item at position `pos` with distance `V`.
     fn put_item(&mut self, pos: K, val: V);
 }
--- a/retworkx-core/src/distancemap.rs
+++ b/retworkx-core/src/distancemap.rs
@@ -23,9 +23,16 @@ use petgraph::graph::IndexType;
 use crate::dictmap::*;
 use hashbrown::HashMap;
 
+/// A mapping for storing the distances of nodes for shortest path algorithms.
 pub trait DistanceMap<K, V> {
+    /// Create mapping with support for items between 0 and `num_elements - 1`.
     fn build(num_elements: usize) -> Self;
+    
+    /// Get the distance to the item at `pos`. If the distance does not exist,
+    /// the function returns `None`.
     fn get_item(&self, pos: K) -> Option<&V>;
+    
+    /// Insert item at position `pos` with distance `V`.
     fn put_item(&mut self, pos: K, val: V);
 }
 
--- a/retworkx-core/src/distancemap.rs
+++ b/retworkx-core/src/distancemap.rs
@@ -1,0 +1,70 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License. You may obtain
+// a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations
+// under the License.
+
+// This module was originally copied and forked from the upstream petgraph
+// repository, specifically:
+// https://github.com/petgraph/petgraph/blob/0.5.1/src/dijkstra.rs
+// this was necessary to modify the error handling to allow python callables
+// to be use for the input functions for edge_cost and return any exceptions
+// raised in Python instead of panicking
+
+//! This module contains the [`DistanceMap`] trait which is used in
+//! [`shortest_path`].
+//!
+//! The trait allows the shortest path functions to support multiple
+//! return types.
+
+use std::hash::Hash;
+
+use petgraph::graph::IndexType;
+
+use crate::dictmap::*;
+
+pub trait DistanceMap<K, V> {
+    fn build(num_elements: usize) -> Self;
+    fn get_item(&self, pos: K) -> Option<&V>;
+    fn put_item(&mut self, pos: K, val: V);
+}
+
+impl<K: IndexType, V: Clone> DistanceMap<K, V> for Vec<Option<V>> {
+    #[inline]
+    fn build(num_elements: usize) -> Self {
+        vec![None; num_elements]
+    }
+
+    #[inline]
+    fn get_item(&self, pos: K) -> Option<&V> {
+        self[pos.index()].as_ref()
+    }
+
+    #[inline]
+    fn put_item(&mut self, pos: K, val: V) {
+        self[pos.index()] = Some(val);
+    }
+}
+
+impl<K: Eq + Hash, V: Clone> DistanceMap<K, V> for DictMap<K, V> {
+    #[inline]
+    fn build(_num_elements: usize) -> Self {
+        DictMap::<K, V>::default()
+    }
+
+    #[inline]
+    fn get_item(&self, pos: K) -> Option<&V> {
+        self.get(&pos)
+    }
+
+    #[inline]
+    fn put_item(&mut self, pos: K, val: V) {
+        self.insert(pos, val);
+    }
+}
--- a/retworkx-core/src/shortest_path/dijkstra.rs
+++ b/retworkx-core/src/shortest_path/dijkstra.rs
@@ -25,48 +25,8 @@ use petgraph::graph::IndexType;
 use petgraph::visit::{EdgeRef, IntoEdges, NodeIndexable, VisitMap, Visitable};
 
 use crate::dictmap::*;
+use crate::distancemap::DistanceMap;
 use crate::min_scored::MinScored;
-
-// Trait for supporting storing
-pub trait DistanceMap<K, V> {
-    fn new_distancemap(num_elements: usize) -> Self;
-    fn get_item(&self, pos: K) -> Option<&V>;
-    fn put_item(&mut self, pos: K, val: V);
-}
-
-impl<K: IndexType, V: Clone> DistanceMap<K, V> for Vec<Option<V>> {
-    #[inline]
-    fn new_distancemap(num_elements: usize) -> Self {
-        vec![None; num_elements]
-    }
-
-    #[inline]
-    fn get_item(&self, pos: K) -> Option<&V> {
-        self[pos.index()].as_ref()
-    }
-
-    #[inline]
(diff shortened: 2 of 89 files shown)


mtreinish:
This makes the overhead worse, adding a sort step looks like it causes an even worse regression:

single_source_shortest_path_2
single_source_shortest_path
all_pairs

IvanIsCoding:
> This makes the overhead worse, adding a sort step looks like it causes an even worse regression:
> 
> single_source_shortest_path_2 single_source_shortest_path all_pairs

Good to know, I will try another idea

IvanIsCoding:
Tried using `Vector` instead of `HashMap`, that might give us more margin.

retworkx-core/src/shortest_path/dijkstra.rs
```diff
@@ -169,5 +170,11 @@ where
...
         visited.visit(node);
     }
-    Ok(scores)
+
+    Ok(graph
+        .node_identifiers()
+        .filter_map(|node_id| {
+            scores[graph.to_index(node_id)].map(|score| (node_id, score))
+        })
+        .collect())
```
georgios-ts, line 179: Iterating directly over scores should be faster.
```suggestion
    Ok(scores
        .into_iter()
        .enumerate()
        .filter_map(|(node, score)| {
            score.map(|val| (graph.from_index(node), val))
        })
        .collect())
```

georgios-ts (commented):
While we are here, we could also see if there is any performance benfit replacing `BinaryHeap` with a [`QuaternaryHeap`](https://docs.rs/dary_heap/0.3.2/dary_heap/type.QuaternaryHeap.html). At least, BGL (that graph-tool depends on) uses a 4-ary heap.

IvanIsCoding:
> While we are here, we could also see if there is any performance benfit replacing `BinaryHeap` with a [`QuaternaryHeap`](https://docs.rs/dary_heap/0.3.2/dary_heap/type.QuaternaryHeap.html). At least, BGL (that graph-tool depends on) uses a 4-ary heap.

That crate seems a bit too young, I will hold on the Quaternary Heap for a bit

IvanIsCoding:
I added the QuaternaryHeap, they had a version without const generics with Rust MSRV 1.31

IvanIsCoding:
@mtreinish I think the biggest culprit is https://github.com/Qiskit/retworkx/blob/main/src/shortest_path/mod.rs#L221-L231, every time we map through a HashMap/DictMap we allocate more memory. Because DictMap uses more memory and the graph had millions of nodes, the results got worse.

The code I have now is not the cleanest, but it creates a DictMap only once. That does lower the performance regression

mtreinish:
@IvanIsCoding yeah that does the trick now. Running the benchmarks again it's faster than 0.10.2 now:

single_source_shortest_path_2
single_source_shortest_path
all_pairs

IvanIsCoding:
I will try to clean the code for this one, and flag other similar cases in #492.

georgios-ts:
It's great that we managed to fixed the regression! Having two separate functions `dijkstra`, `dijkstra_vector` is not the end of the world but it's not the cleanest interface, so I have some guesses/suggestions to make:

1. Given that the biggest regression was in the single source shortest path benchmark between two nodes and hopefully the code https://github.com/Qiskit/retworkx/blob/main/src/shortest_path/mod.rs#L221-L231 does not allocate memory for million of nodes if `goal` is set, I'm wondering if the early return here https://github.com/Qiskit/retworkx/blob/4673692bff01f924f71c6aeba8e8203db7fea6ee/src/shortest_path/mod.rs#L223-L234 that avoids iterating over all nodes in the dict could fix alone the regression.
2. Keep the code from main in `dijkstra` and change the type of `scores` from `DictMap<G::NodeId, K>` to `DictMap<usize, K>`, so we avoid iterate/and allocate a new dict to store inside `PathLengthMapping`. Or, do the opposite and change the dict type inside `PathLengthMapping` to `DictMap<NodeIndex, f64>`.
3. Keep only `dijkstra_vector` in retworkx-core.
4. Define a trait:
    ```rust
   pub trait DistMap<N, K> { 
       fn get(&self,  a: N) -> Option<K>;
       fn put(&mut self, a: N, val: K)
   }
    ```
    similar to [`VisitMap`](https://docs.rs/petgraph/latest/petgraph/visit/trait.VisitMap.html), implement it for `HashMap`/`DictMap`/`Vec` and make `dijkstra` generic like

    ```rust
   pub fn dijkstra<G, F, K, E>(
        graph: G,
        start: G::NodeId,
        goal: Option<G::NodeId>,
        edge_cost: F,
        scores: impl DistMap<G::NodeId, K>
        path: Option<&mut DictMap<G::NodeId, Vec<G::NodeId>>>,
    )
    ```
    so Rust users can choose between `HashMap`/`DictMap`/`Vec`.

IvanIsCoding:
> It's great that we managed to fixed the regression! Having two separate functions `dijkstra`, `dijkstra_vector` is not the end of the world but it's not the cleanest interface, so I have some guesses/suggestions to make:
> 
> 1. Given that the biggest regression was in the single source shortest path benchmark between two nodes and hopefully the code https://github.com/Qiskit/retworkx/blob/main/src/shortest_path/mod.rs#L221-L231 does not allocate memory for million of nodes if `goal` is set, I'm wondering if the early return here https://github.com/Qiskit/retworkx/blob/4673692bff01f924f71c6aeba8e8203db7fea6ee/src/shortest_path/mod.rs#L223-L234
>     that avoids iterating over all nodes in the dict could fix alone the regression.
> 2. Keep the code from main in `dijkstra` and change the type of `scores` from `DictMap<G::NodeId, K>` to `DictMap<usize, K>`, so we avoid iterate/and allocate a new dict to store inside `PathLengthMapping`. Or, do the opposite and change the dict type inside `PathLengthMapping` to `DictMap<NodeIndex, f64>`.
> 3. Keep only `dijkstra_vector` in retworkx-core.
> 4. Define a trait:
>    ```rust
>    pub trait DistMap<N, K> { 
>       fn get(&self,  a: N) -> Option<K>;
>       fn put(&mut self, a: N, val: K)
>    }
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    similar to [`VisitMap`](https://docs.rs/petgraph/latest/petgraph/visit/trait.VisitMap.html), implement it for `HashMap`/`DictMap`/`Vec` and make `dijkstra` generic like
>    ```rust
>    pub fn dijkstra<G, F, K, E>(
>        graph: G,
>        start: G::NodeId,
>        goal: Option<G::NodeId>,
>        edge_cost: F,
>        scores: impl DistMap<G::NodeId, K>
>        path: Option<&mut DictMap<G::NodeId, Vec<G::NodeId>>>,
>    )
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    so Rust users can choose between `HashMap`/`DictMap`/`Vec`.

I like the trait idea, I will try implementing it. It would be nice to have a flexible API that works with all DictMap/HashMap/Vec

IvanIsCoding:
We will need to benchmark again, but I refactored the code such that the function in retworkx-core can return both a Vec and an IndexMap

mtreinish:
> We will need to benchmark again, but I refactored the code such that the function in retworkx-core can return both a Vec and an IndexMap

single_source_shortest_path_2
single_source_shortest_path
all_pairs

retworkx-core/src/shortest_path/dijkstra.rs
```diff
@@ -20,14 +20,54 @@
...
 use petgraph::algo::Measure;
-use petgraph::visit::{EdgeRef, IntoEdges, VisitMap, Visitable};
+use petgraph::graph::IndexType;
+use petgraph::visit::{EdgeRef, IntoEdges, NodeIndexable, VisitMap, Visitable};
 
 use crate::dictmap::*;
 use crate::min_scored::MinScored;
 
+// Trait for supporting storing
+pub trait DistanceMap<K, V> {
```
georgios-ts, line 31: Probably it's better to move this into a separate file since we might need it for other shortest path algorithms too. Also we might need to add documentation on this trait since it'll be in the docs of `retworkx-core` crate. We can even give it another name since we can in fact store any property and not just distances.

retworkx-core/src/shortest_path/dijkstra.rs
```diff
@@ -20,14 +20,54 @@
...
+    fn new_distancemap(num_elements: usize) -> Self;
```
georgios-ts, line 32: I'd prefer naming this method like
```suggestion
    fn distance_map(num_elements: usize) -> Self;
```
or just `new` or `build`

retworkx-core/src/shortest_path/dijkstra.rs
```diff
@@ -20,14 +20,54 @@
...
+        self[pos.index()].as_ref()
+    }
+
+    #[inline]
+    fn put_item(&mut self, pos: K, val: V) {
+        self[pos.index()] = Some(val);
+    }
+}
+
+impl<K: Eq + Hash, V: Clone> DistanceMap<K, V> for DictMap<K, V> {
```
georgios-ts, line 54: We can add a similar implementation for `HashMap`.

retworkx-core/src/shortest_path/dijkstra.rs
```diff
@@ -98,24 +138,25 @@ use crate::min_scored::MinScored;
 /// assert_eq!(res.unwrap(), expected_res);
 /// // z is not inside res because there is not path from b to z.
 /// ```
-pub fn dijkstra<G, F, K, E>(
+pub fn dijkstra<G, F, K, E, S>(
```
georgios-ts, line 141: Can we mention in the docs that you can choose the return type for this function? It currently states that it returns a `DictMap`.

retworkx-core/src/shortest_path/dijkstra.rs
```diff
@@ -98,24 +138,25 @@ use crate::min_scored::MinScored;
...
     goal: Option<G::NodeId>,
     mut edge_cost: F,
     mut path: Option<&mut DictMap<G::NodeId, Vec<G::NodeId>>>,
-) -> Result<DictMap<G::NodeId, K>, E>
+) -> Result<S, E>
 where
-    G: IntoEdges + Visitable,
-    G::NodeId: Eq + Hash,
+    G: IntoEdges + Visitable + NodeIndexable,
+    G::NodeId: Eq + Hash + IndexType,
```
georgios-ts, line 150: I don't think this is needed and actually it'd restrict the graph types that we can use.

```suggestion
    G::NodeId: Eq + Hash,
```

retworkx-core/src/shortest_path/k_shortest_path.rs
```diff
@@ -83,13 +84,13 @@ pub fn k_shortest_path<G, F, E, K>(
...
 where
-    G: IntoEdges + Visitable + NodeCount + NodeIndexable,
+    G: IntoEdges + Visitable + NodeCount + NodeIndexable + IntoNodeIdentifiers,
     G::NodeId: Eq + Hash,
     F: FnMut(G::EdgeRef) -> Result<K, E>,
     K: Measure + Copy,
 {
     let mut counter: Vec<usize> = vec![0; graph.node_bound()];
-    let mut scores = DictMap::with_capacity(graph.node_count());
+    let mut scores: Vec<Option<K>> = vec![None; graph.node_bound()];
```
georgios-ts, line 93: We should also use here `DistanceMap`  for maximum flexibility, right?

src/shortest_path/all_pairs_dijkstra.rs
```diff
@@ -81,23 +81,24 @@ pub fn all_pairs_dijkstra_path_lengths<Ty: EdgeType + Sync>(
...
-                .iter()
-                .filter_map(|(index, cost)| {
-                    if *index == x {
-                        None
-                    } else {
-                        Some((index.index(), *cost))
-                    }
-                })
-                .collect(),
+                    as PyResult<DictMap<NodeIndex, f64>>)
```
georgios-ts, line 91: Type casting here doesn't look great (in terms of readability). IMO, type annotation is nicer:

```rust
let path_lengths: PyResult<DictMap<NodeIndex, f64>> =
     dijkstra(graph, x, None, |e| edge_cost(e.id()), None);

let out_map = PathLengthMapping {
     path_lengths: path_lengths
            .unwrap()
            .iter()
            ...
            .collect(),
};
```

georgios-ts (commented):
I left some inline comments. Apart from that, it's weird that we use internally a vector for the computation in `dijkstra_shortest_path_lengths` but a `DictMap` in `dijkstra_shortest_paths`. For maximum flexibility and control over memory - speed trade-off, it seems a good choice to add a kwarg `sparse` in order to let users choose which data structure best fits their needs.

IvanIsCoding:
> I left some inline comments. Apart from that, it's weird that we use internally a vector for the computation in `dijkstra_shortest_path_lengths` but a `DictMap` in `dijkstra_shortest_paths`. For maximum flexibility and control over memory - speed trade-off, it seems a good choice to add a kwarg `sparse` in order to let users choose which data structure best fits their needs.

I addressed the comments, but I think adding a `sparse` kwarg is not the best choice. We'd need to have a conditional calling the same function with different types, which is a bit cumbersome. I think it's ok to make a choice for the user in this case.

georgios-ts:
Yeah, it's ok to make a choice for the users but then why not use a vector internally for `dijkstra_shortest_paths` (or even in `all_pairs_dijkstra_path_lengths`) too?

IvanIsCoding:
> all_pairs_dijkstra_path_lengths

I've updated the PR to include that as well

retworkx-core/src/distancemap.rs
```diff
@@ -0,0 +1,88 @@
...
+//! return types.
+
+use std::hash::Hash;
+
+use petgraph::graph::IndexType;
+
+use crate::dictmap::*;
+use hashbrown::HashMap;
+
+pub trait DistanceMap<K, V> {
```
mtreinish, line 33: Do we need a docstring for this trait? I'm not sure module docs are sufficient to document how this is used or what the expectations for it are.

retworkx-core/src/distancemap.rs
```diff
@@ -0,0 +1,88 @@
...
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations
+// under the License.
+
+// This module was originally copied and forked from the upstream petgraph
+// repository, specifically:
+// https://github.com/petgraph/petgraph/blob/0.5.1/src/dijkstra.rs
+// this was necessary to modify the error handling to allow python callables
+// to be use for the input functions for edge_cost and return any exceptions
+// raised in Python instead of panicking
```
mtreinish, line 18: Is this true or is it just a copy and paste error.

mtreinish (approved):
Overall LGTM, thanks for keeping with this. The code looks great and the performance definitely is where we want it to be:

all_pairs
single_source_shortest_path
single_source_shortest_path_2

It'd be good if @georgios-ts gave this a look too since he had comments on earlier revisions.

I just had two quick comments inline about the documentation and comments before we merge this around the `DistanceMap` trait. It might be worth manually checking the `cargo doc` output to verify how it renders.

retworkx-core/src/distancemap.rs
IvanIsCoding: This was a copy and paste error

georgios-ts (approved):
LGTM and performance looks great!

retworkx-core/src/distancemap.rs
mtreinish: I took a look at the output for cargo doc, I think we should add a docstring here. When you click `DistanceMap` from the shortest path functions you get this page:

Screenshot_2022-01-14_13-03-00

The module page is visible if you explicitly go to it, but without manually searching for it you probably won't see it. So adding a docstring explaining what this trait is used for and what the methods in the trait do is probably good (just in case someone wants to implement it for their own type).

PR #497, opened 2021-11-27
Add generalized Petersen graphs generator

IvanIsCoding:
Related to #150 

Add a generator for generalized Petersen graphs [1]. That family of graphs has many interesting mathematical properties, and could be useful for both users and retworkx developers. This PR also adds test cases to our isomorphism code based on properties of those Petersen graphs.

[1]: https://mathworld.wolfram.com/GeneralizedPetersenGraph.html

### DIFF
--- a/src/generators.rs
+++ b/src/generators.rs
@@ -2213,6 +2213,45 @@ pub fn lollipop_graph(
     Ok(graph)
 }
 
+#[pyfunction(multigraph = true)]
+#[pyo3(
+    text_signature = "(/, multigraph=True)"
+)]
+pub fn generalized_petersen_graph(
+    py: Python,
+    n: usize,
+    k: usize,
+    multigraph: bool,
+) -> PyResult<graph::PyGraph> {
+    let mut graph = StablePyGraph::<Undirected>::default();
+
+    let star_nodes: Vec<NodeIndex> =
+        (0..n).map(|_| graph.add_node(py.None())).collect();
+
+    let polygon_nodes: Vec<NodeIndex> =
+        (0..n).map(|_| graph.add_node(py.None())).collect();
+        
+    for i in 0..n {
+        graph.add_edge(star_nodes[i], star_nodes[(i + k) % n], py.None());
+    }
+
+    for i in 0..n {
+        graph.add_edge(polygon_nodes[i], polygon_nodes[(i + 1) % n], py.None());
+    }
+
+    for i in 0..n {
+        graph.add_edge(polygon_nodes[i], star_nodes[i], py.None());
+    }    
+
+    Ok(
+        graph::PyGraph {
+            graph,
+            node_removed: false,
+            multigraph
+        }
+    )
+}
+
 #[pymodule]
 pub fn generators(_py: Python, m: &PyModule) -> PyResult<()> {
     m.add_wrapped(wrap_pyfunction!(cycle_graph))?;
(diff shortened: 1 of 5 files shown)


src/generators.rs
```diff
@@ -2213,6 +2213,95 @@ pub fn lollipop_graph(
...
+///   )
+///   mpl_draw(graph, pos=layout)
+///
+/// .. [1] Watkins, Mark E.
+///    "A theorem on tait colorings with an application to the generalized Petersen graphs"
+///    Journal of Combinatorial Theory 6 (2), 152–164 (1969).
+///    https://doi.org/10.1016/S0021-9800(69)80116-X
+///
+#[pyfunction(multigraph = true)]
+#[pyo3(text_signature = "(/, multigraph=True)")]
```
georgios-ts, line 2261:
```suggestion
#[pyo3(text_signature = "(n, k, /, multigraph=True)")]
```

src/generators.rs
```diff
@@ -2213,6 +2213,95 @@ pub fn lollipop_graph(
...
+        return Err(PyIndexError::new_err("n must be at least 3"));
+    }
+
+    if k == 0 || 2 * k >= n {
+        return Err(PyIndexError::new_err(
+            "k is invalid: it must be positive and less than n/2",
+        ));
+    }
+
+    let mut graph = StablePyGraph::<Undirected>::default();
```
georgios-ts, line 2278:
```suggestion
    let mut graph = StablePyGraph::<Undirected>::with_capacity(2*n, 3*n);
```

tests/generators/test_petersen.py
```diff
@@ -0,0 +1,57 @@
...
+        n = 99
+        k = 23
+        graph = retworkx.generators.generalized_petersen_graph(n, k)
+        self.assertEqual(len(graph), 2 * n)
+        self.assertEqual(len(graph.edges()), 3 * n)
+
+    def test_petersen_graph_edge(self):
+        graph = retworkx.generators.generalized_petersen_graph(5, 2)
+        edge_list = graph.edge_list()
+        print(edge_list)
```
georgios-ts, line 29:
```suggestion
```

src/generators.rs
```diff
@@ -2213,6 +2213,95 @@ pub fn lollipop_graph(
...
+#[pyfunction(multigraph = true)]
+#[pyo3(text_signature = "(/, multigraph=True)")]
+pub fn generalized_petersen_graph(
+    py: Python,
+    n: usize,
+    k: usize,
+    multigraph: bool,
+) -> PyResult<graph::PyGraph> {
+    if n < 3 {
+        return Err(PyIndexError::new_err("n must be at least 3"));
```
georgios-ts, line 2269: I think [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError) is more appropriate here but yeah this is something that we're doing in all graph generators.

src/generators.rs
```diff
@@ -2213,6 +2213,95 @@ pub fn lollipop_graph(
...
+/// :param int k: shift that changes the internal star graph.
+/// :param bool multigraph: When set to False the output
+///     :class:`~retworkx.PyGraph` object will not be not be a multigraph and
+///     won't allow parallel edges to be added. Instead
+///     calls which would create a parallel edge will update the existing edge.
+///
+/// :returns: The generated generalized Petersen graph.
+///
+/// :rtype: PyGraph
+/// :raises TypeError: If either ``n`` or ``k`` are
```
georgios-ts, line 2233: Maybe?
```suggestion
/// :raises IndexError: If either ``n`` or ``k`` are
```
Or even better we can mention that `TypeError` will be raised if you pass an arg that can not be converted to a `usize` and `IndexError` if the value is not "valid".

georgios-ts (commented):
Looks good!

src/generators.rs
IvanIsCoding: Done

georgios-ts (approved):
LGTM, thanks for the updates.

PR #499, opened 2021-11-29
Add online topological sorter.

### LINKED ISSUE #495: Online topological sort
I'm interested in replacing [graphlib.TopologicalSorter](https://docs.python.org/3/library/graphlib.html) with something that will also let me do other graph operations (namely pruning subgraphs based on nodes, which it seems like I can do with `PyDAG.substitute_node_with_subgraph`).
In particular, this sort of topological sort allows you to poll for available dependencies as you process them, vs. just giving you a topologically sorted list of all of the nodes in the graph.

georgios-ts:
### Summary

This PR implements a new class `TopologicalSorter` to topologically sort a directed graph.
It gives the ability to process nodes while they are becoming ready (i.e they have no unprocessed
predecessors) and then mark them as done, so more nodes can be freed.

### Details

`TopologicalSorter` wraps the underlying graph inside a `Py<..>` with GIL-independent lifetime
in order to avoid cloning the graph. The downside is that users may modify the graph while using
a `TopologicalSorter`, we do not guard against it and this might result in wrong outputs.

Closes #495.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -65,6 +65,7 @@ Traversal
    retworkx.collect_runs
    retworkx.collect_bicolor_runs
    retworkx.visit.BFSVisitor
+   retworkx.TopologicalSorter
 
 .. _dag-algorithms:
 
--- a/releasenotes/notes/online-topo-sort-f94bcd2064a5cde7.yaml
+++ b/releasenotes/notes/online-topo-sort-f94bcd2064a5cde7.yaml
@@ -1,0 +1,7 @@
+---
+features:
+  - |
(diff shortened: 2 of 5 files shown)


IvanIsCoding (changes requested):
I like the online topological sorting idea. As a replacement for https://docs.python.org/3/library/graphlib.html#graphlib, this looks good.

One thing that is missing in our implementation though is cycle checking. We need to add a test case where a topological sorting does not exist and we throw a cycle error

georgios-ts:
@IvanIsCoding 
Done in d46f800.

IvanIsCoding (approved):

PR #500, opened 2021-12-01
Add LICENSE to MANIFEST

BastianZim:
Currently, the LICENSE is not included in the source distribution on PyPI. This should fix it.


mtreinish (approved):
LGTM, thanks for fixing this.

PR #501, opened 2021-12-01
Add section to README about slack and irc channels

mtreinish:
This commit adds a section to the bottom of the README that mentions our
slack and irc channels for retworkx if people would like to reach out
about the library.


IvanIsCoding (approved):
LGTM. Perhaps you want to add a badge linking to our Slack channel too, but this is good to go

szhorvat:
I wanted to note that when joining the Slack space, one does not automatically join the `#retworkx` channel as well. For a new user, it appears that this space is only about Qiskit, as none of the channels displayed on the left seem to concern retworkx. At first, I did not realize that there are additional channels that I need to search for and join manually.

PR #510, opened 2021-12-03
Improve docstring for lexicographical topological sort

### LINKED ISSUE #506: [joss] lexicographical_topological_sort() docs are unclear
I was unable to tell what `lexicographical_topological_sort()` does after looking at its documentation. As far as I am aware, "lexicographical topological sorting" is not standard terminology. It would be good to explain it.

I looked up the corresponding networkx function, and now I can see that this function is for breaking ties in the ordering based on an additional comparison. This was not clear.

mtreinish:
As was pointed out in #506 the docstring for the
`lexicographical_topological_sort()` function was not really clear as to
what it was or how it differed from the normal `topological_sort()`
function. This commit expands the docstring for the function to outline
specifically how it behaves.

Fixes #506

### DIFF
--- a/src/dag_algo/mod.rs
+++ b/src/dag_algo/mod.rs
@@ -311,7 +311,7 @@ pub fn layers(
 /// This function returns a list of nodes data in a graph lexicographically
 /// topologically sorted using the provided key function. A topological sort
 /// is a linear ordering of vertices such that for every directed edge from
-/// node :math:`u` to node :math:`v` that :math:`u` comes before :math:`v`
+/// node :math:`u` to node :math:`v`, :math:`u` comes before :math:`v`
 /// in the ordering.
 ///
 /// This function differs from :func:`~retworkx.topological_sort` because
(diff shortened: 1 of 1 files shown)


src/dag_algo/mod.rs
```diff
@@ -309,12 +309,21 @@ pub fn layers(
 /// Get the lexicographical topological sorted nodes from the provided DAG
 ///
 /// This function returns a list of nodes data in a graph lexicographically
-/// topologically sorted using the provided key function.
+/// topologically sorted using the provided key function. A topological sort
+/// is a linear ordering of vertices such that for every directed edge from
+/// node :math:`u` to node :math:`v` that :math:`u` comes before :math:`v`
```
szhorvat, line 314: Should "that" be replaced by a comma?

szhorvat:
Much clearer now :-)

src/dag_algo/mod.rs
IvanIsCoding, line 314:
```suggestion
/// node :math:`u` to node :math:`v`, :math:`u` comes before :math:`v`
```

IvanIsCoding (approved):

PR #518, opened 2021-12-04
Fix `PyGraph.degree` if self-loops.

### LINKED ISSUE #517: [joss] Bad result from degree() method with self loops
The `.degree()` method returns incorrect (or at least very unusual) results for undirected graphs with self-loops.

Example:

```
g3 = rx.PyGraph()
g3.extend_from_edge_list([(0,1), (0,0)])

g3.degree(0)
```

```
2
```

The correct result is 3. Note that the degree is not the same as the number of incident edges. It is the number of edge _endpoints_.

Certainly, one may reinterpret the meaning of "degree", but diverging from the de-facto standard textbook definition is not a good idea. It is going to be extremely confusing for users.

georgios-ts:
Closes #517.
In an undirected graph, self-loops should increment by two (and not by one)
the degree of a node.

### DIFF
--- a/releasenotes/notes/fix-degree-with-self-loops-5b40c45cc61fcebe.yaml
+++ b/releasenotes/notes/fix-degree-with-self-loops-5b40c45cc61fcebe.yaml
@@ -1,0 +1,6 @@
+---
+fixes:
+  - |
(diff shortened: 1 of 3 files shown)


IvanIsCoding (approved):
LGTM, thanks for the quick fix

PR #520, opened 2021-12-04
Fix dispatch function of `dfs_edges` and `union`.

### LINKED ISSUE #515: [joss] dfs_edges() does not behave as documented, requires 'source' argument
According to the documentation, the `source` argument of `dfs_edges()` is optional:

https://qiskit.org/documentation/retworkx/stubs/retworkx.dfs_edges.html#retworkx.dfs_edges

> If this is not specified then a source will be chosen arbitrarly and repeated until all components of the graph are searched.

However, in reality the function throws an error:

```
rx.dfs_edges(rx.undirected_gnm_random_graph(10,10))

TypeError: _graph_dfs_edges() missing 1 required positional argument: 'source'
```

This also means that there seems to be no way to traverse an entire non-connected graph with this function.

georgios-ts:
Closes #515.
This commit updates `dfs_edges` to set `source` argument to `None` by default and
fixes an oversight in `union` where user-defined values of `merge_nodes` and `merge_edges`
arguments were being ingored.

### DIFF
--- a/releasenotes/notes/fix-dispatch-3596ef110cc68338.yaml
+++ b/releasenotes/notes/fix-dispatch-3596ef110cc68338.yaml
@@ -1,0 +1,8 @@
+---
+fixes:
+  - |
+    Updates :func:`~retworkx.dfs_edges` to set `source` argument to `None`
+    by default, as already stated in the documentation.
+  - |
+    Fixes an oversight in :func:`~retworkx.union` where user-defined values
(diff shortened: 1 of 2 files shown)


IvanIsCoding:
LGTM. But before we merge, we need a release note about the bug fix

IvanIsCoding (approved):

mtreinish (approved):

PR #531, opened 2021-12-09
Fix `unweighted_average_shortest_path_length` if disconnected graph.

### LINKED ISSUE #524: [joss] unweighted_average_shortest_path_length() mishandles disconnected graphs
`unweighted_average_shortest_path_length()` considers the distance between nodes not reachable from each other to be 0. While this is [documented](https://qiskit.org/documentation/retworkx/stubs/retworkx.unweighted_average_shortest_path_length.html#retworkx.unweighted_average_shortest_path_length), it is simply incorrect, and counter to any reasonable interpretation of the concept. I would consider this a bug because it will confuse users, or worse: lead them to incorrect results in their analysis.

georgios-ts:
### Summary

Fixes the way `unweighted_average_shortest_path_length` handles
disconnected graphs. Disconnected pairs of nodes are now assumed
to have an infinite distance. In addition, an extra kwarg ``disconnected``
was added where, if set to ``True``, the average is taken only over
connected node pairs. By default, it's set to ``False``.

Closes #524.

### DIFF
--- a/src/shortest_path/average_length.rs
+++ b/src/shortest_path/average_length.rs
@@ -10,7 +10,7 @@
 // License for the specific language governing permissions and limitations
 // under the License.
 
-use hashbrown::{HashMap, HashSet};
+use hashbrown::HashSet;
 
 use petgraph::prelude::*;
 use petgraph::EdgeType;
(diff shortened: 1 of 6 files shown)


mtreinish (commented):
The fix looks good to me, but I'm a bit concerned about the change in documented behavior. We probably should have an explicit `upgrade` note here to document that it will return different results for a disconnected graph from previous releases (and maybe have a python example on how to get an identical result after upgrading)

georgios-ts:
@mtreinish  I used an `upgrade` note in 96d3b2d.

> (and maybe have a python example on how to get an identical result after upgrading)

For (strongly) connected graphs the results should always be identical. But for disconnected graphs there is no parameter setting that can give the same result after upgrading since this PR changes the denominator while finding the average.

mtreinish:
> For (strongly) connected graphs the results should always be identical. But for disconnected graphs there is no parameter setting that can give the same result after upgrading since this PR changes the denominator while finding the average.

I was thinking more of documenting if the previously documented behavior was desired showing an example using the rest of the retworkx api to accomplish the same calculation as the previous behavior. Putting something like:

```python
d_mat = rx.distance_matrix(g, null_value=0.0)
n = float(len(graph))
avg_shortest_path = d_mat.sum() / (n * (n - 1.0))
 ```
 
 in the upgrade note

**Comment by georgios-ts** [comment]
@mtreinish Ah, good point. Added in f58617a.

File: src/shortest_path/average_length.rs [hunk 1]
```diff
@@ -61,15 +63,18 @@ pub fn compute_distance_sum<Ty: EdgeType + Sync>(
...
+        node_indices
+            .iter()
+            .map(|index| bfs_traversal(*index))
+            .fold((0, 0), |a, b| (a.0 + b.0, a.1 + b.1))
     } else {
         node_indices
             .par_iter()
             .map(|index| bfs_traversal(*index))
-            .sum()
+            .reduce(|| (0, 0), |a, b| (a.0 + b.0, a.1 + b.1))
```
mtreinish, line 78: Is there a reason you used `reduce()` here vs `fold()` in the serial case?

mtreinish (approved):
Overall LGTM, one small nit in the release note and a question inline about the iterators.

src/shortest_path/average_length.rs
georgios-ts: Yeah, parallel `fold` produces a group of values rather than  a single output value.

PR #533, opened 2021-12-17
Remove pip install setuptools-rust from setup.py

mtreinish:
The setup.py has long carried around a pip install fallback if
setuptools-rust can't be imported. This was necessary as prior to the
introduction of pep-517 there was no way to express a build-system
requirement in python packaging. There is the setup_requires field in
the setuptools metadata, but this would both trigger the install too
late (as we need to import setuptools-rust prior to running
setuptools.setup()) and also fallback to using easy_install instead of
pip which complicates installation. With pep-517 we list setuptools-rust
as a build system requirement in the pyproject.toml and pip will just
install that prior to trying to run the setup.py in the sdist.

But, for users on older versions of pip, before pep-517 was supported,
this caused a cryptic error as there would just be an import error that
setuptools_rust can't be found. The pip install fallback on import error
was added to try and make installation seamless for those users. However,
pep-517 support has been around in pip for sufficiently long at this
point I think we can safely remove this fallback as it provides more
harm than good now. In it's place the README is updated to document
that you need a newer version of pip or to manually install
setuptools-rust prior to trying to build retworkx from source. This
should hopefully be sufficient for users on older versions of pip where
they still might encounter issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,13 @@ pip install retworkx
 will build retworkx for your local system from the source package and install
 it just as it would if there was a prebuilt binary available.
 
+Note: To build from source you will need to ensure you have a newer version
+of pip installed that supports PEP-517, or that you have manually installed
+`setuptools-rust` prior to running `pip install retworkx`. If you recieve an error
+about `setuptools-rust` not being found you should upgrade pip with
+`pip install -U pip` or manually install `setuptools-rust` with
+`pip install setuptools-rust` and try again.
+
 ### Optional dependencies
 
 If you're planning to use the `retworkx.visualization` module you will need to
--- a/README.md
+++ b/README.md
@@ -54,10 +54,10 @@ pip install retworkx
 will build retworkx for your local system from the source package and install
 it just as it would if there was a prebuilt binary available.
 
-Note: To build from source you will need to ensure you have a newer version
-of pip installed that supports PEP-517, or that you have manually installed
-`setuptools-rust` prior to running `pip install retworkx`. If you recieve an error
-about `setuptools-rust` not being found you should upgrade pip with
+Note: To build from source you will need to ensure you have pip >=19.0.0
+installed, which supports PEP-517, or that you have manually installed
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1,6 +1,7 @@
 [build-system]
 requires = ["setuptools", "wheel", "setuptools-rust"]
+build-backend = "setuptools.build_meta"
```
IvanIsCoding, line 3: For those reviewing this PR in the future: `build-backend` refers to https://github.com/pypa/setuptools/blob/main/setuptools/build_meta.py

README.md
```diff
@@ -54,6 +54,13 @@ pip install retworkx
 will build retworkx for your local system from the source package and install
 it just as it would if there was a prebuilt binary available.
 
+Note: To build from source you will need to ensure you have a newer version
+of pip installed that supports PEP-517, or that you have manually installed
+`setuptools-rust` prior to running `pip install retworkx`. If you recieve an error
+about `setuptools-rust` not being found you should upgrade pip with
+`pip install -U pip` or manually install `setuptools-rust` with
+`pip install setuptools-rust` and try again.
```
IvanIsCoding, line 62: I think we could try to be more specific here e.g. `you need pip 19.1 or later to build the wheel from source` (note: I came up with an arbitrary version)

IvanIsCoding (approved):
Overall this LGTM and I am in favor of merging.

However, if we have the time to spin up virtual environments and discover the minimum version of pip that can be helpful to people that build `retworkx` from source. This kind of information can be useful for the more advanced users that build from source or maintain things like https://github.com/conda-forge/staged-recipes/pull/17137

README.md
mtreinish: Your guess is not far off, pep 517 support was added in pip 19.0: https://pip.pypa.io/en/stable/news/#v19-0 I'll update the text here to reflect that

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/533/commits/34c0632c9e0e004d3a97d0f3b03dc622bc0ad713

PR #534, opened 2021-12-17
Remove optional dependency on pydot

### LINKED ISSUE #526: [joss] graphviz_draw() bad output
`graphviz_draw()` always places an additional, non-existent node in the drawing. Example:

```
graphviz_draw(rx.generators.cycle_graph(3))
```

image

I have retworkx 0.10.2 and pydot 1.4.2.

mtreinish:
Previously we were relying on the pydot library to call graphviz to
generate visualizations. This was done for simplicity since pydot is all
about providing a python API wrapper around graphviz. However, it does
this by parsing the dot files into it's own internal representation of
the graph description to provide a python api for interacting with it
prior to calling out to graphviz. For retworkx this is unnecessary as we
have a native function for generating dot files and we just need to pass
that directly to graphviz. Additionally, this parsing adds a place for
bugs to be introduced as was seen in #526 when a new version of
pyparsing (the dependency pydot uses to parse the dot files) started
causing invalid output to be generated. This commit updates the graphviz
drawer to stop using pydot and instead manually calling graphviz
directly with the retworkx generated dot files. The setup.py's optional
extras are then updated to no longer list pydot as required for the
graphviz drawer. This should hopefully reduce the complexity of the code
path and avoid potential future issues.

Fixes #526

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,10 +65,9 @@ installed with `pip install matplotlib` or when you're installing retworkx with
 function `retworkx.visualization.graphviz_drawer` first you will need to install
 graphviz, instructions for this can be found here:
 https://graphviz.org/download/#executable-packages. Then you
-will need to install the [pydot](https://pypi.org/project/pydot/) and
-[pillow](https://python-pillow.org/) Python libraries. This can be done either
-with `pip install pydot pillow` or when installing retworkx with
-`pip install 'retworkx[graphviz]'`.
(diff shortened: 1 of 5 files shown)


IvanIsCoding (approved):
Given that we were importing pydot just to have a [thin layer to call graphviz on the command line](https://github.com/pydot/pydot/blob/master/src/pydot/core.py#L133-L141), and that pydot does not bundle graphviz, I am in favor of removing it.

PR #535, opened 2021-12-17
Fix cargo fmt call in tox.ini

mtreinish:
This commit updates the lint job definition in the tox.ini to work now
that we have workspaces in the rust package. In #459 we added a new
workspace for the dedicated rust package retworkx-core. But having
workspaces means that running cargo fmt outside from the tests/
directory (as we do with tox) means cargo doesn't know which workspace
to run against. To correct this issue this commit adds the --all flag to
the cargo fmt call in the tox job definition to ensure we're able to run
from any directory in the repo and check the formatting on all the rust
files in the project.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Test Build
         run: cargo build
       - name: Rust Format
-        run: cargo fmt -- --check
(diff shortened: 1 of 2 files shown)


IvanIsCoding (approved):
LGTM

PR #536, opened 2021-12-19
Improve dfs_edges documentation

### LINKED ISSUE #516: [joss] dfs_edges() invalid result on multigraphs
`dfs_edges()` does not seem to support multigraphs properly:

```
g = rx.PyGraph()
g.extend_from_edge_list([(0,0), (0,1), (1,2), (1,2), (2,3)])

print(rx.dfs_edges(g, 0))
```

```
EdgeList[(0, 1), (1, 2), (2, 3)]
```

I would expect every edge to be traversed, since this function specifically returns edges (not merely nodes). One possible correct result would be `(0,0), (0,1), (1,2), (2,3), (1,2)`.

IvanIsCoding:
Closes #516 

Improves the documentation of `dfs_edges` by:
* making explicit that only tree edges from the DFS are returned
* adding the pseudocode for the DFS procedure

### DIFF
--- a/retworkx-core/src/traversal/dfs_edges.rs
+++ b/retworkx-core/src/traversal/dfs_edges.rs
@@ -18,7 +18,29 @@ use petgraph::visit::{
     NodeIndexable, VisitMap, Visitable,
 };
 
-/// Return an edge list in depth first order.
+/// Return an edge list of the tree edges from a depth-first traversal.__rust_force_expr!
+///
+/// The pseudo-code for the DFS algorithm is listed below. The output
+/// contains the tree edges found by the procedure.
+///
+/// ```norust
+/// DFS(G, s)
+///   for each vertex u in V
+///       color[u] := WHITE
+///   end for
+///   color[s] := GRAY
+///   STACK := Ø
+///   PUSH(STACK, s)                           start at vertex s
+///   while (STACK != Ø)
+///       u := POP(STACK)
+///       for each vertex v in Adj[u]          (u,v) is a tree edge
+///           if (color[v] = WHITE)
+///               color[v] = GRAY
+///               PUSH(STACK, v)
+///       end for
+///       color[u] := BLACK                    finish vertex u
+///   end while
+/// ```
 ///
 /// Arguments:
 ///
--- a/retworkx-core/src/traversal/dfs_edges.rs
+++ b/retworkx-core/src/traversal/dfs_edges.rs
@@ -24,18 +24,19 @@ use petgraph::visit::{
 /// contains the tree edges found by the procedure.
 ///
 /// ```norust
-/// DFS(G, v) is
-///     let S be a stack
-///     PUSH(S, (v, iterator of G.edges(v)))
-///     while (S != Ø)
-///         let (v, iterator) := LAST(S)
-///         if hasNext(iterator) then
-///             w := next(iterator)
-///             if w is not labeled as discovered then
-///                 label w as discovered                   # (v, w) is a tree edge
-///                 PUSH(S, (w, iterator of G.edges(w)))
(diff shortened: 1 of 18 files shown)


retworkx/__init__.py
```diff
@@ -769,7 +769,34 @@ def _graph_k_shortest_path_lengths(graph, start, k, edge_cost, goal=None):
...
+          while (STACK != Ø)
+              u := POP(STACK)
+              for each vertex v in Adj[u]          (u,v) is a tree edge
+                  if (color[v] = WHITE)
+                      color[v] = GRAY
+                      PUSH(STACK, v)
+              end for
+              color[u] := BLACK                    finish vertex u
+          end while
+
```
georgios-ts, line 795: Actually the pseudocode does not seem to be fully accurate, i.e it doesn't perform a standard DFS.

We can use the pseudocode of a recursive implementation like here https://github.com/Qiskit/retworkx/blob/d84d10372927b9a116cea0e1b73796f315d8270a/retworkx/__init__.py#L1746-L1766

or an iterative one which closely resembles the implementation of this function:

```
DFS(G, v) is
    let S be a stack
    PUSH(S, (v, iterator of G.edges(v)))
    while (S != Ø)
        let (v, iterator) := LAST(S)
        if hasNext(iterator) then
            w := next(iterator)
            if w is not labeled as discovered then
                label w as discovered                       # (u, w) is a tree edge
                PUSH(S, (w, iterator of G.edges(w)))
        else
            POP(S)
```

IvanIsCoding: It's not the standard but if you look at the function, it does a stack based DFS

georgios-ts: But if I run the pseudocode with a triangle as a graph starting from node 0, it outputs (0, 1), (0, 2) as tree edges but the DFS output is (0, 1), (1, 2).

IvanIsCoding: Indeed for some reason that function has a weird DFS, but it is a DFS. I updated the docstring to use the pseudocode you provided.

retworkx-core/src/traversal/dfs_edges.rs
```diff
@@ -18,7 +18,25 @@ use petgraph::visit::{
...
+/// ```norust
+/// DFS(G, v) is
+///     let S be a stack
+///     PUSH(S, (v, iterator of G.edges(v)))
+///     while (S != Ø)
+///         let (v, iterator) := LAST(S)
+///         if hasNext(iterator) then
+///             w := next(iterator)
+///             if w is not labeled as discovered then
+///                 label w as discovered                   # (u, w) is a tree edge
```
georgios-ts, line 35: Sorry, I made a type before:
```suggestion
///                 label w as discovered                   # (v, w) is a tree edge
```

georgios-ts (approved):

szhorvat:
I am curious about the pseudocode, and why it was added.

Is it meant to explain what depth first search is to someone who is not familiar with the concept?

Is it meant to clarify which specific DFS tree will be returned out of the several possible ones?

What does `iterator of G.edges(v)` refer to? All incident edges of `v`? Then is `w` in `w := next(iterator)` a vertex or an edge? (I see it's a vertex, but it's confusing that it's taken from an iterator over the edges.)

I think the pseudocode may be slightly incorrect. It does not label the first node as discovered.

georgios-ts:
> I am curious about the pseudocode, and why it was added.

It's just a sketch of how this function works so we can annotate which edges are being returned in case someone is not familiar, or does  not recall, which edges are classified as tree edges in a DFS traversal.

> What does `iterator of G.edges(v)` refer to?

It's an iterator over all neighbors of nodes `v`.

Issue #532, opened 2021-12-12
Invitation to collaborate on graph library usability project


szhorvat:
I recently set up a usability benchmark project for igraph:

https://github.com/igraph/usability-benchmarks

The idea is to write up a set of small tasks that one might use a graph library for, and compare how easy it is to complete them using various libraries in various high-level programming languages. You are welcome to contribute to the project in any manner. You can write up solutions to the existing tasks using retworkx, or you can propose new tasks. Since the focus of retworkx is somewhat different from that of igraph, perhaps you would be able to come with with different simple tasks which we did not think of yet.

You are welcome to use the information in this repository to improve retworkx.

Note that since this is a new project, things are still in flux, and not yet very well organized.

PR #538, opened 2021-12-19
Add edge utility methods to PyGraph and PyDiGraph

mtreinish:
There were some gaps in working with edges in the graph class API.
Mainly when working with indices there was no method to acquire the
edge's data payload or endpoints easily. Additionally for the PyGraph
class didn't have a simple method to find incident edges directly
instead a user had to rely on neighbors() and get the edges from that
directly. This commit adds methods to fill these API gaps to make
working with edges a bit easier.

### DIFF
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -1554,7 +1554,11 @@ impl PyDiGraph {
         }
     }
 
-    /// Return the list of edge indices incent to a provided node
+    /// Return the list of edge indices incident to a provided node
+    ///
+    /// You can later retrieve the data payload of this edge with
+    /// :meth:`~retworkx.PyDiGraph.get_edge_data_by_index` or its
+    /// endpoints with :meth:`~retworkx.PyDiGraph.get_edge_endpoints_by_index`.
     ///
     /// By default this method will only return the outgoing edges of
     /// the provided ``node``. If you would like to access both the
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -670,7 +670,7 @@ impl PyDiGraph {
     ///
     /// :param int edge_index: The edge index to get the data for
     ///
-    /// :returns: The data object set for the edge
+    /// :returns: The data object for the edge
     /// :raises IndexError: when there is no edge present with the provided
     ///     index
     #[pyo3(text_signature = "(self, edge_index, /)")]
@@ -694,7 +694,8 @@ impl PyDiGraph {
     ///
     /// :param int edge_index: The edge index to get the endpoints for
     ///
-    /// :returns: The data object set for the edge
+    /// :returns: The endpoint tuple for the edge
+    /// :rtype: tuple
     /// :raises IndexError: when there is no edge present with the provided
     ///     index
     #[pyo3(text_signature = "(self, edge_index, /)")]
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -1562,6 +1562,7 @@ impl PyDiGraph {
     ///
     /// :returns: A list of the edge indices incident to a node in the graph
     /// :rtype: EdgeIndices
+    #[pyo3(text_signature = "(self, node, /)")]
     pub fn incident_edges(&self, node: usize) -> EdgeIndices {
         EdgeIndices {
             edges: self
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -666,6 +666,55 @@ impl PyDiGraph {
         Ok(data)
     }
 
+    /// Return the edge data for the edge by it's given index
+    ///
+    /// :param int edge_index: The edge index to get the data for
+    ///
+    /// :returns: The data object set for the edge
+    /// :raises IndexError: when there is no edge present with the provided
+    ///     index
+    #[pyo3(text_signature = "(self, edge_index, /)")]
+    pub fn get_edge_data_by_index(
+        &self,
+        edge_index: usize,
+    ) -> PyResult<&PyObject> {
+        let data = match self.graph.edge_weight(EdgeIndex::new(edge_index)) {
+            Some(data) => data,
+            None => {
+                return Err(PyIndexError::new_err(format!(
+                    "Provided edge index {} is not present in the graph",
+                    edge_index
+                )));
+            }
+        };
+        Ok(data)
+    }
+
+    /// Return the edge endpoints for the edge by it's given index
+    ///
+    /// :param int edge_index: The edge index to get the endpoints for
+    ///
+    /// :returns: The data object set for the edge
+    /// :raises IndexError: when there is no edge present with the provided
+    ///     index
+    #[pyo3(text_signature = "(self, edge_index, /)")]
+    pub fn get_edge_endpoints_by_index(
+        &self,
+        edge_index: usize,
+    ) -> PyResult<(usize, usize)> {
+        let endpoints =
+            match self.graph.edge_endpoints(EdgeIndex::new(edge_index)) {
+                Some(endpoints) => (endpoints.0.index(), endpoints.1.index()),
+                None => {
+                    return Err(PyIndexError::new_err(format!(
+                        "Provided edge index {} is not present in the graph",
+                        edge_index
+                    )));
+                }
+            };
+        Ok(endpoints)
+    }
+
     /// Update an edge's weight/payload inplace
     ///
     /// If there are parallel edges in the graph only one edge will be updated.
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -1600,6 +1600,72 @@ impl PyDiGraph {
         }
     }
 
+    /// Return the index map of edges incident to a provided node
+    ///
+    /// By default this method will only return the outgoing edges of
+    /// the provided ``node``. If you would like to access both the
+    /// incoming and outgoing edges you can set the ``all_edges``
+    /// kwarg to ``True``.
+    ///
+    /// :param int node: The node index to get incident edges from. If
+    ///     this node index is not present in the graph this method will
+    ///     return an empty list and not error.
+    /// :param bool all_edges: If set to ``True`` both incoming and outgoing
+    ///     edges to ``node`` will be returned.
+    ///
+    /// :returns: A mapping of incident edge indices to the tuple
+    ///     ``(source, target, data)``
+    /// :rtype: EdgeIndexMap
+    #[pyo3(text_signature = "(self, node, /, all_edges=False)")]
+    #[args(all_edges = "false")]
+    pub fn incident_edge_index_map(
+        &self,
+        py: Python,
+        node: usize,
+        all_edges: bool,
+    ) -> EdgeIndexMap {
+        let node_index = NodeIndex::new(node);
+        if all_edges {
+            EdgeIndexMap {
+                edge_map: self
+                    .graph
+                    .edges_directed(node_index, petgraph::Direction::Outgoing)
+                    .chain(self.graph.edges_directed(
+                        node_index,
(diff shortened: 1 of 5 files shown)


src/graph.rs
```diff
@@ -378,6 +378,73 @@ impl PyGraph {
...
+    /// :rtype: EdgeIndices
+    pub fn incident_edges(&self, node: usize) -> EdgeIndices {
+        EdgeIndices {
+            edges: self
+                .graph
+                .edges(NodeIndex::new(node))
+                .map(|e| e.id().index())
+                .collect(),
+        }
+    }
```
IvanIsCoding, line 397: For consistency reasons I am against adding this method, as there is no equivalent on `PyDiGraph`.

I would rather have us implement https://github.com/Qiskit/retworkx/blob/main/src/digraph.rs#L1506-L1526 and https://github.com/Qiskit/retworkx/blob/main/src/digraph.rs#L1528-L1547 for consistency. That way we have the same API on both PyGraph and PyDiGraph.

IvanIsCoding (commented):
I would rather implement the equivalents of:
 * [`PyDiGraph.out_edges`](https://qiskit.org/documentation/retworkx/stubs/retworkx.PyDiGraph.out_edges.html#retworkx.PyDiGraph.out_edges): for outgoing edges
* [`PyDiGraph.in_edges`](https://qiskit.org/documentation/retworkx/stubs/retworkx.PyDiGraph.in_edges.html#retworkx.PyDiGraph.in_edges): for incident edges

And have them reuse the same code (as the graph is undirected) than introduce the new method `indicident_edges` only for PyGraph

src/graph.rs
mtreinish: Sure I'll add duals of those `PyDiGraph` methods here. But I'll add this to `PyDiGraph` too because I prefer the index return. I'd like to get the edge index based api a bit more up to snuff.

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/538/commits/3b4ae559fbe15f63de4c7d84e84d3ac1d30867e7

src/digraph.rs
```diff
@@ -666,6 +666,56 @@ impl PyDiGraph {
         Ok(data)
     }
 
+    /// Return the edge data for the edge by it's given index
```
IvanIsCoding, line 669:
```suggestion
    /// Return the edge data for the edge by its given index
```

src/digraph.rs
```diff
@@ -666,6 +666,56 @@ impl PyDiGraph {
...
+                return Err(PyIndexError::new_err(format!(
+                    "Provided edge index {} is not present in the graph",
+                    edge_index
+                )));
+            }
+        };
+        Ok(data)
+    }
+
+    /// Return the edge endpoints for the edge by it's given index
```
IvanIsCoding, line 693:
```suggestion
    /// Return the edge endpoints for the edge by its given index
```

src/graph.rs
```diff
@@ -378,6 +378,120 @@ impl PyGraph {
...
+        let index = NodeIndex::new(node);
+        let dir = petgraph::Direction::Outgoing;
+        let raw_edges = self.graph.edges_directed(index, dir);
+        let out_list: Vec<(usize, usize, PyObject)> = raw_edges
+            .map(|x| (node, x.target().index(), x.weight().clone_ref(py)))
+            .collect();
+        WeightedEdgeList { edges: out_list }
+    }
+
+    /// Return the edge data for the edge by it's given index
```
IvanIsCoding, line 445:
```suggestion
    /// Return the edge data for the edge by its given index
```

src/graph.rs
```diff
@@ -378,6 +378,120 @@ impl PyGraph {
...
+                return Err(PyIndexError::new_err(format!(
+                    "Provided edge index {} is not present in the graph",
+                    edge_index
+                )));
+            }
+        };
+        Ok(data)
+    }
+
+    /// Return the edge endpoints for the edge by it's given index
```
IvanIsCoding, line 469:
```suggestion
    /// Return the edge endpoints for the edge by its given index
```

IvanIsCoding (approved):
Minor grammar comments on the docstring, but apart from that LGTM

src/digraph.rs
```diff
@@ -1503,6 +1553,25 @@ impl PyDiGraph {
...
+
+    /// Return the list of edge indices incent to a provided node
+    ///
+    /// :param int node: The node index to get incident edges from. If
+    ///     this node index is not present in the graph this method will
+    ///     return an empty list and not error.
+    ///
+    /// :returns: A list of the edge indices incident to a node in the graph
+    /// :rtype: EdgeIndices
+    pub fn incident_edges(&self, node: usize) -> EdgeIndices {
```
georgios-ts, line 1565:
```suggestion
    #[pyo3(text_signature = "(self, node, /)")]
    pub fn incident_edges(&self, node: usize) -> EdgeIndices {
```

It should also be mentioned in the docs that this function returns only outgoing incident edges.

src/graph.rs
```diff
@@ -378,6 +378,120 @@ impl PyGraph {
...
 
+    /// Return the list of edge indices incent to a provided node
+    ///
+    /// :param int node: The node index to get incident edges from. If
+    ///     this node index is not present in the graph this method will
+    ///     return an empty list and not error.
+    ///
+    /// :returns: A list of the edge indices incident to a node in the graph
+    /// :rtype: EdgeIndices
+    pub fn incident_edges(&self, node: usize) -> EdgeIndices {
```
georgios-ts, line 389:
```suggestion
    #[pyo3(text_signature = "(self, node, /)")]
    pub fn incident_edges(&self, node: usize) -> EdgeIndices {
```

src/graph.rs
```diff
@@ -378,6 +378,120 @@ impl PyGraph {
...
+                .graph
+                .edges(NodeIndex::new(node))
+                .map(|e| e.id().index())
+                .collect(),
+        }
+    }
+
+    /// Get the endpoint indices and edge data for all edges of a node.
+    ///
+    /// This will return a list of tuples with the parent index the node index
```
georgios-ts, line 401:
```suggestion
    /// This will return a list of tuples with the parent index, the node index
```

src/graph.rs
```diff
@@ -378,6 +378,120 @@ impl PyGraph {
...
+        let raw_edges = self.graph.edges_directed(index, dir);
+        let out_list: Vec<(usize, usize, PyObject)> = raw_edges
+            .map(|x| (x.source().index(), node, x.weight().clone_ref(py)))
+            .collect();
+        WeightedEdgeList { edges: out_list }
+    }
+
+    /// Get the endpoint indices and edge data for all edges of a node.
+    ///
+    /// This will return a list of tuples with the child index the node index
```
georgios-ts, line 424:
```suggestion
    /// This will return a list of tuples with the child index, the node index
```

georgios-ts (approved):

src/digraph.rs
mtreinish: In https://github.com/Qiskit/retworkx/pull/538/commits/e737ca7531995665d1e16f589a6720b52e1e8700 I've updated the digraph method to have an `all_edges` option to return both outgoing and incoming edges and updated the documentation to reflect by default it's only outgoing but if you want incoming you can use that flag.

georgios-ts (commented):
How about adding a new method that returns all information (index, endpoints, data payload) of edges incident to a node similar to [`StableGraph::edges_directed`](https://docs.rs/petgraph/0.6.0/petgraph/stable_graph/struct.StableGraph.html#method.edges_directed)?

With the current API, `PyDiGraph.out_edges` does not return the edge index and with `PyDiGraph.incident_edges` we need two more function calls `PyDiGraph.get_edge_data_by_index`, `PyDiGraph.get_edge_endpoints_by_index`.

mtreinish:
> How about adding a new method that returns all information (index, endpoints, data payload) of edges incident to a node similar to [`StableGraph::edges_directed`](https://docs.rs/petgraph/0.6.0/petgraph/stable_graph/struct.StableGraph.html#method.edges_directed)?
> 
> With the current API, `PyDiGraph.out_edges` does not return the edge index and with `PyDiGraph.incident_edges` we need two more function calls `PyDiGraph.get_edge_data_by_index`, `PyDiGraph.get_edge_endpoints_by_index`.

I added another method `incident_edge_index_map()` in https://github.com/Qiskit/retworkx/pull/538/commits/d441098233db75dda7b490316c281f63f1370daa which returns the `EdgeIndexMap` for the incident edges.

szhorvat:
Two thoughts here:

 - Will it be possible to write generic code that works both for PyGraph and PyDiGraph? That would require a sufficiently similar API.
 - In many (perhaps most) applications for retrieving the edges incident to a vertex, one wants to know which is the _other_ endpoint of the edge. For example, if vertex `a` has incident edges `(a,b)` and `(c,a)`, then the other endpoints are `b` and `c`, respectively. One needs this to perform any kind of graph traversal. With some libraries, getting _the other vertex_ is a bit cumbersome.

mtreinish:
> Two thoughts here:
> 
>     * Will it be possible to write generic code that works both for PyGraph and PyDiGraph? That would require a sufficiently similar API.

The methods being added here are mostly identical between PyGraph and PyDiGraph, the only exception is for directionality with incident_edge* by default for PyDiGraph is only outgoing edges. This is probably an argument for changing the behavior of `all_edges` to by default return all directions and then make the user explicitly request either incoming or outgoing edges if that's what they need. I'll make this change in a bit.

> 
>     * In many (perhaps most) applications for retrieving the edges incident to a vertex in an _undirected_ graph, one wants to know which is the _other_ endpoint of the edge. For example, if vertex `a` has incident edges `(a,b)` and `(c,a)`, then the other endpoints are `b` and `c`, respectively. One needs this to perform any kind of graph traversal. With some libraries, getting _the other vertex_ is a bit cumbersome.

We already have an api for getting the list of indices of the other endpoints, if you call `g.neighbors(a)` it will return `[b, c]`

szhorvat:
What if I want to get all of: (1) edge index (2) edge payload (3) corresponding other endpoint at the same time? Should I use more than one of these functions, and can I rely on them returning the result in the same order?

An example task that requires all three pieces of information is implementing a random walk over the graph, which follows edges with probabilities proportional to their weight. The random walk function would take a starting node as input, and would return the sequence of edges it walked through. The result should be returned in a format that allows one to perform arbitrary operations on these edges, including retrieving their payload or deleting them. It seems to me that the only such format is a sequence of edge indices (anything else would lead to ambiguities in multigraphs). 

In fact I think I am going to add this task to the usability benchmarks (#532), as it's a good way to gauge the flexibility of a graph library. It also touches on the question of how to include self-loops in the list of incident edges: for the undirected random walk self-loops should be included twice (note that this is true for most applications, such as the random walk, but not for _any_ application).

If you can implement such a function (that works on both directed and undirected graphs), then your API is almost certainly flexible enough.

mtreinish:
Hmm, with that use case in mind I think it makes sense to keep the directed graph behavior the same as it is now. If you wanted to write a random walk function that worked with a PyDiGraph or PyGraph without any special casing that kinda reverses my thinking from the earlier comment because you want the outgoing edges only for `PyDiGraph` by default.

As for the undirected case I'll update the `incident_edge_index_map()` to return a fixed endpoint order so that `node` is always the first endpoint index. That way when you call:
```python
g.incident_edge_index_map(3)
```
it will return something like:
```python
{
  1: (3, 4, "value"),
  4: (3, 1, "different edge"),
  5: (3, 1, "parallel edge"),
}
```

PR #539, opened 2021-12-19
Fix future clippy errors

mtreinish:
Nightly rust, for 1.59, has added a couple of new clippy rules (or changed
existing rules) that will start causing an error in CI when rust 1.59 is
released. This commit fixes the issues identified by running nightly
clippy. These errors fall into 2 categories, replacing map().flatten()
with flat_map() and an unecessary late initialization.


IvanIsCoding (approved):
LGTM

PR #541, opened 2021-12-24
Start migration from node_indexes -> node_indices

mtreinish:
Since the earliest retworkx releases the method to return the list of
node indices in the graph was named "node_indexes". However as the
retworkx api evolved we standardized on the use of "indices" everywhere
else in the API. For backwards compatibility we kept the "node_indexes"
method name. This commit starts the process of migrating away from
"node_indexes" a new method "node_indices" is being added. The
"node_indexes" method is now just a backwards compatibility alias for
node_indices. In a future release when we start preparing for the
retworkx 1.0.0 release we will start emitting a deprecation warning
from node_indexes prior to its eventual removal in the 1.0.0 release.

### DIFF
--- a/releasenotes/notes/add-node-indices-3c7c37ed4ea75a2c.yaml
+++ b/releasenotes/notes/add-node-indices-3c7c37ed4ea75a2c.yaml
@@ -1,0 +1,11 @@
+---
+features:
+  - |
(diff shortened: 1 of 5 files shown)


IvanIsCoding (approved):
LGTM

Issue #505, opened 2021-12-02
[joss] cycle_basis() does not work with multigraphs


szhorvat:
`cycle_basis()` does not work with multigraphs. The following never returns:

```
g=rx.PyGraph()
g.extend_from_edge_list([(1,0),(1,2),(1,0)])
```

When using other multigraphs, the computation may finish, but the result does not look meaningful.

In general, there is a fundamental problem with the design of `cycle_basis()`: it returns cycles in term of nodes. However, cycles are in fact a collection of edges, not a collection of nodes. In most applications of cycle bases that I am aware of, one is working directly with edges, which this function does not supply. For example, how would you add two basis elements, in the format returned by this function? And of course, cycles in multigraphs are not representable as sequences of nodes.

Regarding the fact that the function never returns: Are retworkx function in general interruptible? Does the library have facilities for interrupting long calculations?

mtreinish:
Yeah, the algorithm we used for cycle basis doesn't work for multigraphs. It definitely should have a warning in the documentation about that. Ideally we'd also raise an error if a multigraph is passed in, but for backwards compatibility it's tricky to do that because if you create a graph with `g.multigraph == True` but without parallel edges today it works fine and we don't want to break current users.

As for the interrupt handling retworkx doesn't handle general interrupts. There is some complexity in doing it because of the rust/python boundary. To interrupt the rust function we basically have to call back into python during the execution of the rust code to check if there were any signals received and process it (see: https://pyo3.rs/v0.15.1/faq.html#ctrl-c-doesnt-do-anything-while-my-rust-code-is-executing ) otherwise python waits until the rust function returns before it processes the signal. This gets a bit tricky in practice for retworkx because of how the repository is split, we have a native rust library [`retworkx-core`](https://github.com/Qiskit/retworkx/blob/main/retworkx-core/README.md) where pure rust functionality lives. In those functions we don't have access to call python. But for the `cycle_basis()` case we can add a `check_signal` call to handle this.

szhorvat:
Could you please comment on the problem that this function returns lists of vertices (and not lists of edges)? All applications I can think of require the edges. As someone who uses cycle bases for their work, I would consider this a bug (I now see networkx does the same, and I would consider it a bug there as well).

I will open a new issue to discuss edge representation soon.

Regarding the interruptibility, I will also open a new issue, and we can continue the discussion there.

IvanIsCoding:
I'm closing this issue as we documented that the method does not work with multigraphs.

[...]

PR #543, opened 2022-01-05
Document cycle_basis() doesn't work with multigraphs

mtreinish:
As was pointed out in #505 the cycle_basis() function doesn't work with
multigraphs, but this was missing in the documentation. This commit adds
a note to the docstring that it won't work correctly if a multigraph is
used with it.


IvanIsCoding (approved):
LGTM

PR #544, opened 2022-01-06
Documentation fixes and improvements

### LINKED ISSUE #507: [joss] Awkward phrasings in the documentation
Here I will keep posting various documentation problems that are too small to deserve their own GitHub issue.

mtreinish:
This commit goes through the documentation for retworkx and fixes
various typos and mistakes. It also updates the generated api docs to be
put in an 'apiref' path instead of 'stubs' to clean up the urls a bit.
To maintain compatibility with existing links though a redirect is added
to redirect from stubs/* to apiref/*.

Fixes #507


mtreinish:
To start I just went through the issues outlined in #507 that weren't already fixed. But before we merge this I want to do a quick scan through the documentation to see if I catch anything else obvious there.

retworkx/__init__.py
```diff
@@ -648,12 +648,12 @@ def _graph_all_pairs_dijkstra_shortest_path(graph, edge_cost_fn):
...
-    """Find the shortest path from a node
+    """Calculate the shortest path lengths from all nodes in a graph
 
-    This function will generate the shortest path from a source node using
-    Dijkstra's algorithm. This function is multithreaded and will run
-    launch a thread pool with threads equal to the number of CPUs by default.
-    You can tune the number of threads with the ``RAYON_NUM_THREADS``
+    This function will generate the shortest path lengths from all nodes in the
+    graph using Dijkstra's algorithm. This function is multithreaded and will
+    run launch a thread pool with threads equal to the number of CPUs by
```
IvanIsCoding, line 655:
```suggestion
    launch a thread pool with threads equal to the number of CPUs by
```

src/shortest_path/mod.rs
```diff
@@ -297,10 +297,10 @@ pub fn digraph_dijkstra_shortest_path_lengths(
 /// Calculate the the shortest length from all nodes in a
 /// :class:`~retworkx.PyDiGraph` object
 ///
-/// This function will generate the shortest path from a source node using
-/// Dijkstra's algorithm. This function is multithreaded and will run
-/// launch a thread pool with threads equal to the number of CPUs by default.
-/// You can tune the number of threads with the ``RAYON_NUM_THREADS``
+/// This function will calculate the shortest path lengths from all nodes in the
+/// graph using Dijkstra's algorithm. This function is multithreaded and will
+/// run launch a thread pool with threads equal to the number of CPUs by
```
IvanIsCoding, line 302:
```suggestion
/// launch a thread pool with threads equal to the number of CPUs by
```

docs/source/sources.txt
```diff
@@ -0,0 +1,401 @@
+retworkx.adjacency_matrix.html
```
IvanIsCoding, line 1: This file seems autogenerated, should it be included in the source tree or not?

IvanIsCoding (commented):
Mostly looks good, I just want to have a minor discussion about `docs/source/sources.txt `: is it supposed to be autogenerated for all methods? Or is it supposed to be static i.e. only old methods with `stubs/method` will be redirected to the new URLs?

docs/source/sources.txt
mtreinish: It's needed in the tree. There's an ordering problem we don't know what the paths were for `stubs/` were (or what the apiref paths are) so we can't statically build a list of redirects in the configuration. This is going to just be a static list for redirects from old versions of the documentation and it won't ever grow. It's just so we know what to add a html redirect page for at the end of the doc build.

IvanIsCoding (approved):

docs/source/sources.txt
IvanIsCoding: I'm good with a static list, that should be on the tree indeed

retworkx/__init__.py
```diff
@@ -605,9 +605,9 @@ def _graph_dijkstra_shortest_path(
 
 @functools.singledispatch
 def all_pairs_dijkstra_shortest_paths(graph, edge_cost_fn):
-    """Find the shortest path from all nodes
+    """Find the shortest path from all nodes in a graph
```
kevinhartman, line 608:
```suggestion
    """For each node in the graph, finds the shortest paths to all others.
```

Just going off the original doc string, but perhaps this would be clearer.

retworkx/__init__.py
```diff
@@ -605,9 +605,9 @@ def _graph_dijkstra_shortest_path(
...
 
-    This function will generate the shortest path from all nodes int the graph
+    This function will generate the shortest path from all nodes in the graph
```
kevinhartman, line 610: Should we also mention what happens if there's a tie between multiple shortest paths?

retworkx/__init__.py
```diff
@@ -648,12 +648,12 @@ def _graph_all_pairs_dijkstra_shortest_path(graph, edge_cost_fn):
 
 @functools.singledispatch
 def all_pairs_dijkstra_path_lengths(graph, edge_cost_fn):
-    """Find the shortest path from a node
+    """Calculate the shortest path lengths from all nodes in a graph
```
kevinhartman, line 651:
```suggestion
    """For each node in the graph, calculates the lengths of the shortest paths to all others.
```
This wording assumes the key of the result map is the source and the key of the value map is the target.

kevinhartman (approved):
LGTM.

#507 mentions a few things that aren't solved here, like the inconsistency between `indices` versus `indexes` for nodes and edges. That change would require deprecation, I'd expect. Should we track that?

mtreinish:
> #507 mentions a few things that aren't solved here, like the inconsistency between `indices` versus `indexes` for nodes and edges. That change would require deprecation, I'd expect. Should we track that?

The start of that process was actually handled separately in: https://github.com/Qiskit/retworkx/pull/541

retworkx/__init__.py
mtreinish: Yeah, that's how the function works this is definitely more clear than what I had here

retworkx/__init__.py
```diff
@@ -605,9 +605,9 @@ def _graph_dijkstra_shortest_path(
...
 @functools.singledispatch
 def all_pairs_dijkstra_shortest_paths(graph, edge_cost_fn):
-    """Find the shortest path from all nodes
+    """Find the shortest path from all nodes in a graph
 
-    This function will generate the shortest path from all nodes int the graph
+    This function will generate the shortest path from all nodes in the graph
     using Dijkstra's algorithm. This function is multithreaded and will run
     launch a thread pool with threads equal to the number of CPUs by default.
     You can tune the number of threads with the ``RAYON_NUM_THREADS``
```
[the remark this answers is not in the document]
mtreinish: Yeah, that's wrong. It's of the format:

```
{
  source: {target: path, target: path}
}
```

I'll fix this now.

retworkx/__init__.py
mtreinish: Yeah, that's probably useful given the questions in #546. It's not the easiest to describe though because it depends on the exact graph. It will return one of the paths with a bias towards edges added more recently as that's the inner loop in the dijkstra implementation.

retworkx/__init__.py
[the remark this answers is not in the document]
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/544/commits/4f127b9690a11c6e520cbefcb529b8e5e12b64c6

PR #548, opened 2022-01-11
Add flag for attribute conversion when converting from networkx

InnovativeInventor:
- [x] I ran rustfmt locally
---

Currently, converting from a native `networkx` Graph to a `retworkx` PyGraph only preserves adjacency. This PR adds an *optional (and by default disabled) flag* to copy over the node attributes, thereby allowing for a more faithful conversion/translation.

After this PR, it will be possible to do something like this:

```python
In [1]: import networkx as nx

In [2]: import retworkx

In [3]: g = nx.Graph()

In [4]: g.add_node("1", test = True)

In [5]: g.add_node("2", test = True)

In [6]: g.nodes["1"]["test"]
Out[6]: True

In [7]: pygraph = retworkx.networkx_converter(g, keep_attributes=True)

In [8]: pygraph[1]["test"]
Out[8]: True

In [9]: pygraph[1]
Out[9]: {'test': True, '__networkx_name__': '1'}

In [10]: pygraph[0]  # indicies are not always preserved, so it's useful to have the name
Out[10]: {'test': True, '__networkx_name__': '2'}
```

For some context, I am considering adding a `retworkx` backend to a library that I contribute to ([GerryChain](https://github.com/mggg/GerryChain)) to replace our `networkx` usage. `retworkx` seems like an exciting project and may be a viable choice (as opposed to graph-tool or something similar).


tests/test_converters.py
```diff
@@ -18,90 +18,127 @@
...
+        for keep_attributes in [True, False]:
+            out_graph = retworkx.networkx_converter(
+                g, keep_attributes=keep_attributes
+            )
+            self.assertIsInstance(out_graph, retworkx.PyGraph)
+            self.assertEqual(list(out_graph.node_indexes()), list(g.nodes))
+            self.assertEqual(
+                out_graph.weighted_edge_list(), list(g.edges(data=True))
+            )
+            self.assertEqual(out_graph.multigraph, g.is_multigraph())
```
mtreinish, line 30: What do you think about using a `subTest` here (and in all the other test methods) for this? Something like:

```suggestion
        for keep_attributes in [True, False]:
            with self.subTest(keep_attributes=keep_attributes):
                out_graph = retworkx.networkx_converter(
                    g, keep_attributes=keep_attributes
                )
                self.assertIsInstance(out_graph, retworkx.PyGraph)
                self.assertEqual(list(out_graph.node_indexes()), list(g.nodes))
                self.assertEqual(
                    out_graph.weighted_edge_list(), list(g.edges(data=True))
                )
                self.assertEqual(out_graph.multigraph, g.is_multigraph())
```

This would let us see which variant was running in the error output. My other thought would be to use `ddt` to parameterize the methods, but that would require adding it to the test requirements in the tox.ini, so a `subTest` is probably simpler.

retworkx/__init__.py
```diff
@@ -1303,6 +1303,8 @@ def networkx_converter(graph):
         independently.
 
     :param networkx.Graph graph: The networkx graph to convert.
+    :param bool keep_attributes: If `True`, add networkx attributes to retworkx
+        object.
```
mtreinish, line 1307: Could you add a bit to this about what the data payloads for the nodes would look like when this is set to true. Maybe reword this to be something like?

```suggestion
    :param bool keep_attributes: If `True`, add networkx node attributes to the data
        payload to the nodes in the output retworkx graph. When set to ``True`` the
        node data payloads in the output retworkx graph object will be dictionaries
        with the node attributes from the input networkx graph where the
        ``"__networkx_name__"`` key contains the node object from the input networkx
        graph.
```

retworkx/__init__.py
```diff
@@ -1321,6 +1323,13 @@ def networkx_converter(graph):
             for x in graph.edges(data=True)
         ]
     )
+
+    if keep_attributes:
+        for node, node_index in node_indices.items():
+            attributes = graph.nodes[node]
+            attributes["__networkx_name__"] = node
```
mtreinish, line 1330: I'm wondering if networkx ``__network_name__`` is the best choice of key here. This will be the data payload for the node in the networkx graph right? It may not be a string or even have a `__str__()` or `__repr__()` method defined. IIRC the only requirement for a node in a networkx graph is that is hashable. Maybe something like ``__networkx_node__``?

mtreinish (commented):
Thanks for pushing this PR, it definitely looks like this is a useful feature for the networkx converter function. Heh, and we clearly didn't really think about node attributes when we first added it :). I left a few inline comments on some improvements that I think we can make here. Once we make those changes I think this is mostly ready.

Besides the inline comments do you mind adding a release note for this new flag? We're getting ready to push out the [0.11 release](https://github.com/Qiskit/retworkx/milestone/7) soon and making sure this is documented as a new feature in the release will be great. The process we use for release notes is documented here: https://github.com/Qiskit/retworkx/blob/main/CONTRIBUTING.md#release-notes

tests/test_converters.py
InnovativeInventor: Yeah -- I was thinking of ways to make parameterize the tests more prettily, but I didn't want to add more dependencies. `subTest` seems like it'd work reasonably well for now.

retworkx/__init__.py
InnovativeInventor: Fair point. I briefly considered trying to override the `__repr__` of the dict (or something like that), but decided against it.

retworkx/__init__.py
```diff
@@ -1303,6 +1303,14 @@ def networkx_converter(graph):
...
 
     :param networkx.Graph graph: The networkx graph to convert.
+    :param bool keep_attributes: If ``True``, add networkx node attributes to
+        the data payload in the nodes of the output retworkx graph. When set to
+        ``True``, the node data payloads in the output retworkx graph object
+        will be dictionaries with the node attributes from the input networkx
+        graph where the ``"__networkx_node__"`` key contains the node from the
+        input networkx graph.
+    :param bool keep_attributes: If `True`, add networkx attributes to retworkx
+        object.
```
mtreinish, line 1313: Looks like there was a bad merge this line was duplicated:

```suggestion
```

mtreinish (approved):
LGTM, thanks for the quick updates on this.

As for the release note since we're so close to the release I took the liberty of writing one for this in #549 (which is the release prep PR) so it'll be included in the documentation for 0.11.

PR #537, opened 2021-12-19
Add tutorials to documentation

mtreinish:
This commit starts the process of adding tutorials to the documentation.
Tutorials are guides on how you use retworkx to accomplish a specific
task or explain a concept in graph theory. In this commit 3 tutorials
are added to start:

 * An introduction which serves as a both an intro to graph theory
   concepts and on how to interact with retworkx
 * An introduction to directed acyclic graphs and their applications
 * A guide on how to calculate the betweenness centrality and visualize
   it

The expectation is that once we have a tutorials section that we'll
expand it with more tutorials moving forward.

TODO:

- [x] Finish introduction tutorial
- [x] Finish dag tutorial (only the qiskit section left)
- [x] Rebase after #538 merges
- [x] Rebase after #541 merges

PR #554, opened 2022-01-20
Add attribution comment to introduction tutorial

mtreinish:
In #537 we added the tutorials to the retworkx documentation. The
introduction tutorial was heavily based (and used some phrases
verbatim from that document) from the similar tutorial in NetworkX.
While the final result is different as it's specific to retworkx the
basic outline and some of the high level details are the same so we
should provide credit to the original source. I noticed reviewing the
published documentation for the 0.11.0 that I neglected to provide
proper attribution of that as part of #537. This corrects that oversight
and adds a comment to the top of the documentation source to link to the
original networkx document which it was based off of.

### DIFF
--- a/docs/source/tutorial/introduction.rst
+++ b/docs/source/tutorial/introduction.rst
@@ -1,3 +1,7 @@
+.. This document was adapted from and heavily based on the similar
+   introductutorial in  NetworkX's documentation which can be found here:
+   https://networkx.org/documentation/networkx-2.6.2/tutorial.html
+
 ########################
 Introduction to retworkx
 ########################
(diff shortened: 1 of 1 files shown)


docs/source/tutorial/introduction.rst
```diff
@@ -1,3 +1,7 @@
+.. This document was adapted from and heavily based on the similar
+   introductutorial in  NetworkX's documentation which can be found here:
```
mtreinish, line 2:
```suggestion
   introduction tutorial in  NetworkX's documentation which can be found here:
```

docs/source/tutorial/introduction.rst
```diff
@@ -1,3 +1,7 @@
+.. This document was adapted from and heavily based on the similar
```
mtreinish, line 1:
```suggestion
.. This document was adapted from and originally based on the similar
```

PR #556, opened 2022-01-20
Update and unify cibuildwheel configuration and version

mtreinish:
This commit updates the cibuildwheel configuration and version to the
latest release, 2.3.1. At the same time the configuration for
cibuildwheel is unified and simplified by leveraging the pyproject.toml
to store common cross-job options. Job specific options are still set as
environment variables in the job configuration (such as for cross build
settings). This should make the configuration easier to manage moving
forward as we only have to update a single location to adjust for global
setting changes.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -45,20 +45,9 @@ jobs:
           toolchain: stable
       - name: Install cibuildwheel
         run: |
-          python -m pip install cibuildwheel==2.1.2 twine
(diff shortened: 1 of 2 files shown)


mtreinish:
Marking as on hold while https://github.com/Qiskit/retworkx/pull/556/commits/b2724809c8ff1ac630cb917d6490577cc2072ec1 is applied so we can test the new configuration. Once everything is working in the build I'll revert that and remove the label.

PR #557, opened 2022-01-20
Replace most occurrences of "indexes" with "indices"

jlapeyre:
This PR removes most occurrences of "indexes" with "indices" in code, comments, and docs.

* Nothing in ./tests was changed. Perhaps all tests should be changed to use `node_indices` since this just wraps `node_indexes`.

* The two methods `node_indexes` that will likely be deprecated were, of course, not changed.

* release notes mentioning the addition of `node_indexes` were, of course, not changed.

### DIFF
--- a/docs/source/networkx.rst
+++ b/docs/source/networkx.rst
@@ -51,7 +51,7 @@ from the above retworkx example::
     assert 'my_node_a' == graph[node_a]
     assert 'my_node_b' == graph[node_b]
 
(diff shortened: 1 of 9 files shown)


mtreinish (approved):
LGTM, thanks for fixing this.

IvanIsCoding (approved):

IvanIsCoding:
Thanks for replacing the uses of `node_indexes`. If the function is deprecated we shouldn't use internally.

PR #558, opened 2022-01-20
Change style of footnote in DAG tutorial

IvanIsCoding:
Minor detail from #537 

Changes the style of the footnote in the DAG tutorial. Previously, the footnote was mixed with text and was hard to differentiate. This commit wraps it inside a note so it is more explicit the footnote is separate from the main text.

### DIFF
--- a/docs/source/tutorial/dags.rst
+++ b/docs/source/tutorial/dags.rst
@@ -71,8 +71,10 @@ overhead by using :meth:`.PyDiGraph.add_parent` and
 :meth:`.PyDiGraph.add_child` as both add a new node and edge simultaneously and
 neither can introduce a cycle.
 
-.. [#] A trail is a sequence of nodes and edges on the graph where each edge
-   is distinct in the walk.
+.. note::
+
+    .. [#] A trail is a sequence of nodes and edges on the graph where each edge
+       is distinct in the walk.
 
 Applications of DAGs
 ====================


docs/source/tutorial/dags.rst
```diff
@@ -71,8 +71,10 @@ overhead by using :meth:`.PyDiGraph.add_parent` and
 :meth:`.PyDiGraph.add_child` as both add a new node and edge simultaneously and
 neither can introduce a cycle.
 
-.. [#] A trail is a sequence of nodes and edges on the graph where each edge
-   is distinct in the walk.
+.. note::
```
mtreinish, line 74: I'm not sure we want to use a note directive here as it will render as a standalone box that says note. Maybe having a subsubsection header like:

```suggestion
Footnotes:
^^^^^^^^^^
```

Will clearly mark it as separate but avoid highlighting it in a new box

mtreinish (commented):
Sorry I forgot to look at this earlier it fell through the cracks in my review list. While I think we really need to fix the rendering of footnotes in https://github.com/Qiskit/qiskit_sphinx_theme directly coming up with a workaround for right now seems reasonable. That being said I'm not sure the note directive is the best choice here, I left an inline suggestion on an alternative approach we can try. Let me know what you think

IvanIsCoding:
> Sorry I forgot to look at this earlier it fell through the cracks in my review list. While I think we really need to fix the rendering of footnotes in https://github.com/Qiskit/qiskit_sphinx_theme directly coming up with a workaround for right now seems reasonable. That being said I'm not sure the note directive is the best choice here, I left an inline suggestion on an alternative approach we can try. Let me know what you think

I'm closing the PR because I'd rather fix it in the Sphinx theme

PR #559, opened 2022-01-21
Expand on platform support levels

mtreinish:
Following the similar change in Qiskit that defined the levels of
platform support (see:
https://qiskit.org/documentation/getting_started.html#platform-support),
this commit does a similar process for retworkx. This should make it a
bit clearer for users the level of support and expectations when running
retworkx on different platforms.


README.md
```diff
@@ -61,6 +61,81 @@ error about `setuptools-rust` not being found you should upgrade pip with
...
+binaries are built, tested, and published to PyPI as part of the release
+process. These platforms are expected to be installable with just a functioning
+Python environment.
+
+Tier 1 platforms are currently:
+
+- Linux x86_64 (distributions compatible with the manylinux 2010 packaging
+  specification)
+- macOS x86_64 (10.9 or newer)
+- Windows 64 bit
```
IvanIsCoding, line 90: I think a table along the lines of https://julialang.org/downloads/#supported_platforms would work here as well. I think it displays information better than listing it for each Tier

IvanIsCoding (commented):
🥇 This is a must in a project with wheels like retworkx, it is a very good addition. I have only a minor comment about using a table layout to display the Tiers, but appart from that it is ready to go into the documentation.

One thing I disagree however is to put it into the README. I feel our read me would be very long and honestly I'd rather keep it short. Maybe we can create a "Getting started" page like in https://qiskit.org/documentation/getting_started.html? That way we can split some installation details from the README

mtreinish:
> One thing I disagree however is to put it into the README. I feel our read me would be very long and honestly I'd rather keep it short. Maybe we can create a "Getting started" page like in https://qiskit.org/documentation/getting_started.html? That way we can split some installation details from the README

I put it here because we're currently using the readme in the documentation as the equivalent as a getting started page: https://qiskit.org/documentation/retworkx/README.html  this is the same contents as `README.md` and gets set to the "Overview and Installation" section in the table of contents. We probably should split this out as separate documents, but I did it this way originally because the information will basically be duplicated.

README.md
```diff
@@ -61,6 +61,81 @@ error about `setuptools-rust` not being found you should upgrade pip with
...
+
+
+#### Tier 2
+
+Tier 2 platforms are not tested upstream as part of development process.
```
jakelishman, line 95: It's a bit inconsistent whether you say "the development process" or "the development processes" in each tier - probably best to choose one and stick to it.  This one here in tier 2 is missing a definite article whichever way you go.

(I'm aware that this text is copy/pasted from Qiskit's version - I apparently missed it the first time around!)

README.md
```diff
@@ -61,6 +61,81 @@ error about `setuptools-rust` not being found you should upgrade pip with
...
+However, pre-compiled binaries are built, tested, and published to PyPI as part
+of the release process and these packages can be expected to be installed with
+just a functioning Python environment.
+
+Tier 2 platforms are currently:
+
+- Linux i686 (distributions compatible with the manylinux 2010 packaging
```
jakelishman, line 102: We linked to the PEPs for the manylinux specs in the Terra release notes - did you want to do the same in the docs here?

README.md
```diff
@@ -61,6 +61,81 @@ error about `setuptools-rust` not being found you should upgrade pip with
...
+Pre-compiled binaries are built, tested and published to PyPI as
+part of the release process. However, they may not installable with just a
+functioning Python environment and you may be required to build Numpy from
+source, which requires a C/C++ compiler, as part of the installation process.
+
+Tier 3 platforms are currently:
+
+- Linux i686 (distributions compatible with the manylinux 2010 packaging
+  specification) for Python >= 3.10
+- Windows 32bit for Python >= 3.10
```
jakelishman, line 120: Strictly I think you _only_ build for Python 3.10, so the "> 3.10" part isn't entirely true.  Not sure how precise you care about being, though.

README.md
mtreinish: I added links to the new table in: https://github.com/Qiskit/retworkx/pull/559/commits/b365490f698922b09dfd466f5844b76d8a8bbfcc

README.md
mtreinish: I settled on the "the development process" in: https://github.com/Qiskit/retworkx/pull/559/commits/e09e02133053b7a84d99d7fdb7fa608c34742ee5

README.md
mtreinish: I left it this way because I wanted it to be future proof, we're not going to get numpy support for i686 back with 3.11. We might not support 3.11 yet in retworkx but we will at some point and I don't feel like updating this table just to add the `>` in the PR we add 3.11 support.

mtreinish:
Superseded by: https://github.com/Qiskit/rustworkx/pull/657

PR #561, opened 2022-02-03
Set max line length to 100 characters in retworkx-core

IvanIsCoding:
Follow up to #490 

This PR formats retworkx-core with line width of 100 characters. On #490, we accidentally left the line width for the retworkx-core as 80. The PR will fix the inconsistency


mtreinish (approved):
LGTM, thanks for catching this I clearly forgot about retworkx-core when I rebased #490

PR #562, opened 2022-02-04
Add mergify config and document how it's used

mtreinish:
This commit adds a mergify configuration to the repoistory. Mergify will
let us automate PR merging and backporting to stable branches. This
should hopefully simplify the maintainence of the repository as core
reviewers now we only need to tag a PR as "automerge" and the bot will
handle rebasing and merging for us. Additionally, it will automate the
backporting of PRs for us. This will enable us to maintain stable bugfix
releases more easily and hopefully avoid for tension like we had in
0.11.0 between getting a release out for Python 3.10 support and merging
the required features.

To facilitate the use of all these new tools this commit adds detailed
documentation to the contributing guide on code review, the use of
mergify, stable branches, and backporting.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,36 @@
+queue_rules:
+  - name: automerge
+    conditions:
+      - check-success=python3.7-x64 windows-latest
+      - check-success=python3.7-x64 ubuntu-latest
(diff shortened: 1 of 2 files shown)


IvanIsCoding (approved):
LGTM, let's let mergify do most of the backporting work

PR #566, opened 2022-02-17
Run CI on stable branches too

mtreinish:
This commit fixes an oversight in our ci configuration. Starting after the
0.11.0 release we started maintaining a stable branch in parallel to our
main development branch. This branch is used to backport bugfixes from
main so that we can push out bugfix releases independently of our main
development branch. This reduces the pressure pressure to release as
frequently just to get bug or compatibility fixes out and we can develop
new features on our main branch and release when it's ready. However,
as part of this migration we neglected to updated the ci configuration
to run on the new stable branches. This commit fixes this oversight so
that we will run ci on backports in addition to our main development
branch.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,9 +2,9 @@
 name: CI Jobs
 on:
   push:
-    branches: [ main ]
+    branches: [ main, 'stable/*' ]
   pull_request:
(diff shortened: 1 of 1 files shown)


IvanIsCoding (approved):
LGTM

PR #569, opened 2022-03-04
Upgrade PyO3 and numpy to 0.16.2

mtreinish:
This commit updates pyo3 and numpy to the latest release 0.16.0. The
details on these releases can be found here:

https://pyo3.rs/v0.16.0/changelog.html

and

https://github.com/PyO3/rust-numpy/blob/main/CHANGELOG.md

Unfortunately there are a number of breaking changes and deprecations we
will have to update our usage to account for. This commit starts the
process by fixing the compiler errors due to changes in behavior a
subsequent commit will address the deprecation warnings.


IvanIsCoding:
```
Deprecate the #[pyproto] traits. (https://github.com/PyO3/pyo3/pull/2173)
```

I think that will be the most amount of work from moving to PyO3 0.16 as most of our custom return types use `#[pyproto]`. The removal is only going to happen around the PyO3 0.18 release, but it would good to move from #[pyproto]` to `#[pymethod]` before that.

mtreinish:
Yeah, that's what my next step was to tackle all of those pyproto deprecation warnings. They will fail the clippy run in ci without fixing all the deprecation warnings (it's why I tagged it on hold). But I wanted to push up what I had after tracking down all the private pyfunctions.

IvanIsCoding:
> Yeah, that's what my next step was to tackle all of those pyproto deprecation warnings. They will fail the clippy run in ci without fixing all the deprecation warnings (it's why I tagged it on hold). But I wanted to push up what I had after tracking down all the private pyfunctions.

We can temporarily use `#![allow(deprecated)]` (along the lines of https://github.com/PyO3/pyo3/pull/2173) and open a separate issue to migrate away from `#[pyproto]`

georgios-ts:
I just pushed 58ab94e to fix pyproto deprecation warnings. For the most part the changes were straightforward. I just had to redefine our own version of the deprecated `pyo3::gc::PyGCProtocol` since it's the most convenient way to adjust the behavior of each class regarding garbage collection.

Issue #440, opened 2021-09-08
Expand graph operations


georgios-ts:
### What is the expected enhancement?
Graph operations  produce new graphs from initial ones. Right now retworkx can `compose`, take the `union` of two graphs or return the `complement` of a graph. We can expand this list similar to  NetworkX functionality https://networkx.org/documentation/stable/reference/algorithms/operators.html.

nahumsa:
Since that are a lot of graph operations that could be implemented, I think it is useful to have a summary of which operations are already implemented, which could help not to have more than one person working on the same operation. Let me know if I missed something.

Operations:
- [x] Complement
- [ ] Reverse
- [x] Compose
- [x] Union
- [ ] Disjoint union
- [ ] Intersection
- [ ] Difference
- [ ] Symmetric difference (@e-eight)
- [ ] Full_join
- [x] Cartesian product
- [ ] Lexicographic Product
- [ ] Rooted Product
- [ ] Strong Product
- [ ] Tensor Product (#472)
- [ ] Power

PR #571, opened 2022-03-10
Add Graph difference

nahumsa:
Related to #440 

Adds the difference between two graphs for `PyDiGraph` and `PyGraph`.


src/difference.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    let mut final_graph = StablePyGraph::<Ty>::with_capacity(
+        first.node_count(),
+        first.edge_count() - second.edge_count(),
+    );
+
+    for node in first.node_indices() {
+        let weight = &first[node];
+        final_graph.add_node(weight.clone_ref(py));
+    }
```
mtreinish, line 46: What happens if the node weights are different between first and second? Do we care at all that if `first[0] != second[0]` we only preserve the payload for `first[0]`?

src/difference.rs
```diff
@@ -0,0 +1,114 @@
...
+            None => final_graph.add_edge(e.source(), e.target(), e.weight().clone_ref(py)),
+        };
+    }
+
+    Ok(final_graph)
+}
+
+/// Return a new PyGraph that is the difference from two input
+/// PyGraph objects
+///
```
mtreinish, line 62: I think the big thing missing here is the constraints on the input types (they have to have identical node indices). This probably should mention that and also maybe have an example and/or an explanation of how the difference is computed.

mtreinish (commented):
Overall this LGTM I think this is pretty close, I just have a small question inline about the node weights. The only thing i think that's really missing is in the docstrings for the new function is an explanation of what the difference is, an example might go along way to show what the function does and how it works.

src/difference.rs
nahumsa: Do you think we should have a callback to handle weights from both graphs?

georgios-ts:
Users have only indirect control over node indices since we assign them internally. In [`union`](https://github.com/Qiskit/retworkx/blob/main/src/union.rs#L129) we distinguish two nodes/edges based on their data payloads and it's probably better to do the same in this PR too.
In more detail I'm thinking that we should:
1. Verify that we can match every node of the first graph into a unique node in the second graph with the same data payload.
2. Remove an edge `e = (u, v)` from the first graph if the second graph has an edge with endpoints `e' = (m[u], m[v])`, where `m`  is the matching found in step (1.), *and* the edge data payloads agree.

winstxnhdw:
Hey @IvanIsCoding, do you know what else is blocking this PR? I could take over.

IvanIsCoding:
> Hey @IvanIsCoding, do you know what else is blocking this PR? I could take over.

This PR hasn’t been rebased for 3+ years, so feel free to take over but I’d strongly recommend to:
1. Start fresh
2. Use this as a reference 
3. 

The Python code is trivial to copy. The Rust code will be the hard part as PyO3 has changed a lot of leve the years

IvanIsCoding:
> Users have only indirect control over node indices since we assign them internally. In [`union`](https://github.com/Qiskit/retworkx/blob/main/src/union.rs#L129) we distinguish two nodes/edges based on their data payloads and it's probably better to do the same in this PR too. In more detail I'm thinking that we should:
> 
> 1. Verify that we can match every node of the first graph into a unique node in the second graph with the same data payload.
> 2. Remove an edge `e = (u, v)` from the first graph if the second graph has an edge with endpoints `e' = (m[u], m[v])`, where `m`  is the matching found in step (1.), _and_ the edge data payloads agree.

@winstxnhdw those were the blockers I assume

PR #572, opened 2022-03-19
Implement biparition_tree function

InnovativeInventor:
Note: this is a followup from email correspondence that @mtreinish and I had a few months ago. This is a draft PR, so feedback is welcome and it's not ready for merging (yet). It's fairly non-performant (but enough to beat Python/networkx solidly).

This function takes in a graph with population assigned to each node and draws a minimum spanning tree and finds a cut edge that splits the tree into two partitions that have total populations within some epsilon. 

To explain the motivation behind this PR, this function is the main workhorse behind the ReCom algorithm detailed in [this paper](https://hdsr.mitpress.mit.edu/pub/1ds8ptxu/release/4), which has been used in many litigation and civil rights projects to challenge racial and partisan gerrymandered maps and determine VRA compliance (see the recent cases in Alabama, North Carolina, Pennsylvania, etc.). I'm working on a rewrite of [MGGG](https://mggg.org/)'s main gerrymandering analysis software/engine ([GerryChain,](https://github.com/mggg/GerryChain) written in Python) and have achieved a ~15x speedup by naively rewriting the core graph operations in retworkx. 

You can still see my messy debugging, but the rough idea is here.


src/tree.rs
```diff
@@ -135,3 +138,159 @@ pub fn minimum_spanning_tree(
...
+///
+#[pyfunction]
+#[pyo3(text_signature = "(graph, weight_fn, pop, target_pop, epsilon)")]
+pub fn bipartition_tree(
+    py: Python,
+    graph: &graph::PyGraph,
+    weight_fn: PyObject,
+    py_pops: &PyList,
+    py_pop_target: &PyFloat,
+    py_epsilon: &PyFloat,
```
mtreinish, line 279: For these you can let the macro handle the conversion from python to rust for us. Something like:

```suggestion
    py_pops: Vec<f64>,
    py_pop_target: f64,
    py_epsilon: f64,
```

Then we can avoid the unsafe and gil manipulation in the code here and just work purely in rust. I'd probably rename all the args here to drop the `py_` prefix.

src/tree.rs
```diff
@@ -135,3 +138,159 @@ pub fn minimum_spanning_tree(
...
+///
+/// :returns: A set of nodes in one half of the spanning tree
+///
+#[pyfunction]
+#[pyo3(text_signature = "(graph, weight_fn, pop, target_pop, epsilon)")]
+pub fn bipartition_tree(
+    py: Python,
+    graph: &graph::PyGraph,
+    weight_fn: PyObject,
+    py_pops: &PyList,
```
mtreinish, line 277: If this is readonly I would suggest using a numpy array input type here:

```suggestion
    py_pops: PyReadonlyArray1<f64> ,
```

Then we can convert this to a slice with `let pops_slice = py_pops.as_slice()?;`, they we can pass `&[f64]` to the inner `balanced_cut_edge()` call and everything is passed by reference from the inner c array in numpy so we have very little overhead here. Using a list (or a vec as in my other comment) you have iteration and copy/conversion overhead going from a list of floats in python to a Vec<f64> in rust.

src/tree.rs
```diff
@@ -135,3 +138,159 @@ pub fn minimum_spanning_tree(
...
+            }
+            same_partition_tracker[leaf_node.index()].push(leaf_node.index());
+        }
+
+        // eprintln!("leaf nodes: {}", node_queue.len());
+
+        // this process can be multithreaded, if the locking overhead isn't too high
+        // (note: locking may not even be needed given the invariants this is assumed to maintain)
+        let mut balanced_nodes: Vec<(usize, Vec<usize>)> = vec![];
+        let mut seen_nodes: Vec<bool> = vec![false; spanning_tree_graph.node_count()]; // todo: perf test this
```
mtreinish, line 199: Normally for this we use a `HashSet` of `usize` (you can preallocate it with `with_capacity()` too) and then just do a lookup to see if it's been seen before.

mtreinish (commented):
I know this is still a WIP and there is still some work to go on it, but I took a quick look and there are some easy optimizations you can make to improve the performance I commented on inline. I'll wait till it's closer to ready to do a more detailed review.

src/tree.rs
InnovativeInventor: Yep -- I had a version of this with a HashSet -- I think I was testing this out for performance, but forgot to check if it actually improved performance (it probably doesn't). 

I think the idea I had was that it doesn't necessarily make sense to use hash tables if it's just a set of integers from 0 to n that will eventually all be in the set. Not sure if this actually makes a difference, so I was curious to see what happened.

mtreinish: Sure that makes sense, it's worth testing at least.

src/tree.rs
InnovativeInventor: Hmm, I'm getting this error when I try that: 
```
the trait `PyClass` is not implemented for `&[f64]`
```
is there something else I should be passing?

mtreinish: oh that was my bad on the earlier comment I missed that `balance_cut_edge()` was also a pyfunction I don't think you can have `&[f64]` as an input type to a pyfunction because the pyo3 macro can convert to that directly. We can just pass the `PyReadOnlyArray1` directly to `balance_cut_edge()` and then do the slice cast call in `balance_cut_edge()`

InnovativeInventor: Ok. I just realized that this isn't read-only because I re-use the same vector to accumulate populations during each loop (when I do node contraction). Do you have any suggestions on minimizing the number of copies/memory allocs made per loop in this case?

InnovativeInventor: To clarify: I don't seem to be able to pass mutable vector references to pyfunctions, so the current code copies the vector over twice (I think). E.g. I want to get rid of this: https://github.com/Qiskit/retworkx/pull/572/files#diff-9794019b962b882943346745baf52516c9f655c4198f992f64bfdb6bd545f9d2R172

src/tree.rs
InnovativeInventor: Ok, I think switching to `HashSet` is slightly less performant, but is worth the reduced reading headache. I've removed this rather janky code to make it easier to maintain.

src/tree.rs
InnovativeInventor: Ah, figured it out.

InnovativeInventor:
I just gave this a rebase and cleaned up the code a bit -- should be ready for your review. Let me know what you think!

IvanIsCoding (commented):
Thanks for contributing this, we're glad retworkx was useful for your research!

Some things are missing before we are able to merge the PR, an action list would  be:
* Write tests for the functions you added in `tests/`
* Add a release note announcing the new method using `reno new bipartition_tree`
* Add an entry of the new function on [`docs/source/api.rst`](https://github.com/Qiskit/retworkx/blob/main/docs/source/api.rst)


You can find more details of the steps above in [CONTRIBUTING.md](https://github.com/Qiskit/retworkx/blob/main/CONTRIBUTING.md). Feel free to ping me or Matthew if you need help

releasenotes/notes/bipartition_tree-4c1ad080b1fab9e8.yaml
```diff
@@ -0,0 +1,22 @@
...
+    that have population near the population target within some epsilon. The 
+    function returns a list of all such possible cuts, represented as the set 
+    of nodes in one partition/subtree. For example::
+
+          balanced_node_choices = retworkx.bipartition_tree(
+              tree,
+              pops,
+              float(pop_target),
+              float(epsilon)
+          )
```
IvanIsCoding, line 17: Can you wrap it in a code block so it looks nice when we render the release notes?

tests/graph/test_bipartition.py
```diff
@@ -0,0 +1,129 @@
...
+            3.0,
+            0.2,
+        )
+        self.assertEqual(len(balanced_edges), 1)
+
+        # Since this is already a spanning tree, bipartition_graph should
+        # behave identically. That is, it should be invariant to weight_fn
+        graph_balanced_edges = retworkx.bipartition_graph(
+            self.tree,
+            lambda x: random.random(),
```
IvanIsCoding, line 70: Can you avoid using `random.random()` and maybe just return `lambda _: 1.0`? I know it doesn't in this case, but in general we try to always have deterministic callbacks and seed the random when necessary

tests/graph/test_bipartition.py
```diff
@@ -0,0 +1,129 @@
...
+            self.tree,
+            [1.0, 1.0, 1.0, 1.0, 1.0, 1.0],
+            3.0,
+            0.5,
+        )
+        self.assertEqual(len(balanced_edges), 1)
+
+        graph_balanced_edges = retworkx.bipartition_graph(
+            self.tree,
+            lambda x: random.random(),
```
IvanIsCoding, line 88: Recurring comment

tests/graph/test_bipartition.py
```diff
@@ -0,0 +1,129 @@
...
+            self.line,
+            [1.0, 1.0, 1.0, 1.0, 1.0, 1.0],
+            3.0,
+            0.5,
+        )
+        self.assertEqual(len(balanced_edges), 3)
+
+        graph_balanced_edges = retworkx.bipartition_graph(
+            self.line,
+            lambda x: random.random(),
```
IvanIsCoding, line 106: Recurring comment

tests/graph/test_bipartition.py
```diff
@@ -0,0 +1,129 @@
...
+            self.line,
+            [1.0, 1.0, 1.0, 1.0, 1.0, 1.0],
+            3.0,
+            0.01,
+        )
+        self.assertEqual(len(balanced_edges), 1)
+
+        graph_balanced_edges = retworkx.bipartition_graph(
+            self.line,
+            lambda x: random.random(),
```
IvanIsCoding, line 124: Recurring comment

IvanIsCoding (commented):
Left some minor comments but overall the code is looking good

src/tree.rs
```diff
@@ -135,3 +149,125 @@ pub fn minimum_spanning_tree(
...
     Ok(spanning_tree)
 }
+
+/// Bipartition tree by finding balanced cut edges of a spanning tree using
+/// node contraction. Assumes that the tree is connected and is a spanning tree.
+/// A balanced edge is defined as an edge that, when cut, will split the
+/// population of the tree into two connected subtrees that have population near
+/// the population target within some epsilon. The function returns a list of
+/// all such possible cuts, represented as the set of nodes in one
+/// partition/subtree.
```
mtreinish, line 159: Did you have any citations here or other links to the use or source of this algorithm you'd want to add to the documentation?

mtreinish (commented):
Overall this LGTM, I'll have to read the paper in more depth to check the implementation matches what's expected but I trust you did that correctly if I don't have to time to check that. :) I just have a few more mechanical questions and suggestion inline but I think this is getting close. Thanks for sticking with this and continuing to push it forward.

src/tree.rs
```diff
@@ -11,6 +11,8 @@
 // under the License.
 
 use std::cmp::Ordering;
+use std::collections::HashSet;
```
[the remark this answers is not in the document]
InnovativeInventor: Ok, I was under the impression that hashbrown's `HashSet` is now the default `HashSet`.

From the [hashbrown README](https://github.com/rust-lang/hashbrown):
> Since Rust 1.36, this is now the HashMap implementation for the Rust standard library. However you may still want to use this crate instead since it works in environments without std, such as embedded systems and kernels.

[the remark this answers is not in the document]
mtreinish: It is the hashbrown crate is the upstream for the std lib hashset and hashmap. For our purposes what the crate does that's different from the std version is use ahash as the default hash algorithm and also has optional support for rayon (it also has additional options like nostd, etc which we don't leverage here).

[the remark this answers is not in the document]
InnovativeInventor: Ok, that makes sense -- thanks for explaining.

releasenotes/notes/bipartition_tree-4c1ad080b1fab9e8.yaml
InnovativeInventor: Ok, just did this.

tests/graph/test_bipartition.py
InnovativeInventor: Ok, just did this -- determinism is good.

src/tree.rs
```diff
@@ -135,3 +149,125 @@ pub fn minimum_spanning_tree(
...
+    _py: Python,
+    spanning_tree: &graph::PyGraph,
+    pops: Vec<f64>,
+    pop_target: f64,
+    epsilon: f64,
+) -> PyResult<Vec<(usize, Vec<usize>)>> {
+    let mut pops = pops;
+    let spanning_tree_graph = &spanning_tree.graph;
+    let mut same_partition_tracker: Vec<Vec<usize>> =
+        vec![vec![]; spanning_tree_graph.node_count()]; // Keeps track of all all the nodes on the same side of the partition
```
mtreinish, line 182: The other thing I just realized is how do you handle holes in the node indices here? If a node is removed that index will be missing from `graph.node_indices()` with `PyGraph`. For example if you do:

```python
import retworkx
g = retworkx.PyGraph()
g.add_nodes_from(range(5))
g.remove_node(2)
print(g.node_indices())
```

will return `[0, 1, 3, 4]`.

I think you need to adjust this to be `node_bound()` which is the `max(index)` in the node indices list instead of the number of nodes in the graph. There might be other adjustments you need to do to account for this.

src/tree.rs
```diff
@@ -126,6 +128,18 @@ pub fn minimum_spanning_tree(
...
+}
+
+/// Helper function to allow reuse of spanning_tree object to reduce memory allocs
+fn _minimum_spanning_tree<'a>(
+    py: Python,
+    graph: &graph::PyGraph,
+    spanning_tree: &'a mut graph::PyGraph,
+    weight_fn: Option<PyObject>,
+    default_weight: f64,
+) -> PyResult<&'a mut graph::PyGraph> {
```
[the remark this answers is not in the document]
mtreinish: Sorry, I misread the code here when I left the comment. Looking at the changes you just made I didn't realize that making this change would require changing the `bipartition_tree` function to return the spanning tree so that `bipartition_graph` could use the graph on each iteration. When looking at `bipartition_tree` as a standalone python function now it's api seems less than ideal because of that tuple return type.

When I left this comment earlier I was only concerned about the `minimum_spanning_tree()` -> `_minimum_spanning_tree()` call path and it looked odd to me to have a function that takes a mutable reference and then returns a mutable reference to the same object. Normally when you require a mutable borrowed input parameter you don't need to return that reference as part of the function signature because the caller already has the same reference.Maybe instead of passing by value here you can restore the mutable reference here and just change the return type to be `PyResult<()>` so that we return errors. That would retain the user facing pyfunction api for `bipartition_tree` but also avoid the double reference. The other thing I was thinking is maybe make an internal `_bipartition_tree` that works with rust native types so we can develop the python interface in isolation from the internal rust calling of that function. But, I'll defer to what pattern you think is best here as you're the first/primary user of these functions so you're in the best position to evaluate that making sure the API we end up with here is usable.

src/tree.rs
InnovativeInventor: Ok, thanks for pointing this out -- just changed it to be `node_bound()`.

src/tree.rs
```diff
@@ -135,3 +149,125 @@ pub fn minimum_spanning_tree(
...
+///     two partitioned subtrees and the set of nodes making up that subtree.
+#[pyfunction]
+#[pyo3(text_signature = "(spanning_tree, pop, target_pop, epsilon)")]
+pub fn bipartition_tree(
+    _py: Python,
+    spanning_tree: &graph::PyGraph,
+    pops: Vec<f64>,
+    pop_target: f64,
+    epsilon: f64,
+) -> PyResult<Vec<(usize, Vec<usize>)>> {
```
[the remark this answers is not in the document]
InnovativeInventor: Hmm, what would a custom return type look like? Is this implemented in `retworkx` for other funcs?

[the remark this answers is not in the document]
mtreinish: It's pretty easy to do you just add a macro call to `iterators.rs` to define the custom return type which just outlines the name, fields, types, docs, etc for each sequence type built from a vec. For example: https://github.com/Qiskit/retworkx/blob/main/src/iterators.rs#L619-L654 builds the custom `EdgeList` type from a `Vec<(usize, usize)>`. The only complication here is with the `(usize, Vec<usize>)` for each element you'd probably want to nest things so we don't do a copy on the inner vec too and instead return a tuple of `(usize, NodeIndices)` for each `__getitem__` call.

src/tree.rs
```diff
@@ -135,3 +149,125 @@ pub fn minimum_spanning_tree(
...
+///
+/// :returns: A list of tuples, with each tuple representing a distinct
+///     balanced edge that can be cut. The tuple contains the root of one of the
+///     two partitioned subtrees and the set of nodes making up that subtree.
+#[pyfunction]
+#[pyo3(text_signature = "(spanning_tree, pop, target_pop, epsilon)")]
+pub fn bipartition_tree(
+    _py: Python,
+    spanning_tree: &graph::PyGraph,
+    pops: Vec<f64>,
```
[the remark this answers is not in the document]
InnovativeInventor: Ok, that makes sense. But, if `pops` needs to get reset on every loop, isn't copying the behavior that we want?

src/tree.rs
InnovativeInventor: This balanced cut edge finding algo is not actually specified in the literature, I believe (the paper I linked to uses a naive `O(e^2)` algo that calculates the population balance for each possible edge and throws away the unsuitable cut edges, whereas the one I implemented is `O(v+e)` and does BFS traversal). I can link to the original Python implementation, though, if you'd like.

src/tree.rs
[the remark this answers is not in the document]
InnovativeInventor: Ok, if I were to do this:
```python
customvec[100]
```
would the `custom_vec_iter_impl` macro build up/calculate `customvec[0]`, `customvec[1]`, ...,`customvec[100]`? If so, I'm not sure how much performance improvement this will yield.

src/tree.rs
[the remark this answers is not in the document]
InnovativeInventor: Ok, I changed the return type to be `PyResult<()>`, as per your suggestion. I'd like to have `bipartition_tree` exposed, since sometimes we want to use Wilson's Uniform Spanning Tree algo or other spanning tree methods (active area of research), which we already have implemented in Python, so keeping the ability to pass a custom spanning tree would be nice.

If alternative spanning tree methods become more widely used in real-world redistricting efforts, I may want to port them to Rust/retworkx as well. I've renamed `bipartition_tree` to `bipartition_tree_mst` to reflect the possibility that other spanning tree methods may be useful to have in the future.

src/tree.rs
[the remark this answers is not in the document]
mtreinish: It would only return the element at index 100. Basically the way it works is that it builds a `pyclass` struct that has a single `Vec` field. So if you did something like:
```rust
custom_vec_iter_impl!(
    customvec,
    inner_vec,
    (usize, Vec<usize>),
    "My Docstring"
);
```
that expands to something like:

```rust
#[doc = $doc]
#[pyclass(module = "retworkx")]
#[derive(Clone)]
pub struct customvec {
        pub inner_vec: Vec<(usize, Vec<usize>)>,
}
```

With the python sequence protocol implemented so `fn __getitem__(self, idx) -> (usize, Vec<usize>)` just is basically returning `self.inner_vec[idx]` and only does the conversion for that single element.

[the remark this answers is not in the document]
mtreinish: As for the complexity I was briefly talking about before is that using `(usize, Vec<usize>)` as the data type will still do the iteration and conversion on the whole inner `Vec<usize>` inside the tuple. You can avoid that though by nesting custom returns so that you make the data type `(usize, NodeIndices)` instead of `(usize, Vec<usize>)`.

src/tree.rs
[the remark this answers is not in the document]
mtreinish: Oh, I just meant that if it made it easier to work with you could have `bipartition_tree()`/`bipartion_tree_mst()`  and `bipartition_graph()` call `_bipartition_tree()` if that gave you more flexibility around the internal rust types (because we wouldn't need to worry about having types that directly have `FromPyObject` implemented). But it's not strictly necessary, just me thinking out loud.

src/tree.rs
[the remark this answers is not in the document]
InnovativeInventor: Ah, I see. I'm implementing it now, then -- sounds like I should see a speedup with this method.

src/tree.rs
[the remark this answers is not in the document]
mtreinish: For `bipartition_graph` that makes sense but you could still use `PyReadOnlyArray1` there too and avoid the python overhead for the initial copy. So you do something like:

```rust
pub fn bipartition_graph(
    py: Python,
    graph: &graph::PyGraph,
    weight_fn: PyObject,
    pops: PyReadOnly1<f64>
    pop_target: f64,
    epsilon: f64,
) -> PyResult<Vec<(usize, Vec<usize>)>> {
  let pops_slice: &[f64] = pops.as_slice()?;
  while balanced_nodes.is_empty() {
      _bipartition_tree(.., &mut pops.to_vec(), ...);
  }
```
(assuming you split the inner implementation and the python interface into separate functions so you can pass `&mut [f64]`)

I haven't tested the above but something like that should work. So for the standalone `bipartition_tree_mst` pyfunction you use the `unsafe` function to get a mutable slice and for `bipartition_graph` you get an immutable slice and copy it on each iteration as before but you avoid the overhead of the initial python -> rust conversion (which requires iterating over the sequence in python to create the `Vec<f64>`).

src/tree.rs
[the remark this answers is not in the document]
InnovativeInventor: Huh, I don't see a speedup (in fact, it gets slower). Strange.

src/tree.rs
mtreinish: No that's fine. I was thinking it would be nice additional context for the documentation if it was available. But it's not critical and I think the description here is probably sufficient. Maybe having an example to show how it's used and what the expected result looks like would be nice. But not worth blocking over either.

src/tree.rs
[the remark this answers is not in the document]
mtreinish: Hmm, interesting. That probably means that on the python side you're iterating over everything >=1 times. Doing custom return types like this is a tradeoff between returning fast and adding overhead on per element access.

src/tree.rs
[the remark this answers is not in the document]
InnovativeInventor: Huh, I'm getting:
```
error[E0599]: no method named `clone` found for slice `[f64]` in the current scope
   --> src/tree.rs:293:73
    |
293 |             balanced_nodes = _bipartition_tree(&mst, &mut (*pops_slice).clone(), pop_target, epsilon);
    |                                                                         ^^^^^ method not found in `[f64]`
 ```
Strange -- I thought `.clone()` was implemented for all array types now.

File: src/tree.rs (hunk 13 above)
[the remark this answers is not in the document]
Comment (InnovativeInventor) [reply src/tree.rs:179]: Yeah, but all I do is randomly pick (uniformly) one of the balanced edges. Maybe `random.choice` relies on `__len__` which forces it to be fully enumerated.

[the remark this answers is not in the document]
Comment (InnovativeInventor) [reply src/tree.rs:179]: Ah, it does use `__len__`: https://github.com/python/cpython/blob/75eee1d57eb28283a8682a660d9949afc89fd010/Lib/random.py#L371

It looks like `__len__` is automatically/efficiently implemented by the `custom_vec_iter_impl!` macro. This is even stranger.

File: src/tree.rs (hunk 14 above)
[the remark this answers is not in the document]
Comment (InnovativeInventor) [reply src/tree.rs:176]: Import `std::slice` and `std::array` doesn't seem to help.

File: src/tree.rs (hunk 13 above)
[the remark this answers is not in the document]
Comment (InnovativeInventor) [reply src/tree.rs:179]: For context, this is how it's being used:
```python
    balanced_node_choices = retworkx.bipartition_graph_mst(
        graph.pygraph,
        lambda x: random.random(),
        pops,
        float(pop_target),
        float(epsilon)
    )
    balanced_nodes = {graph.retworkx_networkx_mapping[x] for x in choice(balanced_node_choices)[1]}
```

(https://github.com/mggg/GerryChain/pull/381/files?diff=unified&w=0#diff-02d17350b8ad584edb0ba9a458e607c2c9241895e2227bbbaf5d2580d83c52e1R193)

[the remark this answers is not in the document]
InnovativeInventor: There are some cases where there is only one (or zero) balanced edge, so this may be the explanation for the performance hit.

src/tree.rs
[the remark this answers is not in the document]
InnovativeInventor: Ah, figured it out.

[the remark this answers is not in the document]
InnovativeInventor: So, it doesn't look like it's worth the performance overhead (on my end) to convert to a numpy array for `bipartition_graph_mst`. For various architectural reasons, it's fairly painful to keep the pops numpy array floating around on each step of the markov chain. `bipartition_graph_mst` is going to be used much more than `bipartition_tree`, so I'm going to revert my commit and mark this as resolved. Let me know if I should un-mark it.

InnovativeInventor:
@mtreinish @IvanIsCoding I think I addressed all of the feedback you two gave (thanks, by the way!). Let me know if I should re-open any of the comments or address any other concerns/feedback you have.

IvanIsCoding (approved):
LGTM

src/tree.rs
```diff
@@ -126,12 +128,156 @@ pub fn minimum_spanning_tree(
...
+            .collect();
+
+        if unseen_neighbors.len() == 1 {
+            // This will be false if root
+            let neighbor = unseen_neighbors[0];
+            pops[neighbor.index()] += pop;
+            let mut current_partition_tracker = same_partition_tracker[node.index()].clone();
+            same_partition_tracker[neighbor.index()].append(&mut current_partition_tracker);
+
+            if !node_queue.contains(&neighbor) {
```
georgios-ts, line 224: This will have bad performance since we need to traverse the queue to find `neighbor`. I'm wondering if we can invariably push `neighbor` to the queue and skip the logic inside the while loop if `node` is marked as seen.

src/tree.rs
```diff
@@ -126,12 +128,156 @@ pub fn minimum_spanning_tree(
...
+            // Not at the leaves of the unseen subgraph
+            continue;
+        }
+
+        // Check if balanced
+        if pop >= pop_target * (1.0 - epsilon) && pop <= pop_target * (1.0 + epsilon) {
+            balanced_nodes.push((node.index(), same_partition_tracker[node.index()].clone()));
+        }
+
+        seen_nodes.insert(node.index());
```
georgios-ts, line 240: If I read this correctly, this code will run only in `if unseen_neighbors.len() == 1` case so I think it's best to move this inside the `if unseen_neighbors.len() == 1 { .. }` block.

src/tree.rs
```diff
@@ -126,12 +128,156 @@ pub fn minimum_spanning_tree(
...
+    pops: Vec<f64>,
+    pop_target: f64,
+    epsilon: f64,
+) -> PyResult<Vec<(usize, Vec<usize>)>> {
+    let mut balanced_nodes: Vec<(usize, Vec<usize>)> = vec![];
+    let mut mst = (*graph).clone();
+
+    while balanced_nodes.is_empty() {
+        mst.graph.clear_edges();
+        _minimum_spanning_tree(py, graph, &mut mst, Some(weight_fn.clone()), 1.0)?;
```
georgios-ts, line 278: I don't get why we need to recalculate `mst` every time in the while loop (or even why different loops we'll give different results) since the values of all arguments stays the same.

src/tree.rs
```diff
@@ -126,12 +128,156 @@ pub fn minimum_spanning_tree(
...
+///     the graph.
+/// :param float epsilon: The maximum percent deviation from the pop_target
+///     allowed while still being a valid balanced cut edge.
+///
+/// :returns: A list of tuples, with each tuple representing a distinct
+///     balanced edge that can be cut. The tuple contains the root of one of the
+///     two partitioned subtrees and the set of nodes making up that subtree.
+#[pyfunction]
+#[pyo3(text_signature = "(graph, weight_fn, pop, target_pop, epsilon)")]
+pub fn bipartition_graph_mst(
```
georgios-ts, line 265: IMO this function can be omitted from our API since it's just calling `minimum_spanning_tree` and `bipartion_tree` but we already provide these functions as part of our API.

src/tree.rs
```diff
@@ -126,12 +128,156 @@ pub fn minimum_spanning_tree(
...
+        }
+
+        seen_nodes.insert(node.index());
+    }
+
+    balanced_nodes
+}
+
+/// Bipartition graph into two contiguous, population-balanced components using
+/// mst. Assumes that the graph is contiguous. See :func:`~bipartition_tree` for
```
georgios-ts, line 247: What's the meaning of a contiguous graph?

georgios-ts (changes requested):

src/tree.rs
InnovativeInventor: The graph is connected. E.g. only one connected component. Graph partitions are commonly described in terms of contiguity, rather than connected components, especially if the graphs are usually planar. I can reword this if you'd like.

src/tree.rs
InnovativeInventor: We recalculate `mst` every time in the while loop because `bipartition_graph_mst` is a sampling algorithm and reusing `mst` when `_bipartition_tree` fails would change the distribution that we're sampling from.

src/tree.rs
InnovativeInventor: That's a good point! Implementing and benchmarking now.

src/tree.rs
InnovativeInventor: To be clear: `weight_fn` is intended to sample and be non-deterministic (i.e. sampling between the range `[0, 1]`).

src/tree.rs
InnovativeInventor: I disagree. This allows reuse of the spanning tree object if there are no balanced edges detected and reduces the memory allocs (i.e. the amount of times the graph is cloned to create a new spanning tree object).

InnovativeInventor:
FYI, I just rebased this off `main` since this PR appeared to be out of date from `main`.

src/tree.rs
```diff
@@ -188,23 +190,25 @@ fn _bipartition_tree(
 ) -> Vec<(usize, Vec<usize>)> {
     let mut pops = pops;
     let spanning_tree_graph = &spanning_tree.graph;
-    let mut same_partition_tracker: Vec<Vec<usize>> =
-        vec![vec![]; spanning_tree_graph.node_bound()]; // Keeps track of all all the nodes on the same side of the partition
+    let mut same_partition_tracker: Vec<LinkedList<usize>> =
+        vec![LinkedList::new(); spanning_tree_graph.node_bound()]; // Keeps track of all all the nodes on the same side of the partition
```
IvanIsCoding, line 194: I don't think it is worth using LinkedLists in this case, even more given the fact you need to convert them back to a vector at the end

src/tree.rs
```diff
@@ -214,32 +218,40 @@ fn _bipartition_tree(
...
-        }
-
-        seen_nodes.insert(node.index());
     }
 
+    // Convert LinkedList back to vec
     balanced_nodes
+        .iter()
+        .map(|(node, partition_nodes)| (*node, partition_nodes.iter().copied().collect()))
+        .collect()
```
IvanIsCoding, line 254: This offsets most of the gains you get from using a LinkedList (if any)

src/tree.rs
InnovativeInventor: Fair, I just perf tested it, and you're right.

src/tree.rs
InnovativeInventor: Yeah, I just perf tested it, and you're right.

src/tree.rs
IvanIsCoding: We might need to rename weight_fn to something else because on every other function weight_fn takes an edge and returns a weight. Which is not the case here

InnovativeInventor: I'm not sure what you mean by that. This should be the same `weight_fn` that `minimum_spanning_tree` takes (that is, it takes in an edge and return a random weight). E.g. `weight_fn` can be: `lambda _: random.random()` (i.e. uniformly picks between `[0,1]`). However, there are other variants of the ReCom Markov Chain sampling algorithm that require certain edges to be prioritized/deprioritized like so: `lambda x: random.random() if edge_crosses_county(x) else random.random() + 0.5`. This is used to create redistricting plan sampling methods that respect county or city boundaries.

src/tree.rs
```diff
@@ -126,12 +129,161 @@ pub fn minimum_spanning_tree(
...
+}
+
+/// Bipartition graph into two contiguous, population-balanced components using
+/// mst. Assumes that the graph is contiguous. See :func:`~bipartition_tree` for
+/// details on how balance is defined.
+///
+/// :param PyGraph graph: Undirected graph
+/// :param weight_fn: A callable object (function, lambda, etc) which
+///     will be passed the edge object and expected to return a ``float``. See
+///     :func:`~minimum_spanning_tree` for details.
```
IvanIsCoding, line 259: This might be me bing picky, but I'd rename to something along the lines of `weight_sample_fn` and describe it as you described to me in the previous comment. It would make it clearer to the users and maintainers that the intended is non deterministic.

I'm saying so because in general, we expect `weight_fn` to always return the same result given the same edge. Moreover, wherever possible we also try to cache the calls to `weight_fn` to avoid calling Python from Rust more than necessary.

georgios-ts (changes requested):

src/tree.rs
georgios-ts: That's a fair point but I think we can avoid compromising performance and still omit this function with a different design here. We can implement a new pyclass `SpanningTreeSampler` that generates random spanning trees and reuse internally the same memory for stroring the MST and avoid unnecessary allocations. Users can then call:
```python
sampler = retworkx.SpanningTreeSampler(graph)
balanced_nodes = []
while not balanced_nodes:
    tree = sampler.sample()
    balanced_nodes = retworkx.bipartition_tree(tree, ..)
```
to replicate the output of `bipartition_graph_mst` (and it'll be marginally slower).

My main motivation for the above design is two-fold:
1. Sampling a  spanning tree is an interesting problem on its own with different algorithms in the literature (which we can implement at a later point) and more users will benefit from it.
2. The output of `bipartition_graph_mst` feels a bit artificial to me, e.g cutting an output edge will not necessarily cut the input graph into two connected components but it'll only cut the tree that we randomly drew.

PR #576, opened 2022-04-06
Make mapping containers explicitly not sequences

mtreinish:
In PyO3 0.16.x the Python protocol interface was reworked to be more
pythonic and deprecated the pyproto macro and the distinct traits for
implementing different python protocols. As a side effect of that the
overlapping methods for the Python mapping and sequence protocols (they
both have getitem, len, etc). In the 0.16.3 release a pyclass macro
option was added to explicitly mark a container class as a mapping. In
retworkx we have several mapping containers which are not sequences. Now
that we're using PyO3 0.16.3 for retworkx, to make this distinction clear
this commit leverages the new mapping argument to the pyclass for
container classes which are just mappings.


georgios-ts (approved):
LGTM 👍

PR #577, opened 2022-04-07
Add return all simple paths between all pairs of nodes

mtreinish:
This commit adds a new function all_pairs_all_simple_paths
which is used to return all simple paths between all nodes of a
particular length. It basically is a multithreaded loop around the
existing all_simple_paths function that runs the function between all
pairs of nodes in the graph.


mtreinish:
CI here is blocked on the release of Rust 1.60.0 today, I have a fix up for this in #578

taalexander (commented):
Documentation LGTM.

src/connectivity/mod.rs
```diff
@@ -625,6 +626,62 @@ pub fn digraph_all_simple_paths(
...
+///     sett to 0 will behave like the default.
+/// :param int cutoff: The maximum depth of path to include in the output list
+///     of paths. By default includes all paths regardless of depth, setting to
+///     0 will behave like default.
+///
+/// :returns: A mapping of source node indices to a mapping of target node
+///     indices to a list of paths between the source and target nodes.
+/// :rtype: AllPairsMultiplePathMapping
+#[pyfunction]
+#[pyo3(text_signature = "(graph, /, size)")]
```
mtreinish, line 648:
```suggestion
#[pyo3(text_signature = "(graph, /, min_depth=None, cutoff=None)")]
```

mtreinish:
I'm thinking it might be better to change the tests (especially for the undirected graph) to avoid having the hard coded path lists and instead maybe call `all_simple_paths`  in a for loop and compare the output against the all pairs. It's not a perfect test, but at least it means we can eliminate ~2.5k lines of hard coded expected output from the tests

src/connectivity/all_pairs_all_simple_paths.rs
```diff
@@ -0,0 +1,94 @@
...
+                                Some((v.index(), output))
+                            }
+                        })
+                        .collect(),
+                };
+                if out_paths.paths.is_empty() {
+                    None
+                } else {
+                    Some((u.index(), out_paths))
+                }
```
mtreinish, line 90: I'm also debating dropping this so we still have empty dictionaries when there are no paths between a pair of nodes. Not sure what would be better from a user perspective, if you called:
```
res = all_pairs_all_simple_paths(graph)
print(res[0])
```
If there are no paths from 0 and got a key error vs it returning an empty dict

releasenotes/notes/all_pairs_all_simple_paths-9ad2cc3c265f4102.yaml
```diff
@@ -0,0 +1,17 @@
...
+    Added a new function :func:`~.all_pairs_all_simple_paths` which
+    is used to return all simple paths between all pairs of nodes in a graph. It
+    can also be used with a optional ``min_depth`` and ``cutoff`` parameters to
+    filter the results based on path lengths. For example::
+
+        from retworkx.generators import grid_graph
+        from retworkx import find_all_pairs_all_simple_paths_of_size
+
+        g = grid_graph(2, 3)
+        paths = find_all_pairs_all_simple_paths_of_size(g, min_depth=3, cutoff=3)
```
georgios-ts, line 13: It might be good to wrap this code inside a `jupyter-execute` so we avoid these typos. 

```suggestion
        from retworkx import all_pairs_all_simple_paths

        g = grid_graph(2, 3)
        paths = all_pairs_all_simple_paths(g, min_depth=3, cutoff=3)
```

retworkx/__init__.py
```diff
@@ -626,6 +626,40 @@ def _graph_all_pairs_dijkstra_shortest_path(graph, edge_cost_fn):
...
+@functools.singledispatch
+def all_pairs_all_simple_paths(graph, min_depth=None, cutoff=None):
+    """Return all the simple paths between all pairs of nodes in the graph
+
+    This function is multithreaded and will launch a thread pool with threads
+    equal to the number of CPUs by default. You can tune the number of threads
+    with the ``RAYON_NUM_THREADS`` environment variable. For example, setting
+    ``RAYON_NUM_THREADS=4`` would limit the thread pool to 4 threads.
+
+    :param PyGraph graph: The graph to find all simple paths in
```
georgios-ts, line 638:
```suggestion
    :param graph: The graph to find all simple paths in. This can be a :class:`~retworkx.PyGraph`
        or a :class:`~retworkx.PyDiGraph`
```

retworkx/__init__.py
```diff
@@ -626,6 +626,40 @@ def _graph_all_pairs_dijkstra_shortest_path(graph, edge_cost_fn):
...
+    :param int min_depth: The minimum depth of the path to include in the output
+        list of paths. By default all paths are included regardless of depth,
+        sett to 0 will behave like the default.
```
georgios-ts, line 641:
```suggestion
        setting to 0 will behave like the default.
```

src/connectivity/all_pairs_all_simple_paths.rs
```diff
@@ -0,0 +1,94 @@
...
+                    )
+                })
+                .collect(),
+        };
+    }
+    let intermediate_min: usize = match min_depth {
+        Some(depth) => depth - 2,
+        None => 0,
+    };
+    let intermediate_cutoff = cutoff.map(|depth| depth - 2);
```
georgios-ts, line 51: raise a `ValueError` if a user provides `min_depth` or `cutoff` less than 2.

src/connectivity/all_pairs_all_simple_paths.rs
```diff
@@ -0,0 +1,94 @@
...
+                    paths: node_indices
+                        .iter()
+                        .filter_map(|v| {
+                            let output: Vec<Vec<usize>> = algo::all_simple_paths(
+                                graph,
+                                *u,
+                                *v,
+                                intermediate_min,
+                                intermediate_cutoff,
+                            )
```
georgios-ts, line 67: I'm wondering if it's valuable to avoid this loop by modifying petgraph's `algo::all_simple_paths` implementation in order to directly support a list of targets and not just a single target, as NetworkX https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.simple_paths.all_simple_paths.html#networkx.algorithms.simple_paths.all_simple_paths does. This might save us from doing the same work many times.

src/connectivity/mod.rs
```diff
@@ -625,6 +626,62 @@ pub fn digraph_all_simple_paths(
...
 }
 
+/// Return all the simple paths between all pairs of nodes in the graph
+///
+/// This function is multithreaded and will launch a thread pool with threads
+/// equal to the number of CPUs by default. You can tune the number of threads
+/// with the ``RAYON_NUM_THREADS`` environment variable. For example, setting
+/// ``RAYON_NUM_THREADS=4`` would limit the thread pool to 4 threads.
+///
+/// :param PyGraph graph: The graph to find all simple paths in
```
georgios-ts, line 636:
```suggestion
/// :param PyDiGraph graph: The graph to find all simple paths in
```

src/connectivity/mod.rs
```diff
@@ -625,6 +626,62 @@ pub fn digraph_all_simple_paths(
...
+/// :param int min_depth: The minimum depth of the path to include in the output
+///     list of paths. By default all paths are included regardless of depth,
+///     sett to 0 will behave like the default.
```
georgios-ts, line 639:
```suggestion
///     setting to 0 will behave like the default.
```

src/connectivity/mod.rs
```diff
@@ -625,6 +626,62 @@ pub fn digraph_all_simple_paths(
...
```
georgios-ts, line 667:
```suggestion
///     setting to 0 will behave like the default.
```

tests/digraph/test_all_simple_paths.py
```diff
@@ -111,3 +111,214 @@ def test_graph_digraph_all_simple_paths(self):
         dag.add_node(0)
         dag.add_node(1)
         self.assertRaises(TypeError, retworkx.digraph_all_simple_paths, (dag, 0, 1))
+
+
+class TestDAGAllSimplePathsAllPairs(unittest.TestCase):
```
georgios-ts, line 116:
```suggestion
class TestDiGraphAllSimplePathsAllPairs(unittest.TestCase):
```
since it seems that there is a cycle `[(1, 2), (2, 3), (3, 4), (4, 2)]` in the testing graph and hence not a DAG.

src/connectivity/all_pairs_all_simple_paths.rs
mtreinish: Yeah, that's a good idea. It would make it more efficient since we only do a single traversal from each source instead of `n` traversals. I'm not sure how much it'll speed things up in practice but it's worth doing while we're thinking about it. I'll add a custom version of `algo::all_simple_paths` to retworkx-core in a little while to do this.

src/connectivity/all_pairs_all_simple_paths.rs
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/577/commits/f4490cd901b21978d334eaf15ca69ca39d5e6683

src/connectivity/all_pairs_all_simple_paths.rs
mtreinish: I made this change as part of https://github.com/Qiskit/retworkx/pull/577/commits/84f36a823d9008cb6164644a5d875eb4924cd999 it will return an empty inner dict for a source node if there are no simple paths from a node.

mtreinish:
> I'm thinking it might be better to change the tests (especially for the undirected graph) to avoid having the hard coded path lists and instead maybe call `all_simple_paths` in a for loop and compare the output against the all pairs. It's not a perfect test, but at least it means we can eliminate ~2.5k lines of hard coded expected output from the tests

I did this as part of: https://github.com/Qiskit/retworkx/pull/577/commits/84f36a823d9008cb6164644a5d875eb4924cd999

tests/graph/test_all_simple_paths.py
```diff
@@ -176,3 +176,110 @@ def test_digraph_graph_all_simple_paths(self):
...
+            (4, 5),
+            (5, 2),
+            (5, 3),
+        ]
+
+    def test_all_simple_paths(self):
+        dag = retworkx.PyGraph()
+        for i in range(6):
+            dag.add_node(i)
+        dag.add_edges_from_no_data(self.edges)
```
georgios-ts, line 204: Rename dag -> graph, add all nodes in one line.
It's probably better to move these lines in `setUp` since almost all test cases use the same graph.

```suggestion
        graph = retworkx.PyGraph()
        graph.add_nodes_from(range(6))
        graph.add_edges_from_no_data(self.edges)
```

georgios-ts (commented):
Looks good, thanks for the updates.
One thing I don't particularly like is the testing approach in undirected graphs. It might be better to test against a very simple graph (like a simple cycle graph) where we can easily and programmatically generate the expected output and we don't have to use `all_simple_paths` in a for loop.

tests/graph/test_all_simple_paths.py
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/577/commits/0647e15b027c036aa32be382689c5de948af4f61

mtreinish:
> Looks good, thanks for the updates. One thing I don't particularly like is the testing approach in undirected graphs. It might be better to test against a very simple graph (like a simple cycle graph) where we can easily and programmatically generate the expected output and we don't have to use `all_simple_paths` in a for loop.

Yeah, that's a good idea. I rewrote the tests to just use a small cycle graph which was easy to inspect the results by hand in: https://github.com/Qiskit/retworkx/pull/577/commits/0647e15b027c036aa32be382689c5de948af4f61

tests/graph/test_all_simple_paths.py
```diff
@@ -176,3 +176,71 @@ def test_digraph_graph_all_simple_paths(self):
...
+            2: {0: [[2, 3, 0], [2, 1, 0]]},
+            3: {1: [[3, 0, 1], [3, 2, 1]]},
+        }
+        self.assertEqual(paths, expected)
+
+    def test_all_simple_path_no_path(self):
+        dag = retworkx.PyGraph()
+        dag.add_node(0)
+        dag.add_node(1)
+        self.assertEqual({0: {}, 1: {}}, retworkx.all_pairs_all_simple_paths(dag))
```
georgios-ts, line 243:
```suggestion
        graph = retworkx.PyGraph()
        graph.add_node(0)
        graph.add_node(1)
        self.assertEqual({0: {}, 1: {}}, retworkx.all_pairs_all_simple_paths(graph))
```

tests/graph/test_all_simple_paths.py
```diff
@@ -176,3 +176,71 @@ def test_digraph_graph_all_simple_paths(self):
...
+
+    def test_all_simple_paths_empty(self):
+        self.assertEqual({}, retworkx.all_pairs_all_simple_paths(retworkx.PyDiGraph()))
```
georgios-ts, line 246:
```suggestion
        self.assertEqual({}, retworkx.all_pairs_all_simple_paths(retworkx.PyGraph()))
```

docs/source/api.rst
```diff
@@ -173,6 +173,7 @@ Other Algorithm Functions
    retworkx.core_number
    retworkx.graph_greedy_color
    retworkx.metric_closure
+   retworkx.all_pairs_all_simple_paths
```
georgios-ts, line 176: I'm thinking if it's more appropriate to move this (and `retworkx.all_simple_paths`) under `Connectivity` section or even make another separate section.

georgios-ts (approved):

docs/source/api.rst
mtreinish: Sure moving it under connectivity makes sense, I just put it here because that's where `all_simple_paths` was. I'll move this now.

mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/577/commits/edda11dc104ccefd0005624ddde82eb89d5edb59

mtreinish:
I was curious and did a benchmark comparing this function to a for loop with `all_simple_paths` over all permutations of nodes in a graph. I realize it's not entirely a fair comparison since this function is multithreaded and has less internal calls to `all_simple_paths()` because of the new retworkx-core function. But the for loop it is what you'd have to do today if you wanted to get all simple paths from all pairs of nodes:

```python3
from collections import defaultdict
import itertools
import statistics
import time

import numpy as np

import retworkx
import seaborn


seaborn.set()

num_nodes = []
all_pairs_times = []
loop_times = []

#for n in np.logspace(0, 3, dtype=int):
for d in range(3, 101, 2):
    graph = retworkx.generators.heavy_hex_graph(d)
    num_nodes.append(len(graph))
    times = []
    for _ in range(5):
        start = time.perf_counter()
        paths = retworkx.all_pairs_all_simple_paths(graph, min_depth=3, cutoff=3)
        stop = time.perf_counter()
        times.append(stop - start)
    all_pairs_times.append(statistics.mean(times))
    serial_paths = defaultdict(dict)
    serial_start = time.perf_counter()
    permutations = itertools.permutations(range(len(graph)), 2)
    for (u, v) in permutations:
        serial_paths[u][v] = retworkx.all_simple_paths(graph, u, v, min_depth=3,cutoff=3)
    serial_stop = time.perf_counter()
    loop_times.append(serial_stop - serial_start)

plt.loglog(num_nodes, all_pairs_times, label="all_pairs_all_simple_paths()")
plt.loglog(num_nodes, loop_times, label='Loop all_simple_paths')
plt.legend()
plt.xlabel("Number of nodes")
plt.ylabel("Runtime (sec.)")
plt.title("Time to find all simple paths of len(3) in heavy hex graph.")
plt.savefig('/tmp/paths_times.png', dpi=900)
```

This yielded:

paths_times


(note the plot is log log)

PR #578, opened 2022-04-07
Fix clippy warnings from Rust 1.60.0

mtreinish:
Rust 1.60.0 just released and with it came new clippy checks. These new
checks are flagging issues in the retworkx code which is causing issues
in CI. This commit fixes the issues identified by the new clippy
version, mainly replacing `&mut Vec<T>` with `&mut [T]` in a function
signature when a mutable reference to a slice is all that's needed.


IvanIsCoding (approved):
Makes sense, these functions are mostly borrowing the data and slices work fine

PR #579, opened 2022-04-08
Update coverage job to use stable rust

mtreinish:
In the recent Rust 1.60.0 release the instrument-coverage compiler flag
was stabilized. This means it no longer requires a nightly build of the
rust compiler to use it. This commit updates our coverage CI job to use
stable rust so we don't have to rely on a nightly build anymore.


IvanIsCoding (approved):
LGTM, reducing our dependency on Rust nightly is a step forward for the CI. I am glad the feature got stabilized

PR #580, opened 2022-04-08
Simplify and deduplicate multigraph handling in PyGraph

mtreinish:
This commit simplifies and deduplicates the PyGraph method definitions
that add edges to the graph object. When reviewing the multigraph=False
path to debug an issue I found that the PyGraph code around edge
handling was duplicated between every method that adds edges and also
was needlessly returning a PyResult<T>. This commit abstracts the edge
adding to an inner private method that does the multigraph flag handling
and also drops the PyResult return as these functions can't raise an
error. The return type for the functions which add multiple edges is
also updated to be an EdgeIndices object for a faster return time.

### DIFF
--- a/releasenotes/notes/edgeindices-pygraph-523ed0feac553c18.yaml
+++ b/releasenotes/notes/edgeindices-pygraph-523ed0feac553c18.yaml
@@ -1,0 +1,12 @@
+---
+upgrade:
+  - |
+    The return type for the :class:`~.PyGraph` method
+    :meth:`~.PyGraph.add_edges_from` and
+    :meth:`~.PyGraph.add_edges_from_no_data` has changed from a ``list`` of
+    integer edge indices to an :class:`~.EdgeIndices` object. The
+    :class:`~.EdgeIndices` class is a read-only sequence type of integer
+    edge indices. For the most part this should be fully compatible
+    except if you were mutating the output list or were explicitly type
+    checking the return. In these cases you can simply cast the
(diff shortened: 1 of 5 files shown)


mtreinish:
Oh, I should probably add an upgrade release note about changing the return type for `add_edges_from*`

georgios-ts (approved):
LGTM, it's always nice to see some duplicate code being removed.

Not meant to be addressed in this PR but I just noticed that `extend_from_edge_list` does *not* return the indices of the newly created edges which seems a bit inconsistent when comparing to `add_edges_from`. There is also a bug in this function (not sure if this is what you were looking) when calling `extend_from_edge_list` after `remove_node`, e.g the following crashes:
```python
import retworkx

graph = retworkx.PyGraph()
graph.add_nodes_from(range(5))
graph.remove_node(2)
graph.extend_from_edge_list([
    (0, 1), (0, 2)
])
```

mtreinish:
Oh, that's a good catch it is assuming it's a contiguous sequence which is a faulty assumption with removals. I'm wondering if we either should try to fill all the holes in the indices in a case like this, or just raise an exception if nodes have been removed from the graph and document this as a limitation with `extend_from_edge_list`. Do you want to push a PR to address it?

I actually was just looking at the edge list I used in the test as an input with a `PyDigraph` with `multigraph=False` that I misread as `PyGraph` (before morning coffee) and was trying to figure out how both `(2, 0)` and `(0, 2)` were in the undirected graph's edge list.

PR #583, opened 2022-04-11
Update workspace configuration to be tied to specific versions

mtreinish:
This commit updates the retworkx-core crate dependency to have a fixed
version which matches the retworkx crates version. We're not going to be
in a situation where we we release retworkx-core separately from the
main retworkx crate we only want to support the main crate with the core
crate at the same version. This commit makes this relationship
explicitly. At the same time it fixes the version string in the
retworkx-core crate as it wasn't updated in #560 like it should have
been.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -502,7 +502,7 @@ dependencies = [
 
 [[package]]
 name = "retworkx-core"
-version = "0.11.0"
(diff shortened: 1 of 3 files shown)


IvanIsCoding (approved):
LGTM

PR #591, opened 2022-04-21
Add support for slice inputs to custom sequence return types

### LINKED ISSUE #590: Implement slice support for sequence return types

For custom sequence return types we should support slice indexing in addition to just integers. For example if you ran something like:

```python3
import retworkx
retworkx.generators.cycle_graph(24)
g = retworkx.generators.cycle_graph(24)
g.node_indices()[1:3]
```

It should return a `NodeIndices` object  with the slice of 1 to 3. Right now if you want to get a slice you have to cast as a list which is less than ideal because it involves iterating and converting the entire object.

mtreinish:
This commit adds support for dealing with slice inputs to custom
sequence return types such as ``NodeIndices``. Previously, if a slice
was requested from a custom return type it would have raised a
``TypeError``. With this commit now it will return a new container object
with a copy of the data from the requested slice.

Fixes #590

### DIFF
--- a/src/iterators.rs
+++ b/src/iterators.rs
@@ -48,6 +48,7 @@ use pyo3::class::iter::IterNextOutput;
 use pyo3::exceptions::{PyIndexError, PyKeyError, PyNotImplementedError};
 use pyo3::gc::PyVisit;
 use pyo3::prelude::*;
+use pyo3::types::PySlice;
 use pyo3::PyTraverseError;
 
 macro_rules! last_type {
@@ -406,6 +407,12 @@ trait PyGCProtocol {
     fn __clear__(&mut self) {}
 }
 
+#[derive(FromPyObject)]
+enum SliceOrInt<'a> {
+    Slice(&'a PySlice),
+    Int(isize),
+}
+
 macro_rules! custom_vec_iter_impl {
     ($name:ident, $data:ident, $T:ty, $doc:literal) => {
         #[doc = $doc]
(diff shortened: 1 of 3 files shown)


src/iterators.rs
```diff
@@ -471,14 +478,32 @@ macro_rules! custom_vec_iter_impl {
...
-                } else {
-                    Ok(self.$data[idx as usize].clone())
+            fn __getitem__(&self, py: Python, idx: SliceOrInt) -> PyResult<PyObject> {
+                match idx {
+                    SliceOrInt::Slice(slc) => {
+                        let len = self.$data.len().try_into().unwrap();
+                        let indices = slc.indices(len)?;
+                        let start: usize = indices.start.try_into().unwrap();
+                        let stop: usize = indices.stop.try_into().unwrap();
+                        let step: usize = indices.step.try_into().unwrap();
```
georgios-ts, line 488: `unwrap` may panic with negative values. We should either support them or catch the error and return a python exception with a useful error message.

src/iterators.rs
```diff
@@ -471,14 +478,32 @@ macro_rules! custom_vec_iter_impl {
...
+                        let return_vec = $name {
+                            $data: (start..stop)
+                                .step_by(step)
+                                .map(|i| self.$data[i].clone())
+                                .collect(),
+                        };
+                        Ok(return_vec.into_py(py))
+                    }
+                    SliceOrInt::Int(idx) => {
+                        if idx.abs() >= self.$data.len().try_into().unwrap() {
```
georgios-ts, line 498: It's my fault but this check should have been:

```suggestion
                        let len = self.$data.len() as isize;
                        if idx >= len || idx < -len {
```

since it's valid to access the first element of a sequence `xs` of length `len` as `xs[-len]` (previously we'd wrongly raise an index error).

src/iterators.rs
mtreinish: I think that might be ok, I don't think that these can actually be negative. I was looking at that c api docs to confirm https://docs.python.org/3/c-api/slice.html but it doesn't actually say. I just don't know how the api could work if any of the values were negative. (well maybe `step` could be if stop < start). I'll do some more testing to see if I can get this to panic with different slice inputs.

georgios-ts: Yeah I already checked and
```python
import retworkx

graph = retworkx.generators.path_graph(5)
nodes = graph.node_indices()

# panics
sl = nodes[4:1:-1]

# panics
sl = nodes[-1:-3:-1]
```
but if `nodes` was a `list` it'd work.

mtreinish: This was a bit trickier to fix than I was hoping, but I think I've got it working now with negative steps with: https://github.com/Qiskit/retworkx/pull/591/commits/fa549b3ff5854aef94de4c90389624041d9bc386

georgios-ts (approved):
Thanks for the updates.

PR #600, opened 2022-04-30
Fix implementation of tensor product for undirected graphs

derbuihan:
This PR fixes the implementation of tensor product for undirected graphs.

Example: 

``` Python
import networkx as nx

g = nx.path_graph(2)
h = nx.path_graph(2)
r = nx.tensor_product(g, h)

r.edges() # [((0, 0), (1, 1)), ((0, 1), (1, 0))]
```

However, this calculation in retworkx outputs `[((0, 0), (1, 1)), ((1, 1), (0, 0))]`.
I fixed the tensor product in retworkx to output the same result as networkx.

### DIFF
--- a/src/tensor_product.rs
+++ b/src/tensor_product.rs
@@ -70,6 +70,12 @@ fn tensor_product<Ty: EdgeType>(
     if undirected {
         for edge_first in first.edge_references() {
             for edge_second in second.edge_references() {
+                if edge_first.source() == edge_first.target()
+                    || edge_second.source() == edge_second.target()
+                {
+                    continue;
+                }
+
                 let source = hash_nodes
                     .get(&(edge_first.source(), edge_second.target()))
                     .unwrap();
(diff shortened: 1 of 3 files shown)


src/tensor_product.rs
```diff
@@ -65,14 +65,25 @@ fn tensor_product<Ty: EdgeType>(
                 )
                     .into_py(py),
             );
+        }
+    }
+    if undirected {
+        for edge_first in first.edge_references() {
+            for edge_second in second.edge_references() {
+                let source = hash_nodes
```
georgios-ts, line 73: We should skip self loops here, right? Otherwise we'll add them twice in the output graph.

```suggestion
                if edge_second.source() == edge_second.target() {
                    continue;
                }

                let source = hash_nodes
```

georgios-ts (commented):
Nice catch! Can you also add a release note describing the bug fix?

src/tensor_product.rs
derbuihan: Thanks. This do not support multigraph. I fixed it and added its test.

derbuihan:
@georgios-ts 
Even though tensor product has not been released yet, is it necessary to add the bug fix to relaesenotes?

mtreinish (approved):

PR #606, opened 2022-05-07
Temporarily disable typehints in function signatures

### LINKED ISSUE #504: [joss] Function signatures are obscured in the documentation
Part of the function signatures is often obscured in the documentation. For example:

https://qiskit.org/documentation/retworkx/stubs/retworkx.k_shortest_path_lengths.html#retworkx.k_shortest_path_lengths

I see this:

image

I need to make the browser window smaller, but not too small, to see everything:

image

I see the same in Chrome and Firefox.

IvanIsCoding:
Closes #504 

Temporarily disables the typehints in functions signatures until https://github.com/Qiskit/qiskit_sphinx_theme/issues/21 is fixed.

Currently, the hints only appear on the functions with single dispatch and only appear for the graph argument. Because we already document that those functions accept both `PyGraph` and `PyDiGraph`, I argue that we do not lose much by disabling the hints.

Here is what the documentation looks like after the PR is applied:

example_doc

We still get a repetition of the signatures, but without them being obscured

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -60,6 +60,7 @@ master_doc = 'index'
 autosummary_generate = True
 autosummary_generate_overwrite = False
 autoclass_content = 'both'
+autodoc_typehints = 'none' # disabled until https://github.com/Qiskit/qiskit_sphinx_theme/issues/21 is fixed
 
 # Intersphinx configuration
 intersphinx_mapping = {


georgios-ts (commented):
Looks nice! It's an easy and quick fix. I agree we don't lose any information by dropping the type hints.

If we're going down this path, we could also leverage [`autodoc-before-process-signature`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#event-autodoc-before-process-signature) and unregister `PyGraph`, `PyDiGraph` from the single dispatch functions when building the docs. https://stackoverflow.com/questions/25951651/unregister-for-singledispatch. In this way we'll avoid the repetition of the signatures.

IvanIsCoding:
> Looks nice! It's an easy and quick fix. I agree we don't lose any information by dropping the type hints.
> 
> If we're going down this path, we could also leverage [`autodoc-before-process-signature`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#event-autodoc-before-process-signature) and unregister `PyGraph`, `PyDiGraph` from the single dispatch functions when building the docs. https://stackoverflow.com/questions/25951651/unregister-for-singledispatch. In this way we'll avoid the repetition of the signatures.

I tried unregistering but it makes the docs fail because we still need to use some of the methods in jupyter-execute.

I also tried using `autodoc-process-signature` to filter the signatures and I ended up with the normal signature and two blank signatures, it did not filter out the signatures unfortunately.

I pushed a trick that should be harmless to the code and tricks Sphinx. [It is 100% implementation dependant on this check from Sphinx](https://github.com/sphinx-doc/sphinx/blob/96b15e9e66bd4bc1079d983bf6be19d1e9aa7134/sphinx/util/inspect.py), but it gets the job done.

Let me know what you think of the trick. The documentation now looks like this:

Captura de tela de 2022-05-07 09-32-24

PR #609, opened 2022-05-09
Use unsigned_abs() in max_weight_matching

mtreinish:
In the upcoming Rust 1.61.0 release a new clipply lint warning is being
added to cargo clippy. This will be matched on any output from .abs()
calls that immediately cast to an unsigned type [1]. This was being done
in the max weight matching function in retworkx and will cause errors
when 1.61.0 is released. To address this issue from coming up this
commit switches the use of .abs() to .unsigned_abs().

[1] https://rust-lang.github.io/rust-clippy/master/index.html#cast_abs_to_unsigned


mtreinish:
Oh, I guess it wasn't stabilized in 1.48 I'm just going to close this. We'll just have to disable the clippy rule after 1.61 is released I guess since we can't use `unsigned_abs()` with our MSRV.

PR #610, opened 2022-05-09
Support graph attributes in GraphML reader.

### LINKED ISSUE #607: Loading graphml fails if there are any graph attributes
### Information

- **retworkx version**:
- Built from source.
- **Python version**:
- 3.10
- **Rust version**:
- 1.60
- **Operating system**:
- Ubuntu 20.04 on WSL2

Loading graphml with graph attributes results in error.
EG:
rx_graph = rx.read_graphml("manhatten.graphml")
yields:
"Exception: Unsupported 'for' attribute in key with id=d2."

Going into my graphML file, thie seems to be caused by this line:
` <key attr.name="streets_per_node" attr.type="string" for="graph" id="d2" />`
Removing this line, and other lines that contain `for="graph"` allow me to load the graphml file.

I guess this could be broken down to do different things.

a) A bug fix to just ignore graph attributes (or warn) if they're not supported.
b) Support the loading of graph-level attributes (This would be feature-parity with networkx).

Attempt to load the graphml available here:
https://www.kaggle.com/datasets/crailtap/street-network-of-new-york-in-graphml

You should see that the graph attributes are causing this error. Please let me know if you'd like me to post a more in depth series of steps to reproduce the error.

georgios-ts:
### Summary
GraphML documents may declare attributes/keys for `node`, `edge`, `graph` or `all` graph elements.
This commit adds support for attributes with domain `graph` and `all` by populating `Py(Di)Graph.attrs`
field with a dictionary since the first iteration of our GraphML reader would raise if users tried to
load a graphml file with keys declared for these domains.

Closes #607

### DIFF
--- a/src/graphml.rs
+++ b/src/graphml.rs
@@ -110,6 +110,8 @@ fn xml_attribute<'a, B: BufRead>(
 enum Domain {
     Node,
     Edge,
+    Graph,
+    All,
 }
 
 enum Type {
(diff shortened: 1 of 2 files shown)


mtreinish (approved):
LGTM, thanks for the quick update.

PR #615, opened 2022-05-22
Implement `__array__` for sequence types

### LINKED ISSUE #614: Unable to create numpy int array from an `EdgeList`
### Information

- **retworkx version**: main branch (since 71e34d111623d1de2e4870a8227eddacfb3ade4c)
- **Python version**: 3.10
- **Rust version**: 1.60
- **Operating system**: Linux

Trying to create a numpy array from an `EdgeList` object fails

```
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'retworkx.EdgeList'

The above exception was the direct cause of the following exception:

  File "/tmp/test_failure.py", line 5, in <module>
    print(np.asarray(res, dtype=np.uintp))
ValueError: setting an array element with a sequence.
```
I bisected this to 71e34d111623d1de2e4870a8227eddacfb3ade4c (#569). I'm not sure if this is caused by changes in the newer version of pyo3 or our minimal changes to the `custom_vec_iter_impl` macro (I'm leaning towards a pyo3 change).

calling `np.asarray(g.edge_list, dtype=np.uintp)` will create a numpy array for the edge list like:

```
[[0 1]
 [0 2]
 [0 3]
 [0 4]]
```

```python
import retworkx
import numpy as np
g = retworkx.generators.directed_star_graph(5)
res = g.edge_list()
print(np.asarray(res, dtype=np.uintp))
```

IvanIsCoding:
Fixes #614

Implements an explicit conversion from our custom types to Numpy arrays

This a quick fix to support converting using `np.array` and `np.asarray` [using the API specified by NumPy](https://numpy.org/devdocs/user/basics.dispatch.html). The solution is short and not too hard to maintain, the only burden will be remembering to add `py_convert_to_py_array_obj_impl!` for each new sequence return type we create.

Tasks:
* [x] Return 1d array of objects intead of raising not implemented error for some types
* [x] Handle dtype argument
* [x] Test things more thoroughly


src/iterators.rs
```diff
@@ -413,6 +415,50 @@ enum SliceOrInt<'a> {
...
 
+trait PyConvertToPyArray {
+    fn convert_to_pyarray(&self, py: Python) -> PyResult<PyObject>;
+}
+
+macro_rules! py_convert_to_py_array_impl {
+    ($($t:ty)*) => ($(
+        impl PyConvertToPyArray for Vec<$t> {
+            fn convert_to_pyarray(&self, py: Python) -> PyResult<PyObject> {
+                Ok(self.clone().to_pyarray(py).into())
```
georgios-ts, line 426: Is there any downside of using `IntoPyArray` which doesn't allocate more memory? https://docs.rs/numpy/latest/numpy/array/struct.PyArray.html#memory-location

```suggestion
                Ok(self.clone().into_pyarray(py).into())
```

georgios-ts (commented):
I'd rather see this fixed in pyo3 since numpy can automatically handle nested sequence objects but in general the approach here looks good.

src/iterators.rs
georgios-ts: We can not drop `clone()` if we use [`IntoPyArray`](https://docs.rs/numpy/latest/numpy/convert/trait.IntoPyArray.html) since it needs to take ownership of the vector but `__array__` receives a ref `&self` (and pyo3 doesn't allow to receive `self` and for a good reason since you can not move data from Python).

mtreinish: Yeah, the underlying trait method `into_pyarray` (which is part of the `IntoPyArray` @georgios-ts linked to) is doing exactly that for us. It uses an unsafe block to get the pointer: https://github.com/PyO3/rust-numpy/blob/19bfc9d2d0e72bb3ed30c08244ee81abd02d7386/src/convert.rs#L67

jakelishman: I think that's missing the lifetime/read-only ties, though, since it takes `self` not `&mut self`.

georgios-ts: @jakelishman Yeah, this is indeed possible and rust-numpy provides the unsafe [`borrow_from_array`](https://docs.rs/numpy/latest/numpy/array/struct.PyArray.html#method.borrow_from_array) (which results in a writeable array but you can't resize it) so we can avoid cloning the data.

IvanIsCoding: I switched to using `into_pyarray`

mtreinish (approved):
This LGTM, I agree with @georgios-ts having a fix in PyO3 would be the better path forward here. But having this in place now will avoid us from blocking the 0.12 release on this issue. We can revisit if we want this after the PYO3 issue is fixed and included in a release (I meant to try and tackle it but haven't had the time lately). The only thing I was on the fence about is having a release note to document we natively implement `__array__` now on the custom return types. But I think I'm ok without it as I don't think most users will notice vs older releases where having the full sequence protocol implemented implicitly exposed an array form.

Issue #441, opened 2021-09-08
Add more centrality measures


georgios-ts:
### What is the expected enhancement?
Recently a new function that calculates the `betweenness_centrality` was added (https://qiskit.org/documentation/retworkx/api.html#centrality). We can add similar measures like 
- `edge_betweenness_centrality` https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.edge_betweenness_centrality.html#networkx.algorithms.centrality.edge_betweenness_centrality, or 
- `closeness_centrality` https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.closeness_centrality.html#networkx.algorithms.centrality.closeness_centrality

PR #621, opened 2022-06-04
Add eigenvector centrality function

mtreinish:
This commit adds a new function eigenvector_centrality() to compute the
eigenvector centrality of a graph. It uses the same power function
approach that the NetworkX function eigenvector_centrality() [1]
function uses. This is for two reasons, the first is that a more
traditional eigenvector linear algebra/BLAS function would either
require us to link against BLAS at build time (which is a big change
in the build system and a large requirement) or to call out to numpy
via python both of which seemed less than ideal. The second reason was
to make handling holes in node indices bit easier. Using this approach
also enabled us to put the implementation in retworkx-core so it can be
reused with any petgraph graph.

Part of #441

### DIFF
--- a/retworkx-core/src/centrality.rs
+++ b/retworkx-core/src/centrality.rs
@@ -16,8 +16,11 @@ use std::sync::RwLock;
 use hashbrown::HashMap;
 use petgraph::graph::NodeIndex;
 use petgraph::visit::{
+    EdgeRef,
     GraphBase,
     GraphProp, // allows is_directed
+    IntoEdges,
+    IntoNeighbors,
     IntoNeighborsDirected,
     IntoNodeIdentifiers,
     NodeCount,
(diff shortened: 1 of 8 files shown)


retworkx-core/src/centrality.rs
```diff
@@ -297,3 +300,201 @@ where
...
+        + GraphProp
+        + GraphBase<NodeId = NodeIndex>,
+    F: FnMut(G::EdgeRef) -> Result<f64, E>,
+{
+    let tol: f64 = tol.unwrap_or(1e-6);
+    let max_iter = max_iter.unwrap_or(50);
+    let n_start: HashMap<G::NodeId, f64> = graph.node_identifiers().map(|n| (n, 1.)).collect();
+    let n_start_sum: f64 = n_start.len() as f64;
+    let mut x: HashMap<G::NodeId, f64> =
+        n_start.iter().map(|(k, v)| (*k, v / n_start_sum)).collect();
```
georgios-ts, line 371: `n_start` is redundant, we can directly initialize `x`. I removed also the division with `n_start_sum` since my understanding is that NetworkX does it to make sure their initial vector has all entries in [0, 1] but we don't accept a user param for `n_start` as they do, so the all-one vector is fine as a starting choice.

```suggestion
    let mut x: HashMap<G::NodeId, f64> = graph.node_identifiers().map(|n| (n, 1.)).collect();
```

retworkx-core/src/centrality.rs
```diff
@@ -297,3 +300,201 @@ where
...
+        + NodeCount
+        + GraphProp
+        + GraphBase<NodeId = NodeIndex>,
+    F: FnMut(G::EdgeRef) -> Result<f64, E>,
+{
+    let tol: f64 = tol.unwrap_or(1e-6);
+    let max_iter = max_iter.unwrap_or(50);
+    let n_start: HashMap<G::NodeId, f64> = graph.node_identifiers().map(|n| (n, 1.)).collect();
+    let n_start_sum: f64 = n_start.len() as f64;
+    let mut x: HashMap<G::NodeId, f64> =
```
georgios-ts, line 370: I'm wondering if a `Vec` of size `graph.node_bound()` and indexed by [`graph.to_index(n)`](https://docs.rs/petgraph/latest/petgraph/visit/trait.NodeIndexable.html#tymethod.to_index) would be faster for intermediate computations

retworkx-core/src/centrality.rs
```diff
@@ -297,3 +300,201 @@ where
...
+                let w_vec: Vec<G::EdgeRef> = graph
+                    .edges(*node)
+                    .filter(|edge| edge.target() == neighbor)
+                    .collect();
+                let mut w = 0.;
+                for edge in w_vec {
+                    w += weight_fn(edge)?;
+                }
+                *x.get_mut(&neighbor).unwrap() += x_last[node] * w;
+            }
```
georgios-ts, line 386: I'm not sure if this is correct in case of parallel edges: `graph.neighbors` will return a neighbor as many times as the number of parallel edges and each time we sum the weights of all parallel edges, so we are "overcounting".

```suggestion
            for edge in graph.edges(*node) {
                let w = weight_fn(edge)?;
                let neighbor = edge.target();
                *x.get_mut(&neighbor).unwrap() += x_last[node] * w;
            }
```

retworkx-core/src/centrality.rs
```diff
@@ -297,3 +300,201 @@ where
...
+        }
+        let mut norm: f64 = x.values().map(|val| val.powi(2)).sum::<f64>().sqrt();
+        if norm == 0. {
+            norm = 1.;
+        }
```
georgios-ts, line 391: If this ever happens it'd mean our current vector is very close to all-zero and a subsequent iteration might not be able to change it much so we'll think that we converged. But convergence to an all-zero solution is not a right solution so I'm thinking we might want to return `None` here.

retworkx-core/src/centrality.rs
```diff
@@ -297,3 +300,201 @@ where
...
+/// [`eigenvector_centrality()`](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.eigenvector_centrality.html)
+/// function.
+///
+/// In the case of multigraphs the weights of any parallel edges will be
+/// summed when computing the eigenvector centrality.
+///
+/// Arguments:
+///
+/// * `graph` - The graph object to run the algorithm on
+/// * `weight_fn` - An input callable that will be pased the `EdgeRef` for
```
georgios-ts, line 326:
```suggestion
/// * `weight_fn` - An input callable that will be passed the `EdgeRef` for
```

retworkx/__init__.py
```diff
@@ -1592,6 +1592,63 @@ def _graph_betweenness_centrality(graph, normalized=True, endpoints=False, paral
...
+    iterations is reached or when the computed vector between two iterations
+    is smaller than the error tolerance multiplied by the number of nodes.
+    The implementation of this algorithm is based on the NetworkX
+    `eigenvector_centrality() <https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.eigenvector_centrality.html>`__
+    function.
+
+    In the case of multigraphs the weights of any parallel edges will be
+    summed when computing the eigenvector centrality.
+
+    :param PyDigraph graph: The graph object to run the algorithm on
```
georgios-ts, line 1616:
```suggestion
    :param graph: Graph to be used. Can either be a
        :class:`~retworkx.PyGraph` or :class:`~retworkx.PyDiGraph`.
```

retworkx/__init__.py
```diff
@@ -1592,6 +1592,63 @@ def _graph_betweenness_centrality(graph, normalized=True, endpoints=False, paral
     )
 
 
+@functools.singledispatch
+def eigenvector_centrality(graph, weight_fn=None, default_weight=1.0, max_iter=100, tol=1e-6):
+    """Compute the eigenvector centrality of a :class:`~PyGraph`.
```
georgios-ts, line 1597:
```suggestion
    """Compute the eigenvector centrality of a graph.
```

retworkx/__init__.py
```diff
@@ -1592,6 +1592,63 @@ def _graph_betweenness_centrality(graph, normalized=True, endpoints=False, paral
...
+    is smaller than the error tolerance multiplied by the number of nodes.
+    The implementation of this algorithm is based on the NetworkX
+    `eigenvector_centrality() <https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.eigenvector_centrality.html>`__
+    function.
+
+    In the case of multigraphs the weights of any parallel edges will be
+    summed when computing the eigenvector centrality.
+
+    :param PyDigraph graph: The graph object to run the algorithm on
+    :param weight_fn: An optional input callable that will be pased the edge's
```
georgios-ts, line 1617:
```suggestion
    :param weight_fn: An optional input callable that will be passed the edge's
```

src/centrality.rs
```diff
@@ -132,3 +135,127 @@ pub fn digraph_betweenness_centrality(
...
+/// iterations is reached or when the computed vector between two iterations
+/// is smaller than the error tolerance multiplied by the number of nodes.
+/// The implementation of this algorithm is based on the NetworkX
+/// `eigenvector_centrality() <https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.eigenvector_centrality.html>`__
+/// function.
+///
+/// In the case of multigraphs the weights of any parallel edges will be
+/// summed when computing the eigenvector centrality.
+///
+/// :param PyDigraph graph: The graph object to run the algorithm on
```
georgios-ts, line 158:
```suggestion
/// :param PyGraph graph: The graph object to run the algorithm on
```

src/centrality.rs
```diff
@@ -132,3 +135,127 @@ pub fn digraph_betweenness_centrality(
...
+/// :param weight_fn: An optional input callable that will be pased the edge's
```
georgios-ts, line 159:
```suggestion
/// :param weight_fn: An optional input callable that will be passed the edge's
```

src/centrality.rs
```diff
@@ -132,3 +135,127 @@ pub fn digraph_betweenness_centrality(
...
+    graph: &graph::PyGraph,
+    weight_fn: Option<PyObject>,
+    default_weight: f64,
+    max_iter: usize,
+    tol: f64,
+) -> PyResult<CentralityMapping> {
+    let cost_fn = CostFn::try_from((weight_fn, default_weight))?;
+    let ev_centrality = centrality::eigenvector_centrality(
+        &graph.graph,
+        |e| cost_fn.call(py, e.weight()),
```
georgios-ts, line 186: It might be beneficial to precompute and store all edge weights like we do in `all_pairs_dijkstra_path_lengths` so we avoid calling Python overhead each iteration.

src/centrality.rs
```diff
@@ -132,3 +135,127 @@ pub fn digraph_betweenness_centrality(
...
+/// iterations is reached or when the computed vector between two iterations
+/// is smaller than the error tolerance multiplied by the number of nodes.
+/// The implementation of this algorithm is based on the NetworkX
+/// `eigenvector_centrality() <https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.eigenvector_centrality.html>`__
+/// function.
+///
+/// In the case of multigraphs the weights of any parallel edges will be
+/// summed when computing the eigenvector centrality.
+///
+/// :param PyDigraph graph: The graph object to run the algorithm on
```
georgios-ts, line 220:
```suggestion
/// :param PyDiGraph graph: The graph object to run the algorithm on
```

src/centrality.rs
```diff
@@ -132,3 +135,127 @@ pub fn digraph_betweenness_centrality(
...
+/// :param weight_fn: An optional input callable that will be pased the edge's
```
georgios-ts, line 221:
```suggestion
/// :param weight_fn: An optional input callable that will be passed the edge's
```

retworkx-core/src/centrality.rs
```diff
@@ -297,3 +300,201 @@ where
...
+    tol: Option<f64>,
+) -> Result<Option<HashMap<G::NodeId, f64>>, E>
+where
+    G: NodeIndexable
+        + IntoNodeIdentifiers
+        + IntoNeighbors
+        + IntoEdges
+        + NodeCount
+        + GraphProp
+        + GraphBase<NodeId = NodeIndex>,
```
georgios-ts, line 363:
```suggestion
    G: NodeIndexable + IntoNodeIdentifiers + IntoNeighbors + IntoEdges + NodeCount,
    G::NodeId: Eq + std::hash::Hash,
```

retworkx-core/src/centrality.rs
```diff
@@ -297,3 +300,201 @@ where
...
+    F: FnMut(G::EdgeRef) -> Result<f64, E>,
+{
+    let tol: f64 = tol.unwrap_or(1e-6);
+    let max_iter = max_iter.unwrap_or(50);
```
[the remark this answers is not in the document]
mtreinish: Oops, yeah this should be 100 not 50:

```suggestion
    let max_iter = max_iter.unwrap_or(100);
```

retworkx-core/src/centrality.rs
mtreinish: I'll benchmark it's worth testing.

retworkx-core/src/centrality.rs
mtreinish: Well my thinking was for parallel edges the centrality would be higher so it worked out ok. Similarly if we were going to compute this by generating an adjacency matrix and finding the eigenvector for it we'd be doing a similar thing because the adjacency matrix sums the weight of parallel edges.
'
That being said I don't have a super strong opinion on the correct behavior here, so I'll defer to whatever you think is best

retworkx-core/src/centrality.rs
mtreinish: That's fine with me, I'll update this: 
```suggestion
        if norm == 0. {
            return Ok(None);
        }
```

retworkx-core/src/centrality.rs
georgios-ts: Yeah, I agree. It's the right approach to sum the weight of parallel edges but my point was that the total update of the above loop is: `x[neighbor] += x_last[node] * sum(weights of parallel (node, neighbor) edges) * nb. of parallel (node, neighbor) edges` which is incorrect.

retworkx-core/src/centrality.rs
mtreinish: I benchmarked it and it did perform better: 

eigv_centrality_times

so I made the change in https://github.com/Qiskit/retworkx/pull/621/commits/09595a14fa9e5ef5526faea5e936fcf976cb6428

retworkx-core/src/centrality.rs
mtreinish: Ah, my bad I misread your comment. Yeah that makes sense. I'll correct this

mtreinish: Fixed in: https://github.com/Qiskit/retworkx/pull/621/commits/e03856b1f6ee4430d49f45e915e1683074d54884

src/centrality.rs
mtreinish: Let me benchmark this to see if it makes a difference

mtreinish: I benchmarked this and it's basically the same but marginally faster in a couple of cases:

eigv_centrality_times_py_callback

not really a noticeable difference. But I think I'll make the change anyway, mostly because it does open up potential parallelization. I think I can make the inner loop execute in parallel and will test that in bit.

mtreinish: I pushed this up in: https://github.com/Qiskit/retworkx/pull/621/commits/3204e09eb55509860ead214a116755ec86148637 I'm still waiting on the data collection for the run with a python callback. It's been running the sweep for ~30min and the precomputed weights sweep took ~15min. So there is no doubt this is faster. I'll generate the graph when it eventually finishes just for posterity.

mtreinish: Is the graph with a callback set (I just used `float`).

eigv_centrality_times_py_callback

src/centrality.rs
```diff
@@ -132,3 +137,162 @@ pub fn digraph_betweenness_centrality(
...
+        Some(max_iter),
+        Some(tol),
+    )?;
+    match ev_centrality {
+        Some(centrality) => Ok(CentralityMapping {
+            centralities: centrality
+                .iter()
+                .enumerate()
+                .filter_map(|(k, v)| {
+                    if graph.graph.node_weight(NodeIndex::new(k)).is_some() {
```
georgios-ts, line 205:
```suggestion
                    if graph.graph.contains_node(NodeIndex::new(k)) {
```

src/centrality.rs
```diff
@@ -132,3 +137,162 @@ pub fn digraph_betweenness_centrality(
...
+        Some(tol),
+    )?;
+
+    match ev_centrality {
+        Some(centrality) => Ok(CentralityMapping {
+            centralities: centrality
+                .iter()
+                .enumerate()
+                .filter_map(|(k, v)| {
+                    if graph.graph.node_weight(NodeIndex::new(k)).is_some() {
```
georgios-ts, line 285:
```suggestion
                    if graph.graph.contains_node(NodeIndex::new(k)) {
```

georgios-ts (approved):
Thanks for all the updates. Some minor comments but it's good to go.

georgios-ts (changes requested):
Oh, we'll need to update the docs before we merge.

georgios-ts (approved):

Issue #623, opened 2022-06-10
[joss] Graph size limits


szhorvat:
This is a question about the JOSS paper and this sentence:

> retworkx is suitable for modeling graphs ranging from a few nodes scaling up to millions.

What are the actual technical limits (or practical limits) on graph sizes, are these documented, and is there any protection against exceeding them? Is "millions" an actual practical guideline? Any reason it couldn't go to hundreds of millions or even billions?

There seems to be an opportunity for improving the paper here and showing the strengths of retworkx better. I expect this to be a strong point of retworkx. People do work with billion-edge networks these days. Have you tried simple linear-time algorithms such as connected components for such a huge graph?

mtreinish:
Heh, I raised the same point during the work on the initial draft: https://github.com/Qiskit/retworkx/pull/455#discussion_r737374904. Since that was a while ago I just tested with some of the generators (to eliminate the rng time from that comment) and I can generate a billion node star graph in about 16-25 seconds (taking about 38 GB of RAM) and a billion node square grid graph in ~30 seconds (which took about 60GB of RAM).

Since the nodes are stored in a `Vec` internally the theoretical maximum size of a graph is the `std::usize::MAX - 1`. So on a 64 bit platform it's 2^64 - 1 nodes (it's the same limit for edges too). But, obviously you'll be bound by the available ram on the system well before that number of nodes since each one of those nodes will require at least 64 bits of memory to store the pointer to the python object. When that happens you either get a panic exception for trying to allocate too much memory or the kernel will kill the process with an out of memory error depending on how the allocation is done in the function. Some functions like the [binomal tree generator](https://github.com/Qiskit/retworkx/blob/main/src/generators.rs#L891-L895) will preemptively raise an exception if you try to scale too big where you're encroaching on the limits.

There is some checking in petgraph when we add a node that will also raise a `PanicException` if we exceed the max number of nodes. That being said I think I found a bug in petgraph with this check when trying to make a graph with 10 billion node graph that was causing it to be raised, so that will have to be resolved before scaling larger. But at least things on the order of a billion work today though.

For scaling I ran up to a billion nodes on a cycle graph and ran `connected_components()` on it (and also asserted there was as single component for each step) and it yielded this:

![scaling_v2](https://user-images.githubusercontent.com/2447371/173109616-91f17202-31d6-42fe-82f5-bf8b715fdaed.png)

[...]

PR #625, opened 2022-06-10
Optimize performance of some generator functions

mtreinish:
This commit makes some improvements to the internals cycle graph, path
graph, star graph, and  mesh graph generator functions to optimize the
performance and reduce the memory overhead when creating the graphs. The
improvements fall into a couple categories, first the graph memory
allocation is only done once now as opposed to on demand as we create
the graph, and second intermediate index lists are no longer used in
places they weren't required to avoid the overhead of having to maintain
dual `Vec`s.

Related to #623

### DIFF
--- a/src/generators.rs
+++ b/src/generators.rs
@@ -92,29 +92,25 @@ pub fn directed_cycle_graph(
     };
     let mut graph = StablePyGraph::<Directed>::with_capacity(node_len, num_edges);
 
-    let nodes: Vec<NodeIndex> = match weights {
+    match weights {
         Some(weights) => {
-            let mut node_list: Vec<NodeIndex> = Vec::new();
             for weight in weights {
-                let index = graph.add_node(weight);
-                node_list.push(index);
+                graph.add_node(weight);
             }
-            node_list
         }
-        None => (0..num_nodes.unwrap())
-            .map(|_| graph.add_node(py.None()))
-            .collect(),
+        None => {
+            (0..node_len).for_each(|_| {
+                graph.add_node(py.None());
+            });
+        }
     };
-    for (node_a, node_b) in pairwise(nodes) {
-        match node_a {
-            Some(node_a) => {
-                if bidirectional {
-                    graph.add_edge(node_b, node_a, py.None());
-                }
-                graph.add_edge(node_a, node_b, py.None());
-            }
-            None => continue,
-        };
+    for a in 0..node_len - 1 {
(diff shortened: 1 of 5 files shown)


src/generators.rs
```diff
@@ -34,6 +34,15 @@ where
...
 }
 
+#[inline]
+fn get_num_nodes(num_nodes: &Option<usize>, weights: &Option<Vec<PyObject>>) -> usize {
+    if weights.is_some() {
+        weights.as_ref().unwrap().len()
+    } else {
+        num_nodes.unwrap()
+    }
+}
```
georgios-ts, line 44: We might want to ensure the output is non-zero and raise a `ValueError` if not, since later we might substract 1.

src/generators.rs
```diff
@@ -70,29 +79,31 @@ pub fn directed_cycle_graph(
...
             }
             node_list
         }
-        None => {
-            node_len = num_nodes.unwrap();
-            (0..num_nodes.unwrap())
-                .map(|_| graph.add_node(py.None()))
-                .collect()
-        }
+        None => (0..num_nodes.unwrap())
```
georgios-ts, line 104: We have already unwrapped this value (same comment in other places too)
```suggestion
        None => (0..node_len)
```

georgios-ts (commented):
We can drop the dual `Vec`'s in other places too.
For example in `path_graph` we can rewrite
```rust
for (node_a, node_b) in pairwise(nodes) {
    match node_a {
        Some(node_a) => graph.add_edge(node_a, node_b, py.None()),
        None => continue,
    };
}
```

as
```rust
for node_a in 0..node_len - 1 {
    let node_b = node_a + 1;
    graph.add_edge(NodeIndex::new(node_a), NodeIndex::new(node_b), py.None());
}
```

src/generators.rs
mtreinish: Done in: https://github.com/Qiskit/retworkx/pull/625/commits/06c4442e7acb6d63ed37ce0510df9abf2122453f

src/generators.rs
mtreinish: Fixed in: https://github.com/Qiskit/retworkx/pull/625/commits/4d4c9ed740072d7278c4392f7cba16e78ca56bcf I opted to just explicitly return an empty graph if node_len == 0. This is what used to happen if it was a 0 len input for most of them before

georgios-ts (approved):
Looks good!

PR #628, opened 2022-06-13
Use usize for all graph index types

### LINKED ISSUE #623: [joss] Graph size limits

mtreinish:
This commit fixes an issue with scaling retworkx graphs above 2**32 - 1
nodes and/or edges. Originally in retworkx we used a usize for all index
types to ensure the graphs could scale as big as the platform would
allow. However, since those earliest retworkx releases we lost the
explicit usize index typing along the way (likely before 0.1.0 and the
library was still very rough and not really used yet). This commit
corrects this oversight and defines a retworkx crate NodeIndex and
EdgeIndex type which are built around a usize and defines StablePyGraph
to be usize indexed instead of the implicit default of u32. Additionally
to support this several generics bugs in retworkx-core were fixed for
max weight matching and betweenness centrality as they both were only
implementable previously with graphs that used a u32 NodeIndex type.

Fixes #623

### DIFF
--- a/retworkx-core/src/centrality.rs
+++ b/retworkx-core/src/centrality.rs
@@ -11,10 +11,10 @@
 // under the License.
 
 use std::collections::VecDeque;
+use std::hash::Hash;
 use std::sync::RwLock;
 
 use hashbrown::HashMap;
-use petgraph::graph::NodeIndex;
 use petgraph::visit::{
     EdgeRef,
     GraphBase,
@@ -80,8 +80,9 @@ where
         + IntoNeighborsDirected
         + NodeCount
         + GraphProp
-        + GraphBase<NodeId = NodeIndex>
+        + GraphBase
         + std::marker::Sync,
+    <G as GraphBase>::NodeId: std::cmp::Eq + Hash + Send,
     // rustfmt deletes the following comments if placed inline above
     // + IntoNodeIdentifiers // for node_identifiers()
     // + IntoNeighborsDirected // for neighbors()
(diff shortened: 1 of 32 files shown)


mtreinish:
That's fair, I'm not really sure we want it either. I just assumed we were already using `usize` since I have memory of explicitly setting the index to usize in early versions of retworkx (and it was my original intent). I was actually surprised by the 2**32 limit when I was testing the scaling.

Making this a separate type is an interesting idea that might be a better approach. Although there will be some complexity around function typing if we go with that approach. I'm fine with just documenting the 32bit max as the upper bound on node and edge counts for now if that's what we want to do.

georgios-ts:
Yeah, it's certainly more work and will result in more duplicate code if we choose to introduce a new separate type. But this is a tradeoff I'm more inclined to take. Regarding the complexity around function typing, I was thinking about a new dispatch mechanism in Rust in a future 1.0.0 release that will hopefully reduce the boilerplate code and allow us more easily write a single function for all graph types.

Anyhow, for now, I'd go with describing the current 32bit limitation in the documentation.

mtreinish:
Ok, I'll close this and open two new PRs instead. One to fix the generics usage in max weight matching and betweenness centrality and the other to document the upper bound on node and edge bounds in the graph class documentation

PR #629, opened 2022-06-17
Fix generic usage for some retworkx-core functions

mtreinish:
This commit fixes several bugs with the use of generics bugs in
retworkx-core were fixed for max weight matching and betweenness
centrality as they both were only implementable previously with graphs
that used the default Graph/StableGraph u32 NodeIndex type. This was an
artificial bound on the input type as the functions should work with any
valid index type. This commit removes that bound so that functions work
with any indexable graph.

### DIFF
--- a/retworkx-core/src/centrality.rs
+++ b/retworkx-core/src/centrality.rs
@@ -11,10 +11,10 @@
 // under the License.
 
 use std::collections::VecDeque;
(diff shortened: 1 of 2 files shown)


IvanIsCoding (approved):
LGTM, this is the straightforward part from #628

PR #630, opened 2022-06-17
Document maximum number of nodes in edges allowed in graph classes

### LINKED ISSUE #623: [joss] Graph size limits

mtreinish:
The graph classes use a unsigned 32 bit integer as the index type which
means the maximum allowed size for the number of nodes or edges in a
graph object is std::u32::MAX - 1, roughly 4 billion. We should document
this limitation. In theory we can change the index type to use a usize
(see #628 for where this was attempted) which would increase the upper
limit to 2**64 - 1 (on 64 bit platforms), but there is a significant
memory cost for this and to have graphs so large you will need a
significant amount of local memory. In the future we might add a new
large graph type that makes this tradeoff to enable scaling above 4
billion nodes or edges but for now this commit just documents the
limitation with the current graph types.

Fixes #623

### DIFF
--- a/retworkx/__init__.py
+++ b/retworkx/__init__.py
@@ -115,6 +115,10 @@ class PyDAG(PyDiGraph):
     if a method call is made that would add a parallel edge it will instead
     update the existing edge's weight/data payload.
 
+    The maximum number of nodes and edges allowed on a ``PyGraph`` object is
+    :math:`2^{32} - 1` (4,294,967,294) each. Attempting to add more nodes or
+    edges than this will result in an exception being raised.
+
     :param bool check_cycle: When this is set to ``True`` the created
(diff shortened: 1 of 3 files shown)


IvanIsCoding (approved):
LGTM

PR #637, opened 2022-07-01
Set MSRV for clippy in retworkx-core

mtreinish:
Starting in the recently released Rust 1.62.0 clippy enabled a new rule
by default, cast_abs_to_unsigned [1], which warns when an abs() call is
cast to an unsigned type. Instead the rule suggests using unsigned_abs()
instead of abs(). However, with our msrv of 1.48 unsigned_abs() wasn't a
stable api. So trying to fix the places the rule is triggered as
suggested will fail to compile with rust 1.48. To address this and solve
this for future rust releases and potential new rules this commit adds a
clippy.toml file which sets the msrv for the retworkx-core crate. This
will by default disable rules that weren't on by default in rust 1.48.

As an alternative we could set the rust-version in our Cargo.toml to our
msrv and clippy would understand this, however that feature didn't exist
in cargo for rust 1.48 either and would cause a warning to be emitted
when building with 1.48. So this commit opted to just do the minimal
change to avoid the failure blocking further PRs from merging on main
and just condfigured the msrv in clippy for retworkx-core.

[1] https://rust-lang.github.io/rust-clippy/master/index.html#cast_abs_to_unsigned


IvanIsCoding (approved):
LGTM, it is good to keep Clippy working without warnings we cannot fix (yet)

PR #642, opened 2022-07-16
Removed usage of `pydot` in docs

### LINKED ISSUE #555: Replace pydot usage in documentation

There are several places in the documentation where we manually are calling pydot with `graph.to_dot()`'s output to generate visualizations (mainly in the generators modules. For example:

https://qiskit.org/documentation/retworkx/apiref/retworkx.generators.directed_heavy_hex_graph.html

These examples predate the introduction of the `retworkx.visualization` module and at the time was likely the only method of generating a visualization. But, now that we have the visualization modules we should go and update all the examples using pydot and `to_dot()` to generate a visualization with the `graphviz_draw()` function (or maybe `mpl_draw()`).

prakharb10:
Closes #555.

Some other remarks:

- Fixed hyperlinks to paper in a couple of functions.
- The docs for `directed_heavy_square_graph` and `directed_heavy_hex_graph` did not use the respective functions in the example. I believe this was an oversight and have updated those code snippets. Hence, these functions' outputs will differ from the previous ones.
- Haven't updated the occurrences at:

https://github.com/Qiskit/retworkx/blob/657e291ee3e919f5219076ab2d453dc00a3f015b/src/graph.rs#L1129

https://github.com/Qiskit/retworkx/blob/657e291ee3e919f5219076ab2d453dc00a3f015b/src/digraph.rs#L1765

as these specify the usage of the `to_dot()` function specifically.

### DIFF
--- a/src/generators.rs
+++ b/src/generators.rs
@@ -1277,7 +1277,7 @@ pub fn directed_binomial_tree_graph(
 ///        ...   S       ...
 ///
 /// NOTE: This function generates the four-frequency variant of the heavy square code.
-/// This function implements Fig 10.b left of the [paper](https://arxiv.org/abs/1907.09528).
+/// This function implements Fig 10.b left of the `paper <https://arxiv.org/abs/1907.09528>`_.
 /// This function doesn't support the variant Fig 10.b right.
 ///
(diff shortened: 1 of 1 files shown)


mtreinish (approved):
LGTM, thanks for fixing this

PR #644, opened 2022-07-25
Rename retworkx to rustworkx

### LINKED ISSUE #641: Proposal Plan To Rename Library
## Proposal Plan To Rename Library

This is a proposal to give a new name for the library. After our submission to the Journal of Open Source Software (https://github.com/openjournals/joss-reviews/issues/3968), one reviewers and the core developers pointed that the name retworkx is very close to NetworkX. As a courtesy to the NetworkX developers, I suggest to change the name of the library.

### Choose a new name

The first step is to choose a new name for the library. Some suggestions from me and :

* reseaux
* ~qisnet~
* ~networkQIS~
* rustworkx
* ~rustynets~
* ~netrust~

We are also open to suggestions.

### Release the next version package under the new name

The second step is to release the next version of the package (which as of now would be 0.12) under the new name. This entails a PR renaming all of our references from retworkx to the new package name we agree on. 

This will requires us to:
* Create a new package on PyPI under the new name
* Create a new crate on crates.io under the new name
* Create new documentation links in https://qiskit.org/documentation
* Rename our Github repository
* Publish a new version of the pre-print on arXiv
* Update our record on Zenodo

### Release new versions of retworkx packages depending on the new package

The third step is to provide support for retworkx. Ideally, we should just deprecate the name but continue supporting it for at least some releases  (I suggest two relases, which would be 0.12 and 0.13).

This can be done by making retworkx depend on the new package. For example, retworkx 0.12.0 would depend on the 'newpackagename' 0.12.0', retworkx 0.12.1 would depend on 'newpackagename' 0.12.1 and so on. This can be accomplished by importing from 'newpackage'. Because most of our code is in Rust, we just need to change a couple retworkx Python files to something along the lines of:

```python
from newpackagename import *
```

This step also requires us to:
* Add a disclaimer on README and PyPI that retworkx was renamed and that releases under the old name will stop after a couple versions
* Redirect links from https://qiskit.org/documentation/retworkx to the new documentation

mtreinish:
This commit renames the retworkx project to rustworkx. The maintainers of
NetworkX requested that we change the name of the library because they
viewed networkx as being too similar to retworkx. Out of respect for
this request and the NetworkX library which inspired retworkx we are
renaming the library to rustworkx. Rustworkx is a similar name to
retworkx and enables us to keep using the rx abbreviation for imports
which is commonly used.

This commit, migrates all the internal usage and names from retworkx to
rustworkx. This includes the retworkx-core crate which has similarly been
renamed to rustworkx-core. However, for backwards compatibility and to
ease the migration of our existing user base the retworkx package is
still included in rustworkx. An import redirect is added so that
import retworkx.foo will return return rustworkx.foo. Additionally, the
README and documentation are updated to mention the rename.

The migration plan this commit takes is that for the pending 0.12.0
release we'll migrate to rustworkx for everything but keep the backwards
compatibility shim in place with no warning. In the 0.13.0 release we'll
emit a DeprecationWarning (or a FutureWarning) that the name is
deprecated (this waits a release to ensure there is upgrade path where
more than 1 version works without raising a warning to ease migration). We can
then remove the deprecated retworkx name as part of the 1.0.0 release.
During the transition period we'll continue to publish releases to both
the rustworkx and retworkx names through at least 0.14.0 (but with the option
to keep publishing both until 1.0.0) so that users will see the deprecation
warnings when importing retworkx.

Closes #641


tox.ini
```diff
@@ -16,7 +16,7 @@ deps =
   testtools>=2.5.0
   networkx>=2.5
   stestr
-passenv = RETWORKX_TEST_PRESERVE_IMAGES
+passenv = RETWORKX_TEST_PRESERVE_IMAGES RESEAUX_PKG_NAME
```
IvanIsCoding, line 19: We could use [`tox-run-before`](https://pypi.org/project/tox-run-before/) to setup `rustworkx` and `retworkx` in the same environment:

```ini
run_before =
    export RESEAUX_PKG_NAME="rustworkx'
    pip install .
    unset RESEAUX_PKG_NAME
    export RESEAUX_PKG_NAME="retworkx'
    pip install .
    unset RESEAUX_PKG_NAME
```

IvanIsCoding:
I think if we install `rustworkx` locally before installing `retworkx` we will not need to relly on `rustworkx` being on PyPI, as `pip` will find it is already installed. I comment a possible workaround.

IvanIsCoding (approved):

tox.ini
georgios-ts, line 19:
```suggestion
passenv = RETWORKX_TEST_PRESERVE_IMAGES RUSTWORKX_PKG_NAME
```

georgios-ts (approved):

laurencap:
sgtm

PR #645, opened 2022-07-26
[WIP] Add planar layout to retworkx

enavarro51:
This draft PR adds the planar layout to the graph layout methods. This code incorporates and modifies the code from the open PR #475 for the Left/Right Planarity Test. It then creates the embedding that is ultimately converted to position coordinates for displaying the graph layout.

It follows the algorithms from Ulrik Brandes: The Left-Right Planarity Test 2009 http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.217.9208
and M. Chrobak and T.H. Payne: A Linear-time Algorithm for Drawing a Planar Graph on a Grid 1989 http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.51.6677
as well as the python code used in the networkx version of planar layout.

Remaining items include
- [ ] Additional error checking
- [ ] Docs
- [ ] More tests


src/layout/embedding.rs
```diff
@@ -0,0 +1,732 @@
...
+use crate::StablePyGraph;
+use rustworkx_core::connectivity::connected_components;
+use rustworkx_core::planar::lr_planar::{LRState, Sign};
+
+pub type Point = [f64; 2];
+
+#[derive(Debug)]
+pub struct CwCcw<T> {
+    cw: Option<T>,
+    ccw: Option<T>,
```
georgios-ts, line 30: Is `Option` the correct type here? These values will never be `None`. Creating an instance holding a `None` and later updating the weights with the correct values is redundant: store the final values right from the beginning!

IvanIsCoding:
#1206 reminded me that we should revisit this PR. The new Dorogovtsev-Goltsev-Mendes generator would be a perfect example of graph that would benefit from this layout

enavarro51:
@IvanIsCoding I'll take another look at this. I haven't touched this in 2 years (It was my first rustworkx project.) I'll fix the conflicts and run some tests, especially with the new D-G-M generator.

PR #648, opened 2022-08-02
Add retry for flaky tests

IvanIsCoding:
We add retry attemps to the test that writes the edge list

This specific flaky test has made the CI fail multiple times. Instead of manually having to retrigger all of our CI everytime it fails, we attempt to execute the specific test multiple times using `tenacity`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -146,7 +146,7 @@ jobs:
       - name: Download grcov
         run: curl -L https://github.com/mozilla/grcov/releases/download/v0.8.7/grcov-x86_64-unknown-linux-gnu.tar.bz2 | tar jxf -
       - name: Install deps
-        run: pip install -U setuptools-rust networkx testtools fixtures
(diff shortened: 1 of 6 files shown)


mtreinish (commented):
Is a retry actually sufficient for this? The issue we're hitting periodically in CI is python's tempfile usage on the macOS CI nodes isn't creating the file or macOS is flushing it out during the test and it gets deleted before the test is finished. I always assumed it was something more tied to the local run environment that would persist across multiple executions of the test. I've never been able to reproduce these failures locally (but I don't use macOS).

IvanIsCoding:
> Is a retry actually sufficient for this? The issue we're hitting periodically in CI is python's tempfile usage on the macOS CI nodes isn't creating the file or macOS is flushing it out during the test and it gets deleted before the test is finished. I always assumed it was something more tied to the local run environment that would persist across multiple executions of the test. I've never been able to reproduce these failures locally (but I don't use macOS).

That is a fair point, if the system itself is having problem in tempfiles we should retry running the Github action, not the test.

I am closing the PR

PR #679, opened 2022-09-15
Added empty and complete graph generators

prakharb10:
Added generators for `empty` and `complete` graphs.

Partially addresses #150.

### DIFF
--- a/src/generators.rs
+++ b/src/generators.rs
@@ -2420,6 +2420,148 @@ pub fn barbell_graph(
     })
 }
 
+/// Generate an undirected empty graph with ``n`` nodes and no edges.
+///
+/// :param int n: The number of nodes to generate the graph with.
+///
+/// :returns: The generated empty graph
+/// :rtype: PyGraph
+///
+/// .. jupyter-execute::
+///
+///  import rustworkx.generators
+///  from rustworkx.visualization import mpl_draw
+///
+///  graph = rustworkx.generators.empty_graph(5)
+///  mpl_draw(graph)
+///
+#[pyfunction(multigraph = true)]
+#[pyo3(text_signature = "(/, n, multigraph=True)")]
+pub fn empty_graph(py: Python, n: usize, multigraph: bool) -> PyResult<graph::PyGraph> {
+    let mut graph = StableUnGraph::<PyObject, PyObject>::default();
+    for _ in 0..n {
+        graph.add_node(py.None());
+    }
+    Ok(graph::PyGraph {
+        graph,
+        node_removed: false,
+        multigraph,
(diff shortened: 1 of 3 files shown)


src/generators.rs
```diff
@@ -2420,6 +2420,148 @@ pub fn barbell_graph(
...
+    let mut graph = empty_graph(py, n, multigraph)?;
+    if n > 1 {
+        for i in 0..n - 1 {
+            for j in i + 1..n {
+                graph.add_edge(i, j, py.None());
+            }
+        }
+    }
+    Ok(graph)
+}
```
mtreinish, line 2522: I think the complete graph function already exists, it's just called `mesh_graph`: https://github.com/Qiskit/rustworkx/blob/main/src/generators.rs#L553-L627. I think the complete graph name is probably more common. Maybe we should add this and just call `mesh_graph()` internally and document it as an alias.

mtreinish: I don't think there is any real reference to the term, it was added in: https://github.com/Qiskit/rustworkx/pull/191 which was a long time ago. I didn't really think about the name at the time (especially since the library was relatively young then). I agree with you that `complete_graph` is a better name (this has come up a few times I think). For backwards compatibility we can't simply rename `mesh_graph` -> `complete_graph`. But for the first step just having `complete_graph` as an alias for `mesh_graph` would make it easier for people to fine the function they're looking for.

prakharb10: I understand @mtreinish. I'll do that.

prakharb10:
Does this PR also require a `reno`? If yes, could you tell me what sections need to be included?

mtreinish:
Yes I think a reno release note would be appropriate here, I would say add two notes as features, one documenting the addition of the empty graph functions and the second documenting adding the complete graph functions

releasenotes/notes/empty-and-complete-09d569b42cb6b9d5.yaml
```diff
@@ -0,0 +1,29 @@
...
+      from rustworkx.visualization import mpl_draw
+
+      graph = rustworkx.generators.empty_graph(4)
+      mpl_draw(graph)
+
+  - |
+    Added new generator functions, :func:`~rustworkx.generators.complete_graph`,
+    and :func:`~rustworkx.generators.directed_complete_graph` to the 
+    ``rustworkx.generators`` module that will generate a complete graph.
+    For example:
```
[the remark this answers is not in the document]
prakharb10: Updated the reno @mtreinish. Let me know if further changes are required

mtreinish (approved):

PR #683, opened 2022-09-19
Add support for Python 3.11

mtreinish:
For the upcoming 0.12.0 release it would be good to add support for python 3.11. While this hasn't been released yet it's currently a release candidate with a frozen C API. We're likely to release rustworkx 0.12.0 prior to the Python 3.11 release. In order to avoid having to have rustworkx be a blocker for any downstream packages looking to adopt 3.11 this commit proactively enables support for testing and building with Python 3.11 off the latest rc in order to ensure we're building binaries compatible with 3.11 for the 0.12.0 release.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,7 +55,7 @@ jobs:
     strategy:
       matrix:
         rust: [stable]
-        python-version: [3.7, 3.8, 3.9, "3.10"]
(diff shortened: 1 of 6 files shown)


IvanIsCoding (approved):
LGTM and the Python 3.11rc2 tests pass on CI so that's great

mtreinish:
Before we merge this let me push up a test commit to enable the wheel jobs to make sure everything works still

PR #684, opened 2022-09-20
Add index_output option to layers() function

### LINKED ISSUE #646: rx.layers(dag,) return node creation input data, instead of expected index
for working with layers (iterations, etc.), expected nodes indexes, but not data used for creation

it would be nice to have an extension of the functionality, because the reverse index lookup is not efficient

mtreinish:
This commit adds a new keyword argument, index_output, to the layers() function which is used to return a list of list of node indices representing the layers in the DAG instead of a list of list of node data payloads. The default output does not change, but for users that prefer to work with indices (which is probably a more natural return format for the function) this option allows them to work with indices.

Closes #646

### DIFF
--- a/releasenotes/notes/add-index_output-dag-layers-d8329db4a1f76410.yaml
+++ b/releasenotes/notes/add-index_output-dag-layers-d8329db4a1f76410.yaml
@@ -1,0 +1,7 @@
+---
+features:
+  - |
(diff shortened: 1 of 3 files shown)


IvanIsCoding (approved):
Because we have always returned Python objects from the method, I think adding the keyword and the different return type is the right approach. I believe adding another function wouldn't be as convenient.

LGTM

PR #688, opened 2022-09-26
Use ahash with IndexSet in simple_cycles()

mtreinish:
We recently merged a new simple_cycles() function in #633 that added an implementation of johnson's algorithm for finding all the simple cycles in a directed graph. In that function we used IndexSet to have a set with O(1) lookup with a deterministic ordering. As part of that we used the default hashing algorithm from the Rust stdlib. This leaves some performance on the table as the stdlib hashing algorithm is slower in order to have proven resistance to HashDoS attacks. Since this is relevant for the internal usage in Johnson's algorithm this commit switches the hasing algorithm to ahash (which is the default hashbrown uses) which offers better performance.


IvanIsCoding (approved):
LGTM

PR #691, opened 2022-10-04
Fixed stray `reno` and added script to check the same

### LINKED ISSUE #678: Erroneous Reno in tests
### Information

- **retworkx version**: 0.11.0
- **Python version**: 3.9.6
- **Rust version**: 1.63.0
- **Operating system**: macOS 12.5.1

There's a reno placed in the test directories at - https://github.com/Qiskit/rustworkx/blob/main/tests/rustworkx_tests/generators/releasenotes/notes/d-1-heavy-hex-e2e44861dc75009a.yaml

I'm not sure, but I think this is erroneous.

Reno should be placed in the `releasenotes` directory.

prakharb10:
Closes #678.


.github/workflows/main.yml
```diff
@@ -40,6 +40,8 @@ jobs:
         run: black --check --diff retworkx rustworkx retworkx tests
       - name: Python Lint
         run: flake8 --per-file-ignores='retworkx/__init__.py:F405,F403' setup.py retworkx tests rustworkx
+      - name: Check stray release notes
+        run: python3 tools/find_stray_release_notes.py
```
IvanIsCoding, line 44:
```suggestion
        run: python tools/find_stray_release_notes.py
```

It is implied we use Python 3 by the other commands

tox.ini
```diff
@@ -53,6 +53,7 @@ commands =
   black --check --diff {posargs} '../rustworkx' '../tests' '../retworkx'
   flake8 --per-file-ignores='../rustworkx/__init__.py:F405,F403' ../setup.py ../rustworkx ../retworkx .
   cargo fmt --all -- --check
+  python3 {toxinidir}/tools/find_stray_release_notes.py
```
IvanIsCoding, line 56:
```suggestion
  python {toxinidir}/tools/find_stray_release_notes.py
```
Idem

IvanIsCoding (commented):
Thanks for handling this, just left two minor comments

IvanIsCoding (approved):

PR #692, opened 2022-10-06
Fix retworkx compat artifact CI job

mtreinish:
Before pushing the 0.12.0 tag I noticed the CI configuration for the retworkx release artifacts was broken and wouldn't run because of a dependency on the rustworkx jobs that was defined in a separate workflow. This commit updates the CI config by adding the retworkx artifact job to the wheels workflow so we can express the dependency on rustworkx being published first.


mtreinish:
After this merges I'll tag and push 0.12.0 to trigger the release. Glad I caught this so I didn't have to do the retworkx artifacts manually (I was about to push the tag and was double checking before I did).

IvanIsCoding (approved):
Makes sense, `rustworkx` should be in PyPI before `retworkx` such that we can depend on it.

PR #693, opened 2022-10-07
Bump version numbers after 0.12.0 release

mtreinish:
Now that the 0.12.0 release is out the door this commit bumps the version numbers to indicate that the main branch is now ahead of the 0.12.0 release. At the same time this commit fixes an indentation issue in the CI wheel publish jobs that we hit during the 0.12.0 release.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ README = readme()
 if PKG_NAME == "retworkx":
     README = retworkx_readme_compat + README
     PKG_PACKAGES = ["retworkx"]
-    PKG_INSTALL_REQUIRES.append(f"rustworkx>={PKG_VERSION}")
+    PKG_INSTALL_REQUIRES.append(f"rustworkx=={PKG_VERSION}")
     RUST_EXTENSIONS = []
 
 setup(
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -41,7 +41,7 @@ def readme():
 if PKG_NAME == "retworkx":
     README = retworkx_readme_compat + README
     PKG_PACKAGES = ["retworkx"]
-    PKG_INSTALL_REQUIRES.append(f"rustworkx=={PKG_VERSION}")
+    PKG_INSTALL_REQUIRES.append(f"rustworkx>={PKG_VERSION}")
```
IvanIsCoding, line 44:
```suggestion
    PKG_INSTALL_REQUIRES.append(f"rustworkx=={PKG_VERSION}")
```

I am kind of concerned with allowing `retworkx==0.13` and `rustworkx==0.14` in the same system, do you think it is safe?

mtreinish: I do, I've been thinking about our long term strategy towards moving away from retworkx permenantly. We've documented in rustworkx 1.0.0 we'll stop carrying around the retworkx shim, but I was thinking there would be no harm in leaving retworkx on pypi with an uncapped requirement it should never break as long as python doesn't change how import works. But I think for this PR you're right we should leave this as is, I'll open up a dedicated PR for this and document my thinking there.

IvanIsCoding (approved):

PR #694, opened 2022-10-07
Add cargo publish job for rustworkx-core

mtreinish:
This commit adds a CI job to automatically publish rustworkx-core to crates.io on releases. In the past I have just manually run cargo publish locally but this is error prone and depends on me doing the correct thing. Having it run automatically in CI limits the chances of something going wrong.


IvanIsCoding (approved):
LGTM (and the identation is right this time!)

Issue #696, opened 2022-10-10
`NodeIndices` does not implement sequence protocol


polwel:
### Information

- **rustworkx version**: `rustworkx-0.12.0-cp39-cp39-win_amd64.whl` from PyPI
- **Python version**: 3.9.12
- **Rust version**: n/a
- **Operating system**: Windows 10

### What is the current behavior?
Minimal example:

```python
>>> import rustworkx as rx
>>> graph = rx.generators.directed_path_graph(5)
>>> reversed(rx.topological_sort(graph))
Traceback (most recent call last):
  File "C:\Users\polw\venv-dev\lib\site-packages\IPython\core\interactiveshell.py", line 3378, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-4-c1f474a6ec26>", line 1, in <module>
    reversed(rx.topological_sort(graph))
TypeError: rustworkx.NodeIndices is not a sequence
```

From Python, I can use both `len()`  and indexing on `rustworkx.NodeIndices`. So that works.

### What is the expected behavior?

No error :)

`reversed` consumes sequences, and `NodeIndices` is documented to implement the sequence protocol.

### Steps to reproduce the problem
See above.

mtreinish:
Sigh I thought pyo3 fixed all these compatibility issues (we hit similar things earlier in the cycle after upgrading our pyo3 version). Thanks for reporting this I'll see if we're missing something on the newer pyo3 version and get it to act like a sequence properly again, and in the worst case we can just downgrade pyo3 to a version that is doing the correct thing.

IvanIsCoding:
We need to add tests to cover some behaviors and expected use cases of the custom return types, it's similar to #614 that we took some PyO3 "magic" for granted and they changed the API

PR #702, opened 2022-10-18
Update PyO3 to 0.17.2

IvanIsCoding:
Bumps PyO3 and NumPy to 0.17.2

It has been a while since dependabot has sent us updates notifications for PyO3 and we missed a couple. But I think bumping to 0.17.2 should work fine given the Rust MSRV is still the same and we have released 0.12

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -271,13 +271,15 @@ dependencies = [
 
 [[package]]
 name = "numpy"
-version = "0.16.2"
+version = "0.17.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "383ae168529a39fc97cbc1d9d4fa865377731a519bc27553ed96f50594de7c45"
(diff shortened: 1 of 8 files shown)


mtreinish (approved):
LGTM, I had a branch doing this for 0.17.0 but was blocked on some breaking changes. It's good to see that 0.17.2 fixed those.

We should test to see if #696 is fixed by this or if we can fix it now that we're on the newer version. If so we can backport this for 0.12.1

IvanIsCoding:
> LGTM, I had a branch doing this for 0.17.0 but was blocked on some breaking changes. It's good to see that 0.17.2 fixed those.
> 
> We should test to see if #696 is fixed by this or if we can fix it now that we're on the newer version. If so we can backport this for 0.12.1

I think I can write the tests covering the use case and make a PR by end of the week. Hopefully they pass straight away

Issue #704, opened 2022-10-19
Transitive closure


LaurentBergeron:
### What is the expected enhancement?

Implementing a transitive closure function with similar functionality to the one in Networkx:

https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.dag.transitive_closure.html

mtreinish:
I just opened up https://github.com/Qiskit/rustworkx/pull/707 which did the DAG optimized form of this function. Primarily because it looked more fun for me to implement, but I didn't mark that as closing this because it's not the general form which works with any graph. So if someone still wants to work on that they should feel free

Procatv:
Can I be assigned, I'd like to take a look at this issue

PR #707, opened 2022-10-20
Add transitive_closure_dag function

mtreinish:
This commit adds a new function transitive_closure_dag() which is an optimized method for computing the transitive closure for DAGs. In support of this a new function descendants_at_distance() for finding the nodes a fixed distance from a given source to both rustworkx and rustworkx-core.

Related to: #704

TODO:

- [ ] Add tests for descendants_at_distance (both from python and rust)

### DIFF
--- a/src/dag_algo/mod.rs
+++ b/src/dag_algo/mod.rs
@@ -29,6 +29,8 @@ use petgraph::graph::NodeIndex;
 use petgraph::prelude::*;
 use petgraph::visit::NodeCount;
 
+use rustworkx_core::traversal::descendants_at_distance;
+
 /// Find the longest path in a DAG
 ///
 /// :param PyDiGraph graph: The graph to find the longest path on. The input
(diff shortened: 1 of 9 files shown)


src/dag_algo/mod.rs
```diff
@@ -634,3 +636,50 @@ pub fn collect_bicolor_runs(
...
+///     the transitive closure. If one is not provided (or it is explicitly
+///     set to ``None``) a topological order will be computed by this function.
+///
+/// :returns: The transitive closure of ``graph``
+/// :rtype: PyDiGraph
+///
+/// :raises DAGHasCycle: If the input ``graph`` is not acyclic
+#[pyfunction]
+#[pyo3(text_signature = "(graph, / topological_order=None)")]
+pub fn transitive_closure_dag(
```
IvanIsCoding, line 659: Would you mind implementing this as a rustworkx-core function? I think it would make reusing it for #704 easier

IvanIsCoding (commented):
> This commit adds a new function transitive_closure_dag() which is an optimized method for computing the transitive closure for DAGs. In support of this a new function descendants_at_distance() for finding the nodes a fixed distance from a given source to both rustworkx and rustworkx-core.
> 
> Related to: #704
> 
> TODO:
> 
> * [ ]  Add tests for descendants_at_distance (both from python and rust)

Well, computing the transitive closure for a DAG is 80% of the problem for the general case because the full algorithm is:
* Condense the graph into strongly connect components (which we already implement)
* Apply the DAG algorithm and get the transitive closure for the intermediate graph
* Map the answer from the intermediate graph to the original graph

src/dag_algo/mod.rs
georgios-ts: There is [`Build`](https://docs.rs/petgraph/latest/petgraph/data/trait.Build.html) trait.

mtreinish: Thanks @georgios-ts I completely missed the build trait that should do the trick.

I'm working on a local branch right now, but I'm fighting the trait implementations now. Basically `IntoNodeIdentifiers` and the other visit traits are implemented for `&StablePyGraph` but not a mutable borrow or the type itself and `Build` is only available on a mutable borrow or the type itself. I'll see what I can come up with

georgios-ts: I've faced a similar situation before and it's quite tricky (and ugly) to get this right. I think a signature like this
```rust
pub fn transitive_closure_dag<'a, G, F>(
    graph: &'a G,
    topological_order: Option<Vec<G::NodeId>>,
    default_weight_fn: F,
) -> Result<G, algo::Cycle<G::NodeId>>
where
    G: Clone + Build,
    for<'b> &'b G:
        GraphBase<NodeId = G::NodeId> + Visitable + IntoNeighborsDirected + IntoNodeIdentifiers,
    G::NodeId: std::cmp::Eq + std::hash::Hash,
    F: Fn() -> G::EdgeWeight,
{
...
}
```
will do the job.

mtreinish: Awesome, thanks for the hint. That was enough to get it working for me. I didn't realize you could specify lifetimes on generic constraints like that, it was the piece I was missing along with defining `&G: GraphBase<NodeId = G::NodeId>` that was enough to get it moved into rustworkx-core in https://github.com/Qiskit/rustworkx/pull/707/commits/c4c0ab73444c01681402115f2a561b5f320d3f60

PR #713, opened 2022-10-24
Fix QEMU-based Linux wheel jobs.

kevinhartman:
### Summary
There appears to be an issue with memory consumption under the hood when using QEMU with Rust cargo: 
https://github.com/rust-lang/cargo/issues/10583.

This is causing the CI wheel workflow to run out of memory and fail for QEMU-based jobs, e.g.:
https://github.com/Qiskit/rustworkx/actions/runs/3245206601/jobs/5420102493#step:7:325

### Details and comments
From the discussion on the linked issue, the workaround is to tell cargo to use the Git executable directly, rather than `libgit2`.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -110,7 +110,7 @@ jobs:
           python -m cibuildwheel --output-dir wheelhouse
         env:
           CIBW_BEFORE_ALL_LINUX: "yum install -y wget && {package}/tools/install_rust.sh"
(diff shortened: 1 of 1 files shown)


mtreinish (approved):
LGTM, thanks for fixing this

IvanIsCoding:
Genuinely, awesome work on digging through the issues to find a simple solution. I spent a whole day trying to build the library with `cargo-zigbuild` and couldn't fix it.

PR #714, opened 2022-10-25
Switch from Python 3.11rc to official release

IvanIsCoding:
Switches from the 3.11 Python release candidate to the official release.

Also, I profited to remove some references to Pytho 3.6 as we have already specified in the release notes that we will not support anymore.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 minversion = 2.1
-envlist = py36, py37, py38, py39, py310, lint
+envlist = py37, py38, py39, py310, py311, lint
 isolated_build = true
 
 [testenv]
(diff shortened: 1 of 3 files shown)


IvanIsCoding:
Need to wait until 3.11 appears in https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json

tox.ini
```diff
@@ -21,11 +21,6 @@ changedir = {toxinidir}/tests
 commands =
   stestr run {posargs}
 
-[testenv:py36]
-extras =
```
mtreinish, line 25: I think we can add the extras to the root tesyenv here. Iirc, I split it out for py310 because matplotlib didn't work when we first added support for it. Although I guess py311 might have the same issue, it's worth Che though.

IvanIsCoding: Done

PR #717, opened 2022-10-26
Bump MSRV to 1.56.1 and update dependencies and Rust Edition

### LINKED ISSUE #716: Raise MSRV to 1.56

Right now our MSRV for rustworkx is 1.48 to match what the latest rust version that Debian stable ships. This was originally done to support the piwheels build environment which installs rustc via debian packages. However this causes a lot of development friction as our upstream dependencies typically adopt newer MSRVs much faster than debian bumps their MSRV. It also blocks us from using newer rust language features. When weighing that against the raspberry pi user base via piwheels is very small (607 all time downloads compared to our monthly pypi download rate of ~150k) and the rustworkx wheels failed to build so there hasn't been any downloads of it ye I think we should just pick an MSRV that suits our development needs more than trying to maintain compatibility for an unsupported platform.

I'd like to propose we bump the MSRV to 1.56.1. This is the version that adds rust edition 2021 and I think will enable us to use the latest versions of most of our upstream dependencies (it's also the MSRV that qiskit-terra picked for similar reasons).

mtreinish:
The previous minimum support Rust version of 1.48 was originally selected to match the latest version of Rust available through Debian stable. This was done primarily to support the piwheels build environemnt which installs rustc via Debian packages. However, the 1.48 MSRV is causing a lot development friction as it's quite old and is holding back a lot of dependencies. When weighing the relatively tiny piwheels user base (for an unsupported platform) against this it's probably time to decouple our MSRV for rustworkx from that and pick a version that makes the most sense for our development needs.

This commit raises the minimum supported rust version to 1.56.1 and then bumps all our dependencies held back because of the previous MSRV. At the same time the rust edition used for rustworkx is changed to be edition 2021 instead of edition 2018. This will enable us to leverage newer rust syntax now.

Fixes #716


IvanIsCoding (approved):
LGTM> Rust 1.56.1 is already one year old, I think it is worth updating to use the 2021 edition

PR #721, opened 2022-11-01
Add New Paper Details

IvanIsCoding:
Update citation to reflect https://doi.org/10.21105/joss.03968 🍾


README.md
```diff
@@ -7,7 +7,7 @@
 [![](https://img.shields.io/pypi/dm/rustworkx.svg?style=popout-square)](https://pypi.org/project/rustworkx/)
 [![Coverage Status](https://coveralls.io/repos/github/Qiskit/rustworkx/badge.svg?branch=main)](https://coveralls.io/github/Qiskit/rustworkx?branch=main)
 [![Minimum rustc 1.56.1](https://img.shields.io/badge/rustc-1.56.1+-blue.svg)](https://rust-lang.github.io/rfcs/2495-min-rust-version.html)
-[![arXiv](https://img.shields.io/badge/arXiv-2110.15221-b31b1b.svg)](https://arxiv.org/abs/2110.15221)
+[![DOI](https://joss.theoj.org/papers/10.21105/joss.03968/status.svg)](https://doi.org/10.21105/joss.03968)
```
mtreinish, line 10: I think we can probably keep both badges, there isn't any overhead for having both arxiv and joss listed here imk. But, if there is a standard convention for this I'm fine with that

IvanIsCoding: Fair point, I kept both and added a Zenodo badge as well

PR #725, opened 2022-11-03
Fix clippy error with new Rust 1.65 release

mtreinish:
Rust 1.65 was recently released which added some new clippy rules. This commit fixes the failure reported when running clippy with 1.65. This would have caused a CI failure as we start using the new rust version in CI.


IvanIsCoding (approved):
LGTM, seems straightforward

PR #730, opened 2022-11-07
Fix support of sequence protocol for returned lists

### LINKED ISSUE #696: `NodeIndices` does not implement sequence protocol

tuxu:
Add the `sequence` option to PyO3's `pyclass`, so that the `sq_length` slot is implemented [1]. Implementing this method is required for sequence types, or Python C API functions like `PySequence_Size` will fail with an error.

The `reversed` built-in method relies on `PySequence_*` methods. A test reversing `NodeIndices` is added to guard against future violations of the sequence protocol.

Fixes #696.

[1]: https://github.com/PyO3/pyo3/pull/2567

### DIFF
--- a/releasenotes/notes/fix-sequence-protocol-e95246e864cc850a.yaml
+++ b/releasenotes/notes/fix-sequence-protocol-e95246e864cc850a.yaml
@@ -1,0 +1,10 @@
+---
+fixes:
+  - |
+    Fixed an issue with the custom sequence return types,
+    :class:`~.BFSSuccessors`, :class:`~.NodeIndices`, :class:`~.EdgeList`,
+    :class:`~.WeightedEdgeList`, :class:`~.EdgeIndices`, and :class:`~.Chains`
+    where they previosuly were missing certain attributes that prevented them
+    being used as a sequence for certain built-in functions such as
+    ``reversed()``.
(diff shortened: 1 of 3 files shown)


mtreinish (approved):
LGTM, thanks for fixing this! I just pushed up a quick release note to document the fix. Lets merge this, backport it, and we can prepare a 0.12.1 release to get this fix out

PR #731, opened 2022-11-07
Add longest_simple_path() function to rustworkx

mtreinish:
This commit adds a new function longest_simple_path() which is used to find the longest simple path in the graph. It internally is running basically the same algorithm as all_pairs_all_simple_paths but instead of building a giant nested mapping of all the simple paths found in the graph it instead filters the return to just the longest path found. This is useful if you're quickly trying to find the longest path in the graph as it avoids the overhead of building the return object and then iterating over it again.


src/connectivity/mod.rs
```diff
@@ -736,6 +737,134 @@ pub fn graph_all_pairs_all_simple_paths(
...
 
+/// Return a longest simple path in the graph
+///
+/// This function searches computes all pairs of all simple paths and returns
+/// a path of the longest length from that set. It is roughly equivalent to
+/// running something like::
+///
+///     from rustworkx import all_pairs_all_simple_paths
+///
+///     max((y.values for y in all_pairs_all_simple_paths(graph).values()), key=lambda x: len(x))
```
alexanderivrii, line 748: @mtreinish, thank you for implementing this function for me. After running multiple experiments from within Qiskit, I believe that `longest_simple_path` works correctly. However, running the documented equivalent formulation raises an error. First, I believe it should be `y.values()` and not `y.values`, but this is still not fully correct. The following does seem to work:
```
simple_path_pairs = rx.all_pairs_all_simple_paths(graph)
longest_path = max((u for y in simple_path_pairs.values() for z in y.values() for u in z), key=lambda x: len(x))
```

src/connectivity/mod.rs
```diff
@@ -628,6 +629,138 @@ pub fn graph_all_pairs_all_simple_paths(
...
+/// :returns: A sequence of node indices that represent the longest simple graph
+///     found in the graph. If the graph is empty ``None`` will be returned instead.
+/// :rtype: NodeIndices
+#[pyfunction]
+#[pyo3(text_signature = "(graph, /)")]
+pub fn digraph_longest_simple_path(graph: &digraph::PyDiGraph) -> Option<NodeIndices> {
+    if graph.graph.node_count() == 0 {
+        return None;
+    } else if graph.graph.edge_count() == 0 {
+        return Some(NodeIndices {
```
Cryoris, line 666: This is rather a question than a comment: It seems that this line will always return some value, as we checked before that nodes do exist. Is `Some` needed here to comply with the return type of the function (`Option<...>`), which allows `None` in the results?

src/connectivity/mod.rs
```diff
@@ -628,6 +629,138 @@ pub fn graph_all_pairs_all_simple_paths(
...
+                    *u,
+                    &node_index_set,
+                    0,
+                    None,
+                )
+                .values()
+                .flat_map(|paths| {
+                    paths
+                        .iter()
+                        .map(|path| path.iter().map(|v| v.index()).collect::<Vec<usize>>())
```
Cryoris, line 687: I'm a bit confused by this line: wouldn't it be possible to already ask for the length of each path here? Or why do we need to iterate over that path and get the node indices (`v.index`)? (Maybe I also got lost and that's not what's actually happening 😛 )

src/connectivity/mod.rs
```diff
@@ -628,6 +629,138 @@ pub fn graph_all_pairs_all_simple_paths(
...
+/// ``RAYON_NUM_THREADS=4`` would limit the thread pool to 4 threads.
+///
+/// :param PyGraph graph: The graph to find the longest path in
+///
+/// :returns: A sequence of node indices that represent the longest simple graph
+///     found in the graph. If the graph is empty ``None`` will be returned instead.
+/// :rtype: NodeIndices
+#[pyfunction]
+#[pyo3(text_signature = "(graph, /)")]
+pub fn graph_longest_simple_path(graph: &graph::PyGraph) -> Option<NodeIndices> {
```
Cryoris, line 730: Is this function exactly the same as for the `PyDiGraph`? If yes, do we need to dispatch to two functions or couldn't they just be handled by the same?

tests/rustworkx_tests/graph/test_all_simple_paths.py
```diff
@@ -243,3 +243,32 @@ def test_all_simple_path_no_path(self):
...
         self.assertEqual({}, rustworkx.all_pairs_all_simple_paths(rustworkx.PyGraph()))
+
+
+class TestGraphLongestSimplePath(unittest.TestCase):
+    def setUp(self):
+        super().setUp()
+        self.graph = rustworkx.generators.cycle_graph(4)
+
+    def test_all_simple_paths(self):
+        res = rustworkx.longest_simple_path(self.graph)
```
Cryoris, line 254: This seems like a very simple test, as each `n`-node path is a longest path, but since the code is already tested in the DiGraph case I would be fine with this 😄

Cryoris (commented):
Just to small comprehensive questions below, otherwise LGTM 🙂

src/connectivity/mod.rs
mtreinish: Yep, exactly. Since the return type is `Option<NodeIndices>` we have to explicitly say the return is something here and not None. When `Option<T>` is used you need to either be `Some(T)` or `None`. The underlying reason is `Option` is just an `enum` defined like:

```rust
pub enum Option<T> {
    None,
    Some(T),
}
```

so to create an Option object we need to either be `None` or `Some`.

src/connectivity/mod.rs
mtreinish: I can combine the core logic into a combined generic rust function very easily (I'll do this). But, we need two separate entrypoints for the python interface because of the typing constraints around building the rust<->python interface, pyo3 needs to know the concrete type to generate the code for the python interface.

mtreinish: I did this in: https://github.com/Qiskit/rustworkx/pull/731/commits/1838e2d80517864c31aa6e71a523cdd3cc914797

src/connectivity/mod.rs
mtreinish: I've simplified the logic to minimize the intermediate object creation in https://github.com/Qiskit/rustworkx/pull/731/commits/f22f64b44077b292bd3eeaa0ffc780e323b7c185 and https://github.com/Qiskit/rustworkx/pull/731/commits/733e9ecd0e76a03e5a17412d8aebfb3908a7e66d I haven't benchmarked this it might be slightly faster (but I also expect the compiler was optimizing away a bunch in the previous version of the code)

mtreinish: I played with it a little bit more and was able to drop all the copying so it should definitely be faster I think

Cryoris (approved):
Thanks for the explanations and updates -- LGTM!

mtreinish:
I've been testing this locally and I need to tweak the per thread path calculation to avoid storing so many intermediate values. Even removing all the copying there is a huge memory explosion running this even on modestly sized graphs because we're storing more than we need to by generating a mapping of all the simple paths from each source node and filtering that per thread.

Instead to fix this we'll need to add a new function to just store a single path (I have an in progress commit locally that I'm finalizing to do this now) to avoid taking too much memory. This algorithm still doesn't scale well, but it never would because the problem is np complete or np hard (not sure the formal characterization/proof but it explodes in search space size as the graphs get bigger)

IvanIsCoding (approved):
LGTM, the problem is NP-Hard and a brute-force solution is what we would expect from a general graph library

PR #736, opened 2022-11-09
Add summary of the rustworkx benchmarks to the documentation

cometta:
benchmark of https://github.com/mtreinish/retworkx-comparison-benchmarks . Refer to issue https://github.com/Qiskit/rustworkx/issues/550


IvanIsCoding (commented):
Thanks for taking care of this. I have three minor comments regarding the images:
* we might need to increase the axis on the Time to create a weighted directed graph
* we need to update the captions to say `rustworkx` not `retworkx`
* Because we are on the web I think saving the image as an `.svg` might be better than using `.png`

IvanIsCoding:
> I made changes according to your comments. Let me know if anything else pending

Can you run `pip show <package>` for each graph package and post the configuration of your system (e.g. Ubuntu 22.04 with AMD Ryzen XYZ and 128Gb of RAM) on the comments for this PR?

I will take over and add these details and the citations for where we grabbed the datasets, then we should be good to go

IvanIsCoding (approved):

PR #740, opened 2022-11-11
Update priority-queue to 1.3.0

vlvrd:


mtreinish:
Thank you for opening a PR, unfortunately we won't be able to merge this right now. The new version of priority-queue requires a newer version of rust than our current MSRV (minimum supported Rust version) which is 1.56.1. We hit this same CI failure on #740 which was opened by dependabot after priority-queue was released.  We can revisit using the newer version of priority-queue when we bump our MSRV. However, that'll likely be some time because we just recently bumped the version to 1.56 in #717 so we're unlikely to change it again in the short term.

I'm going to close this PR because of this, but please feel free to reopen this PR if I'm missing something or you feel there is more to discuss on this.

PR #753, opened 2022-11-28
Deduplicate wheel job definitions and update cibuildwheel

mtreinish:
This commit simplifies the wheel action configuration and deduplicates much of the job definitions. This is done primarily by unifying the cibuildwheel configuration in the pyproject.toml to set default values for all the common options. Then the separate duplicated install and run job stages are switched over to the native action published by cibuildwheel which deduplicates the configuration further. At the same time this commit bumps the configuration to use the latest version of cibuildwheel.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -58,27 +58,15 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
-      - name: Install cibuildwheel
-        run: |
-          python -m pip install cibuildwheel==2.10.1 twine
       - name: Build wheels
-        run: |
-          python -m cibuildwheel --output-dir wheelhouse
-        env:
-          CIBW_BEFORE_ALL_LINUX: "yum install -y wget && {package}/tools/install_rust.sh"
-          CIBW_ENVIRONMENT_LINUX: 'PATH="$PATH:$HOME/.cargo/bin"'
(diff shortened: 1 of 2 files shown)


mtreinish:
Marking as on hold until I test the new configuration works as expected.

.github/workflows/wheels.yml
```diff
@@ -1,47 +1,9 @@
...
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          override: true
-      - name: Run cargo publish
-        run: |
-          cd rustworkx-core
-          cargo publish
-        env:
-          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
```
IvanIsCoding, line 22: If I understood well, `rustworkx-core` will not be published by the action? Is this a mistake?

IvanIsCoding (commented):
This would remove a good chunk of code duplication, I just left a comment about `rustworkx-core`

.github/workflows/wheels.yml
mtreinish: This pr is still configured for testing, once I've got everything in a good place I will revert https://github.com/Qiskit/rustworkx/pull/753/commits/31ba30ecd16a09115ae0f47f90e038e9b7e42206 and remove the on hold label

mtreinish:
Superseded by #997

PR #754, opened 2022-11-30
Add SECURITY.md to document security policy

mtreinish:
This commit adds a SECURITY.md file to the repository to document the security policy. I recently enabled the private security advisories feature on the repository (which is a relatively new "beta" feature in github). Since we now have a place to privately raise potential security issues (besides email) it is good to have a documented policy on how security vulnerabilities should be reported and our support policy for version we will fix (which is just the latest release). As the project matures we can adjust this policy as needed (likely to support more than one version at a time).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Rustworkx supports one minor version release at a time both for bug and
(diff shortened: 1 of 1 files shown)


IvanIsCoding (approved):
Cool. We haven't been affected by a security issue yet, but it is good to have this in place before an incident

PR #756, opened 2022-12-01
add find_predecessor_node_by_edge.

ewinston:
add function to get immediate predecessor of node by edge.

### DIFF
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -1737,7 +1737,7 @@ impl PyDiGraph {
     /// :returns: The node object that has an edge from it to the provided
     ///     node index which matches the provided condition
     #[pyo3(text_signature = "(self, node, predicate, /)")]
-    pub fn find_adjacent_predecessor_node_by_edge(
+    pub fn find_predecessor_node_by_edge(
         &self,
         py: Python,
         node: usize,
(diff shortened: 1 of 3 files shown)


src/digraph.rs
```diff
@@ -1724,6 +1724,42 @@ impl PyDiGraph {
         Err(NoSuitableNeighbors::new_err("No suitable neighbor"))
     }
 
+    /// Find a target node with a specific edge
+    ///
+    /// This method is used to find a target node that is an adjacent predecessor
```
mtreinish, line 1729:
```suggestion
    /// This method is used to find an adjacent predecessor
```

mtreinish (commented):
Overall I think this looks good, thanks for writing this. The only thing I think missing (besides the two nits on the docstring inline) is a negative test that tests we raise the appropriate error if there is no matching inbound edge found

ewinston:
@mtreinish Thanks for the tip on how to implement this.

mtreinish (commented):
The code LGTM now thanks. I just left inline comments to remove "adjacent" from the function name. I think it's a bit redundant with predecessor in the name already

The other thing I think we should have is a reno release note documenting the new feature.

mtreinish (approved):

PR #762, opened 2022-12-16
Fix clippy with Rust 1.66.0

mtreinish:
The recent release of Rust 1.66.0 [1] added new on by default clippy rules which are causing failures when we run the clippy CI job with the new rust version. This commit updates the code to fix the issues that clippy has identified which will unblock CI.

[1] https://blog.rust-lang.org/2022/12/15/Rust-1.66.0.html


IvanIsCoding (approved):
LGTM, seems straightforward

PR #765, opened 2022-12-24
Add token swapper to rustworkx-core

### LINKED ISSUE #701: Add implementation of approximate token swapper to rustworkx and rustworkx-core

In qiskit-terra we're using an implementation of https://arxiv.org/abs/1602.05150 to handle transforming circuit layouts. This implementation for qiskit-terra is here: https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/transpiler/passes/routing/algorithms/token_swapper.py which is built around a rustworkx graph object implemented in Python.

However, it would be great it retworkx had a native implementation of this algorithm and had it exposed in both rustworkx and rustworkx-core (to expose it to rust consumers of the library too) especially since it's generic and strictly structural.

enavarro51:
Fixes #701

This PR ports the ApproximateTokenSwapper functionality written in python in qiskit-terra to rustworkx-core. It was based on         the paper: Approximation and Hardness for Token Swapping by Miltzow et al. (2016). ArXiV: https://arxiv.org/abs/1602.05150 and work by the qiskit-terra team.

- [x] Connect to rustworkx python interface
- [x] Additional tests
- [x] Release note

Using a jupyter notebook with the python interface to the rust `token_swapper`, the following execution times were noted with a series of 1 to 100 element permutations and creation of swaps to build the inverse permutation,
```
Main: 19.8 sec +- 2.8 sec (edited)
New code: 7.47 sec +- .1 sec (edited)
```
Further rust optimizations may be possible.


enavarro51:
There is a `trials` parameter passed to the `token_swapper` in the `qiskit-terra` code that defaults to 4. I was just wondering what the rationale is for that setting.

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,556 @@
...
+        }
+        // Build the first results Vec until a 0 len result is found
+        let mut first_results: Vec<Vec<Swap>> = Vec::new();
+        for result in trial_results {
+            if result.is_empty() {
+                first_results.push(result);
+                break;
+            }
+            first_results.push(result);
+        }
```
mtreinish, line 147: I would integrate this with my iterator example above and make this just a filter function with a boolean, something like:

```rust
let mut empty_found = false;

...

.filter(|x| {
    if empty_found {
        None
    } else {
        if x.is_empty() {
            empty_found = true;
        }
        Some(x)
    }
})
.collect()
```

If you combine this with the iterator I suggested above this will remove an intermediate allocation and be much more efficient.

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,556 @@
...
+        let mut res_min = MAX;
+        let mut final_result: Vec<Swap> = Vec::new();
+        for res in first_results {
+            let res_len = res.len();
+            if res_len < res_min {
+                final_result = res;
+                res_min = res_len;
+            }
+        }
+        final_result
```
mtreinish, line 158: Heh, this can also be combined into the same iterator chain above. You can use `.min_by_key()` instead of collect and avoid all the Vec allocations for results. This makes the full iterator chain something like:

```rust
     let mut empty_found = false;
     let final_result Vec<Swap> = (0..self.trials)
            .map(|_| {
                let results = self.trial_map(
                    &mut digraph.clone(),
                    &mut sub_digraph.clone(),
                    &mut tokens.clone(),
                    &mut todo_nodes.clone(),
                   &mut rng_seed,
                )
        })
        .filter(|result| {
            if empty_found {
                None
            } else {
               if x.is_empty() {
                    empty_found = true;
               }
                Some(x)
            }
        })
        .min_by_key(|result| result.len())
        .unwrap()
```

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,556 @@
...
+                None,
+            );
+            let dist_node: Result<DictMap<G::NodeId, usize>, &str> = dijkstra(
+                &self.graph,
+                id_node,
+                Some(id_token),
+                |_| Ok::<usize, &str>(1),
+                None,
+            );
+            if dist_neighbor.unwrap()[&id_token] < dist_node.unwrap()[&id_token] {
```
mtreinish, line 196: I would move these unwraps to the variable definitions and remove the `Result<>` from the types on `dist_node` and `dist_neighbor`. I just think that would be easier to read since we're never going to error in `dijkstra()` as our `edge_cost` argument is hard coded to 1.

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,556 @@
...
+        // sub_digraph. Then add new token edges back in.
+        for node in [node1, node2] {
+            let mut edge_nodes = vec![];
+            for successor in digraph.neighbors_directed(node, Outgoing) {
+                edge_nodes.push((node, successor));
+            }
+            for (edge_node1, edge_node2) in edge_nodes {
+                let edge = digraph.find_edge(edge_node1, edge_node2).unwrap();
+                digraph.remove_edge(edge);
+            }
```
mtreinish, line 316: Is there a reason to use `edge_nodes` as a vec here? Can't we just do this as:
```rust
for edge_node2 in digraph.neighbors_directed(node, Outgoing) {
    let edge = digraph.find_edge(node, edge_node2).unwrap();
    digraph.remove_edge(edge);
}
```

This will avoid any memory allocations which will be much faster.

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,556 @@
...
+            let mut edge_nodes = vec![];
+            for successor in sub_digraph.neighbors_directed(node, Outgoing) {
+                edge_nodes.push((node, successor));
+            }
+            for (edge_node1, edge_node2) in edge_nodes {
+                let edge = sub_digraph.find_edge(edge_node1, edge_node2).unwrap();
+                sub_digraph.remove_edge(edge);
+            }
```
mtreinish, line 324: Same comment here as above.

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,556 @@
...
+        }
+    }
+}
+
+/// This module performs an approximately optimal Token Swapping algorithm
+/// Supports partial mappings (i.e. not-permutations) for graphs with missing tokens.
+///
+/// Based on the paper: Approximation and Hardness for Token Swapping by Miltzow et al. (2016)
+/// ArXiV: https://arxiv.org/abs/1602.05150
+/// and generalization based on our own work.
```
mtreinish, line 345: I assume you're just copying this from qiskit-terra. I'm not sure what "and generalization based on our work" means in this context.

mtreinish (commented):
Thanks a bunch for doing this! I'm really looking forward to having this algorithm in rustworkx. I have a few inline comments that I think will clean up the implementation a bit.

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,556 @@
...
+            );
+        }
+        // Do the trials and build a results Vec
+        let mut trial_results: Vec<Vec<Swap>> = Vec::new();
+        for _ in 0..self.trials {
+            let results = self.trial_map(
+                &mut digraph.clone(),
+                &mut sub_digraph.clone(),
+                &mut tokens.clone(),
+                &mut todo_nodes.clone(),
```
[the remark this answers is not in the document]
enavarro51: Not sure how important this is. It gets a little hairy since `add_token_edges` is called before this with the uncloned graphs, etc. and it's also called from within `trial_map` on the cloned graphs.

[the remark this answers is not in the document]
mtreinish: It's not that bad to do, I tested it locally this is the diff needed:

```diff
diff --git a/rustworkx-core/src/token_swapper.rs b/rustworkx-core/src/token_swapper.rs
index ed09648..7e7b8d9 100644
--- a/rustworkx-core/src/token_swapper.rs
+++ b/rustworkx-core/src/token_swapper.rs
@@ -128,10 +128,10 @@ where
         let mut trial_results: Vec<Vec<Swap>> = Vec::new();
         for _ in 0..self.trials {
             let results = self.trial_map(
-                &mut digraph.clone(),
-                &mut sub_digraph.clone(),
-                &mut tokens.clone(),
-                &mut todo_nodes.clone(),
+                digraph.clone(),
+                sub_digraph.clone(),
+                tokens.clone(),
+                todo_nodes.clone(),
                 &mut rng_seed,
             );
             trial_results.push(results);
@@ -202,10 +202,10 @@ where
 
     fn trial_map(
         &mut self,
-        digraph: &mut StableGraph<(), (), Directed>,
-        sub_digraph: &mut StableGraph<(), (), Directed>,
-        tokens: &mut HashMap<NodeIndex, NodeIndex>,
-        todo_nodes: &mut Vec<NodeIndex>,
+        mut digraph: StableGraph<(), (), Directed>,
+        mut sub_digraph: StableGraph<(), (), Directed>,
+        mut tokens: HashMap<NodeIndex, NodeIndex>,
+        mut todo_nodes: Vec<NodeIndex>,
         rng_seed: &mut Pcg64,
     ) -> Vec<Swap> {
         // Create a random trial list of swaps to move tokens to optimal positions
@@ -218,11 +218,11 @@ where
             let todo_node = todo_nodes[random];
 
             // If there's a cycle in sub_digraph, add it to swap_edges and do swap
-            let cycle = find_cycle(&*sub_digraph, Some(todo_node));
+            let cycle = find_cycle(&sub_digraph, Some(todo_node));
             if !cycle.is_empty() {
                 for edge in cycle[1..].iter().rev() {
                     swap_edges.push(*edge);
-                    self.swap(edge.0, edge.1, digraph, sub_digraph, tokens, todo_nodes);
+                    self.swap(edge.0, edge.1, &mut digraph, &mut sub_digraph, &mut tokens, &mut todo_nodes);
                 }
                 steps += cycle.len() - 1;
             // If there's no cycle, see if there's an edge target that matches a token key.
@@ -237,10 +237,10 @@ where
                         self.swap(
                             new_edge.0,
                             new_edge.1,
-                            digraph,
-                            sub_digraph,
-                            tokens,
-                            todo_nodes,
+                            &mut digraph,
+                            &mut sub_digraph,
+                            &mut tokens,
+                            &mut todo_nodes,
                         );
                         steps += 1;
                         found = true;
@@ -251,7 +251,7 @@ where
                 // an unhappy node. Look for a predecessor and add node and pred
                 // to swap_edges and do swap
                 if !found {
-                    let cycle: Vec<Edge> = find_cycle(&*digraph, Some(todo_node));
+                    let cycle: Vec<Edge> = find_cycle(&digraph, Some(todo_node));
                     let unhappy_node = cycle[0].0;
                     let mut found = false;
                     let di2 = &mut digraph.clone();
@@ -261,10 +261,10 @@ where
                             self.swap(
                                 unhappy_node,
                                 predecessor,
-                                digraph,
-                                sub_digraph,
-                                tokens,
-                                todo_nodes,
+                                &mut digraph,
+                                &mut sub_digraph,
+                                &mut tokens,
+                                &mut todo_nodes,
                             );
                             steps += 1;
                             found = true;
```

I think it will be good to do because it's a lot clearer about the ownership model being used. I don't think it'll have a big impact on performance or anything, but it's more idiomatic and easier to read.

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,556 @@
...
+
+use crate::connectivity::find_cycle;
+use crate::dictmap::*;
+use crate::shortest_path::dijkstra;
+use crate::traversal::dfs_edges;
+
+type Swap = (NodeIndex, NodeIndex);
+type Edge = (NodeIndex, NodeIndex);
+
+struct TokenSwapper<G: GraphBase>
```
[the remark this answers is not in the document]
mtreinish: That's fair. I was just wondering because everything seemed to involve the single struct. Either way I'm fine with this structure, if a need comes up to make it public it's easy to add later

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,556 @@
...
+
+/// This module performs an approximately optimal Token Swapping algorithm
+/// Supports partial mappings (i.e. not-permutations) for graphs with missing tokens.
+///
+/// Based on the paper: Approximation and Hardness for Token Swapping by Miltzow et al. (2016)
+/// ArXiV: https://arxiv.org/abs/1602.05150
+/// and generalization based on our own work.
+///
+/// The inputs are a partial `mapping` to be implemented in swaps, and the number of `trials` to
+/// perform the mapping. It's minimized over the trials.
```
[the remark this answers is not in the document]
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/765/commits/2d74c45a8e8c6e763cd7ce32034d9f094f991661

rustworkx-core/src/token_swapper.rs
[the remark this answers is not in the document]
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/765/commits/2d74c45a8e8c6e763cd7ce32034d9f094f991661

rustworkx-core/src/token_swapper.rs
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/765/commits/2d74c45a8e8c6e763cd7ce32034d9f094f991661

rustworkx-core/src/token_swapper.rs
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/765/commits/2d74c45a8e8c6e763cd7ce32034d9f094f991661

rustworkx-core/src/token_swapper.rs
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/765/commits/2d74c45a8e8c6e763cd7ce32034d9f094f991661

rustworkx-core/src/token_swapper.rs
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/765/commits/2d74c45a8e8c6e763cd7ce32034d9f094f991661

rustworkx-core/src/token_swapper.rs
enavarro51: Still working on this. Will need a little refinement.

enavarro51: The closure for the `filter` method has to return a bool. In the previous code, I believe the only reason for checking for `is_empty` is to break out of the `for` early since an empty will be the min. Breaking out of an iterator is not so easy as best I can tell.

This code works and I think it's pretty efficient without the `is_empty` unless `trials` is a very large number,
```
        (0..self.trials)
            .map(|_| {
                self.trial_map(
                    digraph.clone(),
                    sub_digraph.clone(),
                    tokens.clone(),
                    todo_nodes.clone(),
                    &mut rng_seed,
                )
            })
            .min_by_key(|result| result.len())
            .unwrap()
```

mtreinish: Sure, that makes sense especially for small numbers of `trials`. 

The other thing which I think might be worth looking into here is parallelism. Since this iterator is quite compact and has no shared data dependency (because of the `clone()` calls we can easily use rayon here and potentially further speed up this iterator. It should be mostly a matter of adding a `.into_par_iter()` before the map (although we'll have to adjust the seed usage so it's not a shared value across all the threads.

enavarro51: Ok. Will look into the parallel.

enavarro51: Ok. Got it working. Now with my local test,
- Old token_swapper: 16.7 sec
- New without parallel: 7.6 sec
- New with parallel: 2.6 sec

I'll update and push shortly.

enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/765/commits/83ccd8c397727df14e27164e3ede77d84243086b.

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,555 @@
...
+        for node in self.graph.node_identifiers() {
+            self.add_token_edges(
+                self.node_map[&node],
+                &mut digraph,
+                &mut sub_digraph,
+                &mut tokens,
+            );
+        }
+        // Do the trials
+        let seed = self.rng_seed;
```
mtreinish, line 125: This will cause the same seed to be used on each of the trials which will result in each result to be the same. The issue is unlike in the serial case when there was a shared rng instance that would cycle based on the fixed seed, in this case we're initializing a thread local rng with the same seed value in each thread which will result in the same behavior across all the trials. The pattern I normally use to fix this is to create an outer rng to generate `self.trials` list of seeds and then use each of those seeds to initialize a thread local rng. For example in this case it would be:

```diff
diff --git a/rustworkx-core/src/token_swapper.rs b/rustworkx-core/src/token_swapper.rs
index 8184f7f..4239a43 100644
--- a/rustworkx-core/src/token_swapper.rs
+++ b/rustworkx-core/src/token_swapper.rs
@@ -122,10 +122,17 @@ where
             );
         }
         // Do the trials
-        let seed = self.rng_seed;
-        (0..self.trials)
+        let outer_rng: Pcg64 = match self.rng_seed {
+            Some(rng_seed) => Pcg64::seed_from_u64(rng_seed),
+            None => Pcg64::from_entropy(),
+        };
+        let seed_vec: Vec<u64> = outer_rng
+            .sample_iter(&rand::distributions::Standard)
+            .take(self.trials)
+            .collect();
+        seed_vec
             .into_par_iter()
-            .map(|_| {
+            .map(|seed| {
                 self.trial_map(
                     digraph.clone(),
                     sub_digraph.clone(),
@@ -190,15 +197,12 @@ where
         mut sub_digraph: StableGraph<(), (), Directed>,
         mut tokens: HashMap<NodeIndex, NodeIndex>,
         mut todo_nodes: Vec<NodeIndex>,
-        seed: Option<u64>,
+        seed: u64,
     ) -> Vec<Swap> {
         // Create a random trial list of swaps to move tokens to optimal positions
         let mut steps = 0;
         let mut swap_edges: Vec<Swap> = vec![];
-        let mut rng_seed: Pcg64 = match seed {
-            Some(rng_seed) => Pcg64::seed_from_u64(rng_seed),
-            None => Pcg64::from_entropy(),
-        };
+        let mut rng_seed: Pcg64 = Pcg64::seed_from_u64(seed);
         while !todo_nodes.is_empty() && steps <= 4 * digraph.node_count().pow(2) {
             // Choose a random todo_node
             let between = Uniform::new(0, todo_nodes.len());
```

enavarro51: Ah, of course. Will fix.

enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/765/commits/805be9cdba948bf2f22830f844db6c6fd332056f. Still doing 2.6 sec.

src/token_swapper.rs
```diff
@@ -0,0 +1,61 @@
...
+use petgraph::graph::NodeIndex;
+use pyo3::prelude::*;
+use rustworkx_core::token_swapper;
+
+/// This module performs an approximately optimal Token Swapping algorithm
+/// Supports partial mappings (i.e. not-permutations) for graphs with missing tokens.
+///
+/// Based on the paper: Approximation and Hardness for Token Swapping by Miltzow et al. (2016)
+/// ArXiV: https://arxiv.org/abs/1602.05150
+/// and generalization based on our own work.
```
mtreinish, line 26:
```suggestion
```

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,558 @@
...
+        }
+    }
+}
+
+/// Module to perform an approximately optimal Token Swapping algorithm. Supports partial
+/// mappings (i.e. not-permutations) for graphs with missing tokens.
+///
+/// Based on the paper: Approximation and Hardness for Token Swapping by Miltzow et al. (2016)
+/// ArXiV: <https://arxiv.org/abs/1602.05150>
+///
```
mtreinish, line 343: We should also document this function is multithreaded with rayon, since a rust user might need to know this depending on the larger global threading context. I do wonder if we want to add an argument for running in parallel like we do in other rustworkx-core functions (for example `betweenness_centrality` has `parallel_threshold` to control when rayon is used or not).

mtreinish (commented):
Overall this LGTM now, thanks for making the updates. I just have a couple small comments in the docs.

The other thing is the API docs for python aren't going to be built, you'll need to add the function to https://github.com/Qiskit/rustworkx/blob/main/docs/source/api.rst to include it in the python documentation.

rustworkx-core/src/token_swapper.rs
enavarro51: I can add the `parallel_threshold`. Looks like I can use the `CondIterator` to choose which one.

src/token_swapper.rs
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/765/commits/7a7ae3bca4873c0a7166cd005289bde33470de39.

src/token_swapper.rs
```diff
@@ -0,0 +1,61 @@
...
+///
+/// The inputs are a partial ``mapping`` to be implemented in swaps, and the number of ``trials``
+/// to perform the mapping. It's minimized over the trials.
```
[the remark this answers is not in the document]
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/765/commits/7a7ae3bca4873c0a7166cd005289bde33470de39.

mtreinish (commented):
Overall I think the code is ready to merge, but I found 2 more small little things inline. Basically I think we need to discuss the `panic!` and then one more python side test case which I think would be good to have.

tests/rustworkx_tests/test_token_swapper.py
```diff
@@ -0,0 +1,107 @@
...
+        graph = rx.PyGraph()
+        graph.extend_from_edge_list(
+            [(0, 1), (0, 2), (0, 3), (0, 4), (1, 2), (1, 3), (1, 4), (2, 3), (2, 4), (3, 4), (3, 6)]
+        )
+        permutation = {0: 4, 1: 0, 2: 3, 3: 6, 4: 2, 6: 1}
+        swaps = rx.graph_token_swapper(graph, permutation, 4, 4, 1)
+        swap_permutation(permutation, swaps)
+        self.assertEqual({i: i for i in permutation}, permutation)
+
+    def test_partial_simple(self) -> None:
```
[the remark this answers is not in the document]
enavarro51: Will do.

docs/source/api.rst
```diff
@@ -369,6 +369,7 @@ typed API based on the data type.
    rustworkx.graph_complement
    rustworkx.graph_union
    rustworkx.graph_tensor_product
+   rustworkx.graph_token_swapper
```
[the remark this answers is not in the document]
enavarro51: I vote for let it be for now.

tests/rustworkx_tests/test_token_swapper.py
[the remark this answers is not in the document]
enavarro51: Actually this is a problem. When using map `{0: 3}` it generates a swap list of `{(0, 1), (1, 2)}` and a final output of `{2: 3}`. I think when we create the digraph and sub-digraph that we may need to map the node id's. Right now we're just using `add_node` on these which will not match the original graph if a node has been removed in the original. 

I have some code I used in `planar_layout` to put the appropriate holes into a newly created stable graph. I could use that here. Let me know if you think that's a reasonable approach.

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,572 @@
...
+                    }
+                    if !found {
+                        panic!("unexpected stop")
+                    }
+                }
+            }
+        }
+        if !todo_nodes.is_empty() {
+            panic!("got todo nodes");
+        }
```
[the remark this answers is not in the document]
mtreinish: Hmm, do you want to just change the error message than. How about something like:

```suggestion
        assert!(
            todo_nodes.is_empty(),
            "The output final swap map is incomplete, this points to a bug in rustworkx please open an issue."
        );
```

That makes it a bit more clear in the code were doing a final sanity check and that it's pointing to an internal bug if there are leftover nodes.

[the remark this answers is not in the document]
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/765/commits/0a56fd8c6a05ace5d22bfc072783f7c47f822aaf. I also changed the one above it for !found to assert.

tests/rustworkx_tests/test_token_swapper.py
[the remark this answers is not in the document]
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/765/commits/0a56fd8c6a05ace5d22bfc072783f7c47f822aaf. It passes the test now.

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,608 @@
...
+    }
+
+    fn map(&mut self) -> Vec<Swap> {
+        let num_nodes = self.graph.node_bound();
+        let num_edges = self.graph.edge_count();
+
+        // Directed graph with nodes matching ``graph`` and
+        // edges for neighbors closer than nodes
+        let mut digraph = StableGraph::with_capacity(num_nodes, num_edges);
```
mtreinish, line 93: We should make this `self.graph.node_bound()` instead of `num_nodes()` because we're always going to overallocate the way this is being done since we grow the graph to `node_bound` no matter what.

rustworkx-core/src/token_swapper.rs
```diff
@@ -0,0 +1,608 @@
...
+        for gnode in self.graph.node_identifiers() {
+            let gidx = self.graph.to_index(gnode);
+            if gidx != count {
+                for idx in count..gidx {
+                    digraph.remove_node(NodeIndex::new(idx));
+                }
+                count = gidx;
+            }
+            count += 1;
+        }
```
mtreinish, line 110: Instead of doing this to reproduce the holes exactly what if we maintained a mapping so that we can convert `digraph` indices to self.graph indices for output. If we changed the type of `digraph` to `StableGraph<G::NodeId, ()>` then we can store the original index in the graph when we add nodes. Something like:

```rust
for gnode in self.graph.node_identifiers {
    digraph.add_node(gnode)
}
```

Then we can just lookup the original index by looking at the weight of the node.

mtreinish (commented):
Thanks for fixing the issue with the index holes (and sorry for my super slow turnaround, the qiskit release has been dominating my time). I left two inline comments which are mutually exclusive. I'm a bit concerned by the performance overhead of trying to recreate the holes the way you did to fix the issue so I suggested a different approach using a mapping to maintain the original indices. If that doesn't work then my other suggestion is just a small allocation tweak to improve the runtime performance slightly.

rustworkx-core/src/token_swapper.rs
enavarro51: Actually if you look above this, I actually changed `num_nodes` to be `node_bound`. Should have changed the var name.

mtreinish: Ah ok, yeah the variable name tripped me up. If this is already using the upper bound on the node index we're good here.

rustworkx-core/src/token_swapper.rs
mtreinish: Sure, that's reasonable let's take a look in a follow up. I agree it's not really a runtime concern so much. This method really just adds a single node iteration on top of everything which isn't terrible. I was thinking more about memory efficiency than runtime. BUt we can do this in a follow up.

mtreinish (approved):
LGTM, thanks for your hard work on this feature and sticking with it through all the updates and slow review turnaround (sorry about that)! This will be quite useful  both in rustworkx but also qiskit where we're using this function currently to reverse layout permutations in the transpiler with control flow instructions. Having it be significant;y faster will improve performance there, but also having it as a native rust function in rustworkx-core enables us to leverage this functionality in qiskit's rust modules too.

PR #772, opened 2023-01-05
Move complete_graph generator to rustworkx-core

enavarro51:
This PR migrates the complete_graph generator to rustworkx-core. It's based on https://github.com/Qiskit/rustworkx/pull/758.


src/generators.rs
```diff
@@ -2466,7 +2466,22 @@ pub fn complete_graph(
     weights: Option<Vec<PyObject>>,
     multigraph: bool,
 ) -> PyResult<graph::PyGraph> {
-    mesh_graph(py, num_nodes, weights, multigraph)
```
IvanIsCoding, line 2469: We should probably update `mesh_graph` to call `complete_graph` to avoid code duplication

src/generators.rs
```diff
@@ -2499,15 +2514,32 @@ pub fn complete_graph(
...
-#[pyfunction(multigraph = true)]
+#[pyfunction(multigraph = "true")]
 #[pyo3(text_signature = "(/, num_nodes=None, weights=None, multigraph=True)")]
 pub fn directed_complete_graph(
     py: Python,
     num_nodes: Option<usize>,
     weights: Option<Vec<PyObject>>,
     multigraph: bool,
 ) -> PyResult<digraph::PyDiGraph> {
-    directed_mesh_graph(py, num_nodes, weights, multigraph)
```
IvanIsCoding, line 2510: Same applies for the directed version

IvanIsCoding (commented):
For historical reasons we have both the `mesh_graph` and `complete_graph` generators. `complete_graph` is the name found in most libraries and it aliases to the implementation we had first. We need to keep this in mind when porting. it is okay to call the implementation from `rustworkx-core` we just need to keep the functions being aliases of each other.

src/generators.rs
enavarro51: Sure. Done in https://github.com/Qiskit/rustworkx/pull/772/commits/0717ca68f6f99f2aa12a6328106ba9c142db3d31.

src/generators.rs
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/772/commits/0717ca68f6f99f2aa12a6328106ba9c142db3d31.

src/generators.rs
```diff
@@ -474,42 +474,16 @@ pub fn mesh_graph(
...
+    let default_fn = || py.None();
+    let graph: StablePyGraph<Undirected> =
+        match core_generators::complete_graph(num_nodes, weights, default_fn, default_fn) {
+            Ok(graph) => graph,
+            Err(_) => {
+                return Err(PyIndexError::new_err(
+                    "num_nodes and weights list not specified",
+                ))
             }
-        }
```
IvanIsCoding, line 498: I would replace this with `complete_graph(py, num_nodes, weights, multigraph)` to avoid duplicate code

src/generators.rs
```diff
@@ -550,44 +524,16 @@ pub fn directed_mesh_graph(
...
-            node_removed: false,
-            check_cycle: false,
-            cycle_state: algo::DfsSpace::default(),
-            multigraph,
-            attrs: py.None(),
-        });
-    }
-    let num_edges = node_len * (node_len - 1);
-    let mut graph = StablePyGraph::<Directed>::with_capacity(node_len, num_edges);
-    match weights {
```
IvanIsCoding, line 571: Same apllies here, I'd call `directed_mesh_graph(py, num_nodes, weights, multigraph)`

IvanIsCoding (commented):
I added some comments about aliasing `mesh_graph` and `directed_mesh_graph` to avoid code duplication. After the aliases are there I think it should be good to go

src/generators.rs
enavarro51: Ok. Got it. It's on the way...

enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/772/commits/af756265926b98333723ab25192bcb5bfb035828.

src/generators.rs
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/772/commits/af756265926b98333723ab25192bcb5bfb035828.

IvanIsCoding (approved):
Awesome

PR #779, opened 2023-01-10
Move heavy_hex_graph generator to rustworkx-core

enavarro51:
This PR migrates the heavy_hex_graph generator to rustwokx-core. It's based on https://github.com/Qiskit/rustworkx/pull/758.

There was a bug in the rustworkx undirected version of the heavy_hex_graph, which created a different ordering in the edges from the directed version. The undirected tests were changed to fix it.


IvanIsCoding (approved):
LGTM

PR #784, opened 2023-01-15
Docs and consistency cleanup for rustworkx-core generators

enavarro51:
This PR cleans up the generators documentation and makes the code for the generators more consistent. It's based on #780 and should be merged after that one.


rustworkx-core/src/generators/binomial_tree_graph.rs
```diff
@@ -132,30 +115,26 @@ where
...
-                    default_edge_weight(),
-                );
-            }
-            if bidirectional && !find_edge(&mut graph, target_index, source_index) {
-                graph.add_edge(
+            graph.update_edge(
+                graph.from_index(source_index),
+                graph.from_index(target_index),
+                default_edge_weight(),
+            );
```
mtreinish, line 122: This isn't quite the same because if the edge already exists we're replacing the weight with the output of `default_edge_weight()`. This won't matter if `default_edge_weight()` returns the same value every time, but there is no guarantee that will be the case.

rustworkx-core/src/generators/lollipop_graph.rs
```diff
@@ -15,19 +15,18 @@ use std::hash::Hash;
...
-use super::utils::pairwise;
+use super::utils::{get_num_nodes, pairwise};
 use super::InvalidInputError;
 
-/// Generate a lollipop graph
+/// Generate an undirected lollipop graph where a complete graph is connected to a
+/// path.
 ///
-/// Arguments:
+/// .. note::
```
mtreinish, line 24: I don't think this syntax will work in rustdoc since it's markdown based. I'm not sure what the correct syntax is off the top of my head, we'll have to check the docs.

rustworkx-core/src/generators/mod.rs
```diff
@@ -26,7 +26,7 @@ mod path_graph;
 mod petersen_graph;
 mod star_graph;
 
-mod utils;
+pub mod utils;
```
mtreinish, line 29: Doesn't this expose `rustworkx_core::generators::mod::utils` as a public facing interface. Do we want to commit to the interface on the util functions there?

enavarro51: This is just there since `rustworkx/src/steiner_tree.rs` was using the `pairwise` fn in `generators.rs`. I moved `pairwise` to utils since it was being used by a couple of the generators, but no longer used in `generators.rs`. If you don't want pub here, the simplest would be to just duplicate `pairwise` in `steiner_tree`.

rustworkx-core/src/generators/binomial_tree_graph.rs
enavarro51: Ah, missed that. Was thinking of the defaults as fixed. I can return to doing the find and then the add if not there.

rustworkx-core/src/generators/lollipop_graph.rs
enavarro51: This passes the tests. Is there somewhere this would fail and not be tested?

rustworkx-core/src/generators/mod.rs
mtreinish: Ah, I missed that when I went through this earlier. I think that's fine but I just wonder about the rest of the utils module. If we only need `pairwise` maybe we should just export only that one function. I'm just worried about committing the API surface for what were previously internal functions in the python side.

rustworkx-core/src/generators/binomial_tree_graph.rs
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/784/commits/63d2307d1fd505b22ef802a407d05b26b27e08a9.

rustworkx-core/src/generators/lollipop_graph.rs
enavarro51: Thanks for that info. I went back through all the docs in core and I think everything should be clean now. I removed all the markdown and made things prettier. Turns out the trick for displaying Ascii is to use `text` instead of `rust` for code-like sections.
Should be done in https://github.com/Qiskit/rustworkx/pull/784/commits/63d2307d1fd505b22ef802a407d05b26b27e08a9 and https://github.com/Qiskit/rustworkx/pull/784/commits/b0e4dfc044c1db120755be7d4a4c82b6c574b412.

rustworkx-core/src/generators/mod.rs
enavarro51: So where should we put `pairwise` so it's accessible from both `rustworkx` and `rustworkx-core`?

src/generators.rs
```diff
@@ -579,62 +569,64 @@ pub fn directed_grid_graph(
...
+///            |   |   |
+///        ... F-S-F-S-F-...
+///            |   |   |
+///            .........
+///            |   |   |
+///        ... D   D   D ...
+///             \ /     \
+///        ...   S       ...
+///
+/// NOTE: This function generates the four-frequency variant of the heavy square code.
```
mtreinish, line 594: Should we use `.. note::` here? Since this will be rendered in a sphinx context we can use the directive.

src/generators.rs
```diff
@@ -644,66 +636,69 @@ pub fn binomial_tree_graph(
...
```
mtreinish, line 661: Same question about `.. note::` here.

src/generators.rs
```diff
@@ -907,156 +868,193 @@ pub fn directed_heavy_square_graph(
...
-///     ...-F   F-S-F ...
-///         |   |   |
-///     ... D   D   D ...
-///         |   |   |
-///     ... F-S-F   F-...
-///         |   |   |
-///     ... D   D-S-D ...
-///
+/// Creates a full r-ary tree of `n` nodes.
+/// Sometimes called a k-ary, n-ary, or m-ary tree.
```
mtreinish, line 1007: This might need a blank line, I think sphinx might treat these 2 lines as a single summary without it.

mtreinish (commented):
Overall this LGTM, thanks for fixing all this. I just left a couple of inline comments on the python side docs. The only other thing is a small change to making `rustworkx_core::generators::utils` public so we limit it to just the `pairwise` function.

rustworkx-core/src/generators/mod.rs
enavarro51: I added a top-level utils with pairwise in it. In https://github.com/Qiskit/rustworkx/pull/784/commits/5c3ea62972b3d9adcf62367ae55c14c030a8a270.

src/generators.rs
enavarro51: In https://github.com/Qiskit/rustworkx/pull/784/commits/5c3ea62972b3d9adcf62367ae55c14c030a8a270.

enavarro51:
Had a fairly major merge conflict with the pyo3 changes, which I think is all fixed, but you might want to take another look at generators.rs.

mtreinish (commented):
Overall I think this is mostly ready to go, just a couple small inline comments and questions. It's a bit tricky to review because the github webui shows some functions shifting positions, but I think it's all good

src/generators.rs
```diff
@@ -875,40 +831,53 @@ pub fn heavy_square_graph(py: Python, d: usize, multigraph: bool) -> PyResult<gr
...
 
-/// Generate an directed heavy square graph. Fig. 6 of
-/// https://arxiv.org/abs/1907.09528.
+/// Generate a directed heavy hex graph.
+///
+/// Fig. 2 of https://arxiv.org/abs/1907.09528
 /// An ASCII diagram of the graph is given by:
 ///
-/// .. code-block:: console
+/// .. code-block:: text
```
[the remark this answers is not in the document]
enavarro51: Don't actually know why there is a difference. I did run into a problem on the rust docs side with the ``\``'s on the heavy square drawing, but not sure that's related.

mtreinish (approved):
LGTM, thanks for making the updates

PR #787, opened 2023-01-20
Update PyO3 and Rust Numpy to 0.18.0

mtreinish:
The latest releases of PyO3 and rust numpy were just released. The changelogs for both can be found at:

https://pyo3.rs/v0.18.0/changelog

and

https://github.com/PyO3/rust-numpy/blob/main/CHANGELOG.md

Since the two releases are tightly coupled this commit opts to update them together instead of the normal dependabot workflow. Similarly some of our usage (mainly around signatures and argument ordering) for building a python interface using PyO3 has been deprecated. That usage is also updated in this PR to avoid compilation warnings, which will be treated as errors by clippy in CI.


src/layout/mod.rs
```diff
@@ -82,13 +83,13 @@ pub fn graph_spring_layout(
...
-    num_iter: Option<usize>,
-    tol: Option<f64>,
+    repulsive_exponent: i32,
+    adaptive_cooling: bool,
+    num_iter: usize,
+    tol: f64,
     weight_fn: Option<PyObject>,
     default_weight: f64,
-    scale: Option<f64>,
+    scale: f64,
```
IvanIsCoding, line 92: Is this a consequence of https://github.com/PyO3/pyo3/pull/2703?

IvanIsCoding (commented):
I think this is a not ordinary bump, so I will take some time to review it

src/layout/mod.rs
mtreinish: It's related to the signature changes more generally, but I don't think because of that pr directly (the changes to the visitor function were directly because of that pr). The changes here were because of the type mismatch between the default values set in the signature and these types, `2. != Option<f64>`

mtreinish:
> I think this is a not ordinary bump, so I will take some time to review it

Fwiw, almost all of the changes here are just using the new `#[pyo3(signature)] ` interface instead of `#[args]` or `#[pyfunction]`. I tried to keep the diff minimal besides that, but do check it carefully because it would have been easy for me to make a mistake.

src/layout/mod.rs
IvanIsCoding: I guess we could have used `Some(2.0)` instead. But because 99% call `spring_layout` from our `__init__.py` which never provides `None` to `graph_spring_layout`, I think this is fine

src/traversal/mod.rs
```diff
@@ -608,15 +625,21 @@ pub fn graph_dfs_search(
...
     graph: &digraph::PyDiGraph,
     source: Option<Vec<usize>>,
     weight_fn: Option<PyObject>,
-    visitor: PyDijkstraVisitor,
+    visitor: Option<PyDijkstraVisitor>,
 ) -> PyResult<()> {
+    if visitor.is_none() {
+        return Err(PyTypeError::new_err("Missing required argument visitor"));
+    }
+    let v = visitor.unwrap();
```
IvanIsCoding, line 642:
```suggestion
    let visitor = visitor.unwrap();
```
I'd rather do this than change the variable name

src/traversal/mod.rs
```diff
@@ -627,7 +650,7 @@ pub fn digraph_dijkstra_search(
         &graph.graph,
         starts,
         |e| edge_cost_fn.call(py, e.weight()),
-        |event| dijkstra_handler(py, &visitor, event),
+        |event| dijkstra_handler(py, &v, event),
```
IvanIsCoding, line 653:
```suggestion
        |event| dijkstra_handler(py, &visitor, event),
```

src/traversal/mod.rs
```diff
@@ -683,15 +706,21 @@ pub fn digraph_dijkstra_search(
...
     graph: &graph::PyGraph,
     source: Option<Vec<usize>>,
     weight_fn: Option<PyObject>,
-    visitor: PyDijkstraVisitor,
+    visitor: Option<PyDijkstraVisitor>,
 ) -> PyResult<()> {
+    if visitor.is_none() {
+        return Err(PyTypeError::new_err("Missing required argument visitor"));
+    }
+    let v = visitor.unwrap();
```
IvanIsCoding, line 723:
```suggestion
    let visitor = visitor.unwrap();
```

src/traversal/mod.rs
```diff
@@ -702,7 +731,7 @@ pub fn graph_dijkstra_search(
         &graph.graph,
         starts,
         |e| edge_cost_fn.call(py, e.weight()),
-        |event| dijkstra_handler(py, &visitor, event),
+        |event| dijkstra_handler(py, &v, event),
```
IvanIsCoding, line 734:
```suggestion
        |event| dijkstra_handler(py, &visitor, event),
```

src/traversal/mod.rs
```diff
@@ -284,22 +285,26 @@ pub fn descendants(graph: &digraph::PyDiGraph, node: usize) -> HashSet<usize> {
...
     py: Python,
     graph: &digraph::PyDiGraph,
     source: Option<Vec<usize>>,
-    visitor: PyBfsVisitor,
+    visitor: Option<PyBfsVisitor>,
 ) -> PyResult<()> {
+    if visitor.is_none() {
+        return Err(PyTypeError::new_err("Missing required argument visitor"));
+    }
+    let v = visitor.unwrap();
```
IvanIsCoding, line 301:
```suggestion
    let visitor = visitor.unwrap();
```

src/traversal/mod.rs
```diff
@@ -284,22 +285,26 @@ pub fn descendants(graph: &digraph::PyDiGraph, node: usize) -> HashSet<usize> {
...
     let starts: Vec<_> = match source {
         Some(nx) => nx.into_iter().map(NodeIndex::new).collect(),
         None => graph.graph.node_indices().collect(),
     };
 
-    breadth_first_search(&graph.graph, starts, |event| {
-        bfs_handler(py, &visitor, event)
-    })?;
+    breadth_first_search(&graph.graph, starts, |event| bfs_handler(py, &v, event))?;
```
IvanIsCoding, line 307:
```suggestion
    breadth_first_search(&graph.graph, starts, |event| bfs_handler(py, &visitor, event))?;
```

src/traversal/mod.rs
```diff
@@ -370,22 +375,26 @@ pub fn digraph_bfs_search(
...
     py: Python,
     graph: &graph::PyGraph,
     source: Option<Vec<usize>>,
-    visitor: PyBfsVisitor,
+    visitor: Option<PyBfsVisitor>,
 ) -> PyResult<()> {
+    if visitor.is_none() {
+        return Err(PyTypeError::new_err("Missing required argument visitor"));
+    }
+    let v = visitor.unwrap();
```
IvanIsCoding, line 391:
```suggestion
    let visitor = visitor.unwrap();
```

src/traversal/mod.rs
```diff
@@ -538,22 +551,26 @@ pub fn digraph_dfs_search(
...
     py: Python,
     graph: &graph::PyGraph,
     source: Option<Vec<usize>>,
-    visitor: PyDfsVisitor,
+    visitor: Option<PyDfsVisitor>,
 ) -> PyResult<()> {
+    if visitor.is_none() {
+        return Err(PyTypeError::new_err("Missing required argument visitor"));
+    }
+    let v = visitor.unwrap();
```
IvanIsCoding, line 567:
```suggestion
    let visitor = visitor.unwrap();
```

src/traversal/mod.rs
```diff
@@ -538,22 +551,26 @@ pub fn digraph_dfs_search(
...
     let starts: Vec<_> = match source {
         Some(nx) => nx.into_iter().map(NodeIndex::new).collect(),
         None => graph.graph.node_indices().collect(),
     };
 
-    depth_first_search(&graph.graph, starts, |event| {
-        dfs_handler(py, &visitor, event)
-    })?;
+    depth_first_search(&graph.graph, starts, |event| dfs_handler(py, &v, event))?;
```
IvanIsCoding, line 573:
```suggestion
    depth_first_search(&graph.graph, starts, |event| dfs_handler(py, &visitor, event))?;
```

src/traversal/mod.rs
```diff
@@ -370,22 +375,26 @@ pub fn digraph_bfs_search(
...
+    let v = visitor.unwrap();
     let starts: Vec<_> = match source {
         Some(nx) => nx.into_iter().map(NodeIndex::new).collect(),
         None => graph.graph.node_indices().collect(),
     };
 
-    breadth_first_search(&graph.graph, starts, |event| {
-        bfs_handler(py, &visitor, event)
-    })?;
+    breadth_first_search(&graph.graph, starts, |event| bfs_handler(py, &v, event))?;
```
IvanIsCoding, line 397:
```suggestion
    breadth_first_search(&graph.graph, starts, |event| bfs_handler(py, &visitor, event))?;
```

src/traversal/mod.rs
```diff
@@ -454,22 +463,26 @@ pub fn graph_bfs_search(
...
     py: Python,
     graph: &digraph::PyDiGraph,
     source: Option<Vec<usize>>,
-    visitor: PyDfsVisitor,
+    visitor: Option<PyDfsVisitor>,
 ) -> PyResult<()> {
+    if visitor.is_none() {
+        return Err(PyTypeError::new_err("Missing required argument visitor"));
+    }
+    let v = visitor.unwrap();
```
IvanIsCoding, line 479:
```suggestion
    let visitor = visitor.unwrap();
```

src/traversal/mod.rs
```diff
@@ -454,22 +463,26 @@ pub fn graph_bfs_search(
...
     let starts: Vec<_> = match source {
         Some(nx) => nx.into_iter().map(NodeIndex::new).collect(),
         None => graph.graph.node_indices().collect(),
     };
 
-    depth_first_search(&graph.graph, starts, |event| {
-        dfs_handler(py, &visitor, event)
-    })?;
+    depth_first_search(&graph.graph, starts, |event| dfs_handler(py, &v, event))?;
```
IvanIsCoding, line 485:
```suggestion
    depth_first_search(&graph.graph, starts, |event| dfs_handler(py, &visitor, event))?;
```

src/traversal/mod.rs
```diff
@@ -284,22 +285,26 @@ pub fn descendants(graph: &digraph::PyDiGraph, node: usize) -> HashSet<usize> {
...
 pub fn digraph_bfs_search(
     py: Python,
     graph: &digraph::PyDiGraph,
     source: Option<Vec<usize>>,
-    visitor: PyBfsVisitor,
+    visitor: Option<PyBfsVisitor>,
 ) -> PyResult<()> {
+    if visitor.is_none() {
+        return Err(PyTypeError::new_err("Missing required argument visitor"));
+    }
```
IvanIsCoding, line 300: I think this is the only problematic part of the PR... Why can't `digraph_bfs_search` accept `PyBfsVisitor`? I think PyO3 should check that the argument is valid, not us

IvanIsCoding (commented):
Overall this is looking good, the only minor comments are on the traversal and layout modules

src/traversal/mod.rs
```diff
@@ -284,22 +285,26 @@ pub fn descendants(graph: &digraph::PyDiGraph, node: usize) -> HashSet<usize> {
...
+    let visitor = visitor.unwrap();
     let starts: Vec<_> = match source {
         Some(nx) => nx.into_iter().map(NodeIndex::new).collect(),
         None => graph.graph.node_indices().collect(),
     };
 
-    breadth_first_search(&graph.graph, starts, |event| {
-        bfs_handler(py, &visitor, event)
-    })?;
+    breadth_first_search(&graph.graph, starts, |event| bfs_handler(py, &visitor, event))?;
```
IvanIsCoding, line 307:
```suggestion
    breadth_first_search(&graph.graph, starts, |event| {
        bfs_handler(py, &visitor, event)
    })?;
```

src/traversal/mod.rs
```diff
@@ -370,22 +375,26 @@ pub fn digraph_bfs_search(
...
```
IvanIsCoding, line 397:
```suggestion
    breadth_first_search(&graph.graph, starts, |event| {
        bfs_handler(py, &visitor, event)
    })?;
```

src/traversal/mod.rs
```diff
@@ -454,22 +463,26 @@ pub fn graph_bfs_search(
...
+    let visitor = visitor.unwrap();
     let starts: Vec<_> = match source {
         Some(nx) => nx.into_iter().map(NodeIndex::new).collect(),
         None => graph.graph.node_indices().collect(),
     };
 
-    depth_first_search(&graph.graph, starts, |event| {
-        dfs_handler(py, &visitor, event)
-    })?;
+    depth_first_search(&graph.graph, starts, |event| dfs_handler(py, &visitor, event))?;
```
IvanIsCoding, line 485:
```suggestion
    depth_first_search(&graph.graph, starts, |event| {
        dfs_handler(py, &visitor, event)
    })?;
```

src/traversal/mod.rs
```diff
@@ -538,22 +551,26 @@ pub fn digraph_dfs_search(
...
```
IvanIsCoding, line 573:
```suggestion
    depth_first_search(&graph.graph, starts, |event| {
        dfs_handler(py, &visitor, event)
    })?;
```

src/traversal/mod.rs
mtreinish: Basically we had two options to fix this, either give visitor a default option or move `visitor` before `source`. I opted to give it a default here because moving `visitor` would be a breaking change because it would break existing callers of this function using positional arguments

IvanIsCoding: Well, I guess we’ll have to deal with it

IvanIsCoding (approved):
LGTM

PR #793, opened 2023-01-26
Set MSRV for clippy in rustworkx

mtreinish:
Starting in the recently released Rust 1.67.0 clippy enabled a new rule by default, uninlined_format_args [1], which warns when format!() is used and a variable is not put in the string inline but isntead as a separate argument in the macro. For example, `format!("{}", foo)` Instead the rule suggests using the variable inline the format call (e.g. `format!("{foo}")`. However, with our msrv of 1.56.1 this syntax was not supported. So trying to fix the places the rule is triggered as suggested will fail to compile with rust 1.56.1. To address this and solve this for future rust releases and potential new rules this commit adds a clippy.toml file which sets the msrv for the rustworkx crate This will by default disable rules that weren't on by default in rust 1.56.1.

[1] https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args


IvanIsCoding (approved):
Seems straightfoward to me.

IvanIsCoding (commented):
Also, for future reference: when we upgrade our MSRV, we should delete that file and just use `rust-version` in `Cargo.toml`: https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

Both the compiler and Clippy detect that field. I find too many config files a bit confusing.

Issue #719, opened 2022-10-28
Fix Deprecation Warnings in Github Actions


IvanIsCoding:
### What is the expected enhancement?

Right now we have [134 warnings on Github Actions telling us that the actions we are using depend on functionality that will be deprecated](https://github.com/Qiskit/rustworkx/actions/runs/3342047005).

Apparently they will be removed sometime in late 2023. It would be good to fix them earlier than when they stop completely.

IvanIsCoding:
Perhaps the most concerning part is that `actions-rs/toolchain` is affected by it and is currently unmaintained (https://github.com/actions-rs/toolchain/issues/221) so we need to find a replacement for it

PR #794, opened 2023-01-27
Update versions of GitHub Actions

obrienpja:
In this PR, I am updating GitHub Actions, setup-python -> v2 to v4, checkout -> v2 to v3. I think this should address many of the warnings mentioned in the issue #719.


mtreinish (approved):
LGTM, thanks for doing this. We'll still need to figure out what to do about the action to install rust, but this is great to fix the warnings for most of the actions.

PR #796, opened 2023-01-28
Add configuration model

tapanih:
Added a possibility to generate a random graph using a configuration model. The result is a random graph with a given degree sequence. The Rust library can additionally generate degree sequences from any `Distribution`.

### DIFF
--- a/rustworkx-core/src/generators/configuration_model.rs
+++ b/rustworkx-core/src/generators/configuration_model.rs
@@ -1,0 +1,344 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License. You may obtain
+// a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations
+// under the License.
+
+use crate::generators::InvalidInputError;
+use num_traits::{Num, ToPrimitive};
+use petgraph::data::{Build, Create};
+use petgraph::visit::{Data, GraphProp, NodeIndexable};
+use petgraph::Undirected;
+use rand::distributions::Distribution;
+use rand::prelude::SliceRandom;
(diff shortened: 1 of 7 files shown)


src/random_graph.rs
```diff
@@ -493,3 +494,67 @@ pub fn random_geometric_graph(
...
+    degree_sequence: Vec<usize>,
+    weights: Option<Vec<PyObject>>,
+    seed: Option<u64>,
+) -> PyResult<graph::PyGraph> {
+    let default_fn = || py.None();
+    let mut rng: Pcg64 = match seed {
+        Some(seed) => Pcg64::seed_from_u64(seed),
+        None => Pcg64::from_entropy(),
+    };
+    let degree_sequence = match core_generators::DegreeSequence::new(degree_sequence) {
```
IvanIsCoding, line 531: Do you need to check this degree sequence with [Erdős–Gallai](https://en.wikipedia.org/wiki/Erd%C5%91s%E2%80%93Gallai_theorem)? Or do you return a graph that might not be a simple graph?

IvanIsCoding (commented):
Thanks for contributing this! This is neat.

I will have to read the paper to give a better review of the code, but I do have a question about graph realization right off the bat

IvanIsCoding:
Also, you might want to update `api.rst` to include the new function you added

src/random_graph.rs
tapanih: The graph may not be simple. The generator in NetworkX works in a similar fashion: https://networkx.org/documentation/stable/reference/generated/networkx.generators.degree_seq.configuration_model.html#configuration-model

rustworkx-core/src/generators/configuration_model.rs
```diff
@@ -0,0 +1,344 @@
...
+            let value = match value.to_usize() {
+                Some(value) => value,
+                None => return Err(DegreeSequenceError::Conversion),
+            };
+            degree_sequence.push(value);
+            sum += value;
+        }
+        if sum % 2 == 1 {
+            let index = rng.gen_range(0..length);
+            degree_sequence[index] += 1;
```
tapanih, line 88: This felt like a "pragmatic" way to ensuring that the sum is even. However, this could lead to out-of-distribution values in the sequence which is probably confusing for the user. Not sure how to fix it. Probably just retry n times?

PR #797, opened 2023-01-31
Add Katz Centrality

IvanIsCoding:
Related to #441

Add the Katz centrality to both rustworkx and rustworkx-core. I mostly forked the code from `eigenvector_centrality` and made the adaptations


src/centrality.rs
```diff
@@ -557,3 +562,264 @@ pub fn digraph_eigenvector_centrality(
         ))),
     }
 }
+
+/// Compute the Katz centrality of a :class:`~PyGraph`.
+///
+/// For details on the eigenvector centrality refer to:
```
mtreinish, line 568:
```suggestion
/// For details on the Katz centrality refer to:
```

src/centrality.rs
```diff
@@ -557,3 +562,264 @@ pub fn digraph_eigenvector_centrality(
...
+        None => Err(FailedToConverge::new_err(format!(
+            "Function failed to converge on a solution in {} iterations",
+            max_iter
+        ))),
+    }
+}
+
+/// Compute the Katz centrality of a :class:`~PyDiGraph`.
+///
+/// For details on the eigenvector centrality refer to:
```
mtreinish, line 698:
```suggestion
/// For details on the Katz centrality refer to:
```

rustworkx-core/src/centrality.rs
```diff
@@ -672,6 +672,130 @@ where
...
+/// use rustworkx_core::Result;
+/// use rustworkx_core::petgraph;
+/// use rustworkx_core::petgraph::visit::{IntoEdges, IntoNodeIdentifiers};
+/// use rustworkx_core::centrality::katz_centrality;
+///
+/// let g = petgraph::graph::UnGraph::<i32, ()>::from_edges(&[
+///     (0, 1), (1, 2)
+/// ]);
+/// // Calculate the eigenvector centrality
+/// let output: Result<Option<Vec<f64>>> = katz_centrality(&g, |_| {Ok(1.)}, None, None, None, None, None);
```
mtreinish, line 720: I think it'd be good to have an assert on this so the doc test is asserting the output is as expected

rustworkx-core/src/centrality.rs
```diff
@@ -672,6 +672,130 @@ where
...
+            let node = graph.to_index(node_index);
+            beta.insert(node, beta_scalar);
+        }
+    }
+
+    // Check if beta contains all node indices
+    for node_index in graph.node_identifiers() {
+        let node = graph.to_index(node_index);
+        if !beta.contains_key(&node) {
+            return Ok(None); // beta_map was provided but did not include all nodes
```
mtreinish, line 754: Should this be an error instead of `None`?  It feels like a user error if they pass in a mapping that's incomplete here.

rustworkx-core/src/centrality.rs
```diff
@@ -672,6 +672,130 @@ where
...
+        }
+    }
```
mtreinish, line 756: I think putting this in an if/else block would be good here because we don't need to validate `beta` has all the node indices if it was `None`.
```suggestion
    if beta.is_empty() {
        // beta_map was none
        // populate hashmap with default value
        let beta_scalar = beta_scalar.unwrap_or(1.0);
        for node_index in graph.node_identifiers() {
            let node = graph.to_index(node_index);
            beta.insert(node, beta_scalar);
        }
    } else {
        // Check if beta contains all node indices
        for node_index in graph.node_identifiers() {
            let node = graph.to_index(node_index);
            if !beta.contains_key(&node) {
                return Ok(None); // beta_map was provided but did not include all nodes
            }
        }
    }
```

mtreinish (commented):
Overall this LGTM, it's a simple modification to the eigenvector centrality to compute the katz centrality. Just a couple comments inline.

rustworkx-core/src/centrality.rs
IvanIsCoding: None means it’s not defined… it map to an error in Python but in Rust we let the caller handle it

rustworkx-core/src/centrality.rs
IvanIsCoding: Done

rustworkx-core/src/centrality.rs
IvanIsCoding: Done

mtreinish (approved):
LGTM, thanks for making the updates

PR #799, opened 2023-02-01
Add edge betweenness centrality

tapanih:
Related to #441

Added edge betweenness centrality to rustworkx and rustworkx-core. It is mostly based on betweenness centrality with some modifications.


jlapeyre:
Thanks for the contribution!
I think it would be a good idea to cross-reference (vertex) betweeness centrality somehow. networkx has quite a bit more of this. But I don't want to overload this PR. Also, rustworkx is less complete and has less need at present. networkx implements something like [forty-two centralities](https://networkx.org/documentation/stable/reference/algorithms/centrality.html). With this PR, rustworkx will have three.

But I think including something like a "See also" like [this](https://github.com/networkx/networkx/blob/226de67121e8f7d0b44cdce6f053a43f70c034d6/networkx/algorithms/centrality/betweenness.py#L68-L71) would be useful.

releasenotes/notes/add-edge-betweenness-centrality-8de06bf716caece0.yaml
```diff
@@ -0,0 +1,41 @@
...
+      btw = rustworkx.edge_betweenness_centrality(graph)
+      # Color edges in graph visualization with edge betweenness centrality
+      colors = []
+      for i in graph.edge_indices():
+          colors.append(btw[i])
+      mpl_draw(graph, edge_color=colors)
+  - |
+    Added a new function to rustworkx-core ``edge_betweenness_centrality`` to
+    the ``rustworkx_core:centrality`` module which computes the edge betweenness
+    centrality of all edges in a given graph.
```
jlapeyre, line 41: * There are several lines above that consist only of whitespace. This should be removed (only a single newline for each such line.)
* There should be a newline at the end of the file.

jlapeyre, line 41: File needs a newline.

src/centrality.rs
```diff
@@ -326,3 +326,119 @@ pub fn digraph_eigenvector_centrality(
         ))),
     }
 }
+
+/// Compute the edge betweenness centrality of all edges in a :class:`~PyGraph`.
```
jlapeyre, line 330: The edge and vertex centrality functions should be next to each other in this file, as they are more closely related to each other than to the eigenvector centrality... unless there is another good reason for the current order.

jlapeyre (changes requested):

jlapeyre (approved):
Looks good.

mtreinish (approved):
This LGTM, thanks for pushing this! (and sorrt for the delay in final review)

Issue #811, opened 2023-02-10
test_graphviz / test_image_type fails


jlapeyre:
When running `tox -epy` exactly one test fails:
https://github.com/Qiskit/rustworkx/blob/a5178e28a3921fd6372c523f7997187b27180b04/tests/rustworkx_tests/visualization/test_graphviz.py#L120-L124
Failure occurs already on line 122.
Running a script with copies of lines 121 and 122 produces the same error

#### traceback
```
      File "/home/lapeyre/qiskit/rustworkx/tests/rustworkx_tests/visualization/test_graphviz.py", line 122, in test_image_type
    image = graphviz_draw(graph, image_type="jpg")
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

      File "/home/lapeyre/qiskit/rustworkx/.tox/py/lib64/python3.11/site-packages/rustworkx/visualization/graphviz.py", line 187, in graphviz_draw
    dot_result = subprocess.run(
                 ^^^^^^^^^^^^^^^

      File "/usr/lib64/python3.11/subprocess.py", line 569, in run
    raise CalledProcessError(retcode, process.args,

    subprocess.CalledProcessError: Command '['dot', '-T', 'jpg']' returned non-zero exit status 1.
```

#### versions
rustworkx  main Feb 8, 2023  e4170ef65aa6
fedora linux
`/usr/bin/dot -V`  => dot - graphviz version 5.0.0 (0)
Python 3.11.0

PR #813, opened 2023-02-10
Add commas between args to passenv in tox.ini

### LINKED ISSUE #812: arguments to `pass_env` in tox.ini are not comma separated
`tox -epy` fails with

```
py: failed with pass_env values cannot contain whitespace, use comma to have multiple values in a single line, invalid values found 'RETWORKX_TEST_PRESERVE_IMAGES RUSTWORKX_PKG_NAME'
  py: FAIL code 1 (0.00 seconds)
```

Requiring commas  was introduced in tox 4.0.6 :  https://github.com/tox-dev/tox/pull/2671

See also https://github.com/Qiskit/qiskit-terra/pull/9276

jlapeyre:
The presence of commas is enforced in tox 4.0.6

Closes #812

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -19,7 +19,9 @@ deps =
 extras =
   mpl
   graphviz
-passenv = RETWORKX_TEST_PRESERVE_IMAGES, RUSTWORKX_PKG_NAME
+passenv =
+  RETWORKX_TEST_PRESERVE_IMAGES
+  RUSTWORKX_PKG_NAME
 changedir = {toxinidir}/tests
 commands =
   stestr run {posargs}
--- a/tox.ini
+++ b/tox.ini
@@ -46,7 +46,10 @@ setenv =
   {[testenv]setenv}
 deps =
   -r {toxinidir}/docs/source/requirements.txt
-passenv = {[testenv]passenv}, RETWORKX_DEV_DOCS, RETWORKX_LEGACY_DOCS
+passenv =
+  {[testenv]passenv}
+  RETWORKX_DEV_DOCS
+  RETWORKX_LEGACY_DOCS
 changedir = {toxinidir}/docs
 commands =
   sphinx-build -W -b html source build/html {posargs}
--- a/tox.ini
+++ b/tox.ini
@@ -19,7 +19,7 @@ deps =
 extras =
   mpl
   graphviz
-passenv = RETWORKX_TEST_PRESERVE_IMAGES RUSTWORKX_PKG_NAME
+passenv = RETWORKX_TEST_PRESERVE_IMAGES, RUSTWORKX_PKG_NAME
 changedir = {toxinidir}/tests
 commands =
   stestr run {posargs}
@@ -44,7 +44,7 @@ setenv =
   {[testenv]setenv}
 deps =
   -r {toxinidir}/docs/source/requirements.txt
-passenv = {[testenv]passenv} RETWORKX_DEV_DOCS RETWORKX_LEGACY_DOCS
+passenv = {[testenv]passenv}, RETWORKX_DEV_DOCS, RETWORKX_LEGACY_DOCS
 changedir = {toxinidir}/docs
 commands =
   sphinx-build -W -b html source build/html {posargs}
--- a/.github/workflows/docs_dev.yml
+++ b/.github/workflows/docs_dev.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -U virtualenv setuptools wheel 'tox<4'
+        pip install -U virtualenv setuptools wheel tox
         sudo apt-get install graphviz pandoc
     - name: Build and publish
       env:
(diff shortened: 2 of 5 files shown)


IvanIsCoding (approved):
LGTM

IvanIsCoding:
One thing I would like to ping @mtreinish before merging this is our policy for using Tox 4: https://github.com/Qiskit/rustworkx/blob/main/.github/workflows/main.yml#L87

Right now we use Tox 3. Clearly this change passes the CI on Tox 3. But I wonder if we should profit from this to upgrade to Tox 4

jlapeyre:
Is `tox < 4` specified elsewhere? It would be useful for developing if it were mentioned in say `CONTRIBUTING.md`. Or some Python requirements file, rather than just `main.yaml`.

IvanIsCoding:
Can you revert https://github.com/Qiskit/rustworkx/pull/761 and bump Tox to 4.0 in CI? Then we can check that it works with the latest version which most people will start installing

jlapeyre:
Yes, I can do that. I got the latest, `4.0.6`, and the only problem locally was that this comma thing. Don't know about CI.

jlapeyre:
CI failed. I wonder what `ERROR: Operation cancelled by user` means.

IvanIsCoding:
> Yes, I can do that. I got the latest, `4.0.6`, and the only problem locally was that this comma thing. Don't know about CI.

The Ubuntu retworkx build is failing, I am going to investigate why it worked with Tox 3 but not 4

jlapeyre:
> The Ubuntu retworkx build is failing, I am going to investigate why it worked with Tox 3 but not 4

I 'm digging through the logs

jlapeyre:
Looks like testing fails. But there is a lack of detail.  Can't see the python stack trace.
I do see ` subprocess-exited-with-error`, which makes me think it might be #811

mtreinish:
We'll likely need deeper changes to the tox configuration to enable it to work with tox >4.0.0. Tox 4 made a number of breaking changes, including critically how the build process works. I capped it to tox<4.0.0 at the time to give tox a chance to stabilize after their big breaking release before circling back to update the configuration. So far I don't think they've really slowed down making releases looking at https://pypi.org/project/tox/#history but I agree we should remove the cap since people are getting tripped up on this.

The relevant section in the upgrade guide around where I expect the issues are: https://tox.wiki/en/latest/upgrading.html#packaging-environments

mtreinish:
Looking at the failure in particular it seems to be with the `retworkx` backwards compat side: https://github.com/Qiskit/rustworkx/blob/main/.github/workflows/main.yml#L93-L131 it looks like tox isn't happy trying to build the retworkx package like it was in tox 3 just by setting the env variable

jlapeyre:
> we should remove the cap since people are getting tripped up on this.

Yes, but if I'm not mistaken we don't mention anywhere that you need tox 3. It's only in the CI config files. And the instructions we give are explicitly incorrect `pip install -U tox`.
Furthermore, isn't it possible to put the version restriction in `requirements.txt` or `setup.py`, etc. ?

mtreinish:
> > we should remove the cap since people are getting tripped up on this.
> 
> Yes, but if I'm not mistaken we don't mention anywhere that you need tox 3. It's only in the CI config files. And the instructions we give are explicitly incorrect `pip install -U tox`. Furthermore, isn't it possible to put the version restriction in `requirements.txt` or `setup.py`, etc. ?

It was an oversight, the tox pin wasn't meant to be there long term, but it was intended to be a short term workaround to unblock CI. But it's ended up persisting longer than I expected and the level of compatibility between tox 3 and tox 4 has decreased even further since the earlier releases in the tox 4.x series.

There isn't a good way to specify a tox version cap automatically since tox triggers the package build and installs the requirements and test requirements for us. If we put `tox<4` in the `requirements-dev.txt` list that would end up having tox install tox in the test venvs it creates. Similarly, putting it in one of the other packaging files like `setup.py` or `pyproject.toml` would install tox too late because it would either get installed during build or install time for the rustworkx package (which is both not desired and also too late since tox is what installs rustworkx in a venv for local testing)

tox.ini
```diff
@@ -19,7 +19,7 @@ deps =
 extras =
   mpl
   graphviz
-passenv = RETWORKX_TEST_PRESERVE_IMAGES RUSTWORKX_PKG_NAME
+passenv = RETWORKX_TEST_PRESERVE_IMAGES, RUSTWORKX_PKG_NAME
```
mtreinish, line 22: Instead of the comma syntax I tend to prefer the newline approach since it's more broadly compatible with more versions of tox:

```suggestion
passenv =
  RETWORKX_TEST_PRESERVE_IMAGES
  RUSTWORKX_PKG_NAME
```

tox.ini
```diff
@@ -44,7 +44,7 @@ setenv =
   {[testenv]setenv}
 deps =
   -r {toxinidir}/docs/source/requirements.txt
-passenv = {[testenv]passenv} RETWORKX_DEV_DOCS RETWORKX_LEGACY_DOCS
+passenv = {[testenv]passenv}, RETWORKX_DEV_DOCS, RETWORKX_LEGACY_DOCS
```
mtreinish, line 47:
```suggestion
passenv =
  {[testenv]passenv}
  RETWORKX_DEV_DOCS
  RETWORKX_LEGACY_DOCS
```

IvanIsCoding:
Closing after #851 got merged

PR #814, opened 2023-02-10
Add PyDiGraph method to make edges symmetric

mtreinish:
This commit adds a new method make_symmetric() to PyDiGraph which will modify the graph and add a reverse edge to each edge in the graph if it is not already present.

TODO:

- [x] Add tests
- [x] Add release notes


src/digraph.rs
```diff
@@ -2606,6 +2606,43 @@ impl PyDiGraph {
         edges.is_empty()
     }
 
+    /// Make edges in graph symmetric
+    ///
+    /// This function iterates over all the edges in the graph and for each edge if a reverse
+    /// edge is not present in the graph one will be added.
```
jlapeyre, line 2612: Following is perhaps a bit more concise:
"This function iterates over all edges in the graph, adding for each edge the reversed edge, unless it is already present."

mtreinish: Fixed in: https://github.com/Qiskit/rustworkx/pull/814/commits/acae85f05623956f2cc11b928337205a8d641aba

src/digraph.rs
```diff
@@ -2711,6 +2711,44 @@ impl PyDiGraph {
...
+            .collect();
+        for (edge_endpoints, edge_index) in edges.iter() {
+            let reverse_edge = [edge_endpoints[1], edge_endpoints[0]];
+            if !edges.contains_key(&reverse_edge) {
+                let forward_weight = self.graph.edge_weight(*edge_index).unwrap();
+                let weight: PyObject = match edge_payload_fn.as_ref() {
+                    Some(callback) => callback.call1(py, (forward_weight,))?,
+                    None => forward_weight.clone_ref(py),
+                };
+                self._add_edge(reverse_edge[0], reverse_edge[1], weight)?;
```
jlapeyre, line 2746: I tried to find something a bit simpler. The following is what I came up with. It saves a line. But I'm not convinced it's an improvement.
```suggestion
       for ([edge_source, edge_target], edge_index) in edges.iter() {
            if !edges.contains_key(&[*edge_target, *edge_source]) {
                let forward_weight = self.graph.edge_weight(*edge_index).unwrap();
                let weight: PyObject = match edge_payload_fn.as_ref() {
                    Some(callback) => callback.call1(py, (forward_weight,))?,
                    None => forward_weight.clone_ref(py),
                };
                self._add_edge(*edge_target, *edge_source, weight)?;
```

jlapeyre:
Other than the comment above, this looks really thorough. LGTM.

src/digraph.rs
mtreinish: Let's go with it, I think it's about the same one way or the other and I expect the compiler will generate about the same code in both cases. From a legibility PoV they're equivalent to me so lets save a line.

jlapeyre: I agree on all points.

jlapeyre (approved):
Good to go.

enavarro51 (approved):

PR #815, opened 2023-02-10
Simplifying betweenness_centrality (for vertices)

jlapeyre:
#799 "Add edge betweenness centrality" spurred me to try to reduce code repetition and complexity in betweenness centrality (for vertices). I recall trying `rayon-cond` in the past. In any case, it seems to work here now.


rustworkx-core/src/centrality.rs
```diff
@@ -75,7 +74,7 @@ use rayon_cond::CondIterator;
 /// [`edge_betweenness_centrality`]
 pub fn betweenness_centrality<G>(
     graph: G,
-    endpoints: bool,
+    include_endpoints: bool,
```
mtreinish, line 77: I think this ok from a backwards compatibility perspective since in rust the arguments are all passed positionally (this would be breaking in Python so I wanted to just note this and check myself).

mtreinish (approved):
LGTM, thanks for reviving your old PR. I always felt bad that it fell through the cracks so it's good to get this simplification merged now.

rustworkx-core/src/centrality.rs
jlapeyre: Oh yeah, that's a good point.

PR #818, opened 2023-02-18
Add random generators to rustworkx-core

### LINKED ISSUE #816: Move random generators to rustworkx-core

There are 3 random generators in `rustworkx` - `gnp_random_graph`, `gnm_random_graph` and `random_geometric_graph`. It would be helpful to have these available in `rustworkx-core`, especially for testing purposes.

enavarro51:
Fixes #816. This pull request adds `gnp_random_graph`, `gnm_random_graph`, and `random_geometric_graph` to rustworkx-core. The first 2 are called from the `rustworkx PyGraph` interface. `random_geometric_graph` was not modified on the `rustworkx PyGraph` side due to some issues with the node weight implementation. This may be fixable in the future.


tests/rustworkx_tests/test_random.py
```diff
@@ -143,10 +143,10 @@ def test_random_gnm_undirected_complete_graph(self):
 
     def test_random_gnm_undirected_invalid_num_nodes(self):
         with self.assertRaises(ValueError):
-            rustworkx.undirected_gnm_random_graph(-23, 5)
+            rustworkx.undirected_gnm_random_graph(0, 5)
 
     def test_random_gnm_undirected_invalid_probability(self):
-        with self.assertRaises(ValueError):
+        with self.assertRaises(OverflowError):
```
mtreinish, line 149: PyO3 returns an overflow error for passing a negative to an unsigned type input I would have done a type error, but I guess this makes sense too. But anyway I think we need a release note documenting this as a potentially breaking change. I don't think it's a severe API change and something we should do to simplify the code, but we should just document it as a change.

mtreinish (commented):
This LGTM, it's straightforward enough of a change. But, just one inline note I think we need to document the subtle api change on the exception types for the functions that were previously taking in an isize with negative values. I think it's super minor in practice but something we should probably document

tests/rustworkx_tests/test_random.py
enavarro51: I can change it to a TypeError. What category in the release note should this go under? Upgrade? And you want to document all the input changes from isize to usize, like num_nodes and num_edges as well?

mtreinish: Nah this is fine, upgrade note is the right category for this because it's something an end user might need to change when upgrading. We don't need to document the input change, it should be mostly invisible, it technically would allow them to pass in bigger bigger integers but that would fail anyway because of the max index size is a 32bit int and will error if you try to use >~4 billion nodes or edges.

enavarro51: Actually I just looked at this and this test shouldn't be here. It's testing probability for the `gnm` graph which doesn't take probability as an input. I think we can just remove this test. I can still do the reno for probability on the `gnp` graph.

releasenotes/notes/add-random-generators-9f99e57b5e4188f2.yaml
```diff
@@ -0,0 +1,16 @@
...
+    the number of nodes and a probability for adding edges. The ``gnp_random_graph``
+    takes inputs of the number of nodes and number of edges. The
+    ``random_geometric_graph`` creates a random graph within an n-dimensional
+    cube.
+upgrade:
+  - |
+    Passing a negative value to the ``probability`` argument to the
+    :func:`~rustworkx.gnp_directed_random_graph` or the
+    :func:`~rustworkx.gnp_undirected_random_graph` function will now cause
+    an OverflowError to be raised.
```
mtreinish, line 16:
```suggestion
    an ``OverflowError`` to be raised. Previously, a ``ValueError`` would be raised in this situation.
    This was changed to be consistent with other similar error conditions in other functions in
    the library.
```

enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/818/commits/363fb65950a7685ae302929fb5141eba7b22a125. Also fixed the `token_swapper` test failure that was using `gnm_random_graph`.

enavarro51:
This should be ready to go. I also ported the large random `token_swapper` test to core `test_token_swapper `which uses the core `gnm_random_graph`. I can push that once this is merged.

mtreinish (approved):
LGTM, thanks for the updates

PR #819, opened 2023-02-21
add bfs_predecessors

ewinston:
This adds a bfs_predecessors method to complement bfs_successors.


src/traversal/mod.rs
```diff
@@ -166,6 +166,50 @@ pub fn bfs_successors(
...
+    let mut bfs = Bfs::new(reverse_graph, index);
+    let mut out_list: Vec<(PyObject, Vec<PyObject>)> = Vec::with_capacity(graph.node_count());
+    while let Some(nx) = bfs.next(reverse_graph) {
+        let children = graph
+            .graph
+            .neighbors_directed(nx, petgraph::Direction::Incoming);
+        let mut predecessors: Vec<PyObject> = Vec::new();
+        for pred in children {
+            predecessors.push(graph.graph.node_weight(pred).unwrap().clone_ref(py));
+        }
```
mtreinish, line 200: I realize this was just copied and pasted from the `bfs_successors` function, but we can improve the efficiency here by using the iterator to build the vector all at once. 

```suggestion
        let predecessors: Vec<PyObject> = graph
            .graph
            .neighbors_directed(nx, petgraph::Direction::Incoming)
            .map(|pred| graph.graph.node_weight(pred).unwrap().clone_ref(py))
            .collect();
```

Tthe `bfs_successors` function was one of the first functions in retworkx so I wasn't super experienced with things when I wrote it. Could you also push a follow up to make this improvement to `bfs_successors` too.

mtreinish (commented):
Overall this LGTM, it's pretty straightforward just a copy of the `bfs_successors` code with the direction reversed. I left one inline comment about an improvement we can make (which is also something we should do to `bfs_successors`).

Besides that the only thing that I think is missing is a release note for the new function.

src/traversal/mod.rs
ewinston: Yes, I can update it there too. Thanks!

releasenotes/notes/add_bfs_predecessors-751b468d1b3c01de.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    This PR adds a method to return an iterator of predecessors in a
+    bread-first traversal from a specified node. This is analogous to
+    the existing bfs_successors method.
```
mtreinish, line 6:
```suggestion
    Added a new function, :func:`~.bfs_predecessors`, which is used
    to return a list of predecessors in a reversed bread-first traversal
    from a specified node. This is analogous to the existing
    :func:`~.bfs_successors` method.
```

mtreinish (commented):
This LGTM now, thanks for the updates. I just left an inline suggestion on the release note text. Also I just noticed `bfs_predecessors` is not in the api documentation, it just needs to be added in: https://github.com/Qiskit/rustworkx/blob/main/docs/source/api.rst to the traversal section.

mtreinish (approved):

PR #821, opened 2023-02-21
Fix all deprecation warnings in Rust and Upload actions

### LINKED ISSUE #719: Fix Deprecation Warnings in Github Actions

IvanIsCoding:
Closes #719 

We switch the Github Actions to install Rust as `actions-rs/toolchain` is unmaintained (https://github.com/actions-rs/toolchain/issues/221)

We also bump `upload-artifact` to `v3`, fixing all the remaining warnings


IvanIsCoding:
This is failing until #820 is merged (as any PR, really) and should be good to go

.github/workflows/main.yml
```diff
@@ -77,12 +75,10 @@ jobs:
           python-version: ${{ matrix.python-version }}
           architecture: ${{ matrix.platform.python-architecture }}
       - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@master
```
mtreinish, line 78: Why is this `master` while other jobs use `stable` below?

IvanIsCoding: So, this Github action is a little bit funny because you can specify the version in the `@`. `@stable` installs Rust stable, `@1.57.1` installs 1.57.1 etc. But if you want to have the version as an input (like from the matrix we have), the author suggests using the `toolchain` field: https://github.com/dtolnay/rust-toolchain#inputs

mtreinish (approved):

IvanIsCoding:
#401 added a new step using the old Github action so I need to update that

PR #838, opened 2023-03-18
Implement `check_cycle` in `__getstate__` and `__setstate__`. Expand python copy tests.

### LINKED ISSUE #836: Python copy does not replicate PyDiGraph.copy()
### Information

- **Python version**: 3.10.6
- **Rust version**: 1.64.0
- **Operating system**: Ubuntu 22.04.2 LTS

`check_cycles` is not copied when a `PyDiGraph` is copied using the python native `copy.copy`.

`copy.copy(PyDiGraph)` should produce identical results to `PyDiGraph.copy()`.

```python
import rustworkx as rx
import copy


a = rx.PyDiGraph(check_cycle=True)
print(a.check_cycle)

b = a.copy()
print(b.check_cycle)

c = copy.copy(a)
print(c.check_cycle)

d = copy.deepcopy(a)
print(d.check_cycle)
```
```bash
True
True
False
False
```

Edit: My original test used an object that `copy.deepcopy()` does not copy into a new object. Upon further testing, it behaves as expected.

lukepmccombs:
Adds `check_cycle` to `__getstate__` and `__setstate__` so that python copy and deepcopy should be exact duplicates of the original objects.
Expands the digraph copy and deepcopy tests to include `check_cycle` tests and python object instance tests.

Should resolve #836.

### DIFF
--- a/releasenotes/notes/fix-check-cycle-on-copy-d060a1781976f728.yaml
+++ b/releasenotes/notes/fix-check-cycle-on-copy-d060a1781976f728.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixed the :attr:`~.PyDiGraph.check_cycle` attribute not being preserved
+    when copying :class:`~.PyDiGraph` with `copy.copy()` and `copy.deepcopy()`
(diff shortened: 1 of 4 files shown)


IvanIsCoding (approved):
LGTM, we just need to add the release notes for the bugfix and it should be good to go

PR #842, opened 2023-03-20
Remove stray debug prints from tests

mtreinish:
This commit removes some stray debug print messages that were in some test files. This was causing spurious output that wasn't actionable during test runs both locally and in CI.


IvanIsCoding (approved):
LGTM, I agree that those were left by accident

PR #843, opened 2023-03-21
deploy documentation in qiskit.org/ecosystem

1ucian0:
Similar to https://github.com/Qiskit/qiskit-aer/pull/1748, see https://github.com/Qiskit/qiskit.org/issues/3038 for more context

### Summary

The Qiskit Ecosystem lives in qiskit.org/ecosystem and new place for qiskit related projects that are not technically Qiskit. This PR adds that destination for documentation deployment.

### Details and comments

For now, a copy of the same docs in ecosystem would allow to have both q.o/documentation and q.o/ecosystem for testing that everything is fine. Once that's done, the CDN will redirects to the new location and the deploy to q.o/documentation can be removed.

### DIFF
--- a/tools/deploy_documentation.sh
+++ b/tools/deploy_documentation.sh
@@ -2,7 +2,7 @@
 
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2018, 2019.
+# (C) Copyright IBM 2018, 2023.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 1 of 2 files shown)


mtreinish:
>For now, a copy of the same docs in ecosystem would allow to have both q.o/documentation and q.o/ecosystem for testing that everything is fine. Once that's done, the metapackage can create the redirects and the deploy to q.o/documentation can be removed.

I'm not sure I follow this step, is the plan you're doing to centralize the building of the redirects and remove the excludes here?: https://github.com/Qiskit/qiskit/blob/master/tools/other-builds.txt#L10 I assumed the plan was that we'd upload a redirect pages to `q.o/documentation/rustworkx` here and have them all point to `q.o/ecosystem/rustworkx`. Especially since the builds locally in this repository are able to build the redirect pages easily since they have the full source list available.

1ucian0:
Yes. My plan was to redirect `q.o/documentation/rustworkx/<something> -> q.o/ecosystem/rustworkx/<something>` from the metapackage. If I understood you correctly, it might be possible to be done at a different point (was it at CDN-level?).

1ucian0:
updating initial comment to talk about CDN instead of metapackage. Thanks!

mtreinish (approved):
This LGTM. I'm thinking we should also remember to add the CDN edge function redirect for retworkx -> rustworkx too while we're updating it.

PR #850, opened 2023-03-29
Fixup docs publish job

mtreinish:
This commit fixes two separate issues in the docs publishing job. The first is a failure that happens in the docs deployment on releases. The IFS (internal field separator) variable was getting set too broadly at the the level of the entire script instead of just the tag parsing. This was causing the passing of the secrets for decrypting the credential files to fail. This commit updates the IFS usage so it's not affecting the other commands after we split the tag version.

The second change is updating the default sphinx-build invocation to include the -d flag to change the default output directory for the .doctrees dir.
(see: https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-d for details on the flag). The .doctrees dir is a local cache of pickle files for parsed source files and doesn't need to be included in the build. By moving the default location for this outside our html build dir this will remove a lot of overhead from our uploaded docs as this cache can end up being quite large.

### DIFF
--- a/tools/deploy_documentation.sh
+++ b/tools/deploy_documentation.sh
@@ -27,9 +27,9 @@ echo "show current dir: "
 pwd
 
 CURRENT_TAG=`git describe --abbrev=0`
-IFS='.'
-read -ra VERSION <<< "$CURRENT_TAG"
+IFS=. read -ra VERSION <<< "$CURRENT_TAG"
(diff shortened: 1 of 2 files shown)


IvanIsCoding (approved):
LGTM, after #843 we are uploading the docs to 3 different addresses so tbat doc parsing cache can really add up

PR #851, opened 2023-03-30
Update tox configuration to use tox >= 4.4.0

### LINKED ISSUE #812: arguments to `pass_env` in tox.ini are not comma separated

mtreinish:
Tox 4.0.0 was released in December 2022 [1] and was a major rewrite of the internals of the package that included numerous backwards incompatible changes [2]. Along with that major rewrite was a long period of instability in the package with a flurry of 47 releases [3] since 4.0.0 (which has only been 3-4 months). At the time of the 4.0.0 release we pinned the tox version in CI with #761 to avoid this instability as our tox configuration was not compatible with tox 4.x.y and tox was actually not compatible with how we had things configured more generally. The hope was that tox would stabilize more, fix the issues that plagued the tox 4 release series and we'd be able to relax that pin without requiring bumping our minimum tox version to ensure users could use either the old version or the new version locally. However, since #761 that hope hasn't been realized the divergence between tox 3 and tox 4 has only widened and at least personally I'm not convinced of an improvement in stability to the tox 4 release series. That being said however, this is becoming a developer pain as by default when setting up a new build environment pip will install the latest version of tox and we don't have an effective mechanism to pin the tox version for users as you need to install tox manually as it's the primary python development entrypoint we use. The only only avenue to address this would be documentation updates in the CONTRIBUTING.md file, which we didn't update at the time in #761 because it was meant to be a version temporary pin that has turned out to not be so temporary.

Since it's been >3 months since we first pinned the tox version and that pin was meant to be temporary this commit removes that pin and bumps our minimum supported tox version to be 4.4.0, which despite not being compatible with tox < 4 as we originally hoped, at least seems to work fine with install rustworkx after updating the configuration file. This should hopefully ease the onboarding experience for developers when working with rustworkx and trying to bootstrap a local development environment. Longer term I expect we'll look at moving off of tox, as it no longer seems like a project we can rely on (especially as a key component for our development and CI infrastructure) for rustworkx and instead look at something like `nox` which I've heard good things about and know that PyO3 had moved to it a year or two ago.

Fixes #812

[1] https://pypi.org/project/tox/4.0.0/
[2] https://tox.wiki/en/latest/upgrading.html
[3] https://pypi.org/project/tox/#history

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,10 +1,10 @@
 [tox]
-minversion = 2.1
+minversion = 4.4.0
 envlist = py37, py38, py39, py310, py311, lint
 isolated_build = true
 
 [testenv]
-install_command = pip install -c{toxinidir}/constraints.txt -U {opts} {packages}
+install_command = python -I -m pip install -c{toxinidir}/constraints.txt -U {opts} {packages}
 setenv =
   VIRTUAL_ENV={envdir}
   LANGUAGE=en_US
@@ -19,14 +19,15 @@ deps =
 extras =
   mpl
   graphviz
-passenv = RETWORKX_TEST_PRESERVE_IMAGES RUSTWORKX_PKG_NAME
+passenv =
+  RETWORKX_TEST_PRESERVE_IMAGES
+  RUSTWORKX_PKG_NAME
 changedir = {toxinidir}/tests
 commands =
   stestr run {posargs}
 
 [testenv:lint]
 basepython = python3
-envdir = .tox/lint
 deps =
   black~=22.0
   flake8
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -19,14 +19,15 @@ deps =
...
+passenv =
+  RETWORKX_TEST_PRESERVE_IMAGES
+  RUSTWORKX_PKG_NAME
 changedir = {toxinidir}/tests
 commands =
   stestr run {posargs}
 
 [testenv:lint]
 basepython = python3
-envdir = .tox/lint
```
IvanIsCoding, line 29: I think this is the only difference from #813, really

IvanIsCoding (approved):
LGTM, this is almost identical to #813 but it passes the CI so let's stick to it

PR #853, opened 2023-04-02
Add reverse inplace function for digraph

matanco64:
I added a reverse function for a graph, it takes all the edges and reverses them,

also added tests and documentation for the function


src/digraph.rs
```diff
@@ -2720,6 +2720,56 @@ impl PyDiGraph {
...
+    ///
+    /// let mut graph = DiGraph::<(), i32>::new();
+    /// graph.add_edge(0, 1, 3).unwrap();
+    /// graph.add_edge(1, 2, 5).unwrap();
+    /// graph.add_edge(2, 3, 2).unwrap();
+    ///
+    /// graph.reverse_inplace(py);
+    ///
+    /// assert_eq!(graph.edges().collect::<Vec<_>>(), vec![(3, 2), (2, 1), (1, 0)]);
+    /// ```
```
mtreinish, line 2754: One of the weird quirks of rustworkx is that docstrings for these functions are actually built using sphinx and not rustdoc. So while this is a valid rust docstring to document this function it's actually going to have issues when sphinx tries to parse it. Mostly because sphinx uses restructured text instead of markdown but also some of the directives are correct. I'd rewrite this as:

```suggestion
    ///  .. note:: The edge indices will change if this method is used to reverse the edges
    ///
    /// For example::
    ///
    ///     import rustworkx as rx
    ///
    ///     graph = rx.PyDiGraph()
    ///     graph.add_nodes_from(range(4))
    ///     graph.add_edge(0, 1, 3)
    ///     graph.add_edge(1, 2, 5)
    ///     graph.add_edge(2, 3, 2)
    ///
    ///     graph.reverse_inplace()
    ///
    ///     assert graph.edge_list() == [(3, 2), (2, 1), (1, 0)]
```

I removed the panic detail because I don't think the function can ever panic in practice.

tests/rustworkx_tests/digraph/test_edges.py
```diff
@@ -962,3 +962,29 @@ def test_extend_from_weighted_edge_list(self):
...
+            (1, 2, "b"),
+            (0, 2, "c"),
+            (2, 3, "d"),
+            (0, 3, "e"),
+        ]
+        graph.add_edges_from(edge_list)
+        graph.reverse_inplace()
+        self.assertEqual([(1, 0), (2, 1), (2, 0), (3, 2), (3, 0)], graph.edge_list())
+
+    def test_reverse_large_graph(self):
```
mtreinish, line 980: Two small test cases that I think would be good to add is one with an empty graph. The other is for a graph where there is an edge removed in the middle (which will leave a hole in the edge indices). I don't expect there are any problems with the method's behavior with these, but they're common edge cases that are good to verify (because a lot of other methods have had issues in the past with both).

mtreinish (commented):
Overall this looks great. I think your approach here makes the most sense to do an inplace edge reversal. I left a couple of small inline comments, the thing blocking CI I expect is my comment about the docs. The other thing which is missing is a release note to document the addition of the new feature. If you don't mind could you add one, the procedure for doing this is documented here:  https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md#release-notes

src/digraph.rs
```diff
@@ -2720,6 +2720,45 @@ impl PyDiGraph {
...
+    ///     graph.add_nodes_from(range(4))
+    ///     graph.add_edge(0, 1, 3)
+    ///     graph.add_edge(1, 2, 5)
+    ///     graph.add_edge(2, 3, 2)
+    ///
+    ///     graph.reverse_inplace()
+    ///
+    ///     assert graph.edge_list() == [(3, 2), (2, 1), (1, 0)]
+    #[pyo3(text_signature = "(self)")]
+    pub fn reverse_inplace(&mut self, py: Python) {
```
IvanIsCoding, line 2745: Because we do not have a reverse method, you can call it reverse

IvanIsCoding, line 2745:
```suggestion
    pub fn reverse(&mut self, py: Python) {
```

tests/rustworkx_tests/digraph/test_edges.py
```diff
@@ -962,3 +962,34 @@ def test_extend_from_weighted_edge_list(self):
...
+        graph.add_nodes_from([i for i in range(4)])
+        edge_list = [
+            (0, 1, "a"),
+            (1, 2, "b"),
+            (0, 2, "c"),
+            (2, 3, "d"),
+            (0, 3, "e"),
+        ]
+        graph.add_edges_from(edge_list)
+        graph.reverse_inplace()
```
IvanIsCoding, line 977:
```suggestion
        graph.reverse()
```

src/digraph.rs
```diff
@@ -2720,6 +2720,45 @@ impl PyDiGraph {
...
+    ///
+    ///     import rustworkx as rx
+    ///
+    ///     graph = rx.PyDiGraph()
+    ///     graph.add_nodes_from(range(4))
+    ///     graph.add_edge(0, 1, 3)
+    ///     graph.add_edge(1, 2, 5)
+    ///     graph.add_edge(2, 3, 2)
+    ///
+    ///     graph.reverse_inplace()
```
IvanIsCoding, line 2741:
```suggestion
    ///     graph.reverse()
```

releasenotes/notes/added-reverse-inplace-for-digraph-c59759a1d54edb1a.yaml
```diff
@@ -0,0 +1,17 @@
...
+    For example:
+    import rustworkx as rx
+    
+    graph = rx.PyDiGraph()
+    graph.add_nodes_from(range(4))
+    graph.add_edge(0, 1, 3)
+    graph.add_edge(1, 2, 5)
+    graph.add_edge(2, 3, 2)
+    
+    graph.reverse_inplace()
```
IvanIsCoding, line 15:
```suggestion
    graph.reverse()
```

tests/rustworkx_tests/digraph/test_edges.py
```diff
@@ -962,3 +962,34 @@ def test_extend_from_weighted_edge_list(self):
...
+        edge_list = list(zip(range(LARGE_AMOUNT_OF_NODES), range(1, LARGE_AMOUNT_OF_NODES)))
+        weighted_edge_list = [(s, d, "a") for s, d in edge_list]
+        graph.add_edges_from(weighted_edge_list)
+        graph.reverse_inplace()
+        reversed_edge_list = [(d, s) for s, d in edge_list]
+        self.assertEqual(reversed_edge_list, graph.edge_list())
+
+    def test_empty_graph_reversal(self):
+        graph = rustworkx.PyDiGraph()
+        graph.reverse_inplace()
```
IvanIsCoding, line 994:
```suggestion
        graph.reverse()
```

src/digraph.rs
```diff
@@ -2720,6 +2720,45 @@ impl PyDiGraph {
...
+        let indices = self.graph.edge_indices().collect::<Vec<EdgeIndex>>();
+        for idx in indices {
+            let (source_node, dest_node) = self
+                .graph
+                .edge_endpoints(idx)
+                .expect("Edge Index received from iterator");
+            let weight = self
+                .graph
+                .edge_weight(idx)
+                .expect("Edge Index received from iterator")
```
IvanIsCoding, line 2755:
```suggestion
                .unwrap()
```

If this fails is because of PyO3.  It panics and there is not much point in printing a message

src/digraph.rs
```diff
@@ -2720,6 +2720,45 @@ impl PyDiGraph {
...
+    ///
+    ///     assert graph.edge_list() == [(3, 2), (2, 1), (1, 0)]
+    #[pyo3(text_signature = "(self)")]
+    pub fn reverse_inplace(&mut self, py: Python) {
+        let indices = self.graph.edge_indices().collect::<Vec<EdgeIndex>>();
+        for idx in indices {
+            let (source_node, dest_node) = self
+                .graph
+                .edge_endpoints(idx)
+                .expect("Edge Index received from iterator");
```
IvanIsCoding, line 2751:
```suggestion
                .unwrap();
```
If this fails is because of PyO3. It panics and there is not much point in printing a message

IvanIsCoding (commented):
Thanks for contributing this. Overall LGTM, I just have one suggestion. Also, I second Matthew's comment that we should add some extra tests

My main suggestion is naming the method `reverse` because [NetworkX does it that way](https://networkx.org/documentation/stable/reference/classes/generated/networkx.DiGraph.reverse.html). It is not a rule in the project but it does avoid people asking what the equivalent method is. For now, we can ignore the `copy` keyword and just always do it in place.

tests/rustworkx_tests/digraph/test_edges.py
matanco64: Hey Sorry for the late response and update, I've added the tests

src/digraph.rs
matanco64: Good Idea, I thought that because it behaves differently to not name it the same way (the copy option which I found hard to implement easily) 
if you want I can add it (maybe in a separate PR)

matanco64: done

tests/rustworkx_tests/digraph/test_edges.py
matanco64: done

src/digraph.rs
matanco64: done

releasenotes/notes/added-reverse-inplace-for-digraph-c59759a1d54edb1a.yaml
matanco64: done

tests/rustworkx_tests/digraph/test_edges.py
matanco64: done

src/digraph.rs
matanco64: yup :( all hell breaks loose

src/digraph.rs
matanco64: changed

matanco64 (commented):
I've gone over all the comments
1. renamed
2. added tests
3. changed to unwraps

I also added code to the interface at digraph.pyi so it will be easier to develop in pycharm and such

IvanIsCoding (approved):
LGTM, you just need to run `cargo fmt` because you forgot but it should be good to go

matanco64:
> 

Thank You! ran cargo fmt, should be good now :)

IvanIsCoding:
You’ll need to fix your docstring too, it’s giving `Inline literal start-string without end-string.` when building the docs

mtreinish:
I pushed up a quick fix to update the docs syntax here to get this merged as it LGTM besides the docs failure. The underlying issue was the docstring for rustworkx. The docstrings get exported to python as rst that sphinx will need to read but the docs were constructed using rust doc so I just converted that to use python/sphinx rst.

PR #857, opened 2023-04-12
Improve documentation for graph_greedy_color

mtreinish:
This commit improves the documentation for the graph_greedy_color function. Previously, the details on the function and the algorithm it implemented where a bit sparse. This commit expands it by explaining the source for the algorithm, making it clear it's not always going to return an optimal solution, and also adding an example.


src/coloring.rs
```diff
@@ -26,13 +26,36 @@ use petgraph::visit::NodeCount;
 
 use rayon::prelude::*;
 
-/// Color a PyGraph using a largest_first strategy greedy graph coloring.
+/// Color a :class:`~.PyGraph` object using a greedy graph coloring algorithm
+///
+/// This function uses a `largest-first` strategy as described in [1]_ and colors
+/// the nodes with higher degree first. As the coloring problem is NP-hard
+/// this is a heuristic algorithm and may not return an optimal solution.
```
Cryoris, line 33: How about putting the warning into a `..note::` to make it better visible? 🙂 We should also add it to `SparsePauliOp.group_commuting` in Qiskit Terra, which might be where users are coming from (I can add it there 🙂)

src/coloring.rs
```diff
@@ -26,13 +26,36 @@ use petgraph::visit::NodeCount;
...
 /// :param PyGraph: The input PyGraph object to color
 ///
 /// :returns: A dictionary where keys are node indices and the value is
 ///     the color
 /// :rtype: dict
+///
+/// .. jupyter-execute::
+///
+///     import matplotlib.pyplot as plt
+///
```
mtreinish, line 44:
```suggestion
```

src/coloring.rs
mtreinish: Done in: https://github.com/Qiskit/rustworkx/pull/857/commits/9243086e21e2d9556a40aa17dcd7d0efacfe4595

Cryoris:
There's a missing period in the first sentence, I think, otherwise LGTM!

Cryoris (approved):

IvanIsCoding (approved):

PR #861, opened 2023-04-14
Replace Tox with Nox

### LINKED ISSUE #852: Investigate tox alternatives

As a follow up to #851 and the tox 4 release series more generally along with the instability and backwards compatibly issues associated with that release. We should investigate looking at an alternative tool for managing our CI and local development environments and running of tests and other checks. The top contender in my mind is https://github.com/wntrblm/nox which I know pyo3 has moved to and it seems to work well for them. Personally I've not tried it so this issue is more track an investigation into alternative tooling and figure out what our migration plan away from tox would look like.

IvanIsCoding:
Closes #852 

We replace Tox with Nox, overall it seems promising. I do think customizing Nox seems much, much easier given that we have the full power of Python. In fact, I feel we could profit and replace many items in `tools/` too with Nox in the future


noxfile.py
```diff
@@ -0,0 +1,62 @@
...
+  "networkx>=2.5",
+  "stestr",
+]
+
+extras = [
+  "mpl",
+  "graphviz",
+]
+
+@nox.session(python=["3"])
```
IvanIsCoding, line 19: This is for @mtreinish: we could add all Python versions here, but I feel the way our CI is structured we prefer to spawn different jobs rather than create multiple Python environments in the same job

noxfile.py
```diff
@@ -0,0 +1,62 @@
...
+  "fixtures",
+  "testtools>=2.5.0",
+  "networkx>=2.5",
+  "stestr",
+]
+
+extras = [
+  "mpl",
+  "graphviz",
+]
```
IvanIsCoding, line 17: To be honest, I haven't integrated this well yet. I need to check how to pass the extras for the right environments

CONTRIBUTING.md
```diff
@@ -131,7 +131,7 @@ the compiled extension).
 
 If you just want to run a subset of tests you can pass a selection regex to the
 test runner. For example, if you want to run all tests that have "dag" in the
-test id you can run: `tox -epy -- dag`. You can pass arguments directly to the
+test id you can run: `nox -e test -- dag`. You can pass arguments directly to the
```
IvanIsCoding, line 134: I still need to test this on local, but the general tests passed on CI

.flake8
```diff
@@ -0,0 +1,10 @@
+[flake8]
```
IvanIsCoding, line 1: We got rid of the `tox.ini` file, so I had to put flake8 settings somewhere else

IvanIsCoding (commented):
Some minor details we need to handle before merging this. But overall, it shouldn't be too hard

CONTRIBUTING.md
```diff
@@ -110,19 +110,19 @@ also run successfully. Before you open a new pull request for your change,
...
-[**tox**](https://tox.readthedocs.io/en/latest/#). You can install tox
-with pip: `pip install -U tox`. Tox provides several advantages, but the
+[**Nox**](https://nox.thea.codes/en/stable/). You can install Nox
+with pip: `pip install -U nox`. Nox provides several advantages, but the
 biggest one is that it builds an isolated virtualenv for running tests. This
 means it does not pollute your system python when running. However, by default
-tox will recompile rustworkx from source every time it is run even if there
+Nox will recompile rustworkx from source every time it is run even if there
 are no changes made to the rust code. To avoid this you can use the
 `--skip-pkg-install` package if you'd like to rerun tests without recompiling.
```
IvanIsCoding, line 119: I need to check if this is the correct flag in Nox too

CONTRIBUTING.md
IvanIsCoding: Because we forward the arguments, it works. Which is nice

.flake8
mtreinish: Can we put this in the `pyproject.toml` just to avoid the proliferation of config files?

CONTRIBUTING.md
```diff
@@ -142,42 +142,42 @@ you can do this faster with the `-n`/`--no-discover` option. For example:
...
-image to the current working directory (which if running tests with `tox` is
+image to the current working directory (which if running tests with `nox` is
 `tests/`) to ensure the generated image is usable. However to not clutter the
 system each test cleans up this generated image and by default a test run does
 not include any way to view the images from the visualization tests.
 
 If you want to inspect the output from the visualization tests (which is common
 if you're working on visualizations) you can set the
-`RETWORKX_TEST_PRESERVE_IMAGES` environment variable to any value and this will
+`RUSTWORKX_TEST_PRESERVE_IMAGES` environment variable to any value and this will
```
mtreinish, line 175: Did we update the env variables in the tests to reflect this and just forgot to update the docs?

noxfile.py
mtreinish: For me a common use case is I do things like `tox -epy38 && tox -epy311` to test min and max versions. As long as this is supported in some way it's fine. I've never been a fan of the tox envlist approach where you call `tox` bare and it runs the list of jobs.

mtreinish (commented):
Sorry for the super slow review on this, I'm super keen to get us off of tox. I took a quick look through the config and it looks reasonable to me and this looks great, and is much nicer than what we had with tox. I still need to checkout the branch locally and play with it to see how it works in practice for me.

The one thing I think we should avoid doing for this commit is to keep the tox.ini file around for a little bit. Just because we've been using it since the creation of retworkx so leaving the config around for a while and maybe adding `python -c print("The use of tox for running tests is no longer supported in a future version of rustworkx you should use nox instead. Refer to the contributing guide for more details on it's usage")`  to the tox jobs. I'm just worried about the occasional contributor that comes back after 5 months and tries to run tox and gets frustrated.

.flake8
IvanIsCoding: You would be surprised by https://github.com/PyCQA/flake8/issues/234, they really don't want to support `pyproject.toml`! Maybe with https://github.com/john-hen/Flake8-pyproject

IvanIsCoding:
> Sorry for the super slow review on this, I'm super keen to get us off of tox. I took a quick look through the config and it looks reasonable to me and this looks great, and is much nicer than what we had with tox. I still need to checkout the branch locally and play with it to see how it works in practice for me.
> 
> The one thing I think we should avoid doing for this commit is to keep the tox.ini file around for a little bit. Just because we've been using it since the creation of retworkx so leaving the config around for a while and maybe adding `python -c print("The use of tox for running tests is no longer supported in a future version of rustworkx you should use nox instead. Refer to the contributing guide for more details on it's usage")` to the tox jobs. I'm just worried about the occasional contributor that comes back after 5 months and tries to run tox and gets frustrated.

I'll delete the `tox.ini` file and add a ["legacy" tox.ini](https://tox.wiki/en/stable/config.html#pyproject-toml) to pyproject.toml with the error messages you said

CONTRIBUTING.md
IvanIsCoding: I'll send it in a separate PR

.flake8
IvanIsCoding: Done

CONTRIBUTING.md
IvanIsCoding: Flag is `--no-install`, I updated the doc

noxfile.py
IvanIsCoding: I added support for that with `nox --python 3.11 -e test_with_version`. That way `nox -e test` triggers the version specified in CI and the `test_with_version` launches a specific version (or all of them)

noxfile.py
IvanIsCoding: This was a Tox thing. I switched all the pip install to use `.[all]` to include the deps

IvanIsCoding:
@mtreinish I think this ready for review, I addressed the comments

mtreinish (approved):
I pushed up a small commit to add back the tox.ini file in it's previous form and add the warning to the end of the ouptut. I think it's better to not actively break developer's existing workflows, but just make it explicit we're not supporting it anymore. The cost of keeping the tox.ini file around in the short to medium term is minimal.

IvanIsCoding:
> I pushed up a small commit to add back the tox.ini file in it's previous form and add the warning to the end of the ouptut. I think it's better to not actively break developer's existing workflows, but just make it explicit we're not supporting it anymore. The cost of keeping the tox.ini file around in the short to medium term is minimal.

I am thinking about wrap the current workflows around nox them i.e. make tox only install nox and then call `nox -e test` for example

PR #862, opened 2023-04-14
Check if nodes exist on `add_edge` methods

### LINKED ISSUE #855: `PyDiGraph.add_edge()` and `PyGraph.add_edge` panic if an endpoint doesn't exist
### Information

- **Python version**: 3.11
- **Rust version**: 1.68.2
- **Operating system**: Linux

If you try to add an edge to a graph object with endpoints that aren't present in the graph petgraph panics and that causes an catchable exception to be raised from rustworkx without a user actionable message.
 For example:
 
 ```
 pyo3_runtime.PanicException: StableGraph::add_edge: node index 3 is not a node in the graph
```

I think it would be better to raise an `IndexError` or a custom rustworkx exception class. The issue is a `PanicException` isn't something users can catch easily (as it doesn't inherit from `Exception`) and is supposed to be treated as irrecoverable. Also the error message isn't something a python end user will understand. We should do checks that the indices are present in the graph before calling petgraph and return an appropriate error instead.

```python
g = rx.PyDiGraph()
g.add_edge(2, 3)
```
or

```python
g = rx.PyGraph()
g.add_edge(2, 3)
```

IvanIsCoding:
Close #855 

This checks if the nodes exist and if they do not, it raises an `IndexError` instead of panicking


src/digraph.rs
```diff
@@ -984,6 +984,11 @@ impl PyDiGraph {
     pub fn add_edge(&mut self, parent: usize, child: usize, edge: PyObject) -> PyResult<usize> {
         let p_index = NodeIndex::new(parent);
         let c_index = NodeIndex::new(child);
+        if !self.graph.contains_node(p_index) || !self.graph.contains_node(c_index) {
```
mtreinish, line 987: I'm wondering if we should move this check to `_add_edge()` so that it covers all the edge addition methods. Like if we did:

```
graph = rx.PyDiGraph()
graph.add_edges_from([(0, 1, None), (0, 2, None)])
```
we'd get the same panic.

releasenotes/notes/handle-non-existent-edge-15d70cfe60c89ac2.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    :meth:`rustworkx.PyGraph.add_edge` and :meth:`rustworkx.PyDiGraph.add_edge` and now raises an 
+    `IndexError` when one of the nodes does not exist in the graph. Previously, it caused the Python 
+    interpreter to exit with a `PanicException`
```
mtreinish, line 6:
```suggestion
    interpreter to exit with a ``PanicException``
```

releasenotes/notes/handle-non-existent-edge-15d70cfe60c89ac2.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    :meth:`rustworkx.PyGraph.add_edge` and :meth:`rustworkx.PyDiGraph.add_edge` and now raises an 
+    `IndexError` when one of the nodes does not exist in the graph. Previously, it caused the Python
```
mtreinish, line 5:
```suggestion
    ``IndexError`` when one of the nodes does not exist in the graph. Previously, it caused the Python 
```

mtreinish (commented):
Thanks for fixing this. I'm wondering though if we have the same underlying issue with all the edge addition methods (except for the `extend*` ones).

src/digraph.rs
IvanIsCoding: Done

mtreinish (approved):
LGTM, thanks for the quick update

PR #866, opened 2023-05-08
[blocked] Switch to Furo Sphinx theme

Eric-Arellano:
Rustworkx is not really a part of the Qiskit ecosystem. While Qiskit uses it, it is independent of the Qiskit / QC ecosystem.

There has been tension with using qiskit_sphinx_theme because that theme is optimized exclusively for the Qiskit Ecosystem. For example, it adds a top nav bar to navigate around Qiskit.org. Rustworkx doesn't want that, so it's been overriding the HTML template to remove the top nav bar. But that risks changes to qiskit_sphinx_theme breaking the override, such as CI recently failing.

Instead, it's simpler to use Furo directly. 

(Why Furo? This is the direction qiskit_sphinx_theme is going too. See https://github.com/Qiskit/qiskit_sphinx_theme/issues/232 for the motivation.)


IvanIsCoding (commented):
Well I don't oppose this, PyTorch's theme is ok but it has its quirks (see #606). So if you say that Qiskit's Sphinx Theme is no longer going to support us that is fine.

I do recommend checking [PyData's Sphinx Theme](https://pydata-sphinx-theme.readthedocs.io/en/stable/index.html) though. Many scientific software libraries use the theme including NetworkX, NumPy, and Matplotlib. So it should be a theme that many of our users have seen before which puts it above Furo in my opinion

Eric-Arellano:
We had our first design session for the Ecosystem Theme of qiskit_sphinx_theme and I think it's plausible it will be good enough for this repo to want to use it. For example, we are thinking we should probably not show the top nav bar, which is a complaint with the current theme. See https://github.com/Qiskit/rustworkx/pull/866

So I'm going to close https://github.com/Qiskit/rustworkx/pull/866 for now. We can always revive the PR.

PR #867, opened 2023-05-08
Fix docs to work with Sphinx Theme 1.11

Eric-Arellano:
The HTML template overrides in this repo now diverge too much from qiskit_sphinx_theme. This caused docs to fail to build with the release 1.11. So, this PR re-vendors the base templates. As before, the HTML overrides do two things:

1. Remove the top nav bar from layout.html
2. Fix the URL redirect for previous versions. This will become unnecessary in Theme 1.12 thanks to https://github.com/Qiskit/qiskit_sphinx_theme/pull/267.

This also makes some improvements:
1. Pins qiskit_sphinx_theme to 1.11 for better stability, especially since this repo is overriding HTML templates.
2. Activates jQuery, which Sphinx stopped activating in Sphinx 6.
3. Removes outdated templates.

### DIFF
--- a/docs/source/_templates/sidebar.html
+++ b/docs/source/_templates/sidebar.html
@@ -18,7 +18,7 @@
             <div class="sidebar-l1-expandable">Previous Releases</div>
             <div class="sidebar-subheadings">
                 {% for version in version_list | sort(reverse=True) %}
-                    <div class="sidebar-l2"><a href="/ecosystem/retworkx/stable/{{ version }}/index.html">{{ version }}</a></div>
+                    <div class="sidebar-l2"><a href="/ecosystem/rustworkx/stable/{{ version }}/index.html">{{ version }}</a></div>
                 {% endfor %}
             </div>
         {% endif %}


docs/source/_templates/layout.html
Eric-Arellano, line 1: This file is copied from the 1.11 branch of qiskit_sphinx_theme. The only changes are deleting the code about qiskit-ui-shell (top nav bar).

docs/source/_templates/sidebar.html
```diff
@@ -0,0 +1,99 @@
...
+    {% if global_toc %}
+        {{ global_toc }}
+    {% endif %}
+
+    <!-- PREVIOUS RELEASES -->
+        {% if version_list %}
+            <div class="sidebar-l1-expandable">Previous Releases</div>
+            <div class="sidebar-subheadings">
+                {% for version in version_list | sort(reverse=True) %}
+                    <div class="sidebar-l2"><a href="/ecosystem/retworkx/stable/{{ version }}/index.html">{{ version }}</a></div>
```
Eric-Arellano, line 21: The href is changed.

docs/source/_templates/sidebar.html
```diff
@@ -0,0 +1,99 @@
...
+    var i;
+    for (i = 0; i < toc_rows.length; i++) {
+        if (toc_rows[i].children[0].className === "reference external") {
+            toc_rows[i].className = "toctree-l1-external"
+            }
+    }
+
+    // expand and unexpand previous releases dropdown when clicked
+    var release_rows = document.getElementsByClassName("sidebar-l1-expandable");
+    for (i = 0; i < release_rows.length; i++) {
```
Eric-Arellano, line 88: @javabster found a bug while helping me to figure out why things weren't working. The variable was set to `expandable_rows`. We will fix this upstream.

docs/source/conf.py
```diff
@@ -40,6 +40,7 @@
               'sphinx.ext.intersphinx',
               'sphinxemoji.sphinxemoji',
               'sphinx_reredirects',
+              'qiskit_sphinx_theme',
```
Eric-Arellano, line 43: This activates jQuery, which is no longer done by default with Sphinx 6. The Pytorch theme requires jQuery.

docs/source/_templates/sidebar.html
mtreinish: This actually should be:

```suggestion
                    <div class="sidebar-l2"><a href="/ecosystem/rustworkx/stable/{{ version }}/index.html">{{ version }}</a></div>
```

Thanks to the project rename and then the ecosystem migration we've had 3 copies of the docs out there: `documentation/retworkx`, `documentation/rustworkx`, and now `ecosystem/rustworkx` the last one should be the canonical copy moving forward.

mtreinish (commented):
Thanks for doing this, LGTM just one inline comment (that was probably an issue before this PR too, but it's good to fix it now)

mtreinish (approved):

PR #875, opened 2023-05-16
moving greedy-color to rustworkx-core

alexanderivrii:
This PR moves the `greedy_color` algorithm to `rustworkx_core`. The algorithm is exactly the same modulo some minor renaming and replacement of `node_indices` by `node_identifiers`.

The rust tests in `rustworkx-core/tests/coloring.rs` are adapted from python tests in `rustworkx_test/graph/test_coloring.py`.

---

I am wondering if there is a way to replace 
```
let mut expected_colors = DictMap::new();
expected_colors.insert(NodeIndex::new(0), 0);
expected_colors.insert(NodeIndex::new(1), 1);
expected_colors.insert(NodeIndex::new(2), 1);
```
by something like `DictMap::from` (I was not able to make this work).


releasenotes/notes/migrate-greedy-color-c3239f35840eec18.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    The function ``greedy_color``has been migrated to the ``rustworkx-core``
+    crate in the ``coloring`` module. It colors a graph using a greedy 
+    graph coloring algorithm.
```
mtreinish, line 6:
```suggestion
    Added a new function, ``greedy_color``, to ``rustworkx-core`` in a new
    ``coloring`` module. It colors a graph using a greedy 
    graph coloring algorithm.
```

mtreinish (commented):
This LGTM, thanks for doing this. Just a couple small inline comments and questions, but then I think this is ready to merge.

rustworkx-core/tests/coloring.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use petgraph::graph::Graph;
+use petgraph::graph::NodeIndex;
+use petgraph::Undirected;
+use rustworkx_core::dictmap::*;
+
+use rustworkx_core::coloring::greedy_color;
+
+#[test]
+fn test_greedy_color_empty_graph() {
```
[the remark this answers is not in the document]
enavarro51: I think in our previous discussions, we want to put tests that are for only one module inside the module as you suggest. Putting new modules in `tests` is generally reserved for tests that are testing interactions of multiple modules.

rustworkx-core/tests/coloring.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+#[test]
+fn test_greedy_color_simple_graph() {
+    let graph = Graph::<(), (), Undirected>::from_edges(&[(0, 1), (0, 2)]);
+    let colors = greedy_color(&graph);
+    let mut expected_colors = DictMap::new();
+    expected_colors.insert(NodeIndex::new(0), 0);
+    expected_colors.insert(NodeIndex::new(1), 1);
+    expected_colors.insert(NodeIndex::new(2), 1);
```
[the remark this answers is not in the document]
alexanderivrii: Thanks! I like the suggested approach (and in this case we also don't need the `mut` keyword). I have updated the tests accordingly (and kept the original way in the docs example).

rustworkx-core/tests/coloring.rs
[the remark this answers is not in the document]
alexanderivrii: Thanks! I have moved the tests to the coloring module (and I can see these running, now that I have learned to run `cargo test`).

releasenotes/notes/migrate-greedy-color-c3239f35840eec18.yaml
alexanderivrii: Updated.

rustworkx-core/src/coloring.rs
```diff
@@ -0,0 +1,86 @@
...
+/// expected_colors.insert(NodeIndex::new(0), 0);
+/// expected_colors.insert(NodeIndex::new(1), 1);
+/// expected_colors.insert(NodeIndex::new(2), 1);
+/// assert_eq!(colors, expected_colors);
+/// ```
+///
+///
+/// .. [1] Adrian Kosowski, and Krzysztof Manuszewski, Classical Coloring of Graphs,
+///     Graph Colorings, 2-19, 2004. ISBN 0-8218-3458-4.
+pub fn greedy_color<G>(graph: G) -> DictMap<G::NodeId, usize>
```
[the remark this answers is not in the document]
alexanderivrii: I have changed the name of the internal rust function to `greedy_node_color`. I don't feel strongly one way or another, but it might be slightly more clear if we are planning to add `greedy_edge_color` later.

rustworkx-core/src/coloring.rs
```diff
@@ -0,0 +1,144 @@
...
+use std::hash::Hash;
+
+use crate::dictmap::*;
+use hashbrown::{HashMap, HashSet};
+use petgraph::visit::{EdgeRef, IntoEdges, IntoNodeIdentifiers, NodeCount};
+use rayon::prelude::*;
+
+/// Color a graph using a greedy graph coloring algorithm.
+///
+/// This function uses a `largest-first` strategy as described in [1]_ and colors
```
mtreinish, line 23: I don't think this footnote format works in rustdoc. Looking at the CI run of `cargo doc` it looks like it wasn't happy with this either:

https://github.com/Qiskit/rustworkx/actions/runs/5010849131/jobs/8981094523#step:14:46

Looking at the rustdoc documentation the syntax for footnotes is: https://doc.rust-lang.org/stable/rustdoc/how-to-write-documentation.html?highlight=footnote#footnotes but it might make sense to just put it inline instead of using a footnote here.

rustworkx-core/src/coloring.rs
```diff
@@ -0,0 +1,144 @@
...
+/// the nodes with higher degree first.
```
mtreinish, line 24: Did you want to add the note from #857 here too?

mtreinish (commented):
This LGTM now. Just a couple small inline comments about the documentation for the new function.

rustworkx-core/src/coloring.rs
mtreinish: I pushed up https://github.com/Qiskit/rustworkx/pull/875/commits/4cd566503af77ba7bea56b1c28c090cd5e4bbec0 to resolve this.

mtreinish (approved):

rustworkx-core/src/coloring.rs
alexanderivrii: @mtreinish, thanks for taking care of this.

PR #881, opened 2023-05-25
Prepare 0.13.0 release

mtreinish:
This commit prepares the 0.13.0 release. It moves the release notes that for the release to a separate directory and adds a prelude to the release notes. This should be the last commit to merge and after it merges it'll that commit will be tagged as the 0.13.0 release.

TODO:

- [x] Wait for the last PRs in the milestone to merge (or be deferred to the next release)
- [x] Expand some release notes with more detail


mtreinish:
I decided to defer #823 since it wasn't critical to include in 0.13.0. So with that I think this is unblocked now. So if people are good with the release notes here we can merge this and then tag the release.

IvanIsCoding:
> I decided to defer #823 since it wasn't critical to include in 0.13.0. So with that I think this is unblocked now. So if people are good with the release notes here we can merge this and then tag the release.

I don’t have bandwidth to review it now, but I think we can include it in 0.13.1 before a 0.14 release 

I have more Python annotation in the works so we could backport #823 and release 0.13.1 with deferred from release + annotations + fixes if we find bugs

IvanIsCoding (approved):

PR #882, opened 2023-05-27
[WIP] Shortern boilerplate for `singledispatch`

IvanIsCoding:
Shortern boilerplate code for `singledispatch`. This is convenient becase we don't need repeat the function signature

I need to investigate why the Spiral Layout and Bipartite Layout tests are failling but everything else looks healthy


mtreinish (approved):
This is an excellent cleanup. It reduces the LoC and removes a layer of indirection which added like 100ns to each of the dispatch calls.

PR #883, opened 2023-05-27
Automagic dispatch for universal functions

IvanIsCoding:
Follow up of #882, albeit more agressive on shortening the boilerplate amount of code for universal functions

This PR introduces a new `_rustworkx_dispatch` decorator that:
* Creates the singledispatch for `func`
* Automatically registers `digraph_func` for `PyDiGraph`
* Automatically registers `graph_func` for `PyGraph`

The decorator is based on the original `functools.singledispatch`. It should simplify the amount of code we need to write an universal function. It does add a little bit of "magic" with functools and importlib, but overall it should be maintainable


IvanIsCoding:
This is ready for review

rustworkx/__init__.py
```diff
@@ -15,7 +16,8 @@
 # flake8: noqa
 import rustworkx.visit
 
-sys.modules["rustworkx.generators"] = generators
+__rustworkx_mod__ = importlib.import_module(".rustworkx", package="rustworkx")
+sys.modules["rustworkx.generators"] = __rustworkx_mod__.generators
```
jakelishman, line 20: Out of curiosity, why `importlib` over the regular `from . import rustworkx as __rustworkx_mod__`?

jakelishman (approved):
I scanned through all this and all the re-registrations look right to me, including the `is_isomorphic_node_match` rewrite to not be a direct dispatcher itself.

I personally avoid magic name inference like this, but this will fail loudly and immediately at import time if something's not available, so I'd agree it should be maintainable enough (and the line-count reduction is nice for sure).

My ticks don't count on rustworkx, but you can have one in principle (with or without the `importlib` change).

rustworkx/__init__.py
IvanIsCoding: It is a matter of personal preference, but in my opinion it pairs better with `sys.modules`. For `rustworkx.generators`, the old version was not readable at all. `importlib.import_module` is a common pair with `sys.modules`

jakelishman: I'm not really sure I follow - `importlib.import_module` is kind of just a slightly less efficient way of doing retrieving the exact same module object when all the names are literals - but it was mostly idle curiosity and not meant to be an issue. I'd just never have thought to use the dynamic system when the static system would work.

IvanIsCoding: It is just that:
```
__rustworkx_mod__ = importlib.import_module(".rustworkx", package="rustworkx")
sys.modules["rustworkx.generators"] = __rustworkx_mod__.generators # clear where this came from
```
Is less confusing them:
```
from .rustworkx import *
sys.modules["rustworkx.generators"] = generators  # where did this variable come from???
```

jakelishman: That second one isn't quite what I'm suggesting - I'm saying the second would be
```python
from . import rustworkx as __rustworkx_mod__
from .rustworkx import *
sys.modules["rustworkx.generators"] = __rustworkx_mod__.generators
```

No `importlib` needed.

IvanIsCoding: I will close as `from . import rustworkx` is equally as cryptic and we came to the same conclusion

mtreinish (approved):
LGTM, we end up doing this a lot so abstract away the boilerplate makes a lot of sense here.

PR #885, opened 2023-05-31
Implementation of cycle_basis_edges

### LINKED ISSUE #551: Add equivalent of cycle_basis that returns an edge list

Create an equivalent of the `retworkx.cycle_basis` method but returning edges instead of nodes.

https://github.com/Qiskit/retworkx/blob/5b90ee7821398625b16aec2e4d782dd2ec26daf2/src/connectivity/mod.rs#L67-L135

The original paper (http://www.cs.kent.edu/~dragan/GraphAn/CycleBasis/p514-paton.pdf) which we implement returns a set of nodes instead a set of edges. #505 pointed out that returning a set of edges is more useful in many cases. The expected enhancement is to have a method for each such that users can get a cycle basis both with nodes and with edges.

raynelfss:
Fixes #551

### Summary

The following commits implement the function `cycle_basis_edges` which draws inspiration from the existent `cycle_basis` function. This version modifies the existing pipeline to return a List containing a vector of `EdgeId` in the rustworx-core version of the function.


raynelfss:
Results from testing in Python:

image

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -117,9 +117,160 @@ where
...
+/// # Example
+/// ```rust
+/// use petgraph::prelude::*;
+/// use rustworkx_core::connectivity::cycle_basis_edges;
+///
+/// let edge_list = [(0, 1), (0, 3), (0, 5), (1, 2), (2, 3), (3, 4), (4, 5)];
+/// let graph = UnGraph::<i32, i32>::from_edges(&edge_list);
+/// let mut res: Vec<Vec<(NodeIndex, NodeIndex)>> = cycle_basis_edges(&graph, Some(NodeIndex::new(0)));
+/// ```
+pub fn cycle_basis_edges<G>(graph: G, root: Option<G::NodeId>) -> Vec<Vec<(G::NodeId, G::NodeId)>>
```
mtreinish, line 151: Instead of returning `Vec<(G::NodeId, G::NodeId)>` here it might be better to return `Vec<G::EdgeId>`. Then if we want to return the edge endpoint or weight we can just look it up from the index.

mtreinish (commented):
So overall this looks like a good start. The big things that I think are missing are python side tests to validate the python api is working as expected and also a release note (https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md#release-notes).

The other small concern I have is the amount of duplication between `cycle_basis_edges` and `cycle_basis`. The code is mostly the same between the function and I'm thinking it would be good to combine them into a single code path.

As for the edge weight for the python side you can call `weight.clone_ref(py)` as it's a `PyObject` type and we can clone it with a GIL handle (which is what the `Python` type gives us).

rustworkx-core/src/connectivity/cycle_basis.rs
raynelfss: It's funny you mention this, as in an earlier version, I had the function return the `EdgeId` attribute of each `EdgeRef` but replaced it with the tuple thinking that it should instead return the edge objects. 

Also, I'm assuming that by returning the edge endpoint you mean returning it using the wrapper located in `rustworkx/src/connectivity/mod.rs` lines 100 to 110. Should I use the graph.edges() method fetch the edges by index?

mtreinish: By edge endpoints I meant what you're doing now. For each edge you find you're doing `(edge.source(), edge.target())` (the endpoints being the tuple of nodes for the edge), but instead if you did `edge.id()` that would give you the unique `G::EdgeId` for the edge. Then if you returned a `Vec<G::EdgeId>` callers of this function could look up any attributes of the edge they needed from that list. For examples if you wanted a list of endpoint tuples you could do something like:
```rust
let edge_endpoints = cycle_basis(graph, None)
    .into_iter()
    .filter_map(|e| graph.edge_endpoints(e))
    .collect()
```

but it would also give you the flexibility to get the weights, or other details. It basically is just returning a reference to the edge in `graph` instead of one of properties of the edge.

raynelfss: I see, I will make those changes accordingly.

raynelfss:
> if there a reason to declare both a cycles_edges and a cycles_nodes if from what I understand we only return one or the other? Is it possible in Rust to declare variables like:
if want_edge: let mut cycles: Vec<Vec<G::EdgeId>> = Vec::new(); else: let mut cycles: Vec<Vec<G::NodeId>> = Vec::new();

Well, there might be a better way of doing so that doesn't involve two vectors. The reason I did it this way is that I couldn't conditionally assign this variable with different types (`G::EdgeId || G::NodeId`) without having to wrap two similar processes under if and else statements, which could be a bit redundant.

I could, however, change the enum so that it represents the individual nodes/edges (instead of `Vec<Vec<T>>` for each type) and make the function return an instance of `Vec<Vec<EdgeOrNodes<E, N>>>` and then unwrap the values of said `Vec` individually.

The problem happens when importing this enum into the wrapper function, which complains about the types in use. I could try and get this to work with enough time.

danielleodigie (commented):
LGTM!

mtreinish (commented):
Thanks for the fast update. This is look better from a code duplication standpoint and is making good progress. I left some inline comments mainly about the interface split. I have some concerns that the current version of the code has some breaking API changes to rustworkx-core that we can't make without potentially impacting users. I left some inline suggestions on how we can tweak this a bit to avoid that kind of impact.

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -43,18 +45,52 @@ use std::hash::Hash;
...
-/// let mut res: Vec<Vec<NodeIndex>> = cycle_basis(&graph, Some(NodeIndex::new(0)));
+/// let mut res: Vec<Vec<NodeIndex>> = cycle_basis(&graph, Some(NodeIndex::new(0)), Some(false)).unwrap_nodes();
 /// ```
-pub fn cycle_basis<G>(graph: G, root: Option<G::NodeId>) -> Vec<Vec<G::NodeId>>
+
+pub fn cycle_basis<G>(
+    graph: G,
+    root: Option<G::NodeId>,
+    edges: Option<bool>,
+) -> EdgesOrNodes<G::NodeId, G::EdgeId>
```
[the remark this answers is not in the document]
raynelfss: Oh lord, that was a big oversight on my end. I will make that correction as soon as possible. As for the two public-facing options, could we instead have those two methods in the `rustworkx/src/connectivity/mod.rs` file?

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -114,27 +191,67 @@ where
...
+
+/// Enum for custom return types of `cycle_basis()`.
+pub enum EdgesOrNodes<N, E> {
+    Nodes(Vec<Vec<N>>),
+    Edges(Vec<Vec<E>>),
+}
+/// Functions used to unwrap the desired datatype of `EdgesOrNodes`.
+impl<N, E> EdgesOrNodes<N, E> {
+    pub fn unwrap_nodes(self) -> Vec<Vec<N>> {
+        use EdgesOrNodes::*;
```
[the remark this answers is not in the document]
raynelfss: Yes, it was confusing to me as well.  Using `Self` instead seems to work better. Thank you for your input.

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -114,27 +191,67 @@ where
...
+        match self {
+            Nodes(x) => x,
+            Edges(_) => vec![],
```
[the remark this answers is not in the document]
raynelfss: Interesting, I will look into that further. I didn't know how to handle this outcome properly. Thank you!

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -114,27 +191,67 @@ where
...
+        }
+    }
+    pub fn unwrap_edges(self) -> Vec<Vec<E>> {
+        use EdgesOrNodes::*;
+        match self {
+            Edges(x) => x,
+            Nodes(_) => vec![],
+        }
+    }
 }
```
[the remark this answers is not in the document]
raynelfss: Yes, since we're using a custom return type that can't be interpreted natively as `Vec<Vec<NodeId>>` or `Vec<Vec<EdgeId>>` the values need to be unwrapped. 

Although this could technically be done inside of the public-facing functions that were part of your review, we'd have to rewrite the same procedure twice over (in testing and in the public-facing functions, which is redundant.

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -114,27 +191,67 @@ where
...
-    cycles
+    if want_edges {
+        EdgesOrNodes::Edges(cycles_edges)
+    } else {
+        EdgesOrNodes::Nodes(cycles_nodes)
+    }
+}
+
+/// Enum for custom return types of `cycle_basis()`.
+pub enum EdgesOrNodes<N, E> {
```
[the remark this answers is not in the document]
raynelfss: Yes, having these work internally would be better. I will make this change accordingly.

releasenotes/notes/add-cycle-basis-edges-5cb31eac7e41096d.yaml
```diff
@@ -0,0 +1,56 @@
...
+      edge_list = list(graph.edge_list())
+      cycle_info = [[edge_list[edge] for edge in cycle] for cycle in cycles_edges]
+      # Print the EdgeID's that form cycles in the graph
+      display(cycles_edges)
+      # Print the data retrieved from the graph.
+      display(cycle_info)
+fixes:
+  - |
+    Support for edges when getting the cycle basis. Refer to 
+    `#551 <https://github.com/Qiskit/rustworkx/issues/551>`__ for more details.
```
mtreinish, line 56: You don't actually need this release note what you have for the feature note is more than sufficient to document the new feature for this release. This is really just a new feature there isn't any bug being fixed here. The github "issues" tracker is really both for tracking bugs and also for feature requests.

releasenotes/notes/add-cycle-basis-edges-5cb31eac7e41096d.yaml
```diff
@@ -0,0 +1,56 @@
+---
+features:
+  - |
+    A function, :func:`~rustworkx.cycle_basis_edges` was added to the crate
+    ``rustworkx-core`` in the ``connectivity`` module. This function returns 
+    the edge indices that form the cycle basis of a graph.
```
mtreinish, line 6: You need to split this into 2 release notes, one for the rustworkx-core function and one for the python side function being added.
 
```suggestion
  - |
    A function, ``cycle_basis_edges`` was added to the crate
    ``rustworkx-core`` in the ``connectivity`` module. This function returns 
    the edge indices that form the cycle basis of a graph.
  - |
    Added a new function :func:`~rustworkx.cycle_basis_edges`  which is similar
    to the existing :func:`~.cycle_basis` function but instead of returning node
    indices it returns a list of edge indices for the cycle.
```

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -79,6 +105,7 @@ where
         while !stack.is_empty() {
             // Use the last element added so that cycles are easier to find
             let z = stack.pop().unwrap();
+            // println!("Length of all edges from: {:?}", edges.len());
```
mtreinish, line 108:
```suggestion
```

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -114,18 +181,140 @@ where
...
+/// let graph = UnGraph::<i32, i32>::from_edges(&edge_list);
+/// let mut res: Vec<Vec<NodeIndex>> = cycle_basis(&graph, Some(NodeIndex::new(0)));
+/// ```
+pub fn cycle_basis<G>(graph: G, root: Option<G::NodeId>) -> Vec<Vec<G::NodeId>>
+where
+    G: NodeCount,
+    G: IntoEdges,
+    G: IntoNodeIdentifiers,
+    G::NodeId: Eq + Hash,
+    G::EdgeId: Eq + Hash,
```
mtreinish, line 256: Hmm, I'm mildly concerned about this change to the trait bounds here. It's also technically a backwards incompatible change because it's adding extra constraints (`G: IntoEdges` and `G::EdgeId: Eq + Hash`) on what is acceptable input for `graph`. I'm not sure how to fix this short of going against my earlier suggestion and re-duplicating the implementation and having 2 copies one for `cycle_basis` and the other for `cycle_basis_edges`. However, I'm also struggling to  come up with a real world example where an input graph type might not support `IntoEdges` given the previous traits. So I'm thinking we should just have an `upgrade` note documenting the potentially breaking API change to this function

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -29,32 +31,56 @@ use std::hash::Hash;
...
+    ) -> Vec<G::EdgeId>
+    where
+        G: IntoEdges,
+    {
+        orig_graph
+            .edges(origin)
+            .filter(|edge: &G::EdgeRef| {
+                if equiv {
+                    edge.source() == target && edge.target() == target
+                } else {
```
mtreinish, line 76: Is this extra equiv argument actually needed in case of checking for self loops? Can you swap `.edges_directed(origin, petgraph::Direction::Outgoing)` in instead of `edges()` the you always know source == origin, and if target == source there isn't an option to return anything else (I think this also true without switching `edges()` to `edges_directed()`).

mtreinish (commented):
This is looking good, thanks for the fast updates on this, I have a few more comments inline.

rustworkx-core/src/connectivity/cycle_basis.rs
raynelfss: Yes, this is a redundancy. Since for the case of a self-loop, we are still passing origin and target arguments,  the origin node can be passed as the desired target and that would pretty much work. I will modify the code accordingly.

rustworkx-core/src/connectivity/cycle_basis.rs
raynelfss: Interesting point. The trait IntoEdges is needed to access Edge information. I was unaware of the implications this change would bring. I will ass the `upgrade` note to the release notes accordingly.

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -29,32 +31,45 @@ use std::hash::Hash;
...
+    G::EdgeId: Eq + Hash,
 {
     let mut root_node = root;
     let mut graph_nodes: HashSet<G::NodeId> = graph.node_identifiers().collect();
-    let mut cycles: Vec<Vec<G::NodeId>> = Vec::new();
+    let mut cycles_edges: Vec<Vec<G::EdgeId>> = Vec::new();
+    let mut cycles_nodes: Vec<Vec<G::NodeId>> = Vec::new();
+
+    // Method used to retrieve all the edges between an origin node and a target node.
+    // Can be used to check if a node loops back to itself by enabling equiv
```
mtreinish, line 61:
```suggestion
```

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -29,32 +31,45 @@ use std::hash::Hash;
...
     G::NodeId: Eq + Hash,
+    G::EdgeId: Eq + Hash,
 {
     let mut root_node = root;
     let mut graph_nodes: HashSet<G::NodeId> = graph.node_identifiers().collect();
-    let mut cycles: Vec<Vec<G::NodeId>> = Vec::new();
+    let mut cycles_edges: Vec<Vec<G::EdgeId>> = Vec::new();
+    let mut cycles_nodes: Vec<Vec<G::NodeId>> = Vec::new();
+
+    // Method used to retrieve all the edges between an origin node and a target node.
```
mtreinish, line 60: It's technically a docstring even though it won't get rendered in any doc builds.

```suggestion
    /// Method used to retrieve all the edges between an origin node and a target node.
```

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -29,32 +31,45 @@ use std::hash::Hash;
...
+    // Can be used to check if a node loops back to itself by enabling equiv
+    fn get_edge_between<G>(orig_graph: G, origin: G::NodeId, target: G::NodeId) -> Vec<G::EdgeId>
+    where
+        G: IntoEdges,
+    {
+        orig_graph
+            .edges(origin)
+            .filter(|edge: &G::EdgeRef| edge.target() == target)
+            .map(|edge: G::EdgeRef| edge.id())
+            .collect()
```
mtreinish, line 70: Instead of returning a Vec I wonder if it is more correct to return just a single edge. If there are any parallel edges in the cycle it will return all those edges in the cycle which will be really unexpected I think. Instead if we changed the return type to `G::EdgeId` and changed this to `.next().unwrap()` this will just return the first added edge between two nodes.

raynelfss: Yeah, that seems valid. Will make those changes.

releasenotes/notes/add-cycle-basis-edges-5cb31eac7e41096d.yaml
```diff
@@ -0,0 +1,56 @@
...
+          (g, h, 0.9), 
+          (g, i, 1.0), 
+          (i, j, 0.8),
+      ])
+
+      mpl_draw(graph, with_labels=True)
+      # Retrieve EdgeIDs by enabling the edges flag.
+      cycles_edges = rustworkx.cycle_basis_edges(graph, a)
+      edge_list = list(graph.edge_list())
+      cycle_info = [[edge_list[edge] for edge in cycle] for cycle in cycles_edges]
```
mtreinish: This actually won't work in the presence of removals. It doesn't come up in this example since you're not removing any edges. But it would be better to rewrite this to be a bit more robust as an example for other applications.


```suggestion
      edge_list = list(graph.edge_list())
      cycle_info = [[graph.get_edge_data_by_index(edge) for edge in cycle] for cycle in cycles_edges]
```

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -29,32 +31,37 @@ use std::hash::Hash;
...
+        G::NodeId,
+    ) -> Vec<T>,
+) -> Vec<Vec<T>>
 where
     G: NodeCount,
     G: IntoNeighbors,
     G: IntoNodeIdentifiers,
+    T: Eq + Hash,
     G::NodeId: Eq + Hash,
 {
```
mtreinish: Instead of inlining this, for legibility I would probably add a new generic `F` and add `F` to the `where` clause:


```suggestion
fn inner_cycle_basis<G, T, F>(
    graph: G,
    root: Option<G::NodeId>,
    self_cycle_filter: impl Fn(G, G::NodeId) -> Vec<T>,
    cycle_filter: F,
) -> Vec<Vec<T>>
where
    G: NodeCount,
    G: IntoNeighbors,
    G: IntoNodeIdentifiers,
    T: Eq + Hash,
    G::NodeId: Eq + Hash,
    F: Fn(G, &HashSet<G::NodeId>, &HashMap<G::NodeId, G::NodeId>, G::NodeId, G::NodeId) -> Vec<T>,
{
```

I didn't add a generic for `self_cycle_filter` since it was pretty short, although it's a bit out of place with an `impl` inline and all the other trait bounds are in the `where`.

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -116,15 +115,166 @@ where
...
+/// is defined as the exclusive-or of the edges.
+///
+/// This is adapted from
+///    Paton, K. An algorithm for finding a fundamental set of
+///    cycles of a graph. Comm. ACM 12, 9 (Sept 1969), 514-518.
+///
+/// The function implicitly assumes that there are no parallel edges.
+/// It may produce incorrect/unexpected results if the input graph has
+/// parallel edges.
+///
```
mtreinish, line 128:
```suggestion
```

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -76,8 +83,8 @@ where
         let mut used: HashMap<G::NodeId, HashSet<G::NodeId>> = HashMap::new();
         used.insert(root_index, HashSet::new());
         // Walk the spanning tree
-        // Use the last element added so that cycles are easier to find
         while let Some(z) = stack.pop() {
+            // Use the last element added so that cycles are easier to find
```
mtreinish: Why move this comment?

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -116,15 +115,166 @@ where
...
+/// let mut res: Vec<Vec<NodeIndex>> = cycle_basis(&graph, Some(NodeIndex::new(0)));
+/// ```
+pub fn cycle_basis<G>(graph: G, root: Option<G::NodeId>) -> Vec<Vec<G::NodeId>>
+where
+    G: NodeCount,
+    G: IntoNeighbors,
+    G: IntoNodeIdentifiers,
+    G::NodeId: Eq + Hash,
+{
+    // inner_cycle_basis(graph, root, false).unwrap_nodes()
```
mtreinish:
```suggestion
```

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -116,15 +115,166 @@ where
...
+        G: IntoEdges,
+    {
+        orig_graph
+            .edges(origin)
+            .filter_map(|edge: G::EdgeRef| (edge.target() == target).then_some(edge.id()))
+            .next()
+            .expect("An edge should exist between origin and target node")
+    }
+
+    // inner_cycle_basis(graph, root, false).unwrap_nodes()
```
mtreinish:
```suggestion
```

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -116,15 +115,166 @@ where
...
+///
+/// This is adapted from
+///    Paton, K. An algorithm for finding a fundamental set of
+///    cycles of a graph. Comm. ACM 12, 9 (Sept 1969), 514-518.
+///
+/// The function implicitly assumes that there are no parallel edges.
+/// It may produce incorrect/unexpected results if the input graph has
+/// parallel edges.
+///
+///
```
mtreinish:
```suggestion
```

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -29,32 +31,37 @@ use std::hash::Hash;
...
-/// use rustworkx_core::connectivity::cycle_basis;
-///
-/// let edge_list = [(0, 1), (0, 3), (0, 5), (1, 2), (2, 3), (3, 4), (4, 5)];
-/// let graph = UnGraph::<i32, i32>::from_edges(&edge_list);
-/// let mut res: Vec<Vec<NodeIndex>> = cycle_basis(&graph, Some(NodeIndex::new(0)));
-/// ```
-pub fn cycle_basis<G>(graph: G, root: Option<G::NodeId>) -> Vec<Vec<G::NodeId>>
+/// * `self_cycle_filter` - Specifies the behavior when a single node cycle
+///   is found.
+/// * `cycle_filter` - Specifies the behavior when a cycle is found
```
mtreinish: A small question are these really filters? Isn't it more that this builds the output given a found cycle?

mtreinish (commented):
Overall this is looking much better, thanks for picking this back up. I just left some small inline comments and questions. But I think this is much closer now.

rustworkx-core/src/connectivity/cycle_basis.rs
raynelfss: It's funny you mention that because the initial implementation had them all be generics, but then I felt weird about having too many generics, but this approach makes more sense to me.

rustworkx-core/src/connectivity/cycle_basis.rs
```diff
@@ -76,8 +78,8 @@ where
         let mut used: HashMap<G::NodeId, HashSet<G::NodeId>> = HashMap::new();
         used.insert(root_index, HashSet::new());
         // Walk the spanning tree
-        // Use the last element added so that cycles are easier to find
         while let Some(z) = stack.pop() {
+            // Use the last element added so that cycles are easier to find
```
raynelfss:
```suggestion
        // Use the last element added so that cycles are easier to find
        while let Some(z) = stack.pop() {
```

rustworkx-core/src/connectivity/cycle_basis.rs
raynelfss: They're not filters at all, it was just the name that came to mind at the time. Would `cycle_builder` work better?

rustworkx-core/src/connectivity/cycle_basis.rs
mtreinish: You can never have too many generics, it just means the function is more generic :D

rustworkx-core/src/connectivity/cycle_basis.rs
mtreinish: Yeah cycle builder would be more descriptive to me

raynelfss: Updated in 70bbdd330c2dc7e1db4479ca7f79e7a3ced14a59

rustworkx-core/src/connectivity/cycle_basis.rs
raynelfss: Fixed in bcf55484065945a7b43ef2e7e4a51cba05fdfe73

PR #886, opened 2023-05-31
Node and Edge Filtering

### LINKED ISSUE #800: Node and Edge Filtering and Selection

It would be great to have the ability to filter a graph's nodes and edges by some abstract criteria conditioned on a node's data payload.

A analogous feature seen in `igraph` would be the methods/functions of `select` and `find`.
```
# G is some arbitrary igraph.Graph

G.vs.select(lambda v: v["weight"] > 4)  # vertex selection
G.vs.find(name="foo")
```

danielleodigie:
Aims to resolve #800 
Adds filter_nodes() and filter_edges() methods to the Graph and DiGraph classes
:)


src/digraph.rs
```diff
@@ -2890,6 +2890,60 @@ impl PyDiGraph {
...
+                .graph
+                .node_indices()
+                .filter_map(|node_index| {
+                    if filter(node_index).ok()? {
+                        Some(node_index)
+                    } else {
+                        None
+                    }
+                })
+                .map(|node_index| node_index.index())
```
mtreinish, line 2915:
```suggestion
                .filter_map(|node_index| {
                    if filter(node_index).ok()? {
                        Some(node_index.index())
                    } else {
                        None
                    }
                })
```

src/digraph.rs
```diff
@@ -2890,6 +2890,60 @@ impl PyDiGraph {
...
+                .graph
+                .edge_indices()
+                .filter_map(|edge_index| {
+                    if filter(edge_index).ok()? {
+                        Some(edge_index)
+                    } else {
+                        None
+                    }
+                })
+                .map(|edge_index| edge_index.index())
```
mtreinish, line 2942:
```suggestion
                .filter_map(|edge_index| {
                    if filter(edge_index).ok()? {
                        Some(edge_index.index())
                    } else {
                        None
                    }
                })
```

src/graph.rs
```diff
@@ -1889,6 +1889,60 @@ impl PyGraph {
...
+                .graph
+                .node_indices()
+                .filter_map(|node_index| {
+                    if filter(node_index).ok()? {
+                        Some(node_index)
+                    } else {
+                        None
+                    }
+                })
+                .map(|node_index| node_index.index())
```
mtreinish, line 1915:
```suggestion
                .filter_map(|node_index| {
                    if filter(node_index).ok()? {
                        Some(node_index.index())
                    } else {
                        None
                    }
                })
```

src/graph.rs
```diff
@@ -1889,6 +1889,60 @@ impl PyGraph {
...
+                .graph
+                .edge_indices()
+                .filter_map(|edge_index| {
+                    if filter(edge_index).ok()? {
+                        Some(edge_index)
+                    } else {
+                        None
+                    }
+                })
+                .map(|edge_index| edge_index.index())
```
mtreinish, line 1942:
```suggestion
                .filter_map(|edge_index| {
                    if filter(edge_index).ok()? {
                        Some(edge_index.index())
                    } else {
                        None
                    }
                })
```

mtreinish (commented):
This is looking like a good start, thanks for doing this! I left a small suggestion inline on how you can remove an unnecessary `map()` call. Besides that can you add some tests to validate it works as expected. Also make sure to add a test that raises an exception when `filter_function` is run to validate the exception gets raised back to python correctly. Besides that if you could add a release note documenting the addition of this new feature (you can see the details on this here: https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md#release-notes )

releasenotes/notes/add-filter-98d00f306b5689ee.yaml
```diff
@@ -0,0 +1,20 @@
...
+features:
+  - |
+    The :class:`~rustworkx.PyGraph` and the :class:`~rustworkx.PyDiGraph` classes have a new method 
+    :meth:`~rustworkx.PyGraph.filter_nodes` (or :meth:`~rustworkx.PyDiGraph.filter_nodes`). 
+    This method returns a NodeIndices object with the resulting nodes that fit some abstract criteria indicated by a filter function.
+    For example::
+      def my_filter_function(node)
+        return isinstance(node, str)
+
+      string_indices = graph.filter_nodes(my_filter_function)
```
raynelfss, line 11: Consider changing these lines as they're causing `tox -edocs` to fail:
```suggestion
    For example:

    .. jupyter-execute::
    
      from rustworkx import PyGraph

      graph = PyGraph()
      graph.add_nodes_from(list(range(5))) # Adds nodes from 0 to 5

      def my_filter_function(node):
        return isinstance(node, str)

      string_indices = graph.filter_nodes(my_filter_function)
```

releasenotes/notes/add-filter-98d00f306b5689ee.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - |
+    The :class:`~rustworkx.PyGraph` and the :class:`~rustworkx.PyDiGraph` classes have a new method 
+    :meth:`~rustworkx.PyGraph.filter_edges` (or :meth:`~rustworkx.PyDiGraph.filter_edges`). 
+    This method returns a EdgeIndices object with the resulting edges that fit some abstract criteria indicated by a filter function.
+    For example::
+      def my_filter_function(edge)
+        return isinstance(edge, str)
+
+      string_indices = graph.filter_edges(my_filter_function)
```
raynelfss, line 20: The same goes for this one, try with this block I made:

```suggestion
    For example:
    
    .. jupyter-execute::

      from rustworkx import PyGraph
      from rustworkx.generators import complete_graph

      graph = complete_graph(4)

      def my_filter_function(edge):
        return isinstance(edge, str)

      string_indices = graph.filter_edges(my_filter_function)
```

raynelfss (commented):
Danielle, your additions seem to work well with `rustworkx` on python. I have personally tested it with both `PyGraph` and `PyDiGraph`. Everything LGTM!

However, there are some changes to be made to your release notes, as they're not passing the required tests. Because of the way the release notes are generated, the examples you provided will not run unless formatted properly. Consider the examples I have left below.

releasenotes/notes/add-filter-98d00f306b5689ee.yaml
```diff
@@ -0,0 +1,43 @@
+---
+features:
+  - |
+    The :class:`~rustworkx.PyGraph` and the :class:`~rustworkx.PyDiGraph` classes have a new method 
+    :meth:`~rustworkx.PyGraph.filter_nodes` (or :meth:`~rustworkx.PyDiGraph.filter_nodes`). 
+    This method returns a NodeIndices object with the resulting nodes that fit some abstract criteria indicated by a filter function.
```
mtreinish, line 6:
```suggestion
    This method returns a :class:`~.NodeIndices` object with the resulting nodes that fit some abstract criteria indicated by a filter function.
```

releasenotes/notes/add-filter-98d00f306b5689ee.yaml
```diff
@@ -0,0 +1,43 @@
...
+        def my_filter_function(node):
+          return node > 2
+
+        indices = graph.filter_nodes(my_filter_function)
+        print(indices)
+
+  - |
+    The :class:`~rustworkx.PyGraph` and the :class:`~rustworkx.PyDiGraph` classes have a new method 
+    :meth:`~rustworkx.PyGraph.filter_edges` (or :meth:`~rustworkx.PyDiGraph.filter_edges`). 
+    This method returns a EdgeIndices object with the resulting edges that fit some abstract criteria indicated by a filter function.
```
mtreinish, line 25:
```suggestion
    This method returns a :class:`~.EdgeIndices` object with the resulting edges that fit some abstract criteria indicated by a filter function.
```

src/digraph.rs
```diff
@@ -2889,6 +2889,48 @@ impl PyDiGraph {
...
+    /// :returns: The node indices that match the filter
+    /// :rtype: NodeIndices
+    #[pyo3(text_signature = "(self, filter_function)")]
+    pub fn filter_nodes(&self, py: Python, filter_function: PyObject) -> PyResult<NodeIndices> {
+        let filter = |nindex: NodeIndex| -> PyResult<bool> {
+            let res = filter_function.call1(py, (&self.graph[nindex],))?;
+            res.extract(py)
+        };
+
+        let mut n = Vec::new();
```
mtreinish, line 2904: A potential performance improvement we can make here is to do:

```suggestion
        let mut n = Vec::with_capacity(self.graph.node_count());
```

This basically allocates up front a `Vec` with enough space for the number of nodes in the graph. THis avoids having to do multiple memory allocations as the `Vec` grows (which are slow). The tradeoff here being that we are likely over allocating memory in the return object (unless `filter_function` returns `True` for every node. I'm thinking this tradeoff is ok in this case though because each element of the output array is just a 64bit integer so except for pathological edge cases where we have millions of nodes and the filter only has a few matches nobody will really notice if we've allocated and extra space for a few integers.

But, I'm curious what your thoughts are on this tradeoff and what you think we should do here?

src/digraph.rs
```diff
@@ -2889,6 +2889,48 @@ impl PyDiGraph {
...
+    /// :returns: The edge indices that match the filter
+    /// :rtype: EdgeIndices
+    #[pyo3(text_signature = "(self, filter_function)")]
+    pub fn filter_edges(&self, py: Python, filter_function: PyObject) -> PyResult<EdgeIndices> {
+        let filter = |eindex: EdgeIndex| -> PyResult<bool> {
+            let res = filter_function.call1(py, (&self.graph[eindex],))?;
+            res.extract(py)
+        };
+
+        let mut e = Vec::new();
```
mtreinish, line 2925: Same comment as above with `filter_nodes` we can do:

```suggestion
        let mut e = Vec::with_capacity(self.graph.edge_count());
```

src/digraph.rs
```diff
@@ -2889,6 +2889,48 @@ impl PyDiGraph {
         }
     }
 
+    /// Filters a graph's nodes by some criteria conditioned on a node's data payload and returns those nodes' indices
+    ///
+    /// :param filter_function: Function with which to filter nodes
```
mtreinish, line 2894: I think this needs a little more detail on how this is used. I think we should include what the format is for the filter function (what it gets passed and the required return type). Maybe having an example in the docs here would be useful too since it might be unclear even with an English description of the behavior.

This comment applies to all the functions being added in both graph and digraph.

mtreinish (commented):
This looks great now, thanks for the fast updates adding the tests and release notes. I just had a few small comments inline mostly with some small documentation improvements we should make.

I also left a question inline about memory allocation, it's not a blocker as one way is not clearly better than the other.

src/digraph.rs
danielleodigie: ahh, this makes sense. I think over allocating is worth it in the long run, especially if reallocating is expensive. I'll make that change!

src/digraph.rs
mtreinish: I think the example goes a long way for clarity thanks for adding that. But I still think we should explicitly document the inputs and required output from `filter_function` so that it's clear to users that it will be passed a node weight and is required to return a boolean value (this applies to all the docstrings for the other 3 methods too).

mtreinish (approved):
LGTM, thanks for updating the docstring

PR #887, opened 2023-05-31
Bump pyo3 and rust numpy version to 0.19.0

mtreinish:
Pyo3 0.19.0 and rust-numpy 0.19.0 were just released. This commit updates the version used in rustworkx to these latest releases. At the same time this updates usage of text signature for classes that was deprecated in 0.19.0 release. While not fatal for normal builds this would have failed clippy in CI because we treat warnings as errors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -228,9 +228,9 @@ dependencies = [
 
 [[package]]
 name = "memoffset"
-version = "0.8.0"
+version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d61c719bcfbcf5d62b3a09efa6088de8c54bc0bfcd3ea7ae39fcc186108b8de1"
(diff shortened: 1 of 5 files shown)


IvanIsCoding:
Oh wow my stubs test is failing, let me investigate

IvanIsCoding:
I ran it locally and I can confirm that changing https://github.com/Qiskit/rustworkx/blob/main/rustworkx/iterators.pyi#L71 to `def __array__(self, _dt: Optional[np.dtype] = ...) -> np.ndarray: ...` and adding the respective import solves the test problem

mtreinish:
It must be caused by an upstream change in rust-numpy causing mypy to pick up more of the numpy api here (or something around that). I'll add the extra stuff to the stub file, as I think it'd be good to try and get this in for 0.13.0 which besides #823 I think is ready.

IvanIsCoding (approved):

PR #888, opened 2023-06-05
Fix pickle/deepcopy node hole handling

mtreinish:
This commit fixes an issue introduced by #589 where in certain cases node holes in a graph would result in a panic being raised. This was caused by a logic bug in trying to recreate the holes. Additionally, there were several places where graph methods removed nodes that the flag to indicate there were removals would no be set. This commit fixes all of these issues so that deepcopy/pickle works as expected.

### DIFF
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -383,7 +383,7 @@ impl PyDiGraph {
             let node_idx: usize = item.get_item(0).unwrap().extract().unwrap();
             let node_w = item.get_item(1).unwrap().extract().unwrap();
 
-            for _i in 0..node_idx + 1 {
+            for _i in 0..node_idx {
                 self.graph.add_node(py.None());
             }
             self.graph.add_node(node_w);
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -29,7 +29,7 @@ use rustworkx_core::dictmap::*;
 use pyo3::exceptions::PyIndexError;
 use pyo3::gc::PyVisit;
 use pyo3::prelude::*;
-use pyo3::types::{PyBool, PyDict, PyList, PyLong, PyString, PyTuple};
+use pyo3::types::{PyBool, PyDict, PyList, PyString, PyTuple};
 use pyo3::PyTraverseError;
 use pyo3::Python;
 
@@ -44,7 +44,7 @@ use petgraph::prelude::*;
 
 use petgraph::visit::{
     EdgeIndexable, GraphBase, IntoEdgeReferences, IntoNodeReferences, NodeCount, NodeFiltered,
-    Visitable,
+    NodeIndexable, Visitable,
 };
 
 use super::dot_utils::build_dot;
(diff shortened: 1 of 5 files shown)


src/digraph.rs
```diff
@@ -384,7 +383,7 @@ impl PyDiGraph {
             let node_idx: usize = item.get_item(0).unwrap().extract().unwrap();
             let node_w = item.get_item(1).unwrap().extract().unwrap();
 
-            for _i in 0..node_idx {
+            for _i in 0..node_idx + 1 {
```
mtreinish, line 386:
```suggestion
            for _i in 0..node_idx {
```

src/graph.rs
```diff
@@ -284,56 +285,24 @@ impl PyGraph {
...
             let mut tmp_nodes: Vec<NodeIndex> =
                 Vec::with_capacity(node_bound_1 + 1 - nodes_lst.len());
 
-            for i in 0..nodes_lst.len() + 1 {
-                if i < next_node_idx {
+            for item in nodes_lst {
+                let item = item.downcast::<PyTuple>().unwrap();
+                let next_index: usize = item.get_item(0).unwrap().extract().unwrap();
+                let weight: PyObject = item.get_item(1).unwrap().extract().unwrap();
+                while next_index > self.graph.node_bound() {
```
enavarro51, line 297: What's the purpose of this change? (This review may be mostly questions and same comments will apply to graph.rs and digraph.rs.)

src/graph.rs
```diff
@@ -1062,8 +1019,8 @@ impl PyGraph {
     ///     the graph
     #[pyo3(text_signature = "(self, index_list, /)")]
     pub fn remove_nodes_from(&mut self, index_list: Vec<usize>) -> PyResult<()> {
-        for node in index_list.iter().map(|x| NodeIndex::new(*x)) {
-            self.graph.remove_node(node);
+        for node in index_list {
+            self.remove_node(node)?;
```
enavarro51, line 1023: Why all the changes from `self.graph.remove_node` to `self.remove_node`?

src/graph.rs
```diff
@@ -1846,7 +1803,10 @@ impl PyGraph {
 
     fn __delitem__(&mut self, idx: usize) -> PyResult<()> {
         match self.graph.remove_node(NodeIndex::new(idx)) {
-            Some(_) => Ok(()),
+            Some(_) => {
+                self.node_removed = true;
+                Ok(())
+            }
```
enavarro51, line 1809: This seems pretty obvious. Was there some reason this wasn't done previously?

src/graph.rs
mtreinish: This is the second order issue (which the release note is about) where we weren't tracking removals correctly in several methods. Using `self.remove_node` does the tracking for us.

src/graph.rs
mtreinish: It was an oversight that's the second fix around hole handling (and what I added the release note for)

enavarro51 (approved):
Ok. LGTM. Thanks.

PR #894, opened 2023-06-07
Implement substitute_node_with_subgraph to Pygraph

### LINKED ISSUE #837: Add `PyGraph.substitute_node_with_subgraph()` method

The directed graph class `PyDiGraph` has a method `substitute_node_with_subgraph()` to replace a node in the graph with an input graph. However, the undirected graph class `PyGraph` doesn't have this method. It would be good if we filled this feature gap and add this method to the `PyGraph` class too.

raynelfss:
- Aims to fix #837
- Added method `substitute_node_with_subgraph()` to the `PyGraph`, class which replaces a node with a provided `PyGraph`.
- Based on existent method in `PyDiGraph` class.

#### TO-DO:
- Extend documentation.
- Further testing.


src/graph.rs
```diff
@@ -1594,6 +1596,133 @@ impl PyGraph {
         Ok(out_dict.into())
     }
 
+    /// Docs Pending
+    ///
```
enavarro51, line 1600: I assume the Docs here can mimic the digraph ones?

src/graph.rs
```diff
@@ -1594,6 +1596,133 @@ impl PyGraph {
...
+        node: usize,
+        other: &PyGraph,
+        edge_map_fn: PyObject,
+        node_filter: Option<PyObject>,
+        edge_weight_map: Option<PyObject>,
+    ) -> PyResult<NodeMap> {
+        let filter_fn = |obj: &PyObject, filter_fn: &Option<PyObject>| -> PyResult<bool> {
+            match filter_fn {
+                Some(filter) => {
+                    let res = filter.call1(py, (obj,)).unwrap();
```
enavarro51, line 1616: Any reason for using `unwrap` here instead of `?` like in the digraph version?

src/graph.rs
```diff
@@ -1594,6 +1596,133 @@ impl PyGraph {
...
+
+        let weight_map_fn = |obj: &PyObject, weight_fn: &Option<PyObject>| -> PyResult<PyObject> {
+            match weight_fn {
+                Some(weight_fn) => weight_fn.call1(py, (obj,)),
+                None => Ok(obj.clone_ref(py)),
+            }
+        };
+
+        let map_fn = |source: usize, target: usize, weight: &PyObject| -> PyResult<Option<usize>> {
+            let res = edge_map_fn.call1(py, (source, target, weight)).unwrap();
```
enavarro51, line 1631: Same here.

src/graph.rs
```diff
@@ -1594,6 +1596,133 @@ impl PyGraph {
...
+            in_edges.iter().map(|edge| (edge.0, edge.1)).collect();
+        // Retrieve outgoing edges. Make sure to not include any incoming edge.
+        let out_edges: Vec<(NodeIndex, NodeIndex, PyObject)> = self
+            .graph
+            .edges(node_index)
+            .filter(|edge| !in_set.contains(&(edge.target(), edge.source())))
+            .map(|edge| (edge.source(), edge.target(), edge.weight().clone_ref(py)))
+            .collect();
+        for (source, target, weight) in in_edges {
+            let old_index: Option<usize> = map_fn(source.index(), target.index(), &weight).unwrap();
```
enavarro51, line 1690: Again `?` or `unwrap`?

src/graph.rs
```diff
@@ -1594,6 +1596,133 @@ impl PyGraph {
...
+                            old_index
+                        )))
+                    }
+                },
+                None => continue,
+            };
+            self._add_edge(source, target_out, weight);
+        }
+        for (source, target, weight) in out_edges {
+            let old_index: Option<usize> = map_fn(source.index(), target.index(), &weight).unwrap();
```
enavarro51, line 1706: Again `?` or `unwrap`?

src/graph.rs
```diff
@@ -1594,6 +1596,133 @@ impl PyGraph {
...
+            let source_out: NodeIndex = match old_index {
+                Some(old_index) => match out_map.get(&old_index) {
+                    Some(new_index) => NodeIndex::new(*new_index),
+                    None => {
+                        return Err(PyIndexError::new_err(format!(
+                            "No matter index {} found",
```
enavarro51, line 1712:
```suggestion
                            "No mapped index {} found",
```

src/graph.rs
```diff
@@ -1594,6 +1596,133 @@ impl PyGraph {
...
+            .map(|edge| (edge.source(), edge.target(), edge.weight().clone_ref(py)))
+            .collect();
+        for (source, target, weight) in in_edges {
+            let old_index: Option<usize> = map_fn(source.index(), target.index(), &weight).unwrap();
+            let target_out: NodeIndex = match old_index {
+                Some(old_index) => match out_map.get(&old_index) {
+                    Some(new_index) => NodeIndex::new(*new_index),
+                    None => {
+                        return Err(PyIndexError::new_err(format!(
+                            "No matter index {} found",
```
enavarro51, line 1696:
```suggestion
                            "No mapped index {} found",
```

enavarro51 (changes requested):
Thanks for submitting this. It's looking good. A few in-line comments and if you add the docs, you can un-draft it.

src/graph.rs
raynelfss: No particular reason. Rust was just complaining about using them without an extra import. I guess it was a local problem because it works well now. I will revert back to using the `?`. Thank you for reviewing.

enavarro51 (approved):
LGTM. Thanks for the contribution.

PR #896, opened 2023-06-08
Apply wheel job fixes from 0.13.0 release

mtreinish:
This commit applies the various fixes needed to the wheel publishing job definitions for the 0.13.0 release. There were several jon errors during the release which were caused by various changes made to rustworkx, upstream dependencies, and CI environment since the 0.12.0 which caused issues during the release process. #753 should still be finished to simplify the job definitions, but that should be rebased to take this more targeted fix. The intent is for this to be a minimal diff for backporting to stable/0.13 for a future 0.13.1 release.

Of particular importance here though is the change in support tier for s390x from 3 to 4. This was caused by repeated issues with timeouts caused by running tests during the s390x linux wheel builds. To ensure we can reliably build the wheels this drops the testing from the s390x so that they can reliably complete in 12 job hours.

### DIFF
--- a/releasenotes/notes/s390x-tier-4-1701a0f044759cd1.yaml
+++ b/releasenotes/notes/s390x-tier-4-1701a0f044759cd1.yaml
@@ -5,7 +5,7 @@ upgrade:
     documented in :ref:`platform-suppport`). This is a result of no longer being
     able to run tests during the pre-compiled wheel publishing jobs due to
     constraints in the available CI infrastructure. There hopefully shouldn't 
-    be any meaningful impact resulting from this change, but as we're no longer tests being
+    be any meaningful impact resulting from this change, but as there are no longer tests being
     run to validate the binaries prior to publishing them there are no longer
     guarantees that the wheels for s390x are fully functional (although the
     likelihood they are is still high as it works on other platforms). If any
--- a/releasenotes/notes/s390x-tier-4-1701a0f044759cd1.yaml
+++ b/releasenotes/notes/s390x-tier-4-1701a0f044759cd1.yaml
@@ -1,0 +1,12 @@
+---
+upgrade:
+  - |
+    Support for the Linux s390x platform has changed from tier 3 to tier 4 (as
+    documented in :ref:`platform-suppport`). This is a result of no longer being
+    able to run tests during the pre-compiled wheel publishing jobs due to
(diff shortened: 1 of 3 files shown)


releasenotes/notes/s390x-tier-4-1701a0f044759cd1.yaml
```diff
@@ -0,0 +1,12 @@
+---
+upgrade:
+  - |
+    Support for the Linux s390x platform has changed from tier 3 to tier 4 (as
+    documented in :ref:`platform-suppport`). This is a result of no longer being
+    able to run tests during the pre-compiled wheel publishing jobs due to
+    constraints in the available CI infrastructure. There hopefully shouldn't 
+    be a change resulting from this change, but as we're no longer tests being
```
mtreinish, line 8:
```suggestion
    be any meaningful impact resulting from this change, but as we're no longer tests being
```

releasenotes/notes/s390x-tier-4-1701a0f044759cd1.yaml
```diff
@@ -0,0 +1,12 @@
+---
+upgrade:
+  - |
+    Support for the Linux s390x platform has changed from tier 3 to tier 4 (as
+    documented in :ref:`platform-suppport`). This is a result of no longer being
+    able to run tests during the pre-compiled wheel publishing jobs due to
+    constraints in the available CI infrastructure. There hopefully shouldn't 
+    be any meaningful impact resulting from this change, but as we're no longer tests being
```
enavarro51, line 8:
```suggestion
    be any meaningful impact resulting from this change, but as there are no longer tests being
```

enavarro51 (changes requested):
Just one in-line change. Also there seem to be some Mac changes. Are these required for this, or is this just some updating that needs to be done?

mtreinish:
The two mac updates I made are to update the arm64 wheel jobs to use the `macos-latest` job image. During the 0.13.0 release this job failed because the `macos-10.15` image it was trying to use no longer exits so it was waiting forever to get a VM allocated. The second update was just a small oversight in the supported platform docs, we have support for arm64 macOS and that just wasn't mentioned in the intro to supported platforms. I changed it in this PR because I already had to update to the docs to downgrade the s390x support tier to 4.

enavarro51 (approved):

enavarro51:
LGTM. Thanks.

PR #897, opened 2023-06-09
Add todo_nodes sort to maintain deterministic output in token_swapper

enavarro51:
This PR adds a sort to the `todo_nodes` Vec in `token_swapper` in order to maintain deterministic output when using a seed. This problem appeared when implementing https://github.com/Qiskit/qiskit-terra/pull/10001 and some of the tests produced correct, but random output.


mtreinish (approved):
This LGTM, thanks for the quick fix

PR #899, opened 2023-06-13
Adding Option to change parallel edge behavior in adjacency_matrix functions

### LINKED ISSUE #413: Add option to adjacency matrix functions to adjust parallel edge behavior

Right now the `adjacency_matrix()` function unconditionally sums the weights for parallel edges in the graph. We should make this configurable so that we can do different operations. Off the top of my head, sum(), min(), max(), avg() are all good candidates.

For maximum flexibility we could just make this a python callback function that get's passed both weights and returns the f64 to use (which is what I think networkx does). However that will have much worse performance than using a f64 operation in rust.

danielleodigie:
- Aims to resolve #413 
- This pull request adds the ability to change the behavior of parallel edges when creating the adjacency matrix.
- Previously, these functions resorted to summing the weights of the edges
- Now, while the functions still default to summing the weights, one can choose to take the minimum, maximum, or average of the weights


src/connectivity/mod.rs
```diff
@@ -344,31 +367,60 @@ pub fn digraph_adjacency_matrix(
...
+                *parallel_edge_count.get_mut(&[i, j]).unwrap() += 1;
+            } else {
+                parallel_edge_count.insert([i, j], 2);
+                matrix[[i, j]] = (matrix[[i, j]] + edge_weight) / 2.0;
+                matrix[[j, i]] = (matrix[[j, i]] + edge_weight) / 2.0;
+            }
+        } else {
+            return Err(PyValueError::new_err("Parallel edges can currently only be dealt with using \"sum\", \"min\", \"max\", or \"avg\"."));
         }
     }
```
raynelfss, line 425: You could also use `match` here:
```suggestion
        } else {
            match parallel_edge {
                "sum" => {
                    matrix[[i, j]] += edge_weight;
                    matrix[[j, i]] += edge_weight;
                }
                "min" => {
                    let weight_min = matrix[[i, j]].min(edge_weight);
                    matrix[[i, j]] = weight_min;
                    matrix[[j, i]] = weight_min;
                }
                "max" => {
                    let weight_max = matrix[[i, j]].max(edge_weight);
                    matrix[[i, j]] = weight_max;
                    matrix[[j, i]] = weight_max;
                }
                "avg" => {
                    if parallel_edge_count.contains_key(&[i, j]) {
                        matrix[[i, j]] = (matrix[[i, j]] * parallel_edge_count[&[i, j]] as f64
                            + edge_weight)
                            / ((parallel_edge_count[&[i, j]] + 1) as f64);
                        matrix[[j, i]] = (matrix[[j, i]] * parallel_edge_count[&[i, j]] as f64
                            + edge_weight)
                            / ((parallel_edge_count[&[i, j]] + 1) as f64);
                        *parallel_edge_count.get_mut(&[i, j]).unwrap() += 1;
                    } else {
                        parallel_edge_count.insert([i, j], 2);
                        matrix[[i, j]] = (matrix[[i, j]] + edge_weight) / 2.0;
                        matrix[[j, i]] = (matrix[[j, i]] + edge_weight) / 2.0;
                    }
                }
                _ => {
                    return Err(PyValueError::new_err("Parallel edges can currently only be dealt with using \"sum\", \"min\", \"max\", or \"avg\"."));
                }
            }
        }
    }

raynelfss (approved):
Everything seems to be working well. I just have a couple of code suggestions. But outside of that, LGTM!

enavarro51:
Thanks for working on this. You should also add 2 tests, 1 for graph and 1 for digraph, to assert that each of the options works correctly.

enavarro51 (approved):
LGTM. Thanks for working on this.

Issue #854, opened 2023-04-04
Add edge coloring function


ihincks:
### What is the expected enhancement?

Edge coloring has natural applications in quantum, because, for example, given a coupling map, the edge colors of an edge coloring can be interpreted as gates which are able to be applied simultaneously.

Vizing's theorem (for simple connected graphs) says that you can always edge color a graph in deg(G) or deg(G)+1 colors, where deg(G) is the maximum degree of any vertex. Deciding which of these two cases a particular graph falls into is NP-complete. However, near-optimal polynomial time algorithms exist that use no more than deg(G)+1 colors, notably [Misra-Gries edge-coloring](https://en.wikipedia.org/wiki/Misra_%26_Gries_edge_coloring_algorithm).

Here is my Python implementation:

```python
from typing import Set, FrozenSet, Dict, Iterable, Tuple, List
from collections import defaultdict
from itertools import count

Node = int
Edge = FrozenSet[Node]
Color = int

edge = lambda u, v: frozenset((u, v))

def misra_gries(edges: Set[Edge]) -> Dict[Edge, Color]:
    r"""Return an edge-coloring of a simple connected graph."""
    colors = {}
    adjacencies = defaultdict(set)
    for u, v in edges:
        adjacencies[u].add(v)
        adjacencies[v].add(u)

    def adjacent_colors(u: Node) -> Iterable[Tuple[Node, Color]]:
        # yield all colored nodes adjacent to u, along with their color
        for n in adjacencies[u]:
            color = colors.get(edge(u, n), None)
            if color is not None:
                yield n, color

    def is_free(c: Color, u: Node) -> bool:
        # return whether node u touches an edge of color c
        return all(color != c for _, color in adjacent_colors(u))

    def free_color(u: Node) -> Color:
        # return the first free color of node u
        used_colors = {color for _, color in adjacent_colors(u)}
        for color in count():
            if color not in used_colors:
                return color
                
    def maximal_fan(u: Node, v: Node) -> List[Node]:
        # return a maximal on {u, v} at u
        fan = [v]
        remaining = set(adjacent_colors(u))
        while remaining:
            for n, c in remaining:
                if is_free(c, fan[-1]):
                    fan.append(n)
                    remaining.discard((n,c))
                    break
            else:
                break
        return fan
    
    def cdu_path(c: Color, d: Color, u: Node) -> Set[Edge]:
        # return a cd_u path
        path = set()
        endpoints = {u}
        while endpoints:
[...]

PR #902, opened 2023-06-15
Add Misra-Gries edge coloring method

alexanderivrii:
This is an attempt to add the Misra-Gries edge coloring algorithm to Rustworkx, following @ihincks's python code in #854 and wiki page https://en.wikipedia.org/wiki/Misra_%26_Gries_edge_coloring_algorithm. This is still raw, but I would be happy to get some initial feedback on which parts of the code could be implemented in more rust-friendly way, whether the names and inputs/outputs of different functions look ok, etc.


src/coloring.rs
```diff
@@ -59,3 +67,209 @@ pub fn graph_greedy_color(py: Python, graph: &graph::PyGraph) -> PyResult<PyObje
...
 }
+
+struct MisraGriesAlgorithm<'a> {
+    graph: &'a StablePyGraph<Undirected>,
+    colors: DictMap<EdgeIndex, usize>,
+}
+
+impl<'a> MisraGriesAlgorithm<'a> {
+    pub fn new(graph: &'a StablePyGraph<Undirected>) -> Self {
+        let colors: DictMap<EdgeIndex, usize> = DictMap::new();
```
mtreinish, line 78: I'd probably use `with_capacity` here instead of `new` since we know this will be `graph.edge_count()`. This will save us multiple memory allocations that would be made otherwise while it grows.

src/coloring.rs
```diff
@@ -59,3 +67,209 @@ pub fn graph_greedy_color(py: Python, graph: &graph::PyGraph) -> PyResult<PyObje
...
+        MisraGriesAlgorithm { graph, colors }
+    }
+
+    // Computes colors used at node u
+    fn get_used_colors(&self, u: NodeIndex) -> Vec<usize> {
```
mtreinish, line 83: I think you'd want to use a `HashSet` for this since `Vec<T>.contains()` will have an `O(n)` lookup.

src/coloring.rs
```diff
@@ -59,3 +67,209 @@ pub fn graph_greedy_color(py: Python, graph: &graph::PyGraph) -> PyResult<PyObje
...
+        let mut used_colors: Vec<usize> = Vec::new();
+        for edge in self.graph.edges(u) {
+            if let Some(color) = self.colors.get(&edge.id()) {
+                used_colors.push(*color);
+            }
+        }
+        used_colors
```
mtreinish, line 90: You can actually do this with a single iterator which will be more efficient (mostly because the compiler has more context around the size of things so it can handle allocations better):

```suggestion
        self.graph.edges(u)
            .filter_map(|edge| {
                self.colors.get(&edge.id())
            })
            .copied()
            .collect()
```

(I didn't test this locally so it might not compile, I was mostly guessing from memory about the `copied()` because you'll have a `&usize` for color.

src/coloring.rs
```diff
@@ -59,3 +67,209 @@ pub fn graph_greedy_color(py: Python, graph: &graph::PyGraph) -> PyResult<PyObje
...
+    }
+
+    // Returns the smallest free (aka unused) color at node u
+    fn get_free_color(&self, u: NodeIndex) -> usize {
+        let used_colors = self.get_used_colors(u);
+        let mut c: usize = 0;
+        while used_colors.contains(&c) {
+            c += 1;
+        }
+        c
```
mtreinish, line 100: Just for fun you could do this as an iterator too, but not sure if is any faster. Something like:

```suggestion
        let used_colors = self.get_used_colors(u);
        (0..)
            .position(|color| used_colors.contains(&color))
            .unwrap()
```

src/coloring.rs
```diff
@@ -59,3 +67,209 @@ pub fn graph_greedy_color(py: Python, graph: &graph::PyGraph) -> PyResult<PyObje
...
+        u: NodeIndex,
+        v: NodeIndex,
+    ) -> Vec<(EdgeIndex, NodeIndex)> {
+        let mut fan: Vec<(EdgeIndex, NodeIndex)> = Vec::new();
+        fan.push((ee, v));
+
+        let mut neighbors: Vec<(EdgeIndex, NodeIndex)> = Vec::new();
+        for edge in self.graph.edges(u) {
+            neighbors.push((edge.id(), edge.target()));
+        }
```
mtreinish, line 122: I know this is a theme now :) But I would use an iterator to construct this just so it can be done in one allocation. Something like:

```suggestion
        let mut neighbors: Vec<(EdgeIndex, NodeIndex)> = self.graph.edges(u)
            .map(|edge| (edge.id(), edge.target()))
            .collect();
```

src/coloring.rs
```diff
@@ -59,3 +67,209 @@ pub fn graph_greedy_color(py: Python, graph: &graph::PyGraph) -> PyResult<PyObje
...
+        while fan_extended {
+            fan_extended = false;
+
+            for (edge_index, z) in &neighbors {
+                if let Some(color) = self.colors.get(edge_index) {
+                    if self.is_free_color(last_node, *color) {
+                        fan_extended = true;
+                        last_node = *z;
+                        fan.push((*edge_index, *z));
+                        let position_z = neighbors.iter().position(|x| x.1 == *z).unwrap();
```
mtreinish, line 138: Instead of doing this it'd probably be more efficient to use `neighbors.iter().enumerate()` for the for loop. Then we'd have the index for every step and `position_z` would have no lookup overhead. Right now this is O(n) for this lookup as you have to re-traverse `neighbors` on each iteration.

src/coloring.rs
```diff
@@ -59,3 +67,209 @@ pub fn graph_greedy_color(py: Python, graph: &graph::PyGraph) -> PyResult<PyObje
...
+                        neighbors.remove(position_z);
```
mtreinish, line 139: I'm actually surprised the borrow checker isn't unhappy here because you're mutating `neighbors` while the for loop has an immutable reference to it.

src/coloring.rs
```diff
@@ -59,3 +67,209 @@ pub fn graph_greedy_color(py: Python, graph: &graph::PyGraph) -> PyResult<PyObje
...
+            for i in 1..w + 1 {
+                let next_color = self.colors.get(&(fan[i].0)).unwrap();
+                self.colors.insert(fan[i - 1].0, *next_color);
+            }
+
+            // fill additional color
+            self.colors.insert(fan[w].0, d);
+        }
+
+        self.check_coloring();
```
mtreinish, line 257: Do you plan to expose this to python users or is it just for testing? If you think python users will want this we can add a flag to the pyfunction and put this behind an if statement. If you think it's only good for testing we can put this behind a `debug_assert!()` (https://doc.rust-lang.org/std/macro.debug_assert.html) or something so it only is run when the package is built in debug mode.

mtreinish (commented):
I left a few quick initial comments from a first scan through the PR. The other question I have is whether you want to add this to rustworkx-core or not? I think it's probably generally useful enough (and you did just add the node coloring function to the core crate) that it might make sense to put this in rustworkx-core from the start.

alexanderivrii:
@mtreinish, thanks for the detailed feedback. I guess putting the functionality in rustworkx-core makes sense. And I really like the iterator-based way of implementing things.

src/coloring.rs
alexanderivrii: Nice, `filter_map` is indeed a useful thing to know about. Yes, I do apparently need `copied()` here (otherwise rust indeed complains about the mismatch between `size` and `&size`).

src/coloring.rs
alexanderivrii: This was only used for testing. I have moved this function under `#[cfg(test)]`.

src/coloring.rs
alexanderivrii: `neighbors` is mutable now

alexanderivrii:
I have made another round of changes, moving the new edge coloring function to rustworkx-core and changing the `colors` to be a vector indexed by `usize` (the internal value of `EdgeIndex`, which we can get using `EdgeIndexable`). Using a vector rather than a hash map leads to about 20% improvement in performance (which is not surprising). I don't understand the cause for CI failures, @mtreinish could you please take a look?

mtreinish:
I think the ci failure was a temporary issue related to the release of a new version of something. I just updated the pr branch and it seemed to have work fine.

rustworkx-core/src/coloring.rs
```diff
@@ -220,8 +221,218 @@ where
...
+            .edges(u)
+            .filter_map(|edge| self.colors[EdgeIndexable::to_index(&self.graph, edge.id())])
+            .collect();
+        used_colors
+    }
+
+    // Returns the smallest free (aka unused) color at node u
+    fn get_free_color(&self, u: G::NodeId) -> usize {
+        let used_colors = self.get_used_colors(u);
+        let free_color: usize = (0..)
```
mtreinish, line 253: Just to avoid unbound iteration as a potential source of bugs:

```suggestion
        let free_color: usize = (0..self.colors.len())
```

rustworkx-core/src/coloring.rs
```diff
@@ -220,8 +221,218 @@ where
...
+            .position(|color| !used_colors.contains(&color))
+            .unwrap();
+        free_color
+    }
+
+    // Returns true iff color c is free at node u
+    fn is_free_color(&self, u: G::NodeId, c: usize) -> bool {
+        let used_colors = self.get_used_colors(u);
+        !used_colors.contains(&c)
```
mtreinish, line 262: Could you just do `self.colors.contains(&Some(c))` (or something like this, I might have got the position of `&` incorrect)? You'd have the same linear overhead. 

More generally though I'm wondering if it makes more sense to store a `HashSet` of used colors in the `MisraGries` struct so you don't have to generate it on the fly every time. Then you could make this a O(1) lookup every time. It would require doing book keeping on adding or changing colors but it should be manageable.

rustworkx-core/src/coloring.rs
```diff
@@ -220,8 +221,218 @@ where
...
+
+    // Returns the maximal fan on edge (u, v) at u
+    fn get_maximal_fan(
+        &self,
+        eid: G::EdgeId,
+        u: G::NodeId,
+        v: G::NodeId,
+    ) -> Vec<(G::EdgeId, G::NodeId)> {
+        let mut fan: Vec<(G::EdgeId, G::NodeId)> = Vec::new();
+        fan.push((eid, v));
```
mtreinish, line 273:
```suggestion
        let mut fan: Vec<(G::EdgeId, G::NodeId)> = vec![(eid, v)];
```

rustworkx-core/src/coloring.rs
```diff
@@ -220,8 +221,218 @@ where
...
+            d
+        } else {
+            c
+        }
+    }
+
+    // Returns the longest path starting at node u with alternating colors c, d, c, d, c, etc.
+    fn get_cdu_path(&self, u: G::NodeId, c: usize, d: usize) -> Vec<(G::EdgeId, usize)> {
+        let mut path: Vec<(G::EdgeId, usize)> = Vec::new();
+        let mut cur_node: G::NodeId = u;
```
mtreinish, line 316: The compiler should be able to infer the type here because it's defined on `u`.

```suggestion
        let mut cur_node = u;
```

mtreinish:
Oh, the other thing is the new function needs to be added to the api docs index here: https://github.com/Qiskit/rustworkx/blob/main/docs/source/api/algorithm_functions/other.rst (although maybe we should add a graph coloring page at this point)

rustworkx-core/src/coloring.rs
alexanderivrii: Fixed in [4e22ad7](https://github.com/Qiskit/rustworkx/pull/902/commits/4e22ad7abdb6c27eac28c0ea88927d94ea50eb93) using the stricter upper bound of `max_node_degree + 1`.

rustworkx-core/src/coloring.rs
alexanderivrii: I have followed the above suggestion in [4e22ad7](https://github.com/Qiskit/rustworkx/pull/902/commits/4e22ad7abdb6c27eac28c0ea88927d94ea50eb93), and indeed it gives a significant speedup on medium/large-sized complete graphs. E.g., on my laptop previously it took `16.95` seconds to color a complete graph over 90 vertices, and now it takes `0.58` seconds. For heavy-hex graphs the difference is less pronounced, but the new implementation is still a bit faster.

rustworkx-core/src/coloring.rs
alexanderivrii: Done in [7900916](https://github.com/Qiskit/rustworkx/pull/902/commits/7900916402fb101ad39724e2d0509619ce10907a). Also the code has been updated to use `G::EdgeRef` instead of `G::EdgeId`, so it's enough to only keep the edge reference.

rustworkx-core/src/coloring.rs
alexanderivrii: Removed in [7e6470d](https://github.com/Qiskit/rustworkx/pull/902/commits/7e6470dc628dc7576f52666e71ce313e362b7e77).

alexanderivrii:
> I am a bit worried about the complexity of some of the inner loops in the helper methods as there is a lot of nested looping and those all get called in an outer loop from run_algorithm().

Agreed, it would be great if we could remove the extra loop in the `get_maximal_fan` function (for example by precomputing something ahead of type). Essentially, given a node `u` and its neighbor node `v` we try to "extend" the fan by finding an edge `(u, w)` such that its color is not used at `v`. If such an edge is found, we add it to the fan, and repeat the above process with `w` instead of `v`. The problem is that the node `v` keeps changing and there are multiple colors that can be unused at a given node. So I don't see a simple way to remove one of the two loops. But would be happy to hear suggestions.

mtreinish:
Yeah, I don't have any concrete suggestions off the top of my head yet either. It was more just my impressions as I was reviewing it. But like I said we can leave it like this and if performance becomes an issue or one of us comes up with something it's easy enough to change and improve the internal logic later.

mtreinish (approved):
LGTM, thanks for doing this.

PR #905, opened 2023-06-19
Pin Python version in CI

mtreinish:
Recently github updated the cached version of python installed for 3.7 in the CI environment to 3.7.17. This new binary was not built with bz2 support which is causing failures in our CI jobs that run with 3.7. While we'll be dropping 3.7 support from the main branch (for 0.14.0) in the near future we still support 3.7 on the stable 0.13.x series. This commit pins the python version to the previous patch release which was known to work.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -82,13 +82,13 @@ jobs:
         with:
           python-version: 3.7
           architecture: ${{ matrix.platform.python-architecture }}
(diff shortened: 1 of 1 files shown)


IvanIsCoding (approved):

IvanIsCoding:
LGTM, we should probably also file a bug somewhere for the maintainers of the action to track it

PR #907, opened 2023-06-20
Drop support for python 3.7 for 0.14.0

raynelfss:
As of `0.14.0`, Python 3.7 is no longer supported. This PR aims to prepare the CI and the rustworkx package setup to drop usage of Python 3.7 and work only with Python 3.8+.

### DIFF
--- a/releasenotes/notes/drop-python-3.7-c71f8b6559beaf86.yaml
+++ b/releasenotes/notes/drop-python-3.7-c71f8b6559beaf86.yaml
@@ -3,6 +3,3 @@ upgrade:
   - |
     The minimum required Python version was raised to Python 3.8.
     To use rustworkx, please ensure you are using Python >= 3.8.
-deprecations:
-  - |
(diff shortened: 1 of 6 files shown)


releasenotes/notes/drop-python-3.7-c71f8b6559beaf86.yaml
```diff
@@ -0,0 +1,8 @@
+---
+upgrade:
+  - |
+    The minimum required Python version was raised to Python 3.8.
+    To use rustworkx, please ensure you are using Python >= 3.8.
+deprecations:
+  - |
+    Support for Python 3.7 has been dropped as of 0.14.0.
```
mtreinish, line 8: The upgrade release note you wrote covers this, we don't need a second one here:

```suggestion
```

mtreinish (approved):
LGTM, thanks for doing this. Just one line suggestion on the release note which I'll apply and then enqueue this for merging.

Issue #893, opened 2023-06-07
Type hints in 0.13.0 incompatible with Python 3.7


mtreinish:
### Information

- **rustworkx version**: 0.13.0
- **Python version**: 3.7
- **Rust version**: N/A
- **Operating system**: Linux

### What is the current behavior?

Running mypy with Python 3.7 in a package that uses rustworkx fails trying to parse the stubs files. For example:

```
/opt/hostedtoolcache/Python/3.7.16/x64/lib/python3.7/site-packages/rustworkx/graph.pyi:34: error: invalid syntax  [syntax]
```
See: https://github.com/qiskit-community/qiskit-nature-pyscf/actions/runs/5200238443/jobs/9378913497?pr=35#step:7:15 for where this happened in CI.

### What is the expected behavior?

The type hint syntax used is compatible with the rustworkx python version

### Steps to reproduce the problem

Run mypy using rustworkx on python 3.7

IvanIsCoding:
I will try to fix it next week, but it looks it is a problem with `__init__`

IvanIsCoding:
It turns out that the error is due to [PEP 570](https://peps.python.org/pep-0570/). So this would be a big rewrite which acutally hurts Python 3.8+ users

mtreinish:
Hmm, then I think considering Python 3.7 is going eol; for 0.13.1 let's just write a known issue release note about the lack of type hint compatibility with python 3.7

IvanIsCoding:
> Hmm, then I think considering Python 3.7 is going eol; for 0.13.1 let's just write a known issue release note about the lack of type hint compatibility with python 3.7

If we could just purge the .pyi files from the 3.7 release and keep it in the other ones that would be the best scenario

IvanIsCoding:
I think we should backport #907 to the main branch, release 0.13.1 for all versions except 3.7. And then have another branch for 3.7 but with a modified MANIFEST.in exluding https://github.com/Qiskit/rustworkx/blob/main/MANIFEST.in#L7 which remove the stub files from Python 3.7

mtreinish:
I don't think it's really worth the trouble. Given that Python 3.7 is already EoL (since 06/27/2023) I think it's probably sufficient to just document the type hints are incompatible with Python 3.7 as it doesn't effect the actual execution of the library on Python 3.7 it's just for using mypy to validate against the stub files. I'm preparing the release notes for the 0.13.1 release right now. I'll include this in that PR.

[...]

PR #909, opened 2023-06-21
Try to fix mypy errors for Python 3.7

IvanIsCoding:
Attempt to solve #893. Because the hints work with 3.8+ I suspect it is more a matter of using new syntax than a matter of being correct


IvanIsCoding:
Closing this because unfortunately it does not solve the problem. The error is related to [PEP 570](https://peps.python.org/pep-0570/)

PR #912, opened 2023-06-24
Add assertions to stub files for Python 3.8+

IvanIsCoding:
New attempt at fixing #893 by removing support for Python 3.7 typehints

[Based on their Python version and platform check](https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks)


IvanIsCoding:
Unfortunately this yield:

```
env/lib64/python3.7/site-packages/rustworkx/graph.pyi:39: error: invalid syntax; you likely need to run mypy using Python 3.11 or newer  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```

It is very hard to skip a syntax error in mypy, the solution needs to remove the file in the 3.7 build

PR #923, opened 2023-07-06
Adding Transitive Reduction Function

### LINKED ISSUE #801: Transitive reduction

A "transitive reduction" algorithm. 

Wikipedia: https://en.wikipedia.org/wiki/Transitive_reduction

Functionality should be similar to networkx: https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.dag.transitive_reduction.html

danielleodigie:
Aims to fix #801 
Adds transitive_reduction function to DAG algorithms module


src/dag_algo/mod.rs
```diff
@@ -637,3 +639,74 @@ pub fn collect_bicolor_runs(
...
+                    descendants.insert(v, dfs);
+                }
+                for desc in &descendants[&v] {
+                    u_nbrs.remove(&NodeIndex::new(desc.1));
+                }
+            }
+            *check_count.get_mut(&v).unwrap() -= 1;
+            if check_count[&v] == 0 {
+                descendants.remove(&v);
+            }
```
raynelfss, line 691: If you want to avoid calling `StableGraph<...>.neighbors()` twice, you could instead make a copy of `u_nbrs` by calling `.clone()` on the loop. You could also pass a reference to said copy as the iterator in the loop, since in most cases you require a reference anyway.

```suggestion
        for v in &u_nbrs.clone() {
            if u_nbrs.contains(v) {
                if !descendants.contains_key(v) {
                    let dfs = dfs_edges(&g, Some(*v));
                    descendants.insert(*v, dfs);
                }
                for desc in &descendants[v] {
                    u_nbrs.remove(&NodeIndex::new(desc.1));
                }
            }
            *check_count.get_mut(v).unwrap() -= 1;
            if check_count[v] == 0 {
                descendants.remove(v);
            }
```

But it's all a question of choice. The code honestly works well regardless.

tests/rustworkx_tests/digraph/test_transitive_reduction.py
raynelfss, line 1: You might want to add more tests, just to make sure all bases are covered.

raynelfss (approved):
Everything here looks great to me. I left a couple comments that are up to you to consider. Great Job!

src/dag_algo/mod.rs
mtreinish: In my mind I would try to avoid a call to clone `clone()` because the extra memory allocation and copying is typically slower. But it's totally worth benchmarking to see which ends up being faster.

src/dag_algo/mod.rs
```diff
@@ -637,3 +639,74 @@ pub fn collect_bicolor_runs(
...
+    let mut tr = StablePyGraph::<Directed>::new();
+    let mut descendants = HashMap::new();
+    let mut check_count = HashMap::new();
+    for (i, node) in g.node_indices().enumerate() {
+        tr.add_node(graph.get_node_data(i).unwrap().clone_ref(py));
+        check_count.insert(node, graph.in_degree(i));
+    }
+
+    for u in g.node_indices() {
+        let mut u_nbrs: HashSet<NodeIndex> = g.neighbors(u).collect();
```
IvanIsCoding, line 677:
```suggestion
        let mut u_nbrs: IndexSet<NodeIndex> = g.neighbors(u).collect();
```

src/dag_algo/mod.rs
```diff
@@ -637,3 +639,74 @@ pub fn collect_bicolor_runs(
...
+    py: Python,
+) -> PyResult<digraph::PyDiGraph> {
+    let g = &graph.graph;
+    if !is_directed_acyclic_graph(graph) {
+        return Err(PyValueError::new_err(
+            "Directed Acyclic Graph required for transitive_reduction",
+        ));
+    }
+    let mut tr = StablePyGraph::<Directed>::new();
+    let mut descendants = HashMap::new();
```
IvanIsCoding, line 669:
```suggestion
    let mut descendants = DictMap::new();
```

src/dag_algo/mod.rs
```diff
@@ -637,3 +639,74 @@ pub fn collect_bicolor_runs(
...
+    let mut check_count = HashMap::new();
```
IvanIsCoding, line 670:
```suggestion
    let mut check_count = DictMap::new();
```

src/dag_algo/mod.rs
```diff
@@ -637,3 +639,74 @@ pub fn collect_bicolor_runs(
...
+                graph
+                    .get_edge_data(u.index(), v.index())
+                    .unwrap()
+                    .clone_ref(py),
+            );
+        }
+    }
+    Ok(digraph::PyDiGraph {
+        graph: tr,
+        node_removed: false,
```
IvanIsCoding, line 706: This might not be true if you call it in a graph with removed noves

IvanIsCoding (commented):
The algorithm LGTM, we just need to address two things:
* replace HashMap and HashSet with [`DictMap`](https://docs.rs/rustworkx-core/latest/rustworkx_core/dictmap/type.DictMap.html) and [`IndexSet`](https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html). It can be confusing to Python users why two calls to the same function yield different results. DictMap helps keep our results more deterministic
* we need more unit test, some worthwhile cases include graphs with deleted nodes

src/dag_algo/mod.rs
```diff
@@ -637,3 +642,86 @@ pub fn collect_bicolor_runs(
...
+/// Returns the transitive reduction of a directed acyclic graph
+///
+/// The transitive reduction of G = (V,E) is a graph G- = (V,E-)
+/// such that for all v,w in V there is an edge (v,w) in E- if and only if (v,w) is in E
+/// and there is no path from v to w in G with length greater than 1.
+///
+/// :param PyDiGraph graph: A directed acyclic graph
+///
+/// :returns: a directed acyclic graph representing the transitive reduction, and a map containing the index of a node in the original graph mapped to its equivalent in the resulting graph.
+/// :rtype: Tuple[:class:`~rustworkx.PyGraph`, dict]
```
mtreinish, line 655: I think sphinx is smart enough to resolve this from `rtype`:

```suggestion
/// :rtype: Tuple[PyGraph, dict]
```

src/dag_algo/mod.rs
```diff
@@ -637,3 +642,86 @@ pub fn collect_bicolor_runs(
 
     Ok(block_list)
 }
+
+/// Returns the transitive reduction of a directed acyclic graph
+///
+/// The transitive reduction of G = (V,E) is a graph G- = (V,E-)
```
mtreinish, line 648:
```suggestion
/// The transitive reduction of :math:`G = (V, E)` is a graph :math:`G\prime = (V, E\prime)`
```

src/dag_algo/mod.rs
```diff
@@ -637,3 +642,86 @@ pub fn collect_bicolor_runs(
...
+/// such that for all v,w in V there is an edge (v,w) in E- if and only if (v,w) is in E
```
mtreinish, line 649:
```suggestion
/// such that for all :math:`v` and :math:`w` in :math:`V` there is an edge :math:`(v, w)` in
/// :math:`E\prime` if and only if :math:`(v, w)` is in :math:`E`
```

src/dag_algo/mod.rs
```diff
@@ -637,3 +642,86 @@ pub fn collect_bicolor_runs(
...
+/// and there is no path from v to w in G with length greater than 1.
```
mtreinish, line 650:
```suggestion
/// and there is no path from :math:`v` to :math:`w` in :math:`G` with length greater than 1.
```

src/dag_algo/mod.rs
```diff
@@ -637,3 +642,86 @@ pub fn collect_bicolor_runs(
...
+///
+/// :param PyDiGraph graph: A directed acyclic graph
+///
+/// :returns: a directed acyclic graph representing the transitive reduction, and a map containing the index of a node in the original graph mapped to its equivalent in the resulting graph.
+/// :rtype: Tuple[:class:`~rustworkx.PyGraph`, dict]
+///
+/// :raises PyValueError: if graph is not a DAG
```
mtreinish, line 657:
```suggestion
/// :raises PyValueError: if ``graph`` is not a DAG
```

src/dag_algo/mod.rs
```diff
@@ -637,3 +642,86 @@ pub fn collect_bicolor_runs(
...
+    let g = &graph.graph;
+    let mut index_map = DictMap::new();
+    if !is_directed_acyclic_graph(graph) {
+        return Err(PyValueError::new_err(
+            "Directed Acyclic Graph required for transitive_reduction",
+        ));
+    }
+    let mut tr = StablePyGraph::<Directed>::new();
+    let mut descendants = DictMap::new();
+    let mut check_count = DictMap::new();
```
mtreinish, line 674: I'd probably use `HashMap` for this, because `DictMap` is only advantageous if you need to return it to python as it retains insertion order for iteration (or if you need to iterate over it). But the use here is only for lookups/edits by key so a `HashMap` will be faster.

```suggestion
    let mut check_count = HashMap::with_capacity(g.node_count());
```

src/dag_algo/mod.rs
```diff
@@ -637,3 +642,86 @@ pub fn collect_bicolor_runs(
...
+/// :raises PyValueError: if graph is not a DAG
+
+#[pyfunction]
+#[pyo3(text_signature = "(graph, /)")]
+pub fn transitive_reduction(
+    graph: &digraph::PyDiGraph,
+    py: Python,
+) -> PyResult<(digraph::PyDiGraph, DictMap<usize, usize>)> {
+    let g = &graph.graph;
+    let mut index_map = DictMap::new();
```
mtreinish, line 666:
```suggestion
    let mut index_map = DictMap::with_capacity(g.node_count());
```

src/dag_algo/mod.rs
```diff
@@ -637,3 +642,86 @@ pub fn collect_bicolor_runs(
...
+    if !is_directed_acyclic_graph(graph) {
+        return Err(PyValueError::new_err(
+            "Directed Acyclic Graph required for transitive_reduction",
+        ));
+    }
+    let mut tr = StablePyGraph::<Directed>::new();
```
mtreinish, line 672:
```suggestion
    let mut tr = StablePyGraph::<Directed>::with_capacity(g.node_count(), 0);
```

tests/rustworkx_tests/digraph/test_transitive_reduction.py
```diff
@@ -0,0 +1,73 @@
...
+        a = graph.add_node("a")
+        b = graph.add_node("b")
+        c = graph.add_node("c")
+        d = graph.add_node("d")
+        e = graph.add_node("e")
+        graph.add_edges_from(
+            [(a, b, 1), (a, d, 1), (a, c, 1), (a, e, 1), (b, d, 1), (c, d, 1), (c, e, 1), (d, e, 1)]
+        )
+        tr, _ = rustworkx.transitive_reduction(graph)
+        self.assertCountEqual(list(tr.edge_list()), [(0, 2), (0, 1), (1, 3), (2, 3), (3, 4)])
```
mtreinish, line 30: Doesn't this only assert that the number of edges is matching the expectation. Shouldn't this assert the edge lists match the expected edges in the reduction?

tests/rustworkx_tests/digraph/test_transitive_reduction.py
```diff
@@ -0,0 +1,73 @@
...
+            ]
+        )
+        tr2, _ = rustworkx.transitive_reduction(graph2)
+        self.assertCountEqual(list(tr2.edge_list()), [(0, 1), (1, 2)])
+
+        graph3 = rustworkx.PyDiGraph()
+        graph3.add_nodes_from([0, 1, 2, 3])
+        graph3.add_edges_from([(0, 1, 1), (0, 2, 1), (0, 3, 1), (1, 2, 1), (1, 3, 1)])
+        tr3, _ = rustworkx.transitive_reduction(graph3)
+        self.assertCountEqual(list(tr3.edge_list()), [(0, 1), (1, 2), (1, 3)])
```
mtreinish, line 50: Should this be 3 tests?

tests/rustworkx_tests/digraph/test_transitive_reduction.py
```diff
@@ -0,0 +1,73 @@
...
+        d = graph.add_node("d")
+        e = graph.add_node("e")
+
+        graph.add_edges_from(
+            [(a, b, 1), (a, d, 1), (a, c, 1), (a, e, 1), (b, d, 1), (c, d, 1), (c, e, 1), (d, e, 1)]
+        )
+
+        graph.remove_node(3)
+
+        tr, _ = rustworkx.transitive_reduction(graph)
```
mtreinish, line 66: I think this should have an assertion on the index map being correct since `c` is removed.

mtreinish (commented):
This is a good start, thanks for doing this. I think most of the implementation looks correct. I left some inline comments though on some improvements that could be made that I think will make it faster.

tests/rustworkx_tests/digraph/test_transitive_reduction.py
danielleodigie: I think `assertCountEqual()` asserts that the list has the same elements, disregarding order. Here are the [docs](https://docs.python.org/3.2/library/unittest.html#unittest.TestCase.assertCountEqual) in case I'm reading it wrong. I'm just trying to see if the lists have the same elements and checking `assertListEqual()` fails because the order differs

mtreinish: Ah, yeah you're correct. I was misremembering what the method does. This does check the element counts are the same in the two lists which is what we want here.

mtreinish (commented):
This LGTM, thanks for the quick update. I think this is all ready to go from the code. THe only thing that I think is missing is we're not including this function in the docs. You just need to add this function to the index in: https://github.com/Qiskit/rustworkx/blob/main/docs/source/api/algorithm_functions/dag_algorithms.rst

mtreinish (approved):
This LGTM now, thanks for all the updates and sorry for the delay in review.

PR #927, opened 2023-07-11
Reorganize API docs into subpages

Eric-Arellano:
Before, we had one page that listed every single function and class. When combined with the new Furo theme from https://github.com/Qiskit/rustworkx/pull/926, this resulted in an incredibly big left sidebar because all entries were shown at the same nesting level.

Now, the API docs are organized into subpages. This is what the top-level page looks like:

Screenshot 2023-07-11 at 12 34 21 PM

An example subpage:

Screenshot 2023-07-11 at 12 35 02 PM

The Algorithms subpage (I'm not sure if this should show all functions on the page or not):

Screenshot 2023-07-11 at 12 39 00 PM

----

This also fixes two issues with the top-level ToC:

1. The home page `<self>` was missing
2. `install.rst` had an improper heading style, so the 4 headers on the page were being treated as top-level pages in the left sidebar.

Now, it looks like this:

Screenshot 2023-07-11 at 12 40 45 PM


docs/source/api/algorithm_functions/index.rst
```diff
@@ -0,0 +1,19 @@
+.. _algorithm_api:
+
+Algorithm Functions
+===================
+
+.. toctree::
+   :maxdepth: 1
```
mtreinish, line 7: I think making this depth 2 would be nice so we still retain the list of functions. Especially in the move to furo depth 2 would be needed to show each function in the side bar right? But also for the current pytorch based theme I think having the full function list on the api page is useful, even if a bit overwhelming at times.

mtreinish (commented):
Overall this LGTM, I think this structure will work better once we update the theme package to be the new furo one. Just one inline comment about the max depth for the top level toctree.

I also wonder if we want to mention the new structure in the CONTRIBUTING.md, I thought we said new functions need to update api.rst there, but I can't find it. This is easy enough to do in a follow up PR though, so let's just move forward with this as is and we can look at the contributing documentation in parallel.

docs/source/api/algorithm_functions/index.rst
Eric-Arellano: `maxdepth` has zero impact on Furo's left sidebar. This is how it renders even with maxdepth set to 1

Screenshot 2023-07-12 at 10 15 48 AM

> I think making this depth 2 would be nice so we still retain the list of functions...I think having the full function list on the api page is useful, even if a bit overwhelming at times.

Sounds reasonable to me. We can always change it with one line if we don't like it.

mtreinish (approved):

PR #929, opened 2023-07-14
Relax constraints in rustworkx-core for hashbrown and indexmap

### LINKED ISSUE #911: Consider replacing hashbrown use in core API

Currently, some structs from `hashbrown` are used in the core API (e.g. the [`mapping` parameter of `token_swapper`](https://github.com/Qiskit/rustworkx/blob/c4f0c2e239dc45c1a5c51fc4fe0cc42be5ea9b78/rustworkx-core/src/token_swapper.rs#L410)). As a result, downstream libraries are required to use the same minor version of `hashbrown`, which has proven to be somewhat cumbersome.

Note that by switching to the standard lib equivalents, we may take _some_ perf hit, but perhaps this is worth the UX improvement.

mtreinish:
This commit updates the version constraints on the hashbrown and indexmap dependencies in rustworkx-core's dependency list. These 2 dependencies are used for I/O to rustworkx-core and we previously had our versions set to the highest version compatible with the msrv. However, for users of rustworkx-core with newer MSRVs this limited version support caused issues if they wanted to use a newer version of these library (which was compatible with their MSRV) as an input or output to rustworkx-core. For example, as described in #911 qiskit-terra was using hashbrown 0.13.x and had errors because the released version of rustworkx-core only was listed as compatible with hashbrown 0.12. This version mismatch meant that the hashbrown objects created in qiskit-terra were not recognized by rustworkx-core because of the version mismatch. This commit attempts to fix this issue by increasing the allowable range of versions to be greater than or equal to our previous fixed version and capped at the last known working version. This should enable users to set a fixed library version that they want even if it's newer and requires a higher MSRV.

This commit is for the stable/0.13 branch as the minimum version is lower on the stable branch as the MSRV for the 0.13.0 is much older than what we're using on the main branch now. A separate PR will be pushed up for main to update it there. This was done in reverse in the interest of ensuring we include this in 0.13.1 and the inevitable backport conflict if we started with main first.

Fixes #911

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -95,6 +95,11 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
           targets: ${{ matrix.platform.rust-target }}
+      - name: 'Set msrv min versions'
+        run: |
+          cargo update -p hashbrown:0.12.3 --precise 0.12.3 --workspace
+          cargo update -p indexmap:0.9.3 --precise 1.9.3 --workspace
+        if: ${{ matrix.msrv == 'MSRV' }}
       - name: 'Install dependencies'
         run: python -m pip install --upgrade tox
       - name: 'Install binary dependencies'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -98,7 +98,7 @@ jobs:
       - name: 'Set msrv min versions'
         run: |
           cargo update -p hashbrown:0.12.3 --precise 0.12.3 --workspace
-          cargo update -p indexmap:0.9.3 --precise 1.9.3 --workspace
+          cargo update -p indexmap:1.9.3 --precise 1.9.3 --workspace
         if: ${{ matrix.msrv == 'MSRV' }}
       - name: 'Install dependencies'
         run: python -m pip install --upgrade tox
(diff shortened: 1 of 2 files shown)


IvanIsCoding:
Looks like we have some Clippy errors that are new due to the 1.71 update?

Apart from that this LGTM

IvanIsCoding (approved):

rustworkx-core/Cargo.toml
```diff
@@ -19,15 +19,12 @@ rand_pcg = "0.3.1"
 rayon = "1.6"
 num-traits = "0.2"
 priority-queue = "1.2"
-rayon-cond = "0.2"
+rayon-cond = "0.3"
```
IvanIsCoding, line 22:
```suggestion
rayon-cond = "0.2"
```

I think this is the culprit

.github/workflows/main.yml
```diff
@@ -95,6 +95,11 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
           targets: ${{ matrix.platform.rust-target }}
+      - name: 'Set msrv min versions'
+        run: |
+          cargo update -p hashbrown:0.12.3 --precise 0.12.3 --workspace
+          cargo update -p indexmap:0.9.3 --precise 1.9.3 --workspace
```
IvanIsCoding, line 101:
```suggestion
          cargo update -p indexmap:1.9.3 --precise 1.9.3 --workspace
```

PR #930, opened 2023-07-17
Adding Clippy Fix

danielleodigie:
Because of a new Rust update, Clippy now warned for a line of code in `cycle_basis.rs` in `rustworkx-core`. This PR fixes that one line so that Clippy passes.


mtreinish (approved):
LGTM, thanks for fixing this.

PR #944, opened 2023-07-26
Add a section to Python docs about page for rustworkx-core

mtreinish:
The main page explaining the rustworkx library in the Python docs previously didn't make any mention of rustworkx-core. While in general the Python side is built on top of rustworkx-core and the vast majority of the users of the Python library don't need to know about the pure Rust library inside, it is still good to highlight its existence in case anyone finds the Python library but is really interested in consuming it from Rust. This commit just adds a small subsection to the about page explaining the existence of rustworkx-core, its purpose, and where you can find mroe information about it.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -38,10 +38,10 @@ to install rustworkx.
 rustworkx-core
 --------------
 
-In addition to the Python library as part of rustworkx a standalone Rust library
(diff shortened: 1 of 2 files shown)


enavarro51 (changes requested):
This is a good addition as we put more and more into core. LGTM, with just a couple of readability comments.

enavarro51 (approved):

PR #954, opened 2023-07-31
Fix directed gnp_random_graph

### LINKED ISSUE #950: rustworkx.directed_gnp_random_graph doesn't produce random graphs
### Information

- **rustworkx version**: 0.13.0
- **Python version**: 3.9
- **Rust version**: 1.71.0
- **Operating system**: FreeBSD 13.2

rx.directed_gnp_random_graph(20, 0.5) consistently produces nodes with only one edge, when with ```2*(19)=38``` possible edges the chance of this is extremely low: ```0.5^38 * 38 = 1.38 * 10^-10```.

19 edges on average should be present for each node.

Run this script:
```
import rustworkx as rx

import os
import tempfile

import pydot
from PIL import Image

G = rx.directed_gnp_random_graph(20, 0.5)

dot_str = G.to_dot(
    lambda node: dict(
        color='black', fillcolor='lightblue', style='filled'))
dot = pydot.graph_from_dot_data(dot_str)[0]

dot.write_png("G.png")
```

image

image

image

enavarro51:
This PR fixes #950. Previously the `directed_gnp_random_graph` did not properly produce in and out edges and also contained a bug that caused the process to terminate too quickly, resulting in some nodes having a very small number of edges.

- [x] Add additional tests
- [x] Release note


rustworkx-core/src/generators/random_graph.rs
```diff
@@ -113,32 +114,46 @@ where
                 }
             }
         } else {
-            let mut v: isize = if directed { 0 } else { 1 };
-            let mut w: isize = -1;
-            let num_nodes: isize = num_nodes as isize;
+            let num_nodes = num_nodes as isize;
```
mtreinish, line 117: I know this is pre-existing code but couldn't this potentially overflow since the `num_nodes` input is of type `usize`? Should we have a guard against `num_nodes > isize::MAX` or use `try_into()` here?

mtreinish (commented):
This change LGTM, just a couple of inline questions. Sorry it took so long to review this.

releasenotes/notes/fix-gnp-random-directed-41f2a436c49b9064.yaml
```diff
@@ -0,0 +1,7 @@
+---
+fixes:
+  - |
+    Fixed an issue where the :func:`~rustworkx.generators.directed_gnp_random_graph`
+    and the :func:`~rustworkx-core.generators.gnp_random_graph` for directed graphs
+    produced a graph where lower node numbers had only a small number of edges
+    compared to what was expected and the output was therefore not random.
```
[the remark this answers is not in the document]
enavarro51: That's fine. I'll remove the last clause.

rustworkx-core/src/generators/random_graph.rs
enavarro51: Yeah, `try_into` is a good idea. If it overflows, should we set it to `isize::MAX` or `isize::MIN` or should we error or both?

enavarro51: Also, should we say something in the docs about num_nodes not exceeding some value?

mtreinish: I think we should probably just error in the case we've exceeded.

We can document it but `isize::MAX` is 9,223,372,036,854,775,807 on 64 bit platforms and ~2 billion on 32 bit platforms, maybe just say it will return an error if the num of nodes exceeds the `isize::MAX` directly (for rustworkx-core). But I think the `InvalidInputError` error struct is probably sufficient to cover this case too.

releasenotes/notes/fix-gnp-random-directed-41f2a436c49b9064.yaml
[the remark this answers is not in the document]
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/954/commits/fd7544a9e10ecded245780b2771910b8c12794da.

rustworkx-core/src/generators/random_graph.rs
enavarro51: Done in https://github.com/Qiskit/rustworkx/pull/954/commits/fd7544a9e10ecded245780b2771910b8c12794da. I believe this does what you're saying. I wasn't sure how to test this, since anything I tried with very large numbers gave either a `capacity overflow` or `memory allocation` error.

mtreinish (approved):

rustworkx-core/src/generators/random_graph.rs
enavarro51: Actually I redid this in https://github.com/Qiskit/rustworkx/pull/954/commits/b41cf8a127fb71a91137e2dcfafcfeaaafc9e59d using match. I was able to successfully test locally if I changed all the `isize` to `i8`, but if I test at `isize` levels, I still get capacity or memory errors. So I did not add any new tests for this.

PR #963, opened 2023-08-07
Add annotations for type-specific rustworkx functions

### LINKED ISSUE #960: mypy cannot find type-specific functions
### Information

- **rustworkx version**: 0.13.1
- **Python version**: 3.9
- **Rust version**: N/A, not installed
- **Operating system**: Ubuntu 20.04

PyCharm will think a method exists, I'll go to use it and then run mypy against it, mypy then says "Module has no attribute "graph_dijkstra_shortest_paths"  [attr-defined]"

Mypy would understand the error

I have not been able to get to the bottom of why this is happening, the mypy version is 3.9 **not** 3.7, so is not a duplicate of that issue. 

I just pip installed rustworkx, added 0.13.1 to my requirements file and then tried to use it in our codebase.

IvanIsCoding:
Closes #960
Related to #352, follow up of #401

This adds type annotations to all type-specific rustworkx functions. It also re-exports all type-specific functions at the top-level, because it is unlikely that users will use `rustworkx.rustworkx` (it's a hidden implementation detail)


IvanIsCoding:
This is ready for review, I hope it makes it to 0.14

rustworkx/cartesian_product.pyi
mtreinish, line 1: Do we have to worry that this adding a module that doesn't exist in the rustworkx package? Like we'll have type hints for `rustworkx.cartesian_product.digraph_cartesian_product` which isn't a valid path in rustworkx.

rustworkx/visit.pyi
```diff
@@ -0,0 +1,39 @@
...
+    def tree_edge(self, e: tuple[int, int, _T]): ...
+    def back_edge(self, e: tuple[int, int, _T]): ...
+    def forward_or_cross_edge(self, e: tuple[int, int, _T]): ...
+
+class DijkstraVisitor(Generic[_T]):
+    def discover_vertex(self, v: int, score: float): ...
+    def finish_vertex(self, v: int): ...
+    def examine_edge(self, e: tuple[int, int, _T]): ...
+    def edge_relaxed(self, e: tuple[int, int, _T]): ...
+    def edge_not_relaxed(self, e: tuple[int, int, _T]): ...
```
mtreinish, line 39: These classes are defined in Python, could we integrate the type hints directly in the source for these instead of using a stub file?

mtreinish (commented):
Overall this looks great, thanks for spending all the time to fix this. Just a couple of questions inline.

src/connectivity/mod.rs
```diff
@@ -527,15 +527,15 @@ pub fn digraph_complement(py: Python, graph: &digraph::PyDiGraph) -> PyResult<di
 /// :returns: A list of lists where each inner list is a path of node indices
 /// :rtype: list
 #[pyfunction]
-#[pyo3(text_signature = "(graph, from, to, /, min_depth=None, cutoff=None)")]
+#[pyo3(text_signature = "(graph, origin, to, /, min_depth=None, cutoff=None)")]
 pub fn graph_all_simple_paths(
     graph: &graph::PyGraph,
-    from: usize,
```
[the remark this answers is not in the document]
IvanIsCoding: I had to change because of the Python keyword. There was no workaround

rustworkx/cartesian_product.pyi
IvanIsCoding: The linter complains but it’s a tradeoff. I created the files to organize our code like in Rust. But in practice our Rust just becomes one big file, while Python doesn’t. I could try adding underscores in the name of the files to prevent them from being imported following PEP-8 conventions

rustworkx/visit.pyi
IvanIsCoding: We could but there are some advantages to use stubs for our case. The first one is consistency with the other files. The second one is [avoiding annotation issues at runtime](https://mypy.readthedocs.io/en/stable/runtime_troubles.html?highlight=TYPE_CHECKING#annotation-issues-at-runtime)

mtreinish: That works for me, I just wanted to check because it's one of the rare places we have class definitions in Python so using native type hints is possible.

mtreinish (approved):

PR #965, opened 2023-08-13
add .so files to gitignore

1ucian0:
When following the instructions in https://github.com/Qiskit/rustworkx#develop-mode after compilation I got a `.so` file not ignored by gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,5 @@ retworkx/*pyd
 *.svg
 !/docs/source/images/*.svg
 *.jpg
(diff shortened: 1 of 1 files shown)


IvanIsCoding (approved):
LGTM

mtreinish:
FWIW, we had a line for this in the .gitignore before https://github.com/Qiskit/rustworkx/blob/main/.gitignore#L14-L15 but it was missed in the rustworkx rename so it wasn't matching anymore

PR #967, opened 2023-08-16
Add missing methods in PyGraph and PyDiGraph stubs

### LINKED ISSUE #1036: missing `node_indices` method
### Information

- **rustworkx version**: 0.13.2
- **Python version**: 3.12.0
- **Rust version**: rustc 1.73.0 (cc66ad468 2023-10-03)
- **Operating system**: Microsoft Windows 11 Pro 10.0.22631 Build 22631

`node_indices` doesn't seem to be defined in the PyO3 `pyi` files.

According to the documentation, this method should exist: 
https://qiskit.org/ecosystem/rustworkx/apiref/rustworkx.PyDiGraph.node_indices.html#rustworkx.PyDiGraph.node_indices

Look at [`graph.pyi`](https://github.com/Qiskit/rustworkx/blob/main/rustworkx/graph.pyi) for example. Files probably just need to be regenerated.

IvanIsCoding:
Closes #1036 

Related to #352

Adds methods that were added after #401 was merged. All of them were checked with `mypy.stubtest`


rustworkx/graph.pyi
```diff
@@ -22,14 +24,19 @@ from typing import (
     List,
     Tuple,
     Sequence,
+    TYPE_CHECKING,
 )
 
+if TYPE_CHECKING:
+    from .digraph import PyDiGraph
+
```
IvanIsCoding, line 32: PyGraph and PyDiGraph have a "cyclic" dependency due to the `to_directed` and `to_undirected` method. This is a hack to make Python agree with the Rust code

rustworkx/digraph.pyi
```diff
@@ -52,11 +57,20 @@ class PyDiGraph(Generic[S, T]):
         node_map_func: Callable[[S], int] | None = ...,
         edge_map_func: Callable[[T], int] | None = ...,
     ) -> dict[int, int]: ...
+    def contract_nodes(
+        self,
+        nodes: list[int],
```
mtreinish, line 62: Technically I think this is

```suggestion
        nodes: Sequence[int],
```

although it also will except a `numpy.ndarray`, which isn't a sequence, for all int dtypes I think.

rustworkx/digraph.pyi
```diff
@@ -52,11 +57,20 @@ class PyDiGraph(Generic[S, T]):
...
+        /,
+        check_cycle: bool | None = ...,
+        weight_combo_fn: Callable[[T, T], T] | None = ...,
+    ) -> int: ...
     def copy(self) -> PyDiGraph[S, T]: ...
     def edge_index_map(self) -> EdgeIndexMap[T]: ...
     def edge_indices(self) -> EdgeIndices: ...
     def edge_list(self) -> EdgeList: ...
     def edges(self) -> list[T]: ...
+    def edge_subgraph(self, edge_list: list[tuple[int, int]], /) -> PyDiGraph[S, T]: ...
```
mtreinish, line 73: Same comment as above with `contract_nodes`.

```suggestion
    def edge_subgraph(self, edge_list: Sequence[tuple[int, int]], /) -> PyDiGraph[S, T]: ...
```

rustworkx/graph.pyi
```diff
@@ -48,12 +55,20 @@ class PyGraph(Generic[S, T]):
         node_map_func: Callable[[S], int] | None = ...,
         edge_map_func: Callable[[T], int] | None = ...,
     ) -> dict[int, int]: ...
+    def contract_nodes(
+        self,
+        nodes: list[int],
```
mtreinish, line 60: Same comment as on digraph here.

mtreinish (commented):
Overall this LGTM, just inline comments about the mapping of `Vec<usize>` to `list[int]` as `Vec<usize>` will take more than just a list.

Also is there a point where we can catch these missing functions in CI? I assume we're not yet because we're still marked as only partially typed. But can we enforce this for new methods on the graph classes while working on the rest of the functions?

IvanIsCoding:
> Overall this LGTM, just inline comments about the mapping of `Vec<usize>` to `list[int]` as `Vec<usize>` will take more than just a list.
> 
> Also is there a point where we can catch these missing functions in CI? I assume we're not yet because we're still marked as only partially typed. But can we enforce this for new methods on the graph classes while working on the rest of the functions?

We need to remove the `—ignore` from https://github.com/Qiskit/rustworkx/blob/75dbca9441a707cf4baccab9d0775a3e88460e9a/tox.ini#L79 but I think it needs to be for our whole codebase

mtreinish (approved):
LGTM, thanks for the quick update.

PR #971, opened 2023-08-21
Handle token_swapper impossible swap mapping with Error

mtreinish:
This commit fixes the handling of invalid mapping requests in the token_swapper rustworkx-core function and the graph_token_swapper function in rustworkx that uses it. Previously if an invalid mapping was requested the function would internally panic because it always assumed there was a path in the graph to fulfill the user requested mapping. However, because the rustworkx-core function didn't support error returns a breaking api change is needed to add a result return type.


IvanIsCoding (approved):
LGTM, we need to break the API but it's a better solution than panicking

enavarro51 (approved):
LGTM as well. Definitely needed.

PR #975, opened 2023-08-29
Add capacity keyword arguments to graph creation

jakelishman:
This is just a simple exposure of the allocation control from Rust space to Python space.

I was playing around with this for DAG construction in Qiskit and it (surprisingly to me) didn't have a meaningful effect on performance for what I was doing, but even so, it just _feels_ like something I'd like to be able to expose there as well.


IvanIsCoding:
I’ll let @mtreinish give his opinion as well but I am against exposing this to end-users.

Mostly because Python users should not worry about it. For example, Python lists don’t let you pick a capacity. I think we are leaking a detail from Rust into Python

Also, I think for most of the Qiskit DAG code we spent more time on Rust <-> Python conversion than anything else. For some algorithms it takes more time to convert objects than to actually run the core logic

jakelishman:
That's fair - it's always going to be the case that the other stuff we're doing is slower, it just seemed unnecessary to me to pay the resize cost even as a papercut when I know how many nodes and edges I'm about to generate to put into my graph.

IvanIsCoding:
> That's fair - it's always going to be the case that the other stuff we're doing is slower, it just seemed unnecessary to me to pay the resize cost even as a papercut when I know how many nodes and edges I'm about to generate to put into my graph.

I could see a constructor for PyGraph/PyDiGraph which accepts nodes/edges, a little bit similar to what `extend_from_edge_list` does now

But if we add `with_capacity` then we need to explain to the end user what is a capacity, why it is different from the number of nodes/edges, etc. Which I would rather not do

mtreinish:
For me I think we probably want to expose this functionality in some way for performance. Saving on memory allocations can save a bunch of runtime especially for very large graphs. But, I also share ivan's concerns around exposing too much internal details to the UI. I'm wondering if there is a compromise here on the interface, like instead of doing `with_capacity` adding kwargs to the constructor like `target_node_count` and `target_edge_count` and document the performance benefit and behavior there

jakelishman:
I'm happy to retarget this PR to whatever interface you guys would prefer, if it's something you'd consider adding.  From my side, my ask is only that I can reserve space for nodes and/or edges before I actually provide them - anything that lets me do that, no matter how wordy / hidden, is fine for me, if you guys agree to add it.

(and I can live with it if it's not added at all, for the above reasons)

enavarro51:
What about `initial_node_count` and `initial_edge_count`? That indicates to a Python user that it's not a limit, but a starting point. I do think it's probably worthwhile, especially since we now have a lot of hybrid Python/Rust developers on the qiskit side and probably elsewhere as well.

jakelishman:
Updated to use the new forms suggested above.

src/graph.rs
```diff
@@ -183,16 +187,36 @@ impl PyGraph {
...
             attrs: attrs.unwrap_or_else(|| py.None()),
         }
     }
 
+    fn __getnewargs_ex__<'py>(&self, py: Python<'py>) -> (Py<PyTuple>, Bound<'py, PyDict>) {
+        (
+            (self.multigraph, self.attrs.clone_ref(py)).into_py(py),
+            [
+                ("initial_node_count", self.graph.node_count()),
+                ("initial_edge_count", self.graph.edge_bound()),
```
alexanderivrii, line 214: Silly question: why do you use `node_count` and `edge_bound` (and not say `node_bound` and `edge_bound`, or `node_count` and `edge_count`)?

alexanderivrii:
FWIW, I also like the ability to specify the number of expected nodes/edges when creating rustworkx graphs, so I am in favor of this PR.

src/graph.rs
mtreinish: It should be `node_bound()` and `edge_bound()` because that's the size of the underlying `Vec`s in the `StableGraph` object.

Yeah that's the purpose since, it preallocates the internal `Vec`s it'll make the unpickling a bit faster.

mtreinish:
```suggestion
                ("initial_node_count", self.graph.node_bound()),
                ("initial_edge_count", self.graph.edge_bound()),
```

src/digraph.rs
```diff
@@ -174,6 +174,10 @@ use super::dag_algo::is_directed_acyclic_graph;
 /// :param attrs: An optional attributes payload to assign to the
 ///     :attr:`~.PyDiGraph.attrs` attribute. This can be any Python object. If
 ///     it is not specified :attr:`~.PyDiGraph.attrs` will be set to ``None``.
+/// :param int initial_node_count: The graph will be allocated with enough capacity to store this
+///     many nodes before needing to grow (default 0: no overallocation).
+/// :param int initial_edge_count: The graph will be allocated with enough capacity to store this
+///     many edges before needing to grow (default 0: no overallocation).
```
mtreinish, line 180: I would add here explicitly that this does not prepopulate any nodes or edges. It's all about runtime performance for when you know the size of the graph ahead of time. I think we should be a bit more explicit here for python users who aren't necessarily familiar with using this kind of pattern.

src/digraph.rs
```diff
@@ -298,6 +309,17 @@ impl PyDiGraph {
         }
     }
 
+    fn __getnewargs_ex__<'py>(&self, py: Python<'py>) -> (Py<PyTuple>, Bound<'py, PyDict>) {
+        (
+            (self.check_cycle, self.multigraph, self.attrs.clone_ref(py)).into_py(py),
+            [
+                ("initial_node_count", self.graph.node_count()),
```
mtreinish, line 316:
```suggestion
                ("initial_node_count", self.graph.node_bound()),
```

mtreinish (commented):
I'm fine with this API addition, it is a bit unusual for Python users to worry about allocations, but the performance benefit for doing this can be substantial enough I think it warrants adding a lower level api for advanced users. I just left a couple of inline comments, mainly I think we should be more explicit in the docs as the target audience is a python user. I also think @alexanderivrii caught a a valid issue with the getnewargs_ex method.

src/digraph.rs
jakelishman: Done in ba13b1f9.

src/digraph.rs
jakelishman: Done in b9f564a4.

src/graph.rs
jakelishman: The mystery is really why I wasn't consistent with myself on consecutive lines... Fixed in ba13b1f9.

src/digraph.rs
```diff
@@ -174,6 +174,14 @@ use super::dag_algo::is_directed_acyclic_graph;
 /// :param attrs: An optional attributes payload to assign to the
 ///     :attr:`~.PyDiGraph.attrs` attribute. This can be any Python object. If
 ///     it is not specified :attr:`~.PyDiGraph.attrs` will be set to ``None``.
+/// :param int initial_node_count: The graph will be allocated with enough capacity to store this
+///     many nodes before needing to grow.  This does not prepopulate any nodes with data, it is
+///     only a potential performance optimization if the complete size of the graph is known in
+///     advance (default 0: no overallocation).
+/// :param int initial_edge_count: The graph will be allocated with enough capacity to store this
+///     many edges before needing to grow.  This does not prepopulate any nodes with data, it is
```
mtreinish, line 182:
```suggestion
///     many edges before needing to grow.  This does not prepopulate any edges with data, it is
```

src/graph.rs
```diff
@@ -138,6 +138,14 @@ use petgraph::visit::{
 /// :param attrs: An optional attributes payload to assign to the
 ///     :attr:`~.PyGraph.attrs` attribute. This can be any Python object. If
 ///     it is not specified :attr:`~.PyGraph.attrs` will be set to ``None``.
+/// :param int initial_node_count: The graph will be allocated with enough capacity to store this
+///     many nodes before needing to grow.  This does not prepopulate any nodes with data, it is
+///     only a potential performance optimization if the complete size of the graph is known in
+///     advance (default 0: no overallocation).
+/// :param int initial_edge_count: The graph will be allocated with enough capacity to store this
+///     many edges before needing to grow.  This does not prepopulate any nodes with data, it is
```
mtreinish, line 146:
```suggestion
///     many edges before needing to grow.  This does not prepopulate any edges with data, it is
```

mtreinish (approved):

PR #983, opened 2023-09-18
Make `stestr` >= 4.1 

### LINKED ISSUE #982: Investigate `stestr` breaking

All our CI actions are broken:
```
stestr: 'run' is not a stestr command. See 'stestr --help'.
Did you mean one of these?
  help
  init
  last
```

`stestr` should work. In fact, we have pinned the `strestr` version in the CI and there have been no new releases, which probably indicates that one of the dependencies broke etc

Just run `tox -epy` and the tests will fail

IvanIsCoding:
Closes #982 

Make `stestr` have version greater than 4.1, containing the fix we need

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -154,7 +154,7 @@ jobs:
         if: runner.os == 'Linux'
       - name: 'Build rustworkx and test dependencies'
         run: |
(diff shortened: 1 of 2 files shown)


mtreinish (approved):
It's not strictly required this way because pip will install the latest of everything for us by default. But this doesn't hurt either so lets go for it

PR #989, opened 2023-09-30
Bump `indexmap` to 2.0.2, `hashbrown` to 0.14.1 & more

IvanIsCoding:
This bumps many of our dependencies, with highlights for:
* `indexmap` to 2.0.2 (replacing #987)
*  `hashbrown` to 0.14.1 (replacing #988)

More importantly, it compiles (which the previous PRs didn't for some reason)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -139,9 +139,9 @@ dependencies = [
 
 [[package]]
 name = "hermit-abi"
-version = "0.3.2"
+version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


mtreinish (approved):
I wonder if this means we can also lower our MSRV to 1.63? Either way this LGTM we can look at the MSRV question in a followup

IvanIsCoding:
> I wonder if this means we can also lower our MSRV to 1.63? Either way this LGTM we can look at the MSRV question in a followup

I tried but we can't! The workspace features introduced in #956 are 1.64+

Issue #986, opened 2023-09-22
Release Python 3.12 wheels


jakelishman:
### What is the expected enhancement?

This isn't especially urgent (the final release isn't even out yet), I'm mostly just making this issue so I've got something to point at in Qiskit's tracking of our own blockers against Python 3.12.

(Stable C API wheels would also be fine: #891.)

IvanIsCoding:
I think the safest option might be releasing a 0.13.1 wheel for 3.12 because 0.14 with the ABI might take a little bit. We only have a week or so before 3.12

cclauss:
Not anymore.

IvanIsCoding:
Closing because the vast majority of the wheels are in https://pypi.org/project/rustworkx/0.13.2/#files

For s390x and ppc64 users, keep an eye on https://github.com/Qiskit/rustworkx/actions/runs/6398491808

mtreinish:
Closing this now as rustworkx 0.13.2 has been released with 3.12 wheels (although we missed the trove classifier in the package metadata): https://pypi.org/project/rustworkx/0.13.2/#files

PR #990, opened 2023-09-30
Support Python 3.12 and bump `rustworkx` to 0.13.2

IvanIsCoding:
Start the work to support #986:
* Bump PyO3 to 0.19.2 to support Python 3.12
* Add CI tests with Python 3.12 release candidate
* Update cibuildwheel to 2.16.1, which should build 3.12 wheels

Because this updates PyO3, I think it should trigger a minor version bump to 0.13.2. In the release notes, we can be clear this has little to no change for 0.13.1 users running Python 3.11 or earlier.


auvipy (commented):
Should we consider dropping python 3.7 from CI beside adding python 3.12?

IvanIsCoding:
> Should we consider dropping python 3.7 from CI beside adding python 3.12?

No, we supported 3.7 in the 0.13.x series and we’ll keep supporting it. It’s already dropped for 0.14 and on-wards

.github/workflows/main.yml
```diff
@@ -57,7 +57,7 @@ jobs:
     strategy:
       matrix:
         rust: [stable]
-        python-version: ['3.7.16', 3.8, 3.9, "3.10", "3.11"]
+        python-version: ['3.7.16', 3.8, 3.9, "3.10", "3.11", "3.12.0-rc.3"]
```
mtreinish, line 60: I assume we'll have to wait a little bit for the default gha image to get 3.12.0.

Cargo.lock
```diff
@@ -341,7 +341,7 @@ checksum = "437213adf41bbccf4aeae535fbfcdad0f6fed241e1ae182ebe97fa1f3ce19389"
 dependencies = [
  "libc",
  "ndarray",
- "num-complex 0.4.3",
+ "num-complex 0.2.4",
```
mtreinish, line 344: Is there a reason num-complex got downgraded to 0.2.4, since we had 0.4.3 here before?

.github/workflows/main.yml
IvanIsCoding: It's not in https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json, but the ABI should be identical to the released one

Cargo.lock
IvanIsCoding: Fixed it

IvanIsCoding:
> This LGTM, two questions inline but nothing major. The other thing is we probably need to do a `cargo build` and update the `Cargo.lock` because rustworkx and rustworkx-core are still at 0.13.1 in the lock file. The other thing is do you want to tag this PR as the 0.13.2, if so we probably should include a release note that notes that this release adds preliminary 3.12 support on top of 0.13.1.

I added a very brief release note for 0.13.2. So I think we should be able to tag this with 0.13.2 for the release

mtreinish (approved):

mtreinish:
Oh, actually before we release this we'll need to update cibuildwheel in wheel builds to a newer version that includes 3.12 support

IvanIsCoding:
> Oh, actually before we release this we'll need to update cibuildwheel in wheel builds to a newer version that includes 3.12 support

Can you tag this commit as 0.13.2? I don't have the permission

PR #995, opened 2023-10-05
mapping is not possible when a node has no neighbors

### LINKED ISSUE #994: Token_swapper fails on graphs with edgeless nodes
### Information

Recently (see #971) the token swapper algorithm has been extended to support disconnected coupling maps, and in particular return an error rather than internally panic when swapping is not possible. 

Here is an example where token swapper still internally panics.

```
let mut g = petgraph::graph::UnGraph::<(), ()>::new_undirected();
let a = g.add_node(());
let b = g.add_node(());
let c = g.add_node(());
let d = g.add_node(());
g.add_edge(c, d, ());
let mapping = HashMap::from([(a, b), (b, a)]);
let out = token_swapper(&g, mapping, Some(10), Some(4), Some(50));
```

Ideally this should return `Err(MapNotPossible)` instead of panicking:
```
panicked at 'index out of bounds: the len is 0 but the index is 0', rustworkx-core\src\token_swapper.rs:305:40
```

alexanderivrii:
Fixes #994.

I believe that the function ``add_token_edges(node)`` should return ``Err(MapNotPossible {})`` when no mapping exists. This was already handled in the case that ``node`` had at least one neighbor.  Now I am also returning the error in the case a node has no neighbors.

### DIFF
--- a/rustworkx-core/src/token_swapper.rs
+++ b/rustworkx-core/src/token_swapper.rs
@@ -205,6 +205,16 @@ where
         }
         let id_node = self.rev_node_map[&node];
         let id_token = self.rev_node_map[&tokens[&node]];
+
+        if self
+            .graph
+            .neighbors(id_node)
+            .collect::<Vec<G::NodeId>>()
+            .is_empty()
+        {
+            return Err(MapNotPossible {});
+        }
+
         for id_neighbor in self.graph.neighbors(id_node) {
             let neighbor = self.node_map[&id_neighbor];
             let dist_neighbor: DictMap<G::NodeId, usize> = dijkstra(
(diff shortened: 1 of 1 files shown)


rustworkx-core/src/token_swapper.rs
```diff
@@ -205,6 +205,16 @@ where
         }
         let id_node = self.rev_node_map[&node];
         let id_token = self.rev_node_map[&tokens[&node]];
+
+        if self
+            .graph
+            .neighbors(id_node)
+            .collect::<Vec<G::NodeId>>()
+            .is_empty()
```
mtreinish, line 213:
```suggestion
            .neighbors(id_node)
            .next()
            .is_some()
```

mtreinish (commented):
This LGTM, just one inline  comment that I think will be more efficient.

rustworkx-core/src/token_swapper.rs
alexanderivrii: Oh, this is how one can directly check whether an iterator is empty or non-empty, I was wondering about that. Fixed in [d682b9b](https://github.com/Qiskit/rustworkx/pull/995/commits/d682b9ba5198e7a1d4c371fb494c41123cbab18b).

mtreinish (approved):
LGTM, thanks for the quick update and also fixing the mistake in my suggestion :D

PR #997, opened 2023-10-06
Add support for Python 3.12 and musllinux to main

mtreinish:
As part of the 0.13.2 release we added support for Python 3.12 and musllinux to rustworkx. However, these changes did not happen on main yet. This commit applies the necessary changes to the main branch for the 0.14.0 and future releases. Also to simplify the configuration of the cibuildwheel jobs this combines #753 into this PR so that the configuration is centralized in the pyproject.toml.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -89,6 +89,14 @@ source.
      - s390x
      - :ref:`tier-4`
      - Distributions compatible with the `manylinux 2014`_ packaging specification
+   * - Linux (musl)
+     - x86_64
+     - :ref:`tier-3`
+     -
+   * - Linux (musl)
+     - aarch64
+     - :ref:`tier-3`
+     -
    * - macOS (10.9 or newer)
      - x86_64
      - :ref:`tier-1`
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -93,7 +93,7 @@ source.
      - x86_64
      - :ref:`tier-3`
      -
-   * - macOS (10.9 or newer)
+   * - macOS (10.12 or newer)
      - x86_64
      - :ref:`tier-1`
      -
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -93,10 +93,6 @@ source.
      - x86_64
      - :ref:`tier-3`
      -
-   * - Linux (musl)
-     - aarch64
-     - :ref:`tier-3`
-     -
    * - macOS (10.9 or newer)
      - x86_64
      - :ref:`tier-1`
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ target-version = ['py38', 'py39', 'py310', 'py311']
 manylinux-x86_64-image = "manylinux2014"
 manylinux-i686-image = "manylinux2014"
 skip = "pp* cp36-* cp37-* *win32"
-test-requires = "networkx testtools fixtures"
+test-requires = "networkx"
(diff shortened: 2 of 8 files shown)


docs/source/install.rst
```diff
@@ -89,6 +89,14 @@ source.
      - s390x
      - :ref:`tier-4`
      - Distributions compatible with the `manylinux 2014`_ packaging specification
+   * - Linux (musl)
+     - x86_64
+     - :ref:`tier-3`
+     -
+   * - Linux (musl)
+     - aarch64
+     - :ref:`tier-3`
```
IvanIsCoding, line 98: I'll say I don't think supporting this architecture now is a priority. NumPy themselves (https://pypi.org/project/numpy/1.26.0/#files) don't support musl with aarch64 as of now, so the user already needs to have a C compiler anyway, I think it's fair to assume users are "advanced" and have to install a Rust compiler

mtreinish: Ah ok, yeah I'll add the exclude for it then. I just assumed it would be in the same class as x86_64 but didn't look.

IvanIsCoding (commented):
I think we might need to split the x86_64 and i686 jobs for Linux. i686 seems particularly more prone to error due to BLAS, so I think we should handle installation without BLAS separately

pyproject.toml
```diff
@@ -4,4 +4,33 @@ build-backend = "setuptools.build_meta"
...
+
+[[tool.cibuildwheel.overrides]]
+select = "*-musllinux*"
+before-all = "apk add apk add --no-cache curl gcc && curl https://sh.rustup.rs -sSf | sh -s -- -y && source $HOME/.cargo/env && rustup install stable && rustup default stable"
+test-skip = "cp37-* cp38-*"
+
+[[tool.cibuildwheel.overrides]]
+select = "*i686"
+skip = "pp* cp36-* cp37-* *win32 *-musllinux*"
+before_test = "yum install openblas-devel"
```
IvanIsCoding, line 29: I think you might want to use `-C-Dallow-noblas=true` argument for pip. I saw it in https://github.com/numpy/numpy/issues/24703

pyproject.toml
```diff
@@ -4,4 +4,33 @@ build-backend = "setuptools.build_meta"
...
+
+[[tool.cibuildwheel.overrides]]
+select = "*ppc64le"
+before_test = "yum install openblas-devel"
```
IvanIsCoding, line 33: Probably we need to use `-C-Dallow-noblas=true` for installing NumPy here as well

pyproject.toml
mtreinish: I gave this a try in: https://github.com/Qiskit/rustworkx/pull/997/commits/7621ca1bec914f7c5d4eac6a5f33d23c955cd775 it didn't work. But I'm not convinced the config override is working here. Like on the current branch HEAD ci run yum wasn't run to install openblas.

IvanIsCoding: Hmmm, maybe we can try adding `CIBW_BUILD_FRONTEND: "pip; args: -C-Dallow-noblas=true"`? That way we'd be sure cibuildwheel wouldn't ignore it

mtreinish: I think I figured it out, the issue was I was previously specifying options which couldn't be overridden per job and that was causing it to be ignored I think.

IvanIsCoding (approved):
LGTM, we just need to remove the trigger on push and it should be good to go

PR #998, opened 2023-10-09
Add isolates() function to rustworkx

mtreinish:
This commit adds a new function, isolates(), which is used to find all the isolates in a graph.

PR #1002, opened 2023-10-11
Add two_color and is_bipartite

mtreinish:
This commit adds new functions to rustworkx, two_color() and
is_bipartite(), which are used to compute a two coloring for a graph and
then determine if a givn graph is bipartite. The two_color() function is
added to rustworkx-core as the python is_bipartite() function just wraps
it and converts the output to a bool if a two coloring is possible or
not.

This commit is based on top of #998 and will need to be rebased after
that merges.

You can see a diff of just the contents of this PR here:

https://github.com/Qiskit/rustworkx/compare/mtreinish:retworkx:add-isolates...mtreinish:retworkx:add-two-color?expand=1


tests/rustworkx_tests/graph/test_bipartite.py
```diff
@@ -0,0 +1,55 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import unittest
+
+import rustworkx
+
+
+class TestBipartite(unittest.TestCase):
```
IvanIsCoding, line 18: Can you add test cases using `rustworkx.generators.generalized_petersen_graph` citing Lemma 3 of https://arxiv.org/pdf/1008.3208.pdf (it's a well known result)? I had in mind:
* Test cases with subcases looping over n, k confirming that `G(2n, 2k + 1)` are bipartite
* Test cases with subcases looping over n, k confirming that `G(2n, 2k)`, `G(2n + 1, 2k)`, `G(2n + 1, 2k + 1)`  are not bipartite

tests/rustworkx_tests/digraph/test_bipartite.py
```diff
@@ -0,0 +1,59 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import unittest
+
+import rustworkx
+
+
+class TestBipartite(unittest.TestCase):
```
IvanIsCoding, line 18: Can you add some test cases using generators:
* `rustworkx.generators.directed_grid_graph` for any `(n, m)` for bipartite case
* `rustworkx.generators.directed_cycle_graph` with odd number of vertices for non bipartite case

IvanIsCoding (commented):
The algorithm LGTM but we could add some test cases using generators to have a more robust test suite.

I left comments for Generalized Petersen Graphs, Grid Graphs, and Odd Cycle Graphs

rustworkx-core/src/coloring.rs
```diff
@@ -13,13 +13,90 @@
...
+        + IntoNodeIdentifiers
+        + IntoNeighborsDirected
+        + GraphBase
+        + GraphProp
+        + NodeCount,
+    <G as GraphBase>::NodeId: std::cmp::Eq + Hash,
+{
+    let mut colors = DictMap::with_capacity(graph.node_count());
+    for node in graph.node_identifiers() {
+        if colors.contains_key(&node) {
```
alexanderivrii, line 73: If I understand correctly, this outer loop allows to handle disconnected graphs. Why do we need an explicit treatments for `isolates` (wouldn't they be assigned color `1`)?

rustworkx-core/src/coloring.rs
```diff
@@ -13,13 +13,90 @@
...
+/// let coloring = two_color(&graph).unwrap();
+/// let mut expected_colors = DictMap::new();
+/// expected_colors.insert(NodeIndex::new(0), 1);
+/// expected_colors.insert(NodeIndex::new(1), 0);
+/// expected_colors.insert(NodeIndex::new(2), 1);
+/// expected_colors.insert(NodeIndex::new(3), 0);
+/// expected_colors.insert(NodeIndex::new(4), 1);
+/// assert_eq!(coloring, expected_colors)
+/// ```
+pub fn two_color<G>(graph: G) -> Option<DictMap<G::NodeId, u8>>
```
alexanderivrii, line 61: A general question about Rust: is there a performance penalty from using `DictMap` vs a vector (indexed by NodeIndex or something), i.e. is lookup / insertion into `DictMap` of order `O(1)` or of `O(log n)`?

mtreinish: `DictMap` is basically just a `IndexMap` [alias](https://docs.rs/rustworkx-core/latest/rustworkx_core/dictmap/type.DictMap.html) with a custom default hash algorithm. So lookup/insertion is `O(1)` but unlike a `HashMap` it retains insertion order. I went for using a dict map here because there is no guarantee that the node indices are contiguous or a sequence. The simplest example of that is if the input graph is a `StableGraph` with nodes removed in the middle. In general I'd expect a `Vec` would be faster here, but since we need this function to be general to handle all sorts of different types of graphs this seemed like the best option for the return type.

rustworkx-core/src/coloring.rs
mtreinish: That's a fair point, we probably don't need to I did it to assign the nodes color 0, but 1 is just a valid answer for an isolated node so we should just simplify the code to make it faster.

tests/rustworkx_tests/digraph/test_bipartite.py
mtreinish: Done in: https://github.com/Qiskit/rustworkx/pull/1002/commits/0b94c19b68cde43a509394b955046950bd98d9ea

tests/rustworkx_tests/graph/test_bipartite.py
mtreinish: Done in: https://github.com/Qiskit/rustworkx/pull/1002/commits/0b94c19b68cde43a509394b955046950bd98d9ea

rustworkx-core/src/coloring.rs
mtreinish: Done in: https://github.com/Qiskit/rustworkx/pull/1002/commits/4c85c9bc0c9404cb4722626913edd3f685f6acd3

rustworkx-core/src/coloring.rs
IvanIsCoding: It’s slower than a vector but still pretty fast (`O(1)`). It keeps the insertion order but that is what users expect from a Python dictionary-like structure

IvanIsCoding (approved):

Issue #1000, opened 2023-10-09
Reserved for 1.0.0 release


IvanIsCoding:
### What is the expected enhancement?

#1000!

PR #1004, opened 2023-10-15
Ramp up deprecation of retworkx package and remove tests

mtreinish:
Since it's been > 1 year since we renamed retworkx to rustworkx, this commit starts actively emitting `DeprecationWarning`s for importing retworkx. Previously we had only documented the name as being deprecated, however we still ~54k downloads per month. The warning should hopefully encourage more people to move to the new name.

While we prepare to drop support for the retworkx name in the lead up to rustworkx 1.0 eventually (see #1000) this commit also prepares us to stop publishing updates in the future. Besides removing the warning it decreases the prominence of the rename warning from the rustworkx package documentation. This commit also remove the retworkx legacy tests from the test suite as we've been running them for > 1 year now and have not enountered any issues with the import redirect. The benefits of keeping around a duplicated test suite for validating the redirects has outlived it's usefulness, and is just wasted CPU resources (both locally and more importantly in CI).

As for the final step of the retworkx name, to eventually drop support for the legacy name, a comment is added to the setup.py about how we make that migration. For the final 0.X release of rustworkx we should change the `rustworkx==` pin in retworkx's setup.py to `rustworkx<1`. We do have the option to do this at anytime because the retworkx package source code is basically frozen and just redirecting imports from retworkx to rustworkx. However, if the retworkx package wasn't pinned to a fixed version then current users would potentially get a different experience when installing the same version of the retworkx package (i.e. if a user installed retworkx 0.14.0 tomorrow and then again 6 months later they could be using different rustworkx versions and encounter different API bugs or deprecation warnings).


IvanIsCoding (approved):
LGTM, it's been a year since we switched the name. The downloads for `retworkx` are steady but they (fortunately) have started becoming a smaller fraction of the total `rustworkx` download count. 

I also suspect that many projects have `retworkx` pinned to an older version, 0.12 specifically. For those cases the warning is a reminder to do the simple change of `retworkx -> rustworkx`

PR #1005, opened 2023-10-15
Update pyo3 and numpy to 0.20.0

mtreinish:
This commit updates PyO3 to the latest release 0.20.0 [1] and also moves rust-numpy to 0.20.0 [2] as well. The major change for rustworkx is official support for Python 3.12.0 final. We were previously using the 3.12 pre-release support in PyO3 to publish rustworkx for 3.12, but moving to the latest PyO3 officially supports the stable releases of 3.12. The are some small changes needed to adapt to API changes in PyO3, mostly around updates to the `PyDict.get_item()` api.

[1] https://github.com/PyO3/pyo3/releases/tag/v0.20.0
[2] https://github.com/PyO3/rust-numpy/releases/tag/v0.20.0


IvanIsCoding (approved):
LGTM, it's not surprising the functions that use Python's APIs the most are the one that change from release to release. Most of our Rust code remains intact fortunately.

Also, on the topic of NumPy, I think we should consider https://github.com/numpy/numpy/issues/24300. We probably should add restrictions to only depend on NumPy 1.0 until we confirm NumPy 2.0 works

Cargo.toml
```diff
@@ -29,7 +29,7 @@ fixedbitset = "0.4.2"
 hashbrown = { version = ">=0.13, <0.15", features = ["rayon"] }
 indexmap = { version = ">=1.9, <3", features = ["rayon"] }
 num-traits = "0.2"
-numpy = "0.19.0"
+numpy = "0.20.0"
```
IvanIsCoding, line 32: Not this file but copy https://github.com/Qiskit/qiskit/pull/10893 for us esssentially

mtreinish: Yeah, that sounds good. We definitely should do this for the 0.14.0 release.

PR #1006, opened 2023-10-15
Pivot to from flake8 to ruff for Python linting

mtreinish:
Since the start of the 0.3.x release of the retworkx project we have been using flake8 for Python linting and this has worked well for us. However, recently a new project, ruff [1][2], has shown increasing popularity becuase it provides the same linting coverage but is signficantly faster. Also just from a language solidatry PoV ruff is also written in rust. This commit migrates our linting jobs to use ruff instead of flake8 and sets up a reasonable set of rules as a starting point, which includes the rules equivalent to what we were using before with flake8, pyupgrade, and flake8-pyi, and flake8-quotes.

[1] https://docs.astral.sh/ruff/
[2] https://github.com/astral-sh/ruff


pyproject.toml
```diff
@@ -6,6 +6,35 @@ build-backend = "setuptools.build_meta"
 line-length = 100
 target-version = ['py38', 'py39', 'py310', 'py311']
 
+[tool.ruff]
```
IvanIsCoding, line 9: This is really convenient and will avoid the hacks we had in #861

rustworkx/digraph.pyi
```diff
@@ -13,7 +13,7 @@ import numpy as np
 from .iterators import *
 from .graph import PyGraph
 
-from typing import Any, Callable, Dict, Generic, TypeVar, Optional, List, Tuple, Sequence
```
IvanIsCoding, line 16: Unfortunately this is 3.9+ only and our stubs support Python 3.8. Is it hard to disable the Ruff warning for those?

pyproject.toml
```diff
@@ -6,6 +6,35 @@ build-backend = "setuptools.build_meta"
...
+    "doc",
+    "build",
+]
+
+[tool.ruff.per-file-ignores]
+"rustworkx/__init__.py" = ["F405", "F403"]
+"rustworkx/__init__.pyi" = ["F403", "F405", "PYI001"]
+"rustworkx/digraph.pyi" = ["F403", "F405", "PYI001"]
+"rustworkx/graph.pyi" = ["F403", "F405", "PYI001"]
+"rustworkx/iterators.pyi" = ["F403", "F405", "PYI001"]
```
IvanIsCoding, line 35: I should fix `PYI001` in those files, it would avoid us having to define `__all__` in them

IvanIsCoding (commented):
This is great and it is a great addition in my opinion. `ruff` seems to work outside-of-the-box with `black` and the fact that it integrates with `pyproject.toml` outside of the box is nice.

Just three opinions:
* For the changes in the `.pyi` to use the non-deprecated aliases, it's worth changing BUT Python 3.8 holds us back
* I agree with `PYI001` warnings and I think we should fix them in a separate PR
* We should put this on hold until #1004 is merged to avoid merge conflict headaches

rustworkx/digraph.pyi
mtreinish: Is it? I have the pyupgrade rules set to target 3.8 and this was flagged by those rules. It also passed all the CI tests.

IvanIsCoding: It is because of [PEP 585](https://peps.python.org/pep-0585/). But we have the future import so it doesn’t matter!

IvanIsCoding (approved):
LGTM, we have the `from __future__ import annotations` so the change to typing stubs are fine

Issue #978, opened 2023-09-07
Add a function to get edge indices from endpoints


lukepmccombs:
### What is the expected enhancement?

Users should be able to get a list of edges when given two nodes. With endpoints, edges can be removed, updated, and accessed, however the specific edge affected is unknown in multigraphs. While it is suggested to use indices instead, there is no easy way to get the list of indices that exist between the two nodes.

mtreinish:
That's a good point, we should have a method on `PyGraph` and `PyDigraph` something like `edge_indices_from_endpoints` to look up all the edge indices from a given endpoint pair. There are ways you can get this today either using `incident_edges()` or using `edge_index_map` and filtering the output. But that'd be inefficient if you only care about the edges between two nodes as you'd have to iterate over all the edges they return to find what you care about.

mario-s:
Just for my understanding, the method signature should look somehow like the following? 
`pub fn edge_indices_from_endpoints(&self, start_node: usize, end_node: usize) -> EdgeIndices`

PR #1011, opened 2023-10-19
Add methods for getting an edge index from endpoints.

lukepmccombs:
As in title, adds methods to `PyDiGraph` and `PyGraph` to get the index of an edge found at the given endpoints. Uses the same functions internally as `update_edge`, so it should return the index of said edge if its deterministic and neither changes.
Not exactly the same as I requested in #978, but it's suitable for non-multigraphs.


src/digraph.rs
```diff
@@ -757,6 +757,27 @@ impl PyDiGraph {
...
+    #[pyo3(text_signature = "(self, node_a, node_b, /)")]
+    pub fn get_edge_index(&self, node_a: usize, node_b: usize) -> PyResult<usize> {
+        let index_a = NodeIndex::new(node_a);
+        let index_b = NodeIndex::new(node_b);
+        let edge_index = match self.graph.find_edge(index_a, index_b) {
+            Some(edge_index) => edge_index,
+            None => return Err(NoEdgeBetweenNodes::new_err("No edge found between nodes")),
+        };
+        Ok(edge_index.index())
+    }
```
enavarro51, line 779: It seems like it would be better to generalize this so that all edges between two endpoints are returned. I think you could do something like
```
self.graph.edges(index_a).filter(|edge| edge.target() == index_b).collect();
```
It's similar to `get_all_edge_data` except you'd return the edge indices instead of the data. There's a `get_edge_data` function also which only gets one edge, so you could do 2 functions if you wanted, `get_edge_index` and `get_all_edge_indices`.

mtreinish:
Thanks for opening a PR and contributing to rustworkx. Since you opened this PR someone else opened up a new PR #1055 which implemented the same feature but that more closely aligned with the API outlined in #978 and also factored in the feedback left on this PR. While we had been waiting for updates to this PR for a few months based on the initial review feedback I went ahead with #1055 and it has been merged to main as it was ready to go and I think this PR is now superseded as the API to find an edge index from an endpoint is now already present. Accordingly I'm going to close this PR as I don't think there is anything left to do here.

 I hope you are ok with this decision, and please feel free to contribute more to rustworkx in the future, your contributions to the project are very appreciated. Also please re-open this PR if you feel that I'm mistaken and there is more to do on this pull request or there is more to discuss here.

PR #1012, opened 2023-10-19
Pin Numpy < 2 in requirements

mtreinish:
We have compiled extensions that are built against the Numpy C API. The Numpy transition guide for developers of downstream packages[1] encourages us to put in this pin until we have wheels built against the newer version, which is not expected to be fully ABI compatible (but extensions built against Numpy 2 _should_ work with older versions of Numpy).

Note that this won't prevent package managers from resolving _older_ versions of rustworkx (which don't have the pin) along with Numpy 2, but there's not a vast amount we can do about that now.


IvanIsCoding (approved):
LGTM

PR #1015, opened 2023-10-22
fix `all_simple_paths`

### LINKED ISSUE #955: rustworkx.all_simple_paths returns empty list when the min_depth argument is 0
### Information

- **rustworkx version**: 0.13.0
- **Python version**: 3.9
- **Rust version**: 1.71.0
- **Operating system**: FreeBSD 13.2

empty list is returned

correct list is returned

```len(rx.all_simple_paths(G, 1, 2, 0, 25))``` for any graph

The description says: ```min_depth (int) – The minimum depth of the path to include in the output list of paths. By default all paths are included regardless of depth, setting to 0 will behave like the default.```

The default is None, but the 0 value doesn't behave as the default value.

prakharb10:
Fixes #955

### DIFF
--- a/releasenotes/notes/fix-all-simple-paths-03f4534438d1068d.yaml
+++ b/releasenotes/notes/fix-all-simple-paths-03f4534438d1068d.yaml
@@ -1,0 +1,7 @@
+---
+fixes:
+  - |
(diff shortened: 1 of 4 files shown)


IvanIsCoding (commented):
LGTM, just add the reno and it''s good to merge!

IvanIsCoding (approved):

Issue #933, opened 2023-07-18
a new algorithm that returns all possible shortest paths for a source target pair


rok24:
At the moment digraph_dijkstra_shortest_paths returns only one shortest path for a source-target pair. It would be nice to see all the possible shortest paths with the same length for a source-target pair. 
Currently networkx has this functionality in all_shortest_paths function and it returns all possible shortest paths.

prpr2770:
@mtreinish I'd like to work on this problem. Please let me know how I may proceed, as I'm new to this project.

mtreinish:
Sure thing, I've assigned you the issue. The contributing guide: https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md has some general guidelines for writing PRs. Besides that in this case you'll want to add 2 new functions to https://github.com/Qiskit/rustworkx/blob/main/src/shortest_path/mod.rs that computes all the shortest paths between two nodes for both graphs and digraphs

prpr2770:
@mtreinish Thank you for assigning this and sharing the resources. I shall send you an update as soon as its completed.

lucasvanmol:
Hi @mtreinish, since this hasn't seen much activity recently, could I take this over? I have something that works, but would like some input on the best way to implement it.

mtreinish:
@lucasvanmol sure, this has been idle for > 2 months so I've reassigned the issue to you. I'll try to take a look at your WIP PR soon.

IvanIsCoding:
Closed by #1017

PR #1017, opened 2023-10-23
`all_shortest_paths` function

lucasvanmol:
- [x] I ran rustfmt locally

This (WIP) pull request aims to implement an `all_shortest_paths` algorithm as discussed in #933.

This is a draft pull request because I would like some input on how this is best implemented. 

The most straightforward way to implement such a function would not require a change to `rustworkx-core`; simply run Dijkstra's then look hard enough at the `scores` DistanceMap returned by Dijkstra's in order to find all possible shortest paths. However, this adds some unnecessary complexity, as we could instead keep track of this information while Dijkstra's is running. 

In this initial commit I have included a way this could be implemented by changing the `dijkstra` function in `rustworkx-core`. At this point, it does change the public rust interface, hence why I'm asking for input. Would the best way forward be a separate function? If so should this be in a separate file? Or is it not worth it, and I should just implement the approach of the previous paragraph?


rustworkx-core/src/shortest_path/dijkstra.rs
```diff
@@ -102,7 +102,7 @@ pub fn dijkstra<G, F, K, E, S>(
     start: G::NodeId,
     goal: Option<G::NodeId>,
     mut edge_cost: F,
-    mut path: Option<&mut DictMap<G::NodeId, Vec<G::NodeId>>>,
+    mut path: Option<&mut DictMap<G::NodeId, Vec<Vec<G::NodeId>>>>,
```
IvanIsCoding, line 105: You should create a new function instead of altering the existing one. This would be a breaking change

IvanIsCoding (commented):
Some thoughts on this early draft. I don't think the approach you chose is the best way to implement this.

`all_shortest_paths` can return an exponential number of paths, hence it needs to be a separate function. It can be possibly much slower, which would impact existing users of our Dijkstra function.

Another detail is that currently the code computes a lot of paths that are not necessary. The simplest way to find if an edge `(u, v)` is in a shortest path from `s` to `t` is to check if `d(s, u) + w(u, v) + d(v, t) == d(s, t)`. You can perform two shortest path calls, one in the graph and another in the graph with reversed direction. Then, you could write either a backtracking code or iteration-based code to generate all shortest paths.

Pseudo-code for the best way to implement would be along the lines:

```
def all_shortest_paths(graph, s, t):
    d_fwrd = dijkstra(graph, s)
    d_bckwrd = dijkstra(graph.reverse(), t)
    w = weights(graph)
     
     shortest_paths = [] # list containing the answer
     current_paths = [[s]] #541 
     for k in range(len(graph) - 1):
         next_paths = []
         for path in current_paths:
             u = path[-1]
             for v in neighbors(u):
                 if d_fwrd[u] + w[u][v] + d_bckwrd[v] == d_fwrd[t]:
                      new_path = path + [v]
                      if v == t:
                          shortest_pathd.append(new_path)
                      else:
                          next_paths.append(new_path)
         current_paths = next_paths
    
   return shortest_paths   
```

There are some optimizations that can be done such as filtering the edges in advance to only include edges that are in at least one shortest path, using persistent data structures to avoid copying the array, etc.

lucasvanmol:
Hi, thanks for the valuable and detailed feedback.

I've updated the PR with your feedback in mind, however I haven't implemented the `digraph` function yet, as I have some more comments before proceeding:

- The algorithm I'm using only calls dijkstra once and then backtracks, as opposed to what you proposed. I'm hoping my implementation is sound?
- I've implemented it such that a source and target are always required. This is a bit different to the optional target argument of, for example, the dijkstra function. I'm wondering if this is ok, or if you think users will want the ability to find every possible shortest path between source and all other nodes.
- I'm not sure about the return type yet (also partly due to the above comment), both for the `rustworkx-core` function and the python function. Additionally, should I return the shortest path length as well as all the paths? Maybe a new class similar to `MulitplePathMapping` is needed?

rustworkx-core/src/shortest_path/all_shortest_paths.rs
```diff
@@ -0,0 +1,207 @@
...
+    let path = VecDeque::from([goal]);
+    let mut queue = vec![(goal, path)];
+    while let Some((curr, curr_path)) = queue.pop() {
+        let curr_dist = *scores.get(&curr).unwrap();
+        for edge in graph.edges_directed(curr, Incoming) {
+            let next_dist = match scores.get(&edge.source()) {
+                Some(x) => *x,
+                None => continue,
+            };
+            if curr_dist == next_dist + edge_cost(edge)? {
```
IvanIsCoding, line 113: This is fine but I wonder if just generating the graph with these edges + calling `all_simple_paths` that we already implemented does the job

IvanIsCoding (commented):
Overall the implementation LGTM, I think the backtracking is fine. Maybe we could try to reuse the `all_simple_paths` code

I left some comments on the implementation. I will revisit the tests later, but we should focus on expanding it

IvanIsCoding:
> Hi, thanks for the valuable and detailed feedback.
> 
> I've updated the PR with your feedback in mind, however I haven't implemented the `digraph` function yet, as I have some more comments before proceeding:
> 
> * The algorithm I'm using only calls dijkstra once and then backtracks, as opposed to what you proposed. I'm hoping my implementation is sound?
> * I've implemented it such that a source and target are always required. This is a bit different to the optional target argument of, for example, the dijkstra function. I'm wondering if this is ok, or if you think users will want the ability to find every possible shortest path between source and all other nodes.
> * I'm not sure about the return type yet (also partly due to the above comment), both for the `rustworkx-core` function and the python function. Additionally, should I return the shortest path length as well as all the paths? Maybe a new class similar to `MulitplePathMapping` is needed?

Also, the return type should be just a list of lists, like `all_simple_paths`. Goal should be a required argument because it's used to generate the paths, `all_pairs_all_shortest_paths` is content for a separate function

rustworkx-core/src/shortest_path/all_shortest_paths.rs
```diff
@@ -0,0 +1,207 @@
...
+    goal: G::NodeId,
+    mut edge_cost: F,
+) -> Result<HashSet<Vec<G::NodeId>>, E>
+where
+    G: IntoEdgesDirected + Visitable + NodeIndexable,
+    G::NodeId: Eq + Hash,
+    F: FnMut(G::EdgeRef) -> Result<K, E>,
+    K: Measure + Copy,
+{
+    let scores: DictMap<G::NodeId, K> = dijkstra(&graph, start, Some(goal), &mut edge_cost, None)?;
```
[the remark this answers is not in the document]
lucasvanmol: Looking briefly at the code and performing some (not quite rigorous) tests, it was my understanding that only scores that aren't part of any shortest path will be left unfilled. I'll try to come up with a proper proof and come back to you

rustworkx-core/src/shortest_path/all_shortest_paths.rs
lucasvanmol: I think this is possible, but since we need to backtrack anyway, we're already doing most of the work needed to generate the paths. Calling `all_simple_paths` would mean enumerating over all shortest path edges once more. But if you prefer to return an `Impl Iterator<>` I can definitely change this.

lucasvanmol (commented):
Thanks for the feedback. Sorry I was away for a little way as I had exams. I've added some comments and started work on more tests on the python side of things.

IvanIsCoding:
> Thanks for the feedback. Sorry I was away for a little way as I had exams. I've added some comments and started work on more tests on the python side of things.

With regards to the scores: it’s a valid assumption if the weights are positive. We don’t support negative weight in Dijkstra. However, you need to keep in mind that there can be edges with weight zero.

It’s worth adding this test case:
```
A -> B with weight 0
A -> C with weight 0
B -> C with weight 0
```

The code could halt and never calculate a distance for C. But the correct answers are A->B and A->C->B

lucasvanmol:
Alright, so I've done some extra digging for the `Some(goal)` argument to `dijkstra`. In general, this is fine *if all edges are strictly positive*. We can prove it as such:

Suppose we have a graph G=(V, E), where each edge $e \in E$ has some associated weight $w_e > 0$. Take some starting node $s \in V$ and some goal $g \in V$, and let there be one or more shortest paths between $s$ and $g$ of length $d$. We can write any such path as a sequence of $n$ vertices:

$$s, x_1, ..., x_{n-2}, g$$

Now `dijkstra` gives us a $\text{score}()$ for each of these vertices $x_i$ which determines the minumum distance from $s$ to $x_i$. Because all edges have a weight $> 0$:

$$\text{score}(s) = 0 < \text{score}(x_1) < .... < \text{score}(x_{n-2}) < \text{score}(g) = d$$

In this Dijkstra implementation, whenever a `(node_score, node)` pair is popped off the heap, this `node_score` is the lowest score possible for this node, i.e. $\text{score}($ `node` $)=$ `node_score`. Hence, the nodes are evaluated in ascending score order thanks to the minheap, meaning any $x_i$ in a shortest path will be evaluated before $g$.

However, this argument breaks down if we allow weights = 0, since a node with equal score to $g$ isn't necessarily explored before $g$. To illustrate using a counter example, the following graph doesn't explore all the nodes of all shortest paths:

```mermaid
graph LR;
  a((a)) -->|1| b((b)) ---->|2| f((f));
  a((a)) -->|2| c((c)) -->|1| d((d)) -->|0| e((e)) -->|0| f((f));
```

This occurs because $d, e, f$ all have the same score, and running `dijkstra` on the graph with the goal of `Some(f)` doesn't fill in a score for `e`, meaning the path `a, c, d, e, f` is not found. 

Therefore, I think there are a few options:

1. Don't allow the function to be called if there are 0-weight edges. This is a bit different from `dijkstra` which allows non-negative weights. This solution could also fix issues with 0 weight cycles that could create an infinite amount of shortest paths. If this option is chosen, should there be some sort of check for 0 weight edges, or simply run the algorithm (returning a subset of all paths), while simply mentioning to the user that 0 weight edges aren't supported?
2. Allow 0-weight edges. This would either force us to use `None` for the goal argument in `dijkstra` which can slow down things down, or tweak `dijkstra` to make sure $g$ is explored after any equal-weight nodes (I'd also need to change the logic a bit for backtracking to keep track of visited edges). Still need to deal with 0-weight cycles somehow.
3. Treat 0-weight edge graphs differently, for example using Bellman-Ford. This increases the scope a bit but could also be used to find all shortest paths in graphs with negative edges.

lucasvanmol:
Indeed, the following graph doesn't explore all the nodes of all shortest paths:

```mermaid
graph LR;
  a((a)) -->|1| b((b)) ---->|2| f((f));
  a((a)) -->|2| c((c)) -->|1| d((d)) -->|0| e((e)) -->|0| f((f));
```

The path `a, c, d, e, f` is not found. 

However, I'm not sure what the best way to fix this is because this means graphs can be created using 0 weight cycles which can create infinite paths.

Therefore, I think there are three options:

1. Don't allow the function to be called if there are 0-weight edges. If this option is chosen, should there be some sort of check for 0 weight edges, or simply run the algorithm (returning a subset of all paths), while documenting to the user that 0 weight edges aren't supported?
2. Allow 0-weight edges. This would either force us to use `None` for the goal argument in `dijkstra` which can slow down things down, or change/copy `dijkstra` to make sure the goal node `f` is explored _after_ any equal-weight nodes (I'd also need to change the logic a bit for backtracking to keep track of visited edges). Still need to deal with 0-weight cycles somehow.
3. Maybe use something like Bellman-Ford instead?

IvanIsCoding:
> Indeed, the following graph doesn't explore all the nodes of all shortest paths:
> 
> ```mermaid
> graph LR;
>   a((a)) -->|1| b((b)) ---->|2| f((f));
>   a((a)) -->|2| c((c)) -->|1| d((d)) -->|0| e((e)) -->|0| f((f));
> ```
> The path `a, c, d, e, f` is not found.
> 
> However, I'm not sure what the best way to fix this is because this means graphs can be created using 0 weight cycles which can create infinite paths.
> 
> I see networkx only returns simple paths with no repeated nodes, so it's pretty easy to do the same. Hence using `None` as the goal arg and returning simple paths works for 0 weight edges (with or without cycles).
> 
> However, I think there could be a substantial performance increase for using a version of the dijkstra algorithm which guarantees all possible shortest path nodes are scored, and terminates early. Essentially the only occasion when passing the argument `Some(goal)` does not find solutions is the situation above, when, in addition to some other shortest path, there is another shortest path ending in a chain of 0 weight edges towards the goal. Such a modification to `dijkstra` would have to make sure some goal node `f` is explored _after_ any equal-weight nodes, as a tiebreaking rule.

I think returning all simple paths is fine. I am not too concerned about optimizations for early return on this function as the number of shortest path could be exponential, so calculating all the shortest distances to all vertices is pretty reasonable

lucasvanmol:
I've implemented your suggestions, and at this point it feels like it's no longer a [WIP] so have updated the PR accordingly. If you'd like me to squash the commits please let me know.

IvanIsCoding (approved):
LGTM! I am excited to see this come out on the next release

PR #1030, opened 2023-11-21
Fix clippy findings for Rust 1.74

IvanIsCoding:
Some minor improvements suggested by clippy. They are currently blocking all CI actions


mtreinish (approved):
LGTM, thanks for fixing this. Although I thought we pinned clippy to our msrv to avoid having to do this every release.

Issue #1035, opened 2023-12-01
Migrate to Github Pages and Regenerate historical docs


Eric-Arellano:
Per https://github.com/Qiskit/ecosystem/issues/611, we accidentally deleted all ecosystem subsites when sunsetting qiskit.org.

Rustworkx is live, but the historical API docs are not yet for 0.8 - 0.13. 

If any of these fail to build, we should modify this to start later and redeploy the docs for the current version, so that the bad release doesn't show up in the left sidebar:

https://github.com/Qiskit/rustworkx/blob/4838d399e4be4623314a3dffcbb0ddd6b8021d20/docs/source/conf.py#L136

IvanIsCoding:
> Hey @IvanIsCoding, how did you deploy to https://qiskit.github.io/rustworkx/ with both the un-versioned docs and also 0.13 docs at https://qiskit.github.io/rustworkx/stable/0.13? I don't see the GH Pages code for it.
> 
> Specifically, I'm wondering if you had to upload the two subsites in one distinct run? Or you could do them separately, like first deploy unversioned, then later add `/stable/0.13`? I imagine adding `/stable/0.13` on its own would overwrite the existing site.

I manually created the folder under the `gh-pages` branch. Essentially, the workflow to deploy the latest version needs to always keep the `stable/` folder. And the workflow to deploy a stable version shouldn't touch the latest code and should just create a new folder

IvanIsCoding:
Concluded with #1037 and #1052

PR #1037, opened 2023-12-07
[WIP] Deploy latest documentation to Github Pages

IvanIsCoding:
Related to #1035

### DIFF
--- a/.github/workflows/docs_dev.yml
+++ b/.github/workflows/docs_dev.yml
@@ -2,11 +2,11 @@ name: Docs Publish
 on:
   workflow_dispatch:
   push:
-    branches: [ main ]
+    branches: [ push-to-gh-pages ]
 
 jobs:
   deploy:
-    if: github.repository_owner == 'Qiskit'
+    if: github.repository_owner == 'IvanIsCoding'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


tools/build_documentation_release.sh
```diff
@@ -0,0 +1,36 @@
...
+# Extract the release version from Cargo.toml
+VERSION=$(grep -Po -m1 'version = "\K[0-9]+\.[0-9]+' Cargo.toml)
+
+TMP_DIR=$(mktemp -d)
+git clone --depth 1 --branch gh-pages https://github.com/Qiskit/rustworkx.git $TMP_DIR
+rm -rf $TMP_DIR/stable/$VERSION # Remove the old version in case of a revision release
+mkdir -p $TMP_DIR/stable/$VERSION
+cp -r docs/build/html/* $TMP_DIR/stable/$VERSION/
+mkdir -p release_docs
+cp -r $TMP_DIR/* release_docs/
```
mtreinish, line 32: Doesn't this mean the root of the docs becomes the dev build and we only see the released documentation on `stable/$VERSION`? The current hosting has the root documentation as the current release, `/dev` for the dev builds and `/stable/*` for all the historical builds (including the current release).

mtreinish (commented):
Overall this looks fine to me, just a quick inline question about the mechanics of the structure we're uploading and what we want our hosting to look like.

tools/build_documentation_release.sh
IvanIsCoding: That is correct, dev would be at the root. I can try changing it to stable at the root later but it gets complicated. Especially because we have no sidebar button to go to the dev version

IvanIsCoding: I am updating the code to mimic the old behaviour, it shouldn't be too hard. I will put it on hold temporarily

IvanIsCoding:
@mtreinish I addressed the comments, it should have the latest stable version at the root now

PR #1040, opened 2023-12-17
Save next nodes in Floyd-Warshall numpy variants for path reconstruction

kazuki0824:
This PR adds "next node" matrix to the Floyd-Warshall implementations' return value.
This helps re-construct the shortest path from the result.

`digraph_floyd_warshall_numpy` and `graph_floyd_warshall_numpy` return **two** variable by applying this patch. In recent version, they returned only a length table.

The next node table is inherently dense, even though the graph is sparse. To not break the advantage of the implementation optimized for the sparse graph, I modified only numpy variants.

### DIFF
--- a/src/shortest_path/floyd_warshall.rs
+++ b/src/shortest_path/floyd_warshall.rs
@@ -149,11 +149,18 @@ pub fn floyd_warshall_numpy<Ty: EdgeType>(
     as_undirected: bool,
     default_weight: f64,
     parallel_threshold: usize,
-) -> PyResult<Array2<f64>> {
+) -> PyResult<(Array2<f64>, Array2<usize>)> {
     let n = graph.node_count();
     // Allocate empty matrix
     let mut mat = Array2::<f64>::from_elem((n, n), std::f64::INFINITY);
+    let mut next = Array2::<usize>::from_elem((n, n), 0);
 
+    // Build `next` matrix
+    for i in 0..n {
+        for j in 0..n {
+            next[[i, j]] = j;
+        }
+    }
     // Build adjacency matrix
     for (i, j, weight) in get_edge_iter_with_weights(graph) {
         let edge_weight = weight_callable(py, &weight_fn, &weight, default_weight)?;
@@ -176,6 +183,7 @@ pub fn floyd_warshall_numpy<Ty: EdgeType>(
                     let d_ijk = mat[[i, k]] + mat[[k, j]];
                     if d_ijk < mat[[i, j]] {
                         mat[[i, j]] = d_ijk;
+                        next[[i, j]] = next[[i, k]]
                     }
                 }
             }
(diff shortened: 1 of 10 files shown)


src/shortest_path/floyd_warshall.rs
```diff
@@ -149,11 +149,18 @@ pub fn floyd_warshall_numpy<Ty: EdgeType>(
     as_undirected: bool,
     default_weight: f64,
     parallel_threshold: usize,
-) -> PyResult<Array2<f64>> {
+) -> PyResult<(Array2<f64>, Array2<usize>)> {
```
IvanIsCoding, line 152: Can you separate this code into a new function?

IvanIsCoding (commented):
Thanks for making your first contribution! I agree that we can add predecessors to `floydwarshall_numpy`.

However, changing the return type is a breaking change. It will also fail all our tests, which should give you a hint that it is not exactly the change you're looking to make.

Can you create a new function `rustworkx.floyd_warshall_predecessor_and_distance` to incorporate the new code? It would be akin to  https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.shortest_paths.dense.floyd_warshall_predecessor_and_distance.html

kazuki0824:
@IvanIsCoding now that the new implementations are separated to not break existing APIs

kazuki0824:
If we choose predecessors to represent shortest paths, it may cause much more change of the parallelism implementation than the case we choose successor.

IvanIsCoding:
> If we choose predecessors to represent shortest paths, it may cause much more change of the parallelism implementation than the case we choose successors. That's why I don't use predecessors.

I won’t have time to review this week but I hope to get a review by January 2. In the meantime, I encourage you to write unit tests. You can simply copy the existing Floyd Warshall test cases and reuse them for the new methods

src/shortest_path/floyd_warshall.rs
IvanIsCoding, line 152: Can you duplicate this code? I think the overall implementation is fine but I don't want to add the extra overhead of keeping the predecessors for the existing `floyd_warshall_numpy`

IvanIsCoding (changes requested):
Overall the code looks correct, thanks for implementing it!

There are two items we need to address however:
* It would be best to split the code path for the predecessors from the code path for Floyd Warshall Numpy. The additional overhead of the extra memory doubles the memory consumption in a function that already uses O(n^2) memory. So I'd rather keep the constant down
* We need to add tests for correctness. See https://github.com/Qiskit/rustworkx/blob/main/tests/rustworkx_tests/graph/test_floyd_warshall.py and https://github.com/Qiskit/rustworkx/blob/main/tests/rustworkx_tests/digraph/test_floyd_warshall.py for examples

kazuki0824:
> It would be best to split the code path for the predecessors from the code path for Floyd Warshall Numpy. The additional overhead of the extra memory doubles the memory consumption in a function that already uses O(n^2) memory. So I'd rather keep the constant down

I got it and resolved by an extra argument `generate_successors`. The new `floyd_warshall_numpy` returns a pair, whose latter part is Option<Array2<usize>>, and it has content only if  `generate_successors` is true.

IvanIsCoding (approved):
LGTM, thanks for making your first contribution!

I am going to merge this soon, but as a last step can you add it to the docs? These are the files you need to change:
```
docs/source/api/algorithm_functions/shortest_paths.rst
docs/source/api/pygraph_api_functions.rst
docs/source/api/pydigraph_api_functions.rst
```

You will also need to write a short docstring adding a description of the function you implemented. I can do it for you if you want

kazuki0824:
@IvanIsCoding Can you please tell me the file location where to add the docstring?

IvanIsCoding:
> @IvanIsCoding Can you please tell me the file location where to add a docstring?

You add a comment in the Rust code. See the Floyd Warshall one for example: https://github.com/Qiskit/rustworkx/blob/99f975662a72dbd879c086d37ea32d429f5ef07d/src/shortest_path/mod.rs#L868

kazuki0824:
@IvanIsCoding Thanks for your advice. It's done.

IvanIsCoding (approved):
LGTM, I will add the release notes and merge it today. Thanks for your contribution!

IvanIsCoding:
I don't have permission to edit your branch so I will merge this and add the release notes in a new PR

PR #1051, opened 2024-01-08
Fix URL in comment around origin of matplotlib drawer

mtreinish:
The rustworkx mpl_draw() function was originally adapted from networkx's matplotlib based drawer and heavily modified to work with rustworkx and also conform more to rustworkx's interface style. Attribution to the original source and the license terms of the original networkx code are present in a comment at a top of the module. This port was originally done when rustworkx was still called retworkx. As part of the name migration in #644 an overzealous find/replace broke the URL to the permalink to the original networkx source code that the mpl_draw() function was based on. This commit corrects this oversight so that the link correctly resolves to the proper url.


IvanIsCoding (approved):
LGTM

PR #1053, opened 2024-01-14
Add the function to get topological generations

### LINKED ISSUE #1014: Add an equivalent to networkx's topological_generations?
I find networkx's topological_generations https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.dag.topological_generations.html
to be very friendly for scheduling the parallelized nodes. Is there a plan to add this feature?

Thanks.

inmzhang:
Closes #1014


IvanIsCoding (approved):
LGTM, thanks for making your first contribution!

Overall the code looks good and I don't think there is a need to change anything for functionality. I left one optional comment about the DAG check.

~You also need to sign the CLA to make the CLA bot accept your PR~

src/dag_algo/mod.rs
```diff
@@ -428,6 +428,60 @@ pub fn lexicographical_topological_sort(
...
+/// :param PyDiGraph graph: The DAG to get the topological generations from
+///
+/// :returns: A list of topological generations.
+/// :rtype: list
+///
+/// :raises DAGHasCycle: if a cycle is encountered while sorting the graph
+#[pyfunction]
+#[pyo3(text_signature = "(dag, /)")]
+pub fn topological_generations(dag: &digraph::PyDiGraph) -> PyResult<Vec<NodeIndices>> {
+    if !is_directed_acyclic_graph(dag) {
```
IvanIsCoding, line 447: This check basically performs another topological sort. You can perform yours with the current code and if at the end it does not include all the nodes in the graph return that it was not a DAG

tests/rustworkx_tests/digraph/test_nodes.py
```diff
@@ -194,6 +194,34 @@ def test_topo_sort_with_cycle(self):
         dag.add_edge(node_b, node_a, {})
         self.assertRaises(rustworkx.DAGHasCycle, rustworkx.topological_sort, dag)
 
+    def test_topo_generations_empty(self):
```
IvanIsCoding, line 197: And maybe include a graph that is not a DAG to trigger the condition e.g. `1 <-> 2 -> 3`

inmzhang: I think `test_topo_generations_has_cycle` has already covered this?

src/dag_algo/mod.rs
inmzhang: Yeah I did not realize this, I made some change to the code based on your suggestion

PR #1054, opened 2024-01-16
Replace call to deprecated np.alltrue with np.all

eendebakpt:
The `np.alltrue` method will be removed in numpy 2.0.


mtreinish (approved):
LGTM, thanks for catching and fixing this. I also checked and this function existed in our minimum supported numpy version of 1.16.0 so there shouldn't be any change needed there either.

PR #1055, opened 2024-01-16
Add method to return edge indices from endpoints

### LINKED ISSUE #978: Add a function to get edge indices from endpoints

eendebakpt:
Fixes #978


mtreinish (commented):
Overall this LGTM, it does overlap with #1011 but that PR has been idle with open review comments for a couple months now and this addresses the feedback on that PR. I think the only thing still to do on this PR is fix the CI failure (it just needs to have black run as there is excess whitespace in the digraph tests) and also this could use a release note to document the new methods being added in this PR. The process for adding a release note is documented here: https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md#release-notes

src/digraph.rs
```diff
@@ -547,6 +547,25 @@ impl PyDiGraph {
         }
     }
 
+    /// Return a list of indices of all directed edges between specified nodes
+    ///
+    /// :returns: A list of all the edge indices connecting the specified start and end node
+    /// :rtype: EdgeIndices
+    #[pyo3(text_signature = "(self)")]
```
mtreinish, line 554: The text signature doesn't match the actual function signature and this is the root cause of the mypy stub failure. Removing the text signature should fix it (as pyo3 should auto generate it now if you don't provide it).
```suggestion
```

src/graph.rs
```diff
@@ -422,6 +422,25 @@ impl PyGraph {
         }
     }
 
+    /// Return a list of indices of all edges between specified nodes
+    ///
+    /// :returns: A list of all the edge indices connecting the specified start and end node
+    /// :rtype: EdgeIndices
+    #[pyo3(text_signature = "(self)")]
```
mtreinish, line 429:
```suggestion
```

mtreinish (approved):

PR #1056, opened 2024-01-16
simplify ruff exclude list

danieleades:


IvanIsCoding (commented):
I am debating about merging this. In one point it is less lines of code.

However, I had to read [ruff's documentation](https://docs.astral.sh/ruff/settings/#exclude) to understand why you used `extend-exclude` instead of `exclude`. So I think extending the defaults from ruff is clever but harder to maintain. I will let @mtreinish weight in

danieleades:
> I am debating about merging this. In one hand it is less lines of code.
> 
> However, I had to read [ruff's documentation](https://docs.astral.sh/ruff/settings/#exclude) to understand why you used `extend-exclude` instead of `exclude`. So I think extending the defaults from ruff is clever but harder to maintain. I will let @mtreinish weight in

I'm very sympathetic to this argument, i totally see where you're coming from.

That said, I note the following from the [ruff docs on the 'exclude' config](https://docs.astral.sh/ruff/settings/#exclude):

> Note that you'll typically want to use [extend-exclude](https://docs.astral.sh/ruff/settings/#extend-exclude) to modify the excluded paths.

The items in the default `exclude` list are super common in the python ecosystem, so i feel that explicitly listing them is unnecessary boilerplate. Using 'extend exclude' instead allows you to document what's *different* between `rustworkx` and a 'typical' python library.

mtreinish:
Personally I'm fine either way, my preference is along with Ivan's to be explicit in the config on all the paths to exclude because I don't know off the top of my head what ruff is excluding by default (I didn't even realize there were default exclude paths). But at the end of the day were talking about a super minor difference in the configuration and both are doing the right thing.

IvanIsCoding (approved):
I think it is fine to merge given we don’t change the list much

If ruff breaks us in the future this was the culprit

PR #1058, opened 2024-01-17
Add function to return the bridges of the graph

### LINKED ISSUE #1039: Feature: find bridges

I currently use networkx for a project, the only missing functionality preventing me from trying out this lib is a method for finding the bridges in a graph!

method in networkx: https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.bridges.bridges.html#networkx.algorithms.bridges.bridges

i found an implementation in this rust lib, but i dont know if youre open to using extra libs: https://docs.rs/graphalgs/latest/graphalgs/connect/fn.find_bridges.html

inmzhang:
Closes #1039


rustworkx-core/src/connectivity/biconnected.rs
```diff
@@ -68,17 +72,20 @@ fn is_root(parent: &[usize], u: usize) -> bool {
...
 /// assert_eq!(bicomp, HashMap::from_iter([
 ///     ((nx(0), nx(2)), 1), ((nx(2), nx(1)), 1), ((nx(1), nx(0)), 1),
 ///     ((nx(1), nx(3)), 0)
 /// ]));
+/// assert_eq!(bridges, HashSet::from_iter([(nx(1), nx(3))]));
 /// ```
 pub fn articulation_points<G>(
     graph: G,
     components: Option<&mut HashMap<Edge<G>, usize>>,
+    bridges: Option<&mut HashSet<Edge<G>>>,
```
IvanIsCoding, line 88: Can you keep the signature of `articulation_point` the same and create an internal function `_articulation_point` that is called by both `articulation_point` and `bridges`?

IvanIsCoding (commented):
Thanks! So overall it looks great, my only concern is that `articulation_point` is a public function in `rustworkx-core`, so changing the signature is a breaking change.

I'd rather not break our existing `rustworkx-core` users, so I am suggesting keeping the old signature while calling an internal function that does the work to find the bridges

IvanIsCoding (approved):
Just add the function to the documentation in https://github.com/Qiskit/rustworkx/blob/main/docs/source/api/algorithm_functions/connectivity_and_cycles.rst and we will be good to go

src/connectivity/mod.rs
```diff
@@ -921,6 +921,32 @@ pub fn articulation_points(graph: &graph::PyGraph) -> HashSet<usize> {
...
+/// A bridge is any edge whose removal increases the number of connected
+/// components of a graph.
+///
+/// .. note::
+///
+///     The function implicitly assumes that there are no parallel edges
+///     or self loops. It may produce incorrect/unexpected results if the
+///     input graph has self loops or parallel edges.
+///
+/// :param PyGraph: The undirected graph to be used.
```
IvanIsCoding, line 935: There is an indentation error probably around here, that is why the docs job is failing. You can reproduce the error locally with `tox -edocs`

IvanIsCoding (approved):

PR #1060, opened 2024-01-19
Add option to specify preset colors to graph_greedy_color

mtreinish:
This commit adds a new argument to the graph_greed_color function, which enables a user to provide a callback function that specifies a color to use for particular nodes.


IvanIsCoding (approved):
LGTM

PR #1061, opened 2024-01-19
Finalize type signatures and enforce for all future functions

mtreinish:
This commit adds the last few missing type hints (primarily the generators module) and switches the CI configuration over to enforce that the library is completely type hinted. This makes it a CI error if a function or method is added that doesn't include type hints. This is all built on the hard work of @IvanIsCoding who built up all the type hinting infrastructure and stub files for the majority of rustworkx. This is just the last piece to enforce the library is fully typed moving forward.

TODO:
- [x] Figure out how to teach mypy to discover the rustworkx.generators stubs correctly (and validate they're correct)
- [x] Fix `.to_dot()` signature so it doesn't error and that `graphviz_draw()`'s use of it passes mypy


rustworkx/visualization/matplotlib.pyi
```diff
@@ -0,0 +1,58 @@
...
+    label: str
+    connectionstyle: str
+
+def mpl_draw(
+    graph: PyGraph[_S, _T] | PyDiGraph[_S, _T],
+    pos: typing.Mapping[int, tuple[float, float]] | None = ...,
+    ax: Axes | None = ...,
+    arrows: bool = ...,
+    with_labels: bool = ...,
+    **kwds: typing.Unpack[_DrawKwargs],
```
mtreinish, line 57: `Unpack` was only added in Python 3.11, we'll have to use typing-extensions to get it backported for 3.8 through 3.10.

mtreinish: Done in: https://github.com/Qiskit/rustworkx/pull/1061/commits/b31d6dc639fb3cefe1714165cc10e9b7412778bb

rustworkx/__init__.pyi
```diff
@@ -15,213 +15,6 @@ import rustworkx.visit as visit
 from .rustworkx import *
 from typing import Generic, TypeVar, Any, Callable, Iterator, overload
 
-from .graph import PyGraph as PyGraph
```
IvanIsCoding, line 18: You are going to need to re-export those with `from .rustworkx import PyGraph as PyGraph` because of  #960

rustworkx/visualization/matplotlib.pyi
```diff
@@ -0,0 +1,61 @@
...
+    labels: typing.Callable[[_S], str]
+    edge_labels: typing.Callable[[_T], str]
+    font_size: int
+    font_color: str
+    font_weight: str
+    font_family: str
+    label: str
+    connectionstyle: str
+
+def mpl_draw(
```
IvanIsCoding, line 54: This is really awesome! I personally would have annotate just with `Any` because even matplotlib still lacks annotation suport nowadays

IvanIsCoding (commented):
Thanks for wrapping this up! Just some thoughts before merging:
* The re-exports are awful but #960 was a real reported issue and mypy does not re-export symbols. None of our users import from `rustworkx.rustworkx` so I guess we need the duplication at the root
* Maybe `from __future__ import annotations` fixes the errors in 3.8?
* We should add a release note to celebrate 😃  🍾

mtreinish:
> The re-exports are awful but https://github.com/Qiskit/rustworkx/issues/960 was a real reported issue and mypy does not re-export symbols. None of our users import from rustworkx.rustworkx so I guess we need the duplication at the root

Do you think we can get away from a `from .rustworkx import *` in the `__init__.pyi` stub? Or will I have to list every single function with an `as symbol`?

IvanIsCoding:
> > The re-exports are awful but #960 was a real reported issue and mypy does not re-export symbols. None of our users import from rustworkx.rustworkx so I guess we need the duplication at the root
> 
> Do you think we can get away from a `from .rustworkx import *` in the `__init__.pyi` stub? Or will I have to list every single function with an `as symbol`?

We need to re-export with the `as symbol`. If you import with the `as` it does not count unfortunately because mypy thinks it is not the intent

mtreinish:
> Maybe from __future__ import annotations fixes the errors in 3.8?

`ruff` didn't like this saying it was unnecessary because it's implicit in a stub file. The only way I found to get this working was to add a `# type: ignore` comment to each of the imports (like I did for the pillow import).

rustworkx/__init__.pyi
mtreinish: Done in: https://github.com/Qiskit/rustworkx/pull/1061/commits/85672db3dd3a87726856983c6996896e69c0f1d6

IvanIsCoding (approved):
LGTM, thanks!

PR #1062, opened 2024-01-21
Unnest rustworkx tests folder

IvanIsCoding:
This shortens our folder structure for rustworkx tests after #1004. Which is what we had on the retworkx days


mtreinish (approved):
Sure this makes sense to me. It didn't really bother me to have a layer of nesting, but I could see it being annoying, especially as we don't have any other test suites in the directory.

PR #1063, opened 2024-01-21
Update contribution guidelines around type annotations

mtreinish:
Since #1061 has merged rustworkx is now fully typed with mypy. This means that adding type annotations is no longer optional and is now required for all new function. This commit updates the contribution guidelines to indicate type annotations are required now. The details are also updated a bit to provide a bit more detail on what's required when adding type annotations to rustworkx.


IvanIsCoding (approved):
LGTM, thanks for updating the guide

PR #1065, opened 2024-01-22
Prepare 0.14.0 release

mtreinish:
This commit prepares the 0.14.0 release. It moves the release notes for the release into a separate directory and adds a prelude to the release notes. This should be the last commit to merge for the 0.14.0 release and after it merges that commit will be what is should be tagged as the 0.14.0 release.

As part of this there are few small changes other changes being made. The release notes are update to have consistent formatting and wording, a missing universal function for `all_shortest_paths` is added, missing doc toctree entries are added, and the ahash version in the lock file is downgraded to 0.8.6 as 0.8.7 has an undocumented MSRV on Rust 1.72 when building for aarch64 CPU architectures which is incompatible with rustworkx's documented MSRV.

TODO:
 - [x] Wait for last milestone target PRs to merge


docs/source/api/algorithm_functions/connectivity_and_cycles.rst
```diff
@@ -25,3 +25,6 @@ Connectivity and Cycles
    rustworkx.all_pairs_all_simple_paths
    rustworkx.stoer_wagner_min_cut
    rustworkx.longest_simple_path
+   rustworkx.is_bipartite
+   rustworkx.isolates
+   rustworkx.has_path
```
IvanIsCoding, line 30: I will try to catch the documentation entries more often, this release we let many slip through

IvanIsCoding (commented):
LGTM, thanks for writing the release notes and compiling all the functions that were not in the documentation! I feel that is harder to keep track of them now that we no longer have a single `api.rst`, I might try to come up with something to remind us of the documentations

IvanIsCoding (approved):

Issue #1041, opened 2023-12-18
Switch from `unittest` to `stestr` runner after Windows 3.12 bug is fixed


IvanIsCoding:
### Information

- **rustworkx version**: main
- **Python version**: 3.12
- **Rust version**: Stable
- **Operating system**: Windows

### What is the current behavior?

The test passes when running with `python -m unittest discover` and fail with `stestr run`. This indicates a possible issue with the test runner

### Steps to reproduce the problem

Compare the result of this Github Action: https://github.com/Qiskit/rustworkx/actions/runs/7243374909
Against this Github Action: https://github.com/Qiskit/rustworkx/actions/runs/7243502875

It looks like skipped tests are failing `stestr` in Windows with Python 3.12

woodsp-ibm:
I had noticed a similar failure in recent qiskit algorithms nightly builds - e.g here https://github.com/qiskit-community/qiskit-algorithms/actions/runs/7242612061/job/19728274840 One difference I had noticed is that when it ran ok it was Python 3.12.0 and when it started failing it was 3.12.1 in the log messages. I had looked at stestr, since that is used there too, but it has not been updated in some while. I had not investigated further

IvanIsCoding:
> I had noticed a similar failure in recent qiskit algorithms nightly builds - e.g here https://github.com/qiskit-community/qiskit-algorithms/actions/runs/7242612061/job/19728274840 One difference I had noticed is that when it ran ok it was Python 3.12.0 and when it started failing it was 3.12.1 in the log messages. I had looked at stestr, since that is used there too, but it has not been updated in some while. I had not investigated further

It could be an issue with Python 3.12.1 and one of the stestr or its dependencies. Minor versions shouldn’t break like this but out of all libraries that use undocumented behaviour from Python, I think this is one of those use cases

mtreinish:
FWIW, this was a breaking api change made in Python 3.12.1's unittest module (which IMO was inappropriate for backporting in a stable point release, but there's no arguing with it now). I pushed up: https://github.com/testing-cabal/testtools/pull/371 which I think should fix it, but the failure mode I was fixing in that PR was slightly different from what's in the failure logs. But I expect it's the same root cause.

IvanIsCoding:
[...]

PR #1077, opened 2024-02-04
Promote arm64 macOS to tier 1

mtreinish:
Github recently added a new macOS runner that is using the m1 CPU that is usable for open source projects. [1] Previously rustworkx had support for arm64 macOS at tier 4 because we were only able to cross compile for the platform and not test the binaries. Now that we can run CI jobs on the platform we're able to run both unit tests and test our binaries on release. This commit adds a new set of test jobs and wheel builds that use the macos-14 runner that mirrors the existing x86_64 macOS jobs we have. This brings arm64 macOS to the same support level as arm64.

[1] https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/


mtreinish:
This failed for Python 3.8 and 3.9 because the `setup-python` action couldn't find those versions on macOS arm64. There is an open issue tracking this here: https://github.com/actions/setup-python/issues/808

.github/workflows/main.yml
```diff
@@ -57,9 +57,10 @@ jobs:
     strategy:
       matrix:
         rust: [stable]
-        python-version: [3.8, 3.9, "3.10", "3.11", "3.12"]
+        python-version: [3.8, 3.9, "3.10", "3.11", "3.12.0"]
```
IvanIsCoding, line 60: I guess this fixes #1041 indirectly

IvanIsCoding (approved):
LGTM.  I just added a comment related to `stestr` and Python 3.12.

After #1064 is merged I wonder if we should have a Linux job cross-compiling the wheels for Linux, Windows and macOS universal because it would really save us time with all these runners. Most of the testing time is spent building the crate

.github/workflows/main.yml
mtreinish: It does, although I only did this as an attempt to fix my typo in an earlier iteration and it's not strictly necessary. I can revert this and the stestr usage if you'd prefer to do that separately.

IvanIsCoding: Maybe let's keep it separate because there will be other 3.12 releases and there could be a regression. Although it is unlikely

mtreinish: Reverted in: https://github.com/Qiskit/rustworkx/pull/1077/commits/ecf713b2246fb0d8fa0415fe6e0594cd1619a6ac

PR #1079, opened 2024-02-05
Fix remaining usages of retworkx

johnthagen:
Fix a few missing references to `retworkx` missed in 

- #644


johnthagen:
I'll pass on signing a CLA for this tiny change, but pinging @IvanIsCoding in case fixing these is desired.

PR #1083, opened 2024-02-08
Add node-removal methods linear in node degree

jakelishman:
The existing method, `PyDiGraph.remove_node_retain_edges` is quadratic in node degree, because the ``condition`` function takes in pairs of edges.  This poses a problem for extreme-degree nodes (for example massive barriers in Qiskit).

This commit adds two methods based on making edge-retention decisions by hashable keys, making it linear in the degree of the node (at least; the MIMO broadcasting can make it quadratic again if all edges have the same key, but that's fundamental to the output, rather than the algorithm).

The ideal situation (for performance) is that the edges can be disambiguated by Python object identity, which doesn't require Python-space calls to retrieve or hash, so can be in pure Rust.  This is `remove_node_retain_edges_by_id`.

The more general situation is that the user wants to supply a Python key function, which naturally returns a Python object that we need to use Python hashing and equality semantics for.  This means using Python collections to do the tracking, which impacts the performance (very casual benchmarking using the implicit identity function as the key shows it's about 2x slower than using the identity).  This method is `remove_node_retain_edges_by_key`.

---

Using a small vec for the edge data (as opposed to `Vec`) kept the performance of the case where in- and out-edges are close to a bijection sufficiently similar in my casual benchmarking to _assuming_ a bijection and not allowing broadacsting at all.  Using a `Vec` (and consequently _always_ having pointer indirection to retrieve the nodes) came with a performance penalty (although I stupidly didn't actually write down the numbers on that).


src/digraph.rs
```diff
@@ -1049,8 +1058,183 @@ impl PyDiGraph {
...
+    /// appear in the output, as there are no successors to pair it with.
+    ///
+    /// :param int node: The index of the node to remove. If the index is not present in the graph
+    ///     it will be ignored and this function will have no effect.
+    #[pyo3(signature=(node, /))]
+    pub fn remove_node_retain_edges_by_id(&mut self, py: Python, node: usize) -> PyResult<()> {
+        // As many indices as will fit inline within the minimum inline size of a `SmallVec`.  Many
+        // use cases of this likely only have one inbound and outbound edge with each id anyway.
+        const INLINE_SIZE: usize =
+            2 * ::std::mem::size_of::<usize>() / ::std::mem::size_of::<NodeIndex>();
```
mtreinish, line 1083: I like that this is future proof and supports 32bit platforms well. For the foreseeable future `NodeIndex` is a `u32`. We looked at making it a `usize` a while ago but the extra memory overhead (on 64bit platforms) to support > 4 billion nodes or edges wasn't worth it.

src/digraph.rs
```diff
@@ -1049,8 +1058,183 @@ impl PyDiGraph {
...
+                .edges_directed(node_index, petgraph::Direction::Outgoing)
+            {
+                let key_value = if let Some(key_fn) = &key {
+                    key_fn.call1(py, (edge.weight(),))?
+                } else {
+                    edge.weight().clone_ref(py)
+                };
+                if let Some(edge_data) = out_edges.get_item(key_value.as_ref(py))? {
+                    let edge_data = edge_data.extract::<RemoveNodeEdgeValue>()?;
+                    edge_data.nodes.as_ref(py).append(edge.target().index())?
```
mtreinish, line 1203: I'm a bit confused why do we need to use a `PyList` here? We're just storing a list of node indices right? Can't that just store that in a rust with a and a `Vec<NodeIndex>`? I get you're using a `PyDict` so we can have the key be return from the key func, but nothing precludes storing `Vec<NodeIndex>` in `RemoveNodeEdgeValue` right. It just needs to be a pyclass so it can go in the pydict but the contents don't have to be useful for Python. This would let us avoid the python overhead for manipulating the nodes list.

mtreinish (commented):
Overall this LGTM. The only real comment I have is about the the struct used for the key function. I think there is a way to avoid some of the python overhead there.

src/digraph.rs
jakelishman: mostly I just don't like magic numbers and I don't trust myself to remember what they mean haha

src/digraph.rs
```diff
@@ -3090,3 +3272,33 @@ where
...
+    }
+}
+impl Eq for PyAnyId {}
+
+/// Internal-only helper class used by `remove_node_retain_edges_by_key` to store its data as a
+/// typed object in a Python dictionary.
+#[pyclass]
+struct RemoveNodeEdgeValue {
+    weight: Py<PyAny>,
+    nodes: Vec<NodeIndex>,
```
mtreinish, line 3303: You used a smallv3c in the identity function does the same optimization hold for this? Or because we end up storing this in Python it doesn't natter?

mtreinish: Nah, it's fine that's a good reason to not bother. It wouldn't make a big difference if there was any in this case.

mtreinish (approved):
LGTM, thanks for the quick updates.

PR #1084, opened 2024-02-08
Improve ergonomics of `tox` lint jobs

jakelishman:
This fixes the `lint` job for `tox>=4`, which uses `whitelist_externals` is removed in favour of `allowlist_externals`.  This also makes the `lint` and `black` runs skip installation of the package, which makes them much faster to use while making changes, and these are static analysis tools, so don't need to install.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ target-version = ['py38', 'py39', 'py310', 'py311']
 [tool.ruff]
 line-length = 105 # more lenient than black due to long function signatures
 src = ["rustworkx", "setup.py", "retworkx", "tests"]
-select = [
+lint.select = [
     "E",   # pycodestyle
     "F",   # pyflakes
     "UP",  # pyupgrade
@@ -19,7 +19,7 @@ select = [
 target-version = "py38"
 extend-exclude = ["doc"]
 
-[tool.ruff.per-file-ignores]
+[tool.ruff.lint.per-file-ignores]
 "rustworkx/__init__.py" = ["F405", "F403"]
 "*.pyi" = ["F403", "F405", "PYI001", "PYI002"]
 
--- a/tox.ini
+++ b/tox.ini
@@ -31,11 +31,11 @@ commands =
 
 [testenv:lint]
 basepython = python3
+skip_install = true
 deps =
   black~=22.0
   ruff
-  setuptools-rust
-whitelist_externals=cargo
+allowlist_externals=cargo
 commands =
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -9,7 +9,7 @@ target-version = ['py38', 'py39', 'py310', 'py311']
 [tool.ruff]
 line-length = 105 # more lenient than black due to long function signatures
 src = ["rustworkx", "setup.py", "retworkx", "tests"]
-select = [
+lint.select = [
```
IvanIsCoding, line 12: Is this really necessary? I don't see the advantage of doing this. We only use ruff for linting so it is implicit

IvanIsCoding (approved):
Overall LGTM, I had the same ideas for #861. Just a minor ruff question

pyproject.toml
jakelishman: Ruff issued a warning asking me to do it when I ran the job (same with the other change to a table name) - I think they were planning to deprecate top-level select maybe?

jakelishman: For context, since I'm sat at my computer now, here's what the tox job says without the two extra `lint.` bits, from a fresh environment rebuild:
```
lint: remove tox env folder /Users/jake/code/retworkx/.tox/lint
lint: install_deps> python -I -m pip install -c/Users/jake/code/retworkx/constraints.txt -U 'black~=22.0' ruff
lint: commands[0] /Users/jake/code/retworkx/tests> black --check --diff ../rustworkx ../tests ../retworkx
All done! ✨ 🍰 ✨
160 files would be left unchanged.
lint: commands[1] /Users/jake/code/retworkx/tests> ruff check ../rustworkx ../retworkx . ../setup.py
warning: The top-level linter settings are deprecated in favour of their counterparts in the `lint` section. Please update the following options in `/Users/jake/code/retworkx/pyproject.toml`:
  - 'select' -> 'lint.select'
  - 'per-file-ignores' -> 'lint.per-file-ignores'
lint: commands[2] /Users/jake/code/retworkx/tests> cargo fmt --all -- --check
lint: commands[3] /Users/jake/code/retworkx/tests> python /Users/jake/code/retworkx/tools/find_stray_release_notes.py
  lint: OK (16.18=setup[5.48]+cmd[4.62,0.56,5.17,0.36] seconds)
  congratulations :) (16.42 seconds)
```

IvanIsCoding: I guess they decided to change the name

Issue #1075, opened 2024-01-31
Add function to compute the maximum bisimulation of a graph


sander-hergarten:
### What is the expected enhancement?
Compute the maximum bisimulation natively in rust. A native rust implementation of algorithms such as Paige-Tarjan would be significantly faster than anything in python. 

There are decent native python implementation i have found but unfortunately they don't scale well to large graphs:
https://github.com/fandreuz/BisPy

sander-hergarten:
Since i need this functionality for a project i am working on i would implement this anyways. This is more to see if this is something i can make a pull request for afterwords.

IvanIsCoding:
Pull requests are welcome! I think if you submit  Paige-Tarjan to `rustworkx` it will get merged.

You already have the algorithm you want to implement in mind. [Check the contributing guide](https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md) and once you have a draft feel free to start the request

IvanIsCoding:
Closed by #1089

PR #1089, opened 2024-02-15
compute maximum bisimulation using Paige-Tarjan's Algorithm

sander-hergarten:
regarding #1075


sander-hergarten:
i had some trouble running `tox -edocs` on my machine. Please let me know if there are any issues with the documentation or release notes

tests/digraph/test_bisimulation.py
```diff
@@ -0,0 +1,92 @@
...
+import rustworkx
+
+
+class TestBisimulation(unittest.TestCase):
+    def test_is_maximum_bisimulation(self):
+        for graph, solution in zip(self.graphs, self.reference_solution):
+            res = rustworkx.digraph_maximum_bisimulation(graph)
+            res_format = [tuple(element) for element in res]
+            for element in res_format:
+                self.assertTrue(all([any([el in sol for sol in solution]) for el in element]))
```
IvanIsCoding, line 24: You can split these tests over multiple functions or at least use `with self.subTest()` to help identify the failures individually

src/iterators.rs
```diff
@@ -905,6 +906,57 @@ custom_vec_iter_impl!(
 );
 impl PyGCProtocol for Chains {}
 
+impl PyHash for NodeIndices {
+    fn hash<H: Hasher>(&self, py: Python, state: &mut H) -> PyResult<()> {
+        PyHash::hash(&self.nodes, py, state)?;
+        Ok(())
+    }
+}
```
IvanIsCoding, line 914: If this is not requried I'd rather not make this type hashable. We reserve the right to make it mutable one day which would make it not hashable

releasenotes/notes/maximum-bisimulation-942a9d0dc9b46ee4.yaml
```diff
@@ -0,0 +1,63 @@
...
+    "Three partition refinement algorithms" by Paige and Tarjan. This function
+    recieves a graph and returns a
+    :class:`~rustworkx.RelationalCoarsestPartition`. This function is in regards
+    to issue `#1075 <https://github.com/Qiskit/rustworkx/issues/1075>`__.
+  - |
+    Added a new class :class:`~rustworkx.RelationalCoarsestPartition` to output
+    the maximum bisimulation or relational coarsest partition of a graph. This
+    class contains instances of :class:`~rustworkx.NodeIndices` and can be
+    iterated over.
+issues:
```
IvanIsCoding, line 15: Thanks for writing the release note! Just remember to remove the sections from the template that are not necessary

IvanIsCoding (commented):
I am going to read more about the algorithm before doing the proper review.

I left some comments about Rust/Python but overall it is in good shape! And the `tox -edocs` failure is common, identation is probably mismatched but it should be easy to fix

sander-hergarten:
As a heads up, when reading "Three Partition Refinement Algorithms", I renamed a lot of the variables.
- **Blocks of Partition $X$**: `CoarseBlock`
- **Blocks of Partition $Q$**: `FineBlock`
- **$C$**: `queue`
- **$S$**: `splitter_block`
Also while the authors say that blocks of $Q$ are contained in blocks of $X$ I interpreted this as the block of $Q$ being a subset of $X$. (I found this confusing so i thought i would mention it)

src/bisimulation.rs
```diff
@@ -0,0 +1,372 @@
...
+                        .subblock
+                        .push(*node);
+                }
+                false => {
+                    counterimage_grouped.insert(
+                        key,
+                        CounterImageGroup {
+                            block: Rc::clone(&block),
+                            subblock: Vec::from([*node]),
+                        },
```
IvanIsCoding, line 180: Minor Rust comment: it is more idiomatic to write [`.entry()` and then handle the Occupied/Vacant case](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html). It would also save a hashmap lookup and be slightly faster

src/bisimulation.rs
```diff
@@ -0,0 +1,372 @@
...
+                .enumerate()
+                .min_by(|(_, x), (_, y)| {
+                    (***x)
+                        .borrow()
+                        .values
+                        .len()
+                        .cmp(&(***y).borrow().values.len())
+                })
+                .map(|(index, _)| index)
+                .unwrap();
```
IvanIsCoding, line 271: Instead of `.unwrap()`ing, make the function return a result and use `?` to return the error to the caller. When Rust panics the experience for Python users is not good

src/bisimulation.rs
```diff
@@ -0,0 +1,372 @@
...
+
+fn maximum_bisimulation(graph: &StablePyGraph<Directed>) -> Vec<Block> {
+    let (fine_block_tuple, initial_coarse_partition, mut node_to_block_vec) = initialization(graph);
+
+    let mut queue: CoarsePartition = initial_coarse_partition;
+    let mut all_fine_blocks = vec![fine_block_tuple.0, fine_block_tuple.1];
+
+    while !queue.is_empty() {
+        let (smaller_component, simple_splitter_block) = {
+            let splitter_block = queue.pop().unwrap();
```
IvanIsCoding, line 254: Same `unwrap()` comment

IvanIsCoding (commented):
Adding some Rust comments this time

src/iterators.rs
sander-hergarten: I implemented this for the `RelationalCoarsestPartition` return type, since `custom_vec_iter_impl!` requires this trait. If you would prefer, I can make a new iterator that will act as a container. 

Since `RelationalCoarsestPartition` is also a pretty specific type name i could change it to something like `NodePartition` containing something such as `NodePartitionBlock` or such.

IvanIsCoding: In this case I would implement a new type and add a macro `custom_vec_iter_impl!` calling the old `custom_vec_iter!` and adding the PyHash. The catch is that sometimes these methods can have unintended side effects on performance (e.g. #1096) so it is better to be conservative

src/bisimulation.rs
```diff
@@ -0,0 +1,373 @@
...
+        queue.extend(coarse_block_that_are_now_compound);
+
+        {
+            let counterimage_splitter_complement =
+                build_counterimage(graph, (*simple_splitter_block).clone());
+
+            counterimage_splitter_complement.keys().for_each(|node| {
+                counterimage.remove(node);
+            });
+        }
```
IvanIsCoding, line 324: Any particular statement this is in its own block and not with the other statements outside the bracket?

IvanIsCoding (approved):
I think the algorithm looks good after skimming the original paper from 1987 and the BisPy implementation. I will also trust the tests and the code coverage essentially touching all lines. 

Given how much data is shared I am also don't mind all `Rc` and `RefCell`. In fact, it is quite impressive that this is correct and Rust's borrow checker approves. Quite an achievement for a first time contribution!

I think after we address the hashing it should be good to merge. Notice that #1096 will introduce a compile error with your change because of the new optimization, but that should be easyd to fix.

src/bisimulation.rs
sander-hergarten: I did this to contextually separate the $E^{-1}(B)-E^{-1}(S-B)$ operation. Should i add a comment or remove it?

src/iterators.rs
sander-hergarten: I am not sure I understood this but the hashing from NodeIndices was removed

IvanIsCoding (approved):

Issue #840, opened 2023-03-20
Add reverse direction serialization


mtreinish:
### What is the expected enhancement?

Right now rustworkx has a json node link serializer (https://qiskit.org/documentation/rustworkx/apiref/rustworkx.node_link_json.html#rustworkx.node_link_json) and a graphml deserializer (https://qiskit.org/documentation/rustworkx/apiref/rustworkx.read_graphml.html#rustworkx.read_graphml) but no  reverse path functions. We should add a node link json deserializer (to read node link json into a graph object) and a graphml serializer (to generate a graphml payload from a graph object) so that you can round-trip a graph over a serialization format.

mtreinish:
Also the dot file from `graph.to_dot()` we should have a `from_dot()` too https://qiskit.org/documentation/retworkx/dev/apiref/rustworkx.PyGraph.to_dot.html

AngeloDanducci:
Feel free to assign this to me @mtreinish have some work in progress on this.

mtreinish:
Sure thing, I've assigned you to this issue

mtreinish:
Hey, @AngeloDanducci I was wondering how this was going and if you needed any help.

troyswanson:
Hello! I would like to bump this ticket to add GraphML serialization to the project

poig:
I would love to give it a try until the unitary hack deadline ended

gluonhiggs:
@mtreinish I'm starting with graphml serialization. Do we need anything else apart from a function like `to_graphml` that serializes graph objects, e.g., `digraph_to_graphml` and `graph_to_graphml`?

1ucian0:
@thierry-martinez , please drop a comment here so we can assign this issue to you.

PR #1091, opened 2024-02-16
Add functions to parse node link JSON

mtreinish:
This commit adds the missing functionality to parse node link json and generate rustworkx graph objects from it. This adds two new functions parse_node_link_json_str() to parse a node link json string and parse_node_link_json_file() to parse a node link json file from a path.

Partial #840


docs/source/api/serialization.rst
```diff
@@ -8,3 +8,5 @@ Serialization
 
    rustworkx.node_link_json
    rustworkx.read_graphml
+   rustworkx.parse_node_link_json_file
+   rustworkx.parse_node_link_json_str
```
IvanIsCoding, line 12: I really dislike the name of the functions. Maybe we can come up with alternatives? They feel ugly

tests/graph/test_node_link_json.py
```diff
@@ -140,3 +140,54 @@ def test_not_multigraph(self):
...
+    def test_round_trip(self):
+        graph = rustworkx.generators.path_graph(3)
+        res = rustworkx.node_link_json(graph)
+        new = rustworkx.parse_node_link_json_str(res)
+        self.assertIsInstance(new, type(graph))
+        self.assertEqual(new.nodes(), graph.nodes())
+        self.assertEqual(new.weighted_edge_list(), graph.weighted_edge_list())
+        self.assertEqual(new.attrs, graph.attrs)
+
+    def test_round_trip_with_file(self):
```
IvanIsCoding, line 153: Overall this is super cool! Do you think we should add a round trip (optional) test where we write from NetworkX and them import it in Rustworkx?

IvanIsCoding (commented):
I think this is a great addition. The code is very succint and I think we leverage `serde` well.

My main concern is with the function naming. Maybe we could have `rustworkx.parse_node_link_json` for the string case and `rustworkx.node_link_json_from_file` for the file case? I just found `parse_node_link_json_str` a name that leaves a bad taste in the mouth

I also left a comment about the NetworkX round-trip test. If the test passes, I wonder how fast it will be compared to [the current NetworkX](https://www.rustworkx.org/stable/0.14/apiref/rustworkx.networkx_converter.html#rustworkx.networkx_converter).

docs/source/api/serialization.rst
mtreinish: Yeah, I wasn't happy with them either tbh. I'll switch them to your suggested `parse_node_link_json` and `node_link_json_from_file` those are better names.

tests/graph/test_node_link_json.py
mtreinish: Sure, that's simple enough to add.

mtreinish: This was a good call to add a test for this. The parser is overly strict because it's assuming there was always an `id` field for edges which networkx doesn't always populate in it's node link json. This isn't required in the format, so I'll tweak the parser to make it optional.

docs/source/api/serialization.rst
mtreinish: I renamed the functions to: `parse_node_link_json` and `from_node_link_json_file` in: https://github.com/Qiskit/rustworkx/pull/1091/commits/52be0a0478de40e1c10ac463485adcb6a5ea78a9

tests/graph/test_node_link_json.py
mtreinish: I fixed this in: https://github.com/Qiskit/rustworkx/pull/1091/commits/8f75081365f7f190d7beabd72251aad2eff42fa5 and added a test.

IvanIsCoding (approved):
LGTM, I am glad we can parse the files with NetworkX and vice-versa

Issue #1068, opened 2024-01-25
Add a function to get all connected subgraphs of a given size


mtreinish:
### What is the expected enhancement?

A feature I was wishing rustworkx exposed natively recently was a function that can return all connected subgraphs of a k nodes. It's simple enough to brute force a solution using `rx.is_connected()` and `itertools.combinations`, but doing this natively in rust (and potential in parallel) would be more efficient. 

Also, I found this paper: https://consystlab.unl.edu/Documents/StudentReports/TR-UNL-CSE-2013-0005.pdf which provides an alternative algorithm for performing this search and we probably should implement for this function as at least an option, if not the sole approach, for solving this problem too.

sbrandhsn:
I looked a bit into this and would like to go for the algorithm in https://www.sciencedirect.com/science/article/pii/S0166218X20302377 [1] instead. The work you referenced has a time and space complexity of at least O(d^{k-1}) for a graph with max degree d and a subgraph of size k, and appears to require some CSP routines to be implemented (yuck!). The work in [1] has delay complexity O(k^{2}*d). The delay complexity gives a notion of the runtime between the output of two connected subgraphs.

I also found that there seems to be a general framework for such enumeration problems called [reverse search](https://en.wikipedia.org/wiki/Reverse-search_algorithm) - if you think that qiskit/rustworkx would need to solve a lot of enumeration problems in the future, maybe it makes sense to look into reverse search at some point. Reverse search performs worse for the k subconnected subgraphs problem but might be better long-term as it is more general and might be therefore more maintainable.

mtreinish:
This was implemented by: https://github.com/Qiskit/rustworkx/pull/1104

PR #1104, opened 2024-02-21
Generate all connected subgraphs of size k

sbrandhsn:
This PR addresses #1068 

Here, a method for generating all connected subgraphs of a certain size `k` is developed based on the work in ["Enumerating Connected Induced Subgraphs: Improved Delay and Experimental Comparison". Christian Komusiewicz and Frank Sommer](https://www.uni-marburg.de/de/fb12/arbeitsgruppen/algorith/paper/enucon.pdf). Compared to the brute-force approach of drawing all combinations of `k=3..6` nodes and then checking whether this combination is a connected subgraph, the method demonstrates a runtime speedup of two orders of magnitude as seen in the following figure. The evaluated graph is a heavy-hex graph of distance `d=1, 3, 5, ..., 99`.

connected_subgraphs

Raw data: [raw_data.txt](https://github.com/Qiskit/rustworkx/files/14359414/raw_data.txt)


IvanIsCoding:
Awesome! Looking forward to review this

tests/graph/test_all_connected_subgraphs.py
```diff
@@ -0,0 +1,168 @@
...
+import unittest
+import rustworkx
+
+
+class TestGraphAllConnectedSubgraphs(unittest.TestCase):
+    def setUp(self):
+        super().setUp()
+        self.edges = [(0, 1), (1, 2), (2, 3), (0, 3), (0, 4), (4, 5), (4, 7), (7, 6), (5, 6)]
+        self.nodes = list(range(8))
+        self.expected_subgraph_nodes = [
```
IvanIsCoding, line 22: If you know the connected components you can use [`itertools.combination`](https://docs.python.org/3/library/itertools.html#itertools.combinations) to shorten this a little bit

sbrandhsn: Thanks, I added the bruteforce method that can also serve as a reference for future work and generated the expected results for the tests that way.

sbrandhsn:
> Awesome! Looking forward to review this

Thanks for reviewing this - this is my first rustworkx and also rust contribution so let me know if you have general remarks! :-)

src/connectivity/subgraphs.rs
```diff
@@ -0,0 +1,82 @@
...
+pub fn k_connected_subgraphs<Ty: EdgeType + Sync>(
+    graph: &StablePyGraph<Ty>,
+    k: usize,
+) -> Vec<Vec<usize>> {
+    let mut connected_subgraphs = Vec::new();
+    let mut graph = graph.clone();
+
+    while graph.node_count() >= max(k, 1) {
+        let mut p: HashSet<NodeIndex> = HashSet::new();
+        let v: NodeIndex = graph.node_indices().next().unwrap();
```
IvanIsCoding, line 34: I'd use [`while let`](https://doc.rust-lang.org/rust-by-example/flow_control/while_let.html) to avoid having the `.unwrap()`. Then you can come the `if` to break the loop early at the end

src/connectivity/subgraphs.rs
```diff
@@ -0,0 +1,82 @@
...
+    subgraphs: &mut Vec<Vec<usize>>,
+    k: usize,
+) -> bool {
+    if p.len() == k {
+        subgraphs.push(p.iter().map(|n| n.index()).collect::<Vec<usize>>());
+        return true;
+    }
+    let mut is_leaf_node: bool = false;
+    while x.is_empty().not() {
+        let u: NodeIndex = x.iter().next().cloned().unwrap();
```
IvanIsCoding, line 56: Another good place for [while let](https://doc.rust-lang.org/rust-by-example/flow_control/while_let.html) to avoid `.unwrap()`

src/connectivity/subgraphs.rs
```diff
@@ -0,0 +1,82 @@
...
+            .cloned()
+            .collect();
+        let mut p_next: HashSet<NodeIndex> = p.clone();
+        p_next.insert(u);
+
+        if simple_enum(&mut p_next, &mut x_next, graph, subgraphs, k) {
+            is_leaf_node = true;
+        } else {
+            return is_leaf_node;
+        }
```
IvanIsCoding, line 79: I'd prefer an iterative version using a Stack/Vector/Queue if possible. Some of users have very small stack sizes, which could lead to stack overflow given we are creating `x_next` everytime

IvanIsCoding (commented):
I left a couple of Rust comments. The iterative vs recursion one is the hardest one to deal with but in general we avoid deeply nested recursion in the library.

IvanIsCoding:
@sbrandhsn I have thought more about the iterative version and I think it is achievable. Here is the pseudocode for it in Python:
```
def simple_iterative(p, x):
	ans = []
	# Third value indicates is_leaf
	# None -> not processed
	# False -> child returned false
	# True -> child returned tru
	stack = [(p, x, None)] # third value is is_leaf
	while len(stack) > 0:
		p, xnext, is_leaf = stack[-1]
		
		# Handle P is equal to K case
		if len(p) == k:
			stack[-2][2] = True # update is_leaf
			ans.push(p)
			stack.pop()
			continue

		if is_leaf is not None and not is_leaf:
			if len(stack) > 1:
				stack[-2][2] = is_leaf or stack[-2][2]
			stack.pop()
			continue

		# Handle X is empty case
		if len(xnext) == 0:
			if len(stack) > 1:
				stack[-2][2] = is_leaf or stack[-2][2]
			stack.pop()
			continue
		

		u = any_vertex(xnext)
		xnext.remove(u)
		xprime = union(xnext, difference(N(u), N(p)))
		stack.push((union(p, [u]), xprime))

	return ans
```

I hope that helps, I know the original paper only gave a recursive version

sbrandhsn:
Thanks for your comments! :-) I wasn't aware of `while let` and will change the code accordingly. Regarding recursive vs iterative function. I can see your point in general. However, in this case the stack depth is bound by `k` and the hash set `x_next` is expected to be allocated on the heap, right? 

For the five 64 bit arguments and a stack size of 1 MB, the stack depth (and `k`) can be up to 25000 before we run into a stack overflow... For such large `k` only trivial subgraphs can be expected to be returned in a reasonable time. Do you think it makes sense to emit a warning in cases where `connected_subgraphs` would have an excessive runtime, i.e. for large `k` or graphs with a high degree?

Lemma 4 in  https://www.uni-marburg.de/de/fb12/arbeitsgruppen/algorith/paper/enucon.pdf actually introduces an iterative version. I will look into that version to see if I can implement it in reasonable time and otherwise revert to your pseudo code. :-)

IvanIsCoding:
> For the five 64 bit arguments and a stack size of 1 MB, the stack depth (and `k`) can be up to 25000 before we run into a stack overflow... For such large `k` only trivial subgraphs can be expected to be returned in a reasonable time. Do you think it makes sense to emit a warning in cases where `connected_subgraphs` would have an excessive runtime, i.e. for large `k` or graphs with a high degree?
> 

I don't think we need to emit a warning. At first before rewriting the algorithm wasn't obvious that the bound in the recursion depth is the number of nodes of the graph. If the bound was the number of subgraphs though reaching 25000 is trivial so I had to ask.

I think for the first version the recursive version is ok. The iterative version would be slightly faster and more consistent with most of the code we had. But the final version will hopefully something that we can parallelize (and it does not need to come in this version!)

src/connectivity/mod.rs
```diff
@@ -659,6 +661,27 @@ pub fn digraph_all_pairs_all_simple_paths(
...
+///
+/// :param PyGraph graph: The graph to find all connected subgraphs in
+/// :param int k: The maximum number of nodes in a returned connected subgraph.
+///
+/// :returns: A list of connected subgraphs with k nodes, represented by their node indices
+///
+/// :raises ValueError: If ``k`` is larger than the number of nodes in ``graph``
+#[pyfunction]
+#[pyo3(text_signature = "(graph, k, /)")]
+pub fn connected_subgraphs(graph: &PyGraph, k: usize) -> PyResult<Vec<Vec<usize>>> {
```
IvanIsCoding, line 675: I wonder how much faster it is to return a Vector of `NodeIndices`, specially after #1096 and #1107 made them faster to use again from Python

IvanIsCoding (approved):
I am going to merge this version, I think it is correct and will bring a lot of benefits. We can think about parallelization and other improvements later

PR #1107, opened 2024-02-21
Add manual iterator implementations for custom vec iterables

### LINKED ISSUE #1090: Performance issue with `custom_vec_iter_impl!` `NodeIndices`
While working on `DAGDependencyV2` in Qiskit, I was using `digraph.predecessor_indices` and `digraph.predecessors` and noticed `predecessors` was significantly faster than `predecessor_indices`. Looking at the `digraph` code, it appeared `predecessor_indices` was simpler and therefore should have been faster.

I noticed that `predecessors` returned a `Vec` and `predecessor_indices` returned `NodeIndices`. I modified the latter function to return `Vec<usize>` and the tests I was running for `DAGDependencyV2` went from 74 sec for `NodeIndices` to 12 sec for `Vec<usize>` constructing an equivalent `DAGDependencyV2`. 

Would appreciate any comments on why this is happening and whether this might be a wider problem.

Main code
```
    #[pyo3(text_signature = "(self, node, /)")]
    pub fn predecessor_indices(&self, node: usize) -> NodeIndices {
        NodeIndices {
            nodes: self
                .graph
                .neighbors_directed(NodeIndex::new(node), petgraph::Direction::Incoming)
                .map(|node| node.index())
                .collect(),
        }
    }
```
Modified code
```
    #[pyo3(text_signature = "(self, node, /)")]
    pub fn predecessor_indices(&self, node: usize) -> Vec<usize> {
        self.graph
            .neighbors_directed(NodeIndex::new(node), petgraph::Direction::Incoming)
            .map(|node| node.index())
            .collect()
    }
```

jakelishman:
This adds manual `Iterator` implementors for the custom vec `Iterable` pyclasses, which were previously using the Python default iterator implementation that attempts calls to `__getitem__` with successive integers until an `IndexError` is raised.

Doing this means that we no longer have any overhead from converting Python-space integer indices to Rust-space ones, because we just store a pointer to the backing vec internally and yield from that using an index we track ourselves.

---

Using the same setup code as discussed in #1096, with a manual iteration through afterwards:
```python
import rustworkx as rx

N = 1_000_000
g = rx.PyGraph()
g.add_nodes_from(list(range(N)))
nodes = g.node_indexes()

%timeit for _ in nodes: pass
```
On `main` (which includes #1096), the `timeit` measured the iteration at ~78ms, and with this PR it dropped to ~48ms on my machine.

I didn't add all the new iterator structs to the typing information concretely, instead just referencing them opaquely as `Iterator[T]`, because these objects only implement magic protocol methods and are not intended to be constructable / picklable / nameable from Python space at all (similar to how `list_iterator` isn't as a builtin).  Open to changing that if preferred.

I'll mark this as:

- Fix #1090 

but feel free to edit this to remove that link if you think there's more places to look for performance here.


jakelishman:
Sure, I can add those.  Flicking quickly through the tests now, I suppose there's a question of whether we want to implement more efficient `reversed` iterators as well.  That's generated separately to the sequence-protocol iterator, so what I've done here doesn't affect `reversed(g.node_indices())` nor `reversed(iter(g.node_indices())` (which continues to not work, which is valid behaviour).

I can throw in a reversed iterator for each of these if you like as well, but it equally might not be worth whatever increase in compilation times it comes with?

jakelishman:
Actually: what kind of `iter` tests were you thinking?  I deliberately made the actual iterator type unnameable from Python, so I can't really `isinstance` check it or anything like that, and there shouldn't be any real testable behaviour that differs from the implicit sequence-protocol iterator.

jakelishman:
Release node added in 8f9a991.

mtreinish:
> Actually: what kind of `iter` tests were you thinking? I deliberately made the actual iterator type unnameable from Python, so I can't really `isinstance` check it or anything like that, and there shouldn't be any real testable behaviour that differs from the implicit sequence-protocol iterator. Any test case that iterates through one of these nodes (which many existing tests do) is already calling `iter` on them, it's just implicit within CPython.

I was just thinking of something that explicitly exercises it, mostly to say we did. I wasn't expecting a type check, I think the existing tests we have for this on mappings do something like:

```python
test = iter(object)
result = list(test)
self.assertEqual(result, expected)
```

It might just be duplicate coverage as that implicitly happens in other tests, but I didn't see the harm in at least validating it explicit.y

jakelishman:
Ok, I've added the reversed iterators in 5f60d9e and explicit tests of both `iter` and `reversed` in b68aa5f.  I tweaked the `iter` tests slightly so they compare `list(x) == list(iter(x))`, which doubles it as a test that the iterable is multiply consumable.

The timings of the reversed iterators are the same as the timings as the forwards iterators: this PR took the million-node `NodeIndices` iteration (with a `reversed()` around it) from about 76ms to about 47ms this time when I ran them.

src/iterators.rs
```diff
@@ -594,11 +609,114 @@ macro_rules! custom_vec_iter_impl {
...
+            }
+
+            fn __length_hint__(&self, py: Python) -> usize {
+                self.inner
+                    .as_ref()
+                    .unwrap()
+                    .borrow(py)
+                    .$data
+                    .len()
+                    .saturating_sub(self.index)
```
mtreinish, line 648: Oh that's a good trick to comply with `__length_hint__` not returning a negative. Although as a usize I guess without this in release mode we would just overflow which wouldn't result in a `ValueError` although the hint would be wildly inaccurate (although PEP424 says the return is "not required to be accurate") for an empty or consumed iterator.

mtreinish (approved):
LGTM, thanks for doing this.

src/iterators.rs
jakelishman: I actually wrote this initially for `__len__` (which is required to be accurate) and reused it for `__length_hint__`, but then I remembered / realised that iterators don't typically implement `__len__` so just renamed the method.

PR #1108, opened 2024-02-21
Add `reverse` argument to topological sorters

### LINKED ISSUE #1105: Add reverse topological ordering

I would like to be able to get a "reverse" topological ordering from `lexicographical_topological_sort` and `TopologicalSorter` in the sense of "a topological order if all the edges reversed direction" - it'd be helpful on Qiskit in a few places where we want to find final operations.  For our purposes, we don't need `topological_sort` to offer the same functionality, which might be easier since that one calls out to `petgraph` within `rustworkx` at the moment.

The logic is very straightforward - we just swap which of `Direction::Outgoing` and `Direction::Incoming` we use in a couple of places in the sorters.  I can provide a PR if it's an acceptable feature.

jakelishman:
This allows `lexicographical_topological_sort` and `TopologicalSorter` to produce a topological sort that would have been found if the edges in the graph had reversed directions, without constructing the edge-flipped graph.

This is useful for graphs where we might want to do topological analysis from both ends of the data flow, for example in Qiskit where we occasionally want to find the "final" operations topologically.

Close #1105


src/dag_algo/mod.rs
```diff
@@ -338,37 +338,50 @@ pub fn layers(
...
         let res = key.call1(py, (a,))?;
         Ok(res.to_object(py))
     };
     // HashMap of node_index indegree
     let node_count = dag.node_count();
+    let (in_dir, out_dir) = if reverse {
+        (petgraph::Direction::Outgoing, petgraph::Direction::Incoming)
+    } else {
+        (petgraph::Direction::Incoming, petgraph::Direction::Outgoing)
+    };
```
IvanIsCoding, line 381: Can you create a public function in `dag_algo` to encapsulate this logic and reuse it in the other files? It is a nitpick but just to avoid the kind of bugs where we flip it

IvanIsCoding (approved):
LGTM, the implementation is very clever and I like the small dif.

My only comment is for encapsulating the logic of the directions to be overprotective. Especially given we revisit topological sorting once in a while

src/dag_algo/mod.rs
jakelishman: Done in 12a91ec.

jakelishman:
Sorry for disappearing for a while there.  I've added the helper in 12a91ec.

PR #1109, opened 2024-02-22
Add missing musl skip for ppc64le builds

mtreinish:
This commit fixes an issue in the ppc64le wheel builds. One of the two ppc64le wheel build jobs was incorrectly attempting to build wheels for musl environments which isn't a supported platform and the jobs failed to configure. This commit just adds the missing skip for those environments to the job configuration to fix this issue.


IvanIsCoding (approved):
LGTM, this should work as I tested it in #1064

PR #1111, opened 2024-02-22
Prepare 0.14.1 release

mtreinish:
This commit prepares for a 0.14.1 release which just includes #1109 to fix ppc64le builds and #1096 to fix the overhead of `__getitem__` on custom sequence return types.

### DIFF
--- a/releasenotes/notes/prepare-0.14.1-e5065553a44eb035.yaml
+++ b/releasenotes/notes/prepare-0.14.1-e5065553a44eb035.yaml
@@ -1,7 +1,7 @@
 ---
 prelude: >
   The rustworkx 0.14.1 release is a small bug fix release that fixes two issues,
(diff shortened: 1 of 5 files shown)


releasenotes/notes/prepare-0.14.1-e5065553a44eb035.yaml
```diff
@@ -0,0 +1,28 @@
+---
+prelude: >
+  The rustworkx 0.14.1 release is a small bug fix release that fixes two issues,
+  first it address a performance regression that was introduced in 0.12.0
```
IvanIsCoding, line 4:
```suggestion
  firstly it address a performance regression that was introduced in 0.12.0
```

IvanIsCoding (approved):
I agree, we should release the fix. It is a significant performance improvement.

PR #1128, opened 2024-03-01
Add `initial` argument to topological sorters

### LINKED ISSUE #1106: Add an `initial_nodes` argument to topological sorters

For both `lexicographical_topological_sort` and `TopologicalSorter`, it would be convenient (and possibly more performant) on Qiskit if we could pass a list of start-point indices to the sorter.  This is for two purposes:
1. we don't need Rustworkx to iterate through the whole graph to find the initial nodes; in the cases I'm thinking of for Qiskit, we already know the node indices we ought to start from.
2. this would let us start the topological ordering from a mid point of the ordering, such as if we only want to consider things that happen in the graph after a certain point, without having to calculate the order up to that point.

Give-or-take, my proposed API would be to expose an `initial_nodes: Optional[Sequence[int]]` argument to `lexicographical_topological_sort` and `TopologicalSorter`.  If not given, the logic remains the same as it is now.  If it is given, the given `initial_nodes` are treated as the start point, even if they do no have zero in degree.  All other nodes are treated identically, so if there are nodes that cannot be visited because they come topologically after nodes that do not strictly dominate the `initial_nodes`, they are simply never returned.

Logically, I think this can be achieved by slightly modifying the population of the `in_degree` maps in the sorters to do some `entry(node).get_or_insert_with(|node| in_degree(node))` type of stuff in the ordering calculations, so the in degree of nodes is calculated lazily as a node first becomes visitable.  In the case that `initial_nodes` was not passed, we'd still eagerly populate the whole `in_degree` map.

Similar to #1105, I'm happy to supply a PR if this is an acceptable feature.

jakelishman:
This adds a way for the topological sorters to recieve a set of initial nodes to begin the search from.  This can allow the topological iteration to begin from a mid-point of the order, or to simply provide Rustworkx with the already-known set of initial nodes, skipping the iteration over each node to find the full set.  Especially in `TopologicalSorter`, where one may be using the on-line search to find only the first handful of nodes in the topological order in an inner loop of a mutating graph, this can be a complexity improvement.

The `initial` nodes are required to have no topological ordering between themselves, and it is an error if they do.  There are other possible conventions to choose here, but this one is the easiest to state / understand and the fastest to execute; most other sensible conventions would require some amount of initial walking of the DAG to determine the partial topological order between the nodes before beginning, which would get in the way of the complexity improvements for inner-loop topological sorts.

For similar complexity reasons, the behaviour surrounding zero-in-degree nodes that are _not_ in `initial` is chosen such that only nodes dominated by the `initial` set will be returned.  In other words, if a node has zero in degree, but is not in the `initial` set, it won't be returned as part of the order.  This is easily understandable and documentable ("the returned topological order is of all nodes dominated by `initial`"), and removes the need for the all-node iteration to locate any potential zero-in-degree nodes.

Close #1106 - the motivation in there is that we end up building and using many `TopologicalSorter`s in a loop while adding nodes to one particular graph, and being able to pass the `initial` nodes lets us avoid a quadratic complexity in doing so.

### DIFF
--- a/releasenotes/notes/initial-topo-cd502f3140500f93.yaml
+++ b/releasenotes/notes/initial-topo-cd502f3140500f93.yaml
@@ -1,0 +1,9 @@
+---
+features:
+  - |
+    :func:`.lexicographical_topological_sort` and :class:`.TopologicalSorter` now both accept an
+    ``initial`` keyword argument, which can be used to limit the returned topological orderings to
+    be only over the nodes that are dominated by the ``initial`` set.  This can provide performance
+    improvements by removing the need for a search over all graph nodes to determine the initial set
+    of nodes with zero in degree; this is particularly relevant to :class:`.TopologicalSorter`,
+    where the user may terminate the search after only examining part of the order.
--- a/rustworkx/rustworkx.pyi
+++ b/rustworkx/rustworkx.pyi
@@ -13,6 +13,7 @@ from .visit import BFSVisitor, DFSVisitor, DijkstraVisitor
 from typing import (
     TypeVar,
     Callable,
+    Iterable,
     Iterator,
     final,
     Sequence,
@@ -271,6 +272,7 @@ def lexicographical_topological_sort(
     key: Callable[[_S], str],
     *,
     reverse: bool = ...,
+    initial: Iterable[int] | None = ...,
 ) -> list[_S]: ...
 def transitive_reduction(graph: PyDiGraph, /) -> tuple[PyDiGraph, dict[int, int]]: ...
 def layers(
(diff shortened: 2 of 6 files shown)


enavarro51 (changes requested):
This looks great. I liked the way you were able to detect the partial topological order between nodes using different methods for `lexicographical_topological_sort` and `TopologicalSorter`

I noted some possible readability changes. There's also a typo in line 188 in toposort.py (not yours) 'to marks as done' s/b 'to mark as done' .

All the tests work fine. I was wondering if some additional random testing with larger graphs might uncover some issues these short graphs would not?

jakelishman:
I've pushed up the docs fixes and an update to make it PyO3-0.21 compatible.  What sorts of randomised tests were you thinking of?  I was having a little bit of a hard time imagining a randomised test that I could a) write (making a randomised DAG that actually samples the space of potential node orders well seemed like a non-trivial problem) and b) verify cleanly in a test seems not super easy to me.

If you've got any ideas, I'm happy to expand the tests.

enavarro51 (approved):
This LGTM. I didn't have any specific random tests in mind. Just thinking there might be paths unexplored, but your tests cover any paths I can think of. Thanks,

Issue #1117, opened 2024-02-27
Shortest Path Panic


ryandeng1:
### Information

- **rustworkx version**: 0.14.1
- **Python version**: 3.10
- **Rust version**: I only installed the package via pip. 
- **Operating system**: 22.04

### What is the current behavior?
I see a panic when running `rustworkx.dijkstra_shortest_paths` and `rustworkx.dijkstra_shortest_path_lengths` on a graph produced by `PyGraph.subgraph`. The subgraphs are quite small, hundreds of nodes while the main graph is 100k nodes. 

Running on the larger graph seems to work fine. I think it may be due to the fact that the node ids remain the same as the one in the original graph, but they might be accessed in a way that goes out of bounds in the subgraph?

### Steps to reproduce the problem
Pseudocode:

```
networkx_graph = ...
rustworkx_graph = rustworkx.networkx_converter(networkx_graph)
nodes = ...
subgraph = rustworkx_graph.subgraph(nodes)
distances = rustworkx.dijkstra_shortest_paths(subgraph, start_node, as_undirected=True)
```

The error occurs on https://github.com/Qiskit/rustworkx/blob/eb896fbd88c8bd086379cd2d46b2a1f74981bc9e/rustworkx-core/src/distancemap.rs#L52 according to what is outputted.

```
thread '<unnamed>' panicked at /project/rustworkx-core/src/distancemap.rs:52:13:
index out of bounds: the len is 44 but the index is 56
stack backtrace:
   0:     0x7fb1c2fbba43 - std::backtrace_rs::backtrace::libunwind::trace::hbee8a7973eeb6c93
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/../../backtrace/src/backtrace/libunwind.rs:104:5
   1:     0x7fb1c2fbba43 - std::backtrace_rs::backtrace::trace_unsynchronized::hc8ac75eea3aa6899
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x7fb1c2fbba43 - std::sys_common::backtrace::_print_fmt::hc7f3e3b5298b1083
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/sys_common/backtrace.rs:68:5
[...]

PR #1134, opened 2024-03-07
Fix panic error at shortest paths 

JPena-code:
In #1117 is presented the presence of an uncaught `pyo3.PanicException` when the index of the source node is no out of bound in the graph. With this naive approach i try fix this issue raising a more understandable `IndexError`.

- [x] I ran rustfmt locally

### DIFF
--- a/src/shortest_path/mod.rs
+++ b/src/shortest_path/mod.rs
@@ -78,6 +78,12 @@ pub fn graph_dijkstra_shortest_paths(
     default_weight: f64,
 ) -> PyResult<PathMapping> {
     let start = NodeIndex::new(source);
+    if !graph.graph.contains_node(start) {
+        return Err(PyIndexError::new_err(format!(
+            "Node source index \"{source}\" out of graph bound"
+        )));
+    }
+
     let goal_index: Option<NodeIndex> = target.map(NodeIndex::new);
     let mut paths: DictMap<NodeIndex, Vec<NodeIndex>> = DictMap::with_capacity(graph.node_count());
 
@@ -217,6 +223,12 @@ pub fn digraph_dijkstra_shortest_paths(
     as_undirected: bool,
 ) -> PyResult<PathMapping> {
     let start = NodeIndex::new(source);
+    if !graph.graph.contains_node(start) {
+        return Err(PyIndexError::new_err(format!(
+            "Node source index \"{source}\" out of graph bound"
+        )));
+    }
+
     let goal_index: Option<NodeIndex> = target.map(NodeIndex::new);
     let mut paths: DictMap<NodeIndex, Vec<NodeIndex>> = DictMap::with_capacity(graph.node_count());
     let cost_fn = CostFn::try_from((weight_fn, default_weight))?;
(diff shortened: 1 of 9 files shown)


src/shortest_path/mod.rs
```diff
@@ -77,6 +77,17 @@ pub fn graph_dijkstra_shortest_paths(
     weight_fn: Option<PyObject>,
     default_weight: f64,
 ) -> PyResult<PathMapping> {
+    if !graph
+        .node_indices()
+        .nodes
+        .iter()
+        .any(|&node| node == source)
```
IvanIsCoding, line 84: I think we can simplify this to:
```suggestion
    if !graph.graph.contains_node(start)
```

IvanIsCoding: Note that you will need to move the if statement after the variable is defined

src/shortest_path/mod.rs
```diff
@@ -371,6 +392,17 @@ pub fn graph_dijkstra_shortest_path_lengths(
...
+        .node_indices()
+        .nodes
+        .iter()
+        .any(|&index| index == node)
+    {
+        return Err(PyIndexError::new_err(format!(
+            "Node source index \"{node}\" out of graph bound"
+        )));
+    }
+
```
IvanIsCoding, line 405: Same

src/shortest_path/mod.rs
```diff
@@ -440,6 +472,17 @@ pub fn digraph_dijkstra_shortest_path_lengths(
     edge_cost_fn: PyObject,
     goal: Option<usize>,
 ) -> PyResult<PathLengthMapping> {
+    if !graph
+        .node_indices()
+        .nodes
+        .iter()
+        .any(|&index| index == node)
+    {
```
IvanIsCoding, line 480: Same

src/shortest_path/mod.rs
```diff
@@ -216,6 +227,16 @@ pub fn digraph_dijkstra_shortest_paths(
     default_weight: f64,
     as_undirected: bool,
 ) -> PyResult<PathMapping> {
+    if !graph
+        .node_indices()
+        .nodes
+        .iter()
+        .any(|&node| node == source)
```
IvanIsCoding, line 234: Same

releasenotes/notes/fix-dijkstra-panic-3962ad36788cab00.yaml
```diff
@@ -0,0 +1,11 @@
+---
+fixes:
+  - |
+    Fixed an issue with the Dijkstra path functions:
+
+      * :func:`rustworkx.dijkstra_shortest_paths`
+      * :func:`rustworkx.dijkstra_shortest_path_lengths`
```
IvanIsCoding, line 7: Can you also expand this to:
* `rustworkx.bellman_ford_shortest_path_lengths`
*  `rustworkx.bellman_ford_shortest_paths`
* `rustworkx.astart_shortest_path_lengths`
*  `rustworkx.astar_shortest_paths`

IvanIsCoding (commented):
Thanks for contributing this! The tests LGTM and so do the error messages. I did propose a simpler method to check if a node is present, though.

Also, you would mind expanding the fix to the Bellman-Ford and A* implementations? They are conceptually similar to Dijkstra's and were not included in the bug report. But they also panic now I believe

releasenotes/notes/fix-dijkstra-panic-3962ad36788cab00.yaml
JPena-code: I didn't find an implementation of the `rustworkx.astart_shortest_path_lengths`, could you please  show me where is that function?

IvanIsCoding: I was mistaken, we only have a path function for A star

IvanIsCoding (approved):
Awesome. This is good to

Issue #1137, opened 2024-03-10
Add different greedy strategies for node coloring


alexanderivrii:
### What is the expected enhancement?

This tracks the rustworkx part of https://github.com/Qiskit/qiskit/issues/11779#issue-2130965930. The request is to add other standard greedy node coloring strategies, such as the "saturation first" strategy: always picking the node that has the largest number of different colors assigned to its neighbors, and, in case of a tie, the node that has the largest number of uncolored neighbors.

This will be immediately followed up by a PR.

mtreinish:
I'm going to close this because #1138 merged. There are a lot of other potential coloring strategies in the literature but if we want to add more we can open new issues for those specific strategies.

PR #1138, opened 2024-03-10
Add "saturation first" and "independent set" greedy node coloring strategies

alexanderivrii:
### Description 

This PR adds two more standard coloring strategies to greedy node/edge coloring functions `graph_greedy_color` and `graph_greedy_edge_color`. 

In the literature the first strategy is known as "saturation", "DSATUR" or "SLF". We dynamically choose the vertex that has the largest number of different colors already assigned to its neighbors, and, in case of a tie, the vertex that has the largest number of uncolored neighbors. 

The second strategy is known as "greedy independent sets" or "GIS". We greedily find independent subsets of the graph, and assign a different color to each of these subsets.

This addresses #1137, modulo the fact that there are quadrillions of different greedy node coloring algorithms, and each comes with trillion variations.

Both new strategies can be combined with `preset_color_fn` (for node coloring).

There is also a new Rust enum `GreedyStrategy` exposed as a python class that allows to specify which of the three currently supported greedy strategies is used. This is, calling the functions from the Python code would be as follows:
```
import rustworkx as rx

graph = rx.generators.generalized_petersen_graph(5, 2)

coloring = rx.graph_greedy_color(graph, greedy_strategy=rx.GreedyStrategy.Degree)
coloring = rx.graph_greedy_color(graph, greedy_strategy=rx.GreedyStrategy.Saturation)
coloring = rx.graph_greedy_color(graph, greedy_strategy=rx.GreedyStrategy.IndependentSet)
coloring = rx.graph_greedy_color(graph)
coloring = rx.graph_greedy_edge_color(graph)
coloring = rx.graph_greedy_edge_color(graph, greedy_strategy=rx.GreedyStrategy.Degree)
coloring = rx.graph_greedy_edge_color(graph, greedy_strategy=rx.GreedyStrategy.Saturation)
coloring = rx.graph_greedy_edge_color(graph, greedy_strategy=rx.GreedyStrategy.IndependentSet)
```

**Update**: the `greedy_graph_edge_color` function has been also extended to support the `preset_color_fn` argument (though in this case it's an optional map from an _edge index_ to either a color or to `None`)


src/coloring.rs
```diff
@@ -23,24 +23,40 @@ use rustworkx_core::coloring::{
...
 
+pub use rustworkx_core::coloring::GreedyStrategyCore;
+
+#[pyclass(module = "rustworkx")]
+#[derive(Clone, PartialEq)]
+pub enum GreedyStrategy {
+    Degree,
+    Saturation,
+}
+
```
alexanderivrii, line 34: I need some help here. I want to define the greedy strategy enum inside rustworkx-core, and re-expose it here to the python interface. However, I wasn't able to get it to work, so I have created two different enums, one in rustworkx-core and one here. Is there a way to improve this?

src/coloring.rs
```diff
@@ -63,22 +79,29 @@ use rustworkx_core::coloring::{
...
                 preset_color_fn
                     .call1(py, (node_idx.index(),))
                     .map(|x| x.extract(py).ok())
             };
-            greedy_node_color_with_preset_colors(&graph.graph, callback)?
+            greedy_node_color_with_preset_colors(&graph.graph, callback, inner_strategy)?
         }
-        None => greedy_node_color(&graph.graph),
+        None => greedy_node_color(&graph.graph, inner_strategy),
     };
```
alexanderivrii, line 105: I was wondering if we really need two different functions `greedy_node_color` and `greedy_node_color_with_preset_colors` which are both simple wrappers on top of `inner_greedy_node_color`. I.e. maybe we could have a single function, with the preset function specified via an `Option<...>`. 

One possible complication is that when the preset function is specified, the coloring algorithm may technically fail as we may specify the same color to two nodes connected by an edge. Though, I don't think we actually check for this, should we? Or maybe this behavior should be allowed, i.e. the user may specify an inconsistent preset fn if the goal is to only greedily color the remaining nodes.

src/coloring.rs
```diff
@@ -104,10 +137,23 @@ pub fn graph_greedy_color(
...
 #[pyfunction]
-#[pyo3(text_signature = "(graph, /)")]
-pub fn graph_greedy_edge_color(py: Python, graph: &graph::PyGraph) -> PyResult<PyObject> {
-    let colors = greedy_edge_color(&graph.graph);
+#[pyo3(text_signature = "(graph, /, greedy_strategy=GreedyStrategy::Degree)")]
+#[pyo3(signature=(graph, /, greedy_strategy=GreedyStrategy::Degree))]
+pub fn graph_greedy_edge_color(
+    py: Python,
+    graph: &graph::PyGraph,
+    greedy_strategy: GreedyStrategy,
```
alexanderivrii, line 149: We could also extend the greedy edge coloring function with preset fn. What do you think?

alexanderivrii:
I am confused about the "stubs" problems: what are these and how should I fix them?
```
rustworkx.GreedyStrategy cannot be subclassed at runtime, but isn't marked with @final in the stub
rustworkx.GreedyStrategy.Degree is not present in stub
rustworkx.GreedyStrategy.Saturation is not present in stub
rustworkx.rustworkx.GreedyStrategy cannot be subclassed at runtime, but isn't marked with @final in the stub
rustworkx.rustworkx.GreedyStrategy.Degree is not present in stub
rustworkx.rustworkx.GreedyStrategy.Saturation is not present in stub
```

src/coloring.rs
alexanderivrii: In addition, what do you think about the names "Degree", "Saturation", "IndependentSet"? Do you prefer instead "LargestFirst", "SaturationLargestFirst" or "IndependentSet", somewhat closer to networkx naming: https://networkx.org/documentation/stable/reference/algorithms/coloring.html.

mtreinish: I can't think of a way to do it with one `enum`. The issue is we need to wrap an enum as a pyo3 pyclass to expose it in the rustworkx side to python. But we can't wrap the definition in rustworkx-core, so the only model that makes sense to me is having a duplicate pyclass version that we map to the rustworkx-core enum. The other option would be a string input I guess that we map to the enum, but that seems more error prone for this case.

src/coloring.rs
mtreinish: This was for backwards compatibility between subsequent rustworkx-core releases. There isn't an easy way to add optional parameters to a rust function, so when I added the `preset_colors` option in #1060 I opted to make it a new function so that any existing users of `graph_node_color` were not broken by needed to pass a new `None` argument to `graph_node_color` when they upgraded to `rustworkx-core` 0.14.0. On the python side I did expose it using the same entry point though, but for the pure rust library compatibility I didn't have a good pattern for doing this. (there might be one that I'm ignorant of though).

As for the soundness, yeah it is likely that the preset color function could result in an incorrect or invalid coloring. The idea was to do as you described to enable greedily coloring the remaining nodes, or at least the was the use case I had in mind (and was asked for an implementation of). But we don't check that the end result coloring between the user provided function provides a globally valid coloring. I tried to document this as a potential issue in the documentation where I wrote:

> Note if you do use a callable there is no validation that the preset values are valid colors. You can generate an invalid coloring if you the specified function returned invalid colors for any nodes.

but if you think we should make the language there stronger I think it's a good idea. TBH, this feature felt like a fairly niche use case I didn't expect it to come up in practice very much.

src/coloring.rs
mtreinish: I think that's a good idea, just for consistency if nothing else.

rustworkx-core/src/coloring.rs
```diff
@@ -141,21 +176,188 @@ where
...
+where
+    G: NodeCount + IntoNodeIdentifiers + IntoEdges + NodeIndexable,
+    G::NodeId: Hash + Eq + Send + Sync,
+    F: FnMut(G::NodeId) -> Result<Option<usize>, E>,
+{
+    let mut colors: DictMap<G::NodeId, usize> = DictMap::with_capacity(graph.node_count());
+
+    let mut preset: HashSet<G::NodeId> = HashSet::new();
+    let mut unprocessed: IndexSet<G::NodeId, RandomState> =
+        IndexSet::with_hasher(RandomState::default());
```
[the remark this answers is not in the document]
mtreinish: If you use `IndexSet` with the `ahash` hasher there isn't much difference in lookup performance between `HashSet` and `IndexSet`, index set does have a bit more memory overhead but beyond that the hash algorithm typically has the biggest impact on lookup performance.  It is a good choice if you need to do fast lookups but also want to iterate over the set with a deterministic ordering.

rustworkx/rustworkx.pyi
```diff
@@ -47,6 +47,7 @@ class JSONSerializationError(Exception): ...
 class FailedToConverge(Exception): ...
 class InvalidMapping(Exception): ...
 class GraphNotBipartite(Exception): ...
+class GreedyStrategy: ...
```
mtreinish, line 50: Since the pyclass isn't configured to allow python subclassing of the enum (which is to correct choice) we need to decorate the stub with `@final`  (https://docs.python.org/3/library/typing.html#typing.final) to indicate this:

```suggestion
@final
class GreedyStrategy: ...
```

mtreinish: Oh, I see the other stub test failure is that the class attributes for the variants aren't included either. I think the syntax might be something like:

```python
@final
class GreedyStrategy: ...
    Degree: ...
    Saturation: ...
```

But that might not work, I'd have to refer to python typing documentation. @IvanIsCoding do you happen to know off the top of your head?

src/coloring.rs
mtreinish: I'm fine with the naming you picked, but I think my initial thought would be to match the descriptions from `Adrian Kosowski, and Krzysztof Manuszewski, Classical Coloring of Graphs, Graph Colorings` which is what I think NetworkX used. Matching the terminology of the book and then citing it in the docstring gives people a good way to look up the exact coloring strategy if they so desire. But I like the more compact naming you used as it's easier to type. Maybe we should have a table at the end of the docstring that maps the strategy names to the book, I know you have it in the text written out, but a table for easy lookup might make it easier.

rustworkx/rustworkx.pyi
IvanIsCoding: @mtreinish that is correct, you need to include each method and attribute of a class. You can look at PyDiGraph with multigraph, attrs etc. If there is no specified type you can use `variable: Any`

src/coloring.rs
alexanderivrii: What is the right way to add tables to rust documentation? I thought one should do this using pipes as explained in https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html, attempting this in [d227ad9](https://github.com/Qiskit/rustworkx/pull/1138/commits/d227ad9f0991c4fe91152457c2627e10b2554413), but this leads to error messages. Should this be done using `` .. list-table::`` or something like this?

src/coloring.rs
alexanderivrii: This is also supported: one can specify an optional callback to `graph_greedy_edge_color` that presets some of the edge colors. I have added tests and updated the release note.

alexanderivrii:
I think I have addressed all of the comments from the previous review, this is ready for another round! :)

rustworkx-core/src/coloring.rs
```diff
@@ -141,21 +176,190 @@ where
...
+    G::NodeId: Hash + Eq + Send + Sync,
+    F: FnMut(G::NodeId) -> Result<Option<usize>, E>,
+{
+    let mut colors: DictMap<G::NodeId, usize> = DictMap::with_capacity(graph.node_count());
+
+    let mut nbd_colors: HashMap<G::NodeId, HashSet<usize>> =
+        HashMap::with_capacity(graph.node_count());
+    for k in graph.node_identifiers() {
+        nbd_colors.insert(k, HashSet::new());
+    }
```
mtreinish, line 221: Not that it really matters since you're pre-allocating, but you could do this as a map collect iterator:
```suggestion
    let mut nbd_colors: HashMap<G::NodeId, HashSet<usize>> = graph.node_identifiers().map(|k| {
        (k, HashSet::new())
    })
    .collect();
```

rustworkx-core/src/coloring.rs
```diff
@@ -141,21 +176,190 @@ where
...
+
+    // Greedily process nodes
+    while let Some((k, _)) = pq.pop() {
+        let neighbor_colors = nbd_colors.get(&k).unwrap();
+        let mut current_color: usize = 0;
+        loop {
+            if !neighbor_colors.contains(&current_color) {
+                break;
+            }
```
mtreinish, line 255: Could you do this instead?

```suggestion
        while !neighbor_colors.contains(&current_color) {
```

rustworkx-core/src/coloring.rs
```diff
@@ -141,21 +176,190 @@ where
...
+        // Remove neighbors of preset nodes with the given color
+        for k in &preset {
+            if colors.get(k) == Some(&current_color) {
+                for u in graph.neighbors(*k) {
+                    if unprocessed.swap_take(&u).is_some() {
+                        remaining.insert(u);
+                    }
+                }
+            }
+        }
```
mtreinish, line 318: I think you can use an iterator for this first loop which I think will be a bit faster because it should give compilers hints about the size of `remaining`. It's something like this I think:
(this is untested so I'm sure I made typos or other mistakes, but I think it's the basic idea)

```suggestion
        // Remove neighbors of preset nodes with the given color
        let mut remaining: IndexSet<G::NodeId, RandomState> = preset.iter().filter(|k| {
            colors.get(k) == Some(&current_color)
        })
        .flat_map(|k| {
                graph.neighbors(*k).filter(|u| {
                    unprocessed.swap_take(&u).is_some()
                }
        })
        .collect();
```

rustworkx-core/src/coloring.rs
```diff
@@ -164,44 +368,61 @@ where
...
 /// expected_colors.insert(NodeIndex::new(0), 0);
 /// expected_colors.insert(NodeIndex::new(1), 1);
 /// expected_colors.insert(NodeIndex::new(2), 1);
 /// assert_eq!(colors, expected_colors);
 /// ```
 ///
-pub fn greedy_node_color<G>(graph: G) -> DictMap<G::NodeId, usize>
+pub fn greedy_node_color<G>(
+    graph: G,
+    greedy_strategy: GreedyStrategyCore,
```
mtreinish, line 384: I worry about the API implications of this, wouldn't this be a breaking API change because you're required to pass a strategy enum when calling this function now?

rustworkx-core/src/coloring.rs
```diff
@@ -164,44 +368,61 @@ where
...
+) -> DictMap<G::NodeId, usize>
 where
-    G: NodeCount + IntoNodeIdentifiers + IntoEdges,
+    G: NodeCount + IntoNodeIdentifiers + IntoEdges + NodeIndexable,
```
mtreinish, line 387: Same question here around backwards compatibility? Couldn't this break existing users of `greedy_node_color` because they now have an additional trait constraint on `graph`?

rustworkx/rustworkx.pyi
```diff
@@ -50,6 +50,12 @@ class FailedToConverge(Exception): ...
 class InvalidMapping(Exception): ...
 class GraphNotBipartite(Exception): ...
 
+@final
+class GreedyStrategy:
```
mtreinish, line 54: I would probably change this to `ColoringStrategy` or `GreedyColoringStrategy` to be a bit more descriptive.

mtreinish (commented):
Overall this is looking good. I left a few inline code suggestions but the basic algorithms look sound. The only things I think we need to settle on is the naming of some of the new interfaces. The other concern I have is around backwards compatibility for the public api for the function in rustworkx-core. The modifications to the interfaces for the existing functions are all potentially backwards incompatible and could cause issues for users upgrading. It might be better to split off the strategy mode to yet another new function to avoid that. (Although having 3 functions isn't ideal either)

Also I think this is missing a release note.

rustworkx-core/src/coloring.rs
```diff
@@ -97,12 +101,43 @@ where
...
+    }
+}
+
+fn inner_greedy_node_color_strategy_degree<G, F, E>(
     graph: G,
     mut preset_color_fn: F,
 ) -> Result<DictMap<G::NodeId, usize>, E>
 where
-    G: NodeCount + IntoNodeIdentifiers + IntoEdges,
+    G: NodeCount + IntoNodeIdentifiers + IntoEdges + NodeIndexable,
```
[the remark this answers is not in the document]
alexanderivrii: Hmmm, I don't know why I introduced it, but apparently it's not needed, removing...

rustworkx-core/src/coloring.rs
alexanderivrii: I have removed that (apparently this was not needed).

rustworkx-core/src/coloring.rs
alexanderivrii: As we have discussed, the existing functions now remain unscathed, and two new functions `greedy_node_color_with_coloring_strategy` and `greedy_edge_color_with_coloring_strategy` have been added. I have also changed their return type to be `Result<Dict<...>, E>` to allow intercepting errors when using the preset function.

rustworkx-core/src/coloring.rs
alexanderivrii: This is the only suggestion that I was not able to incorporate. Thanks to this, I have learned about the `flat_map` and I understand the code intention, it's really clever how you chain everything together. However, after the following minor fixes:
```
let mut remaining: IndexSet<G::NodeId, RandomState> = preset
	.iter()
	.filter(|k| colors.get(*k) == Some(&current_color))
	.flat_map(|k| {
		graph
			.neighbors(*k)
			.filter(|u| unprocessed.swap_take(u).is_some())
	})
	.collect();
```
I am getting the following compilation error
```
error: captured variable cannot escape `FnMut` closure body                                                                                                                            
   --> rustworkx-core\src\coloring.rs:303:17
    |
284 |       let mut unprocessed: IndexSet<G::NodeId, RandomState> =
    |           --------------- variable defined here
...
302 |               .flat_map(|k| {
    |                           - inferred to be a `FnMut` closure
303 | /                 graph
304 | |                     .neighbors(*k)
305 | |                     .filter(|u| unprocessed.swap_take(u).is_some())
    | |_________________________________-----------_______________________^ returns a closure that contains a reference to a captured variable, which then escapes the closure body     
    |                                   |
    |                                   variable captured here
    |
    = note: `FnMut` closures only have access to their captured variables while they are executing...
    = note: ...therefore, they cannot allow references to captured variables to escape
help: consider adding 'move' keyword before the nested closure
    |
305 |                     .filter(move |u| unprocessed.swap_take(u).is_some())
    |                             ++++

error: could not compile `rustworkx-core` (lib) due to previous error                                                                                                                  
```
I think I kind of understand the error but I am not sure how to fix it, the automatic suggestion to use `move` leads to even more problems.

alexanderivrii:
Thanks @mtreinish, as suggested (and later discussed offline) I have retained the older version of the functions in rustworkx-core and removed using the unnecessary `NodeIndexable` trait, making the code backwards compatible, at the expense of introducing two new function `greedy_node_coloring_with_coloring_strategy` and `greedy_edge_color_with_coloring_strategy`. I could not think of shorter names. I have made their return type to be `Result<Dict<...>, E>` instead of `Dict<...>`, which allows error handling. I have renamed the enums to `ColoringStrategy` and the function argument to `strategy`, as per your suggestions, I like these much better.

rustworkx-core/src/coloring.rs
mtreinish: Ah too bad, yeah I missed the ownership issue around mutation of `unprocessed`. Rust won't let us do that because it would require it can't track the mutable borrows with the closure. Let's just leave the loop then it's not worth fighting this when a loop works fine.

rustworkx-core/src/coloring.rs
```diff
@@ -295,6 +569,78 @@ where
...
+/// expected_colors.insert(EdgeIndex::new(0), 2);
+/// expected_colors.insert(EdgeIndex::new(1), 0);
+/// expected_colors.insert(EdgeIndex::new(2), 1);
+/// expected_colors.insert(EdgeIndex::new(3), 2);
+/// assert_eq!(colors, expected_colors);
+/// ```
+///
+pub fn greedy_edge_color_with_coloring_strategy<G, F, E>(
+    graph: G,
+    preset_color_fn: F,
```
mtreinish, line 612: Is there a reason to not make this `Option<F>` since it's not mandatory. It just looks a bit awkward in the `rustworkx` crate side  to handle the case when there isn't preset.

mtreinish (commented):
Overall this looks great, thanks for all the diligent work to update things and fix the backwards compatibility. I just have one inline question about the API for the new functions in rustworkx-core (it applies to both the nodes and edges functions). Otherwise I think this is good to merge.

rustworkx-core/src/coloring.rs
alexanderivrii: Thanks! I definitely agree it's better to make it `Option<F>`. With this change, the function above would look like
```
pub fn greedy_edge_color_with_coloring_strategy<G, F, E>(
    graph: G,
    preset_color_fn: Option<F>,
    strategy: ColoringStrategy,
) -> Result<DictMap<G::EdgeId, usize>, E>
where
    G: EdgeCount + IntoNodeIdentifiers + IntoEdges,
    G::EdgeId: Hash + Eq,
    F: Fn(G::EdgeId) -> Result<Option<usize>, E>,
```
So far so good. I thought that using this function (from say testing) now becomes something like
```
let preset_color_fn = |node_idx: EdgeIndex| -> Result<Option<usize>, Infallible> {
    if node_idx.index() == 1 {
        Ok(Some(1))
    } else {
        Ok(None)
    }
};
let colors = greedy_edge_color_with_coloring_strategy(
        &graph,
        Some(preset_color_fn),
        ColoringStrategy::Degree,
    )
    .unwrap();
```
for the case that we want to specify the preset coloring function, and 
```
let colors = greedy_edge_color_with_coloring_strategy(
        &graph,
        None,
        ColoringStrategy::Degree,
    )
    .unwrap();
```
for the case that we do not. However the compiler does not like the second case:
```
    --> rustworkx-core\src\coloring.rs:1505:13
     |
1502 |         let colors = greedy_edge_color_with_coloring_strategy(
     |                      ---------------------------------------- required by a bound introduced by this call
...
1505 |             None,
     |             ^^^^ cannot infer type of the type parameter `T` declared on the enum `Option`
     |
     = note: multiple `impl`s satisfying `_: Fn<(petgraph::prelude::EdgeIndex,)>` found in the following crates: `alloc`, `core`:
     |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `greedy_edge_color_with_coloring_strategy`
help: consider specifying the generic argument
     |
1505 |             None::<T>,
     |                 +++++

```
and I can't figure out how to make this work.

mtreinish: Ok, yeah I can see why rust isn't happy. It needs to know what the `Some()` branch is in that case so it can generate the code for the generic usage in this case. But `None` doesn't have the context and specifying the type with `None::<T>` would be super ugly, especially for a fn generic. I thought it would be more ergonomic to be able to do `None` here, but I guess if it's not that easy. Let's just leave this as is.

mtreinish (approved):

PR #1142, opened 2024-03-15
Adds support GraphML support for long

### LINKED ISSUE #1140: Support for "long" attribute type in GraphML
The GraphML spec (http://graphml.graphdrawing.org/xmlns/1.0rc/graphml-attributes.xsd) states that the allowed attribute types are:

- boolean
- int
- long
- float
- double
- string

However, loading a GraphML file with an attribute type of long (e.g. `<key id="d0" for="edge" attr.name="weight" attr.type="long"/>` fails with: 

```
Exception: Invalid 'attr.type' attribute in key with id=d0.
```

The issue seems to be that the Type enum doesn't have a Long type. Could this be added?

https://github.com/Qiskit/rustworkx/blob/af0d36caa3d3b3488bdb0297b2014e5f3154f994/src/graphml.rs#L118-L124

jamesdbaker:
Fixes #1140 by adding GraphML support for "long", along with corresponding unit tests

### DIFF
--- a/releasenotes/notes/graphml-long-type-support-05e66d139e685dfe.yaml
+++ b/releasenotes/notes/graphml-long-type-support-05e66d139e685dfe.yaml
@@ -1,0 +1,4 @@
+---
+fixes:
+  - |
+    Adds missing support for the Long attribute type to GraphML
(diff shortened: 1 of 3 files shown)


IvanIsCoding:
Thanks so much for finding the bug AND contributing the fix! Overall the code seems ready and the unit test is appropriate.

I think the only missing item is the CLA signing and then we should be good to go

IvanIsCoding (approved):

PR #1153, opened 2024-04-01
Decrease graphviz visualization test size to speed up tests

mtreinish:
With recent graphviz releases the graphviz_draw() visualization tests for 50 node random graphs were taking an inordinate amount of time. This didn't seem to be an issue in CI but on my local system with graphviz 10.0.1 the time for graphviz to generate a visualization for the 50 node random graph was orders of magnitude slower than the rest of the tests in the rustworkx python tests. This commit decreases the graph size to only be 10 nodes, which reduces the runtime of these tests to be inline with expectations.


IvanIsCoding (approved):
I agree, as long as we cover all the lines we should be fine.

PR #1155, opened 2024-04-01
Bump MSRV to 1.70

mtreinish:
This commit bumps the MSRV for rustworkx in 0.15.0 to 1.70.0. This enables us to use GATs (see #1143) and also enables to update some of our dependencies which require newer versions of rust to compile.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,7 +66,7 @@ jobs:
         ]
         include:
           # Test minimal supported Rust version
-          - rust: 1.64
+          - rust: 1.70
             python-version: 3.8
             platform: { os: "ubuntu-latest", python-architecture: "x64", rust-target: "x86_64-unknown-linux-gnu" }
             msrv: "MSRV"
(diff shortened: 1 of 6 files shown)


IvanIsCoding:
I think you are installing 1.7 instead of 1.70

.github/workflows/main.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         ]
         include:
           # Test minimal supported Rust version
-          - rust: 1.64
+          - rust: 1.70
```
IvanIsCoding, line 69:
```suggestion
          - rust: “1.70”
```

.github/workflows/main.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         ]
         include:
           # Test minimal supported Rust version
-          - rust: 1.64
+          - rust: “1.70”
```
IvanIsCoding, line 69:
```suggestion
          - rust: 1.70.0
```

IvanIsCoding (approved):

PR #1156, opened 2024-04-05
Fix test suite for Numpy 2

jakelishman:
The only change needed was a now-removed alias.  Also fix a typo in a test file's name that otherwise took me literally 6 goes to type even with tab completion.


IvanIsCoding (approved):
Thanks! It looks straightforward, I am glad the diff was small.

PR #1159, opened 2024-04-05
Optimise data transfer in `TopologicalSorter`

### LINKED ISSUE #1157: Allow single node indices in `TopologicalSorter.done`

Following #1128, another thing that would help performance for `TopologicalSorter` for the intended use-cases on Qiskit is if it were to allow single indices in its `done` method.  In practice, all the reasons we use the on-line sorter for involve looking at nodes one at a time and making a decision about them, and the overhead of constructing new Python lists and then unwrapping them into Rust vecs is quite expensive.

I'm happy to make a PR, just before I do: would we prefer to have a separate `done_single` method, or to overload `done`?

jakelishman:
This optimises the creation of the `list` objects returned from `TopologicalSorter.get_ready` (by avoiding a temporary `Vec`) and allows `done` to accept a single index, avoiding the need for the user to allocate many small lists in several real-world uses.

This is the overload version of #1157.  Happy to change it to a separate function if desired.

Close #1157

---

With a setup of:
```python
from qiskit.circuit.library import EfficientSU2
from qiskit.converters import circuit_to_dag

import rustworkx

dag = circuit_to_dag(EfficientSU2(100, reps=5, flatten=True), copy_operations=False)._multi_graph
```
(I was using Qiskit 1.0.2 "ish", but it doesn't matter much) and a timing script:
```python
sorter = rustworkx.TopologicalSorter(dag, check_cycle=False, check_args=False)
while ready := sorter.get_ready():
    for node in ready:
        sorter.done(node)
```
for this PR, and `sorter.done([node])` (with a list) for `main`, I found `main` took ~510µs and this PR took ~275µs on my machine.

That's a common pattern for how we want to use this in Qiskit, though we'll only conditionally be calling `done` on some nodes, which is why the overhead of managing lists really hurts us.


src/toposort.rs
```diff
@@ -167,24 +170,24 @@ impl TopologicalSorter {
...
-        out
+    fn get_ready<'py>(&mut self, py: Python<'py>) -> Bound<'py, PyList> {
+        self.num_passed_out += self.ready_nodes.len();
+        PyList::new_bound(
+            py,
+            self.ready_nodes.drain(..).map(|nx| {
+                self.node2state.insert(nx, NodeState::Ready);
+                nx.index()
+            }),
+        )
```
enavarro51, line 181: This is a nice clean way to go straight to a PyList.

src/toposort.rs
```diff
@@ -193,60 +196,73 @@ impl TopologicalSorter {
...
+            for node in nodes {
+                self.done_single(nodes.py(), NodeIndex::new(node.extract()?))?
+            }
+            Ok(())
+        } else {
+            for node in nodes.iter()? {
+                self.done_single(nodes.py(), NodeIndex::new(node?.extract()?))?
+            }
+            Ok(())
+        }
```
enavarro51, line 212: Not quite sure what's the difference in the 2 else clauses. Is the last one assuming `nodes` can't be downcast to `PyList`? If so, what's an example?

src/toposort.rs
```diff
@@ -193,60 +196,73 @@ impl TopologicalSorter {
...
+                )));
+            }
+            Some(NodeState::Done) => {
+                return Err(PyValueError::new_err(format!(
+                    "node {} was already marked done.",
+                    node.index()
+                )));
+            }
+            Some(state) => {
+                debug_assert_eq!(*state, NodeState::Ready);
```
enavarro51, line 234: Did you intend to leave this here?

enavarro51 (commented):
Just a couple of questions and then this looks about ready to go.

src/toposort.rs
jakelishman: The Rust-space typecast is stricter than a Python-space call to `list` (which copies), so anything iterable that's not a list or a subclass would fail it, like a tuple for example.

src/toposort.rs
jakelishman: I didn't add it - it was just there in the original. It doesn't hurt in release mode (it gets compiled out), and it expands the test coverage a bit in debug.

enavarro51: It's fine to leave it. I know it doesn't affect the release, but we rarely use the `debug_assert` in rustworkx.

enavarro51 (approved):
LGTM.

PR #1160, opened 2024-04-05
Allow `TopologicalSorter` without error checking

### LINKED ISSUE #1158: Version of `TopologicalSorter` without error checking

At the moment, `TopologicalSorter` hashes every index both as it goes through `get_ready` and as it's returned to `done` and builds up a hashmap of the size of all visited nodes, only for the purposes of emitting Python-space errors if the user does something incorrectly.  None of "putting in an invalid node", "putting in an unready node" and "putting in an already done node" should cause a Rust-space panic (though will ofc cause the algorithm to give nonsensical results).

It'd be nice for performance if there was a way to use a `TopologicalSorter` that didn't have that memory and runtime overhead.  Philosophically, I'd like it if it wasn't there at all, so there's two separate Rust `pyclass`es, one with error-checking compiled in and one without (and `new` just builds the correct one), but I think most of my concern would be addressed by making `TopologicalSorter::node2state` an `Option` and deciding whether to check it at runtime.

Again, I'll make the PR if there's a chosen direction.

jakelishman:
For representative examples in Qiskit, this reduces the runtime of interacting with the topological sorter by around 15-20%.

Built on top of #1159.  This is the dynamic-check version of #1158.

Close #1158.

---

Using the same timings as #1159, but adding `check_args=False`, the same test drops to 233µs for me, which is another 15% (though I've seen more in other graphs).  There is maybe a 1% hit for the new `Option` checking of `node2state` in the `check_args=True` case, but that might just have been statistical noise on my machine.


enavarro51 (approved):
LGTM.

PR #1169, opened 2024-04-22
Add method: has_node

haoxins:
- [x] I ran rustfmt locally


src/digraph.rs
```diff
@@ -595,6 +595,18 @@ impl PyDiGraph {
...
+    /// Return True if there is a node in the graph.
+    ///
+    /// :param int node: The node index to check
+    ///
+    /// :returns: True if there is a node false if there is no node
+    /// :rtype: bool
+    #[pyo3(text_signature = "(self, node, /)")]
+    pub fn has_node(&self, node: usize) -> bool {
+        let index = NodeIndex::new(node);
+        return self.graph.node_weight(index).is_some();
```
mtreinish, line 607:
```suggestion
        self.graph.contains_node(index)
```

src/graph.rs
```diff
@@ -468,6 +468,18 @@ impl PyGraph {
...
+    /// Return True if there is a node.
+    ///
+    /// :param int node: The index for the node
+    ///
+    /// :returns: True if there is a node false if there is no node
+    /// :rtype: bool
+    #[pyo3(text_signature = "(self, node, /)")]
+    pub fn has_node(&self, node: usize) -> bool {
+        let index = NodeIndex::new(node);
+        return self.graph.node_weight(index).is_some();
```
mtreinish, line 480:
```suggestion
        self.graph.contains_node(index)
```

mtreinish (commented):
Thanks for doing this, overall this is a nice addition! I left a small inline comment to simplify the code slightly. But besides that could you add a unit test that validates the case when this new method returns false? This also is missing the type annotation stubs and it'll fail the ci job checking those. The process for adding them is documented here: https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md#type-annotations

Also if you don't mind could you add a release note to document the new API. The process for doing this is documented here:  https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md#release-notes

tests/digraph/test_nodes.py
```diff
@@ -76,6 +76,12 @@ def test_remove_nodes_from_with_invalid_index(self):
         self.assertEqual(["a"], res)
         self.assertEqual([0], dag.node_indexes())
 
+    def test_has_node(self):
+        dag = rustworkx.PyDAG()
+        node_a = dag.add_node("a")
+        self.assertTrue(dag.has_node(node_a))
+        self.assertFalse(dag.has_node(node_a + 1))
```
mtreinish, line 83: Can you add an equivalent test for `PyGraph` to `tests/graph/test_nodes.py`?

mtreinish (commented):
Thanks for the quick update! This looks ready to go to me now, the only thing I think missing is test coverage for `PyGraph`. Once that's there I think this is good to merge.

tests/digraph/test_nodes.py
haoxins: done

IvanIsCoding (approved):
LGTM, thanks for adding the tests

Issue #1029, opened 2023-11-13
`minimal_cycle_basis`


quantumjim:
### What is the expected enhancement?

Add `minimal_cycle_basis`, with the same functionality as that of NetworkX.

Will be used in Qiskit QEC's `ArcCircuit`.

mtreinish:
For reference to anyone interested in implementing this the algorithm [networkx is implementing](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.cycles.minimum_cycle_basis.html) is published here:
https://link.springer.com/article/10.1007/s00453-007-9064-z

gluonhiggs:
@quantumjim @mtreinish  I want to try this, but I want to make it clear first. 
1. Do we need to add ``minimal_cycle_basis`` module written in Rust?
2. Should I add this to ``rustworkx-core/src/connectivity/``?

gluonhiggs:
@mtreinish yes, sure!

PR #1170, opened 2024-04-23
Resolve issue #1029

gluonhiggs:
In fact, I do not want to merge my code straightforwardly, because I need some advice!
1. I added the function `minimum_cycle_basis` and some tests. It works, but I believe it it need some modifications to be more well programmed, including the choice of having error handler or not.
2. With the assumption that the Rust core of `minimum_cycle_basis` is working fine, we need to add a Python layer to wrap around it. And there are some errors in `src/connectivity/mod.rs` where we use `#[pyfunction]` annotation. I have no idea to fix this. Perhaps, to fix it, I need to modify even the `minimum_cycle_basis.rs` core in `rustworkx-core/src/connectivity/`.
3. [for fun] By the way, I don't feel like this is a good first issue from 0 knowledge about Rust to 7 months developing this :D 
Link to the issue [https://github.com/Qiskit/rustworkx/issues/1029](url)

### DIFF
--- a/rustworkx/rustworkx.pyi
+++ b/rustworkx/rustworkx.pyi
@@ -15,7 +15,6 @@ from typing import (
     Callable,
     Iterable,
     Iterator,
-    Union,
     final,
     Sequence,
     Any,


IvanIsCoding (commented):
I will try to review this when I have time. I am sorry we mislabeled #1029, maybe we underestimated the effort to implement the algorithm?

Nevertheless, congrats on completing the work and submitting the PR! It is a big accomplishment

gluonhiggs:
For convenience, I think I need to explain my idea why I perform the subgraph creation and the lifted graph as in the `minimum_cycle_basis.rs`.
Our ultimate goal is to return a (minimal) cycle basis which contains the information of the input graph nodes (i.e `NodeIndex(n)` for node n). Because in Rust, if we create a new graph which preserves some structure of another graph, we would have the `NodeIndex()` reset. Therefore, I have to use the name of the original nodes and pass them across the flow to track the node order. We manipulate the subgraphs, the lifted graphs and get the data, using the names to map afterwards to obtain the `NodeIndex`.

rustworkx-core/src/connectivity/minimum_cycle_basis.rs
```diff
@@ -0,0 +1,493 @@
...
+        |_| 0,
+    );
+    // Store the shortest path in a list and translate lifted nodes to original nodes
+    let mut min_path: Vec<usize> = Vec::new();
+    match result {
+        Some((_cost, path)) => {
+            for node in path {
+                let node_name = gi.node_weight(node).unwrap();
+                if node_name.contains("_lifted") {
+                    let original_node_name = node_name.replace("_lifted", "");
```
IvanIsCoding, line 334: If you need to add additional information to the hash you can also hash tuples in Rust. Or you can come up with a custom struct and write a hasher for it

rustworkx-core/src/connectivity/minimum_cycle_basis.rs
```diff
@@ -0,0 +1,493 @@
...
+    let mut shortest_path_map: HashMap<String, i32> = HashMap::new();
+    for nodeid in graph.node_identifiers() {
+        let node_weight = graph.node_weight(nodeid).unwrap();
+        let node_name = format!("{:?}", node_weight).trim_matches('"').to_string();
+        let lifted_node_name = format!("{}_lifted", node_name);
+        let node = gi_name_to_node_index[&node_name];
+        let nodeidx = NodeIndex::new(node);
+        let lifted_node = gi_name_to_node_index[&lifted_node_name];
+        let lifted_nodeidx = NodeIndex::new(lifted_node);
+        let result = astar(
```
IvanIsCoding, line 298: Should probably just be a call for [dijkstra](https://docs.rs/rustworkx-core/latest/rustworkx_core/shortest_path/fn.dijkstra.html) given the A* heuristic is a constant

IvanIsCoding (commented):
I left some comments, but overall I think we will need to rewrite most of the hashmap usage because we should avoid the overhead of creating and hashing intermediate strings.

Also, you should replace A* with Dijkstra

rustworkx-core/src/connectivity/minimum_cycle_basis.rs
```diff
@@ -0,0 +1,493 @@
...
+                let new_node = subgraph.add_node(node_name.clone());
+                // get the index of the node in the subgraph
+                let subgraph_node_index = subgraph.to_index(new_node);
+                name_idx_map.insert(node_name, subgraph_node_index);
+            }
+            // Add edges to the subgraph
+            for edge in graph.edge_references() {
+                if component.contains(&edge.source()) && component.contains(&edge.target()) {
+                    let source_name = format!("{}", graph.to_index(edge.source()));
+                    let target_name = format!("{}", graph.to_index(edge.target()));
```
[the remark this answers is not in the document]
gluonhiggs: I'm wondering without the name, how can we know which node is which after putting these into calculations related to the subgraphs and the lifted graph. Imagine `NodeIndex(1)` in the original graph (user input), which might be `NodeIndex(0)` in a subgraph, and `NodeIndex(2)` in the lifted graph. I can only think of mapping these `NodeIndex(i)` from one graph with respect to another, and apparently we still need have to use `HashMap`. Does it sound good?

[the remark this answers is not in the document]
IvanIsCoding: You can create a custom struct to store data e.g. a pair of integers. And then implement the hash trait with derive https://doc.rust-lang.org/std/hash/trait.Hash.html#implementing-hash.

Strings contain at least 24 bits even if they are empty. So we should find better representations to store intermediate data in a more compact way. You don’t want string hashing to be the bottleneck of the algorithm

gluonhiggs:
I have added some modifications based on your advice. However, I don't think my approach using the new trait `EdgeWeightToNumber` is a good idea. I will try to generalize the edge weight type instead of forcefully using `i32`.

src/connectivity/mod.rs
```diff
@@ -918,6 +918,36 @@ pub fn stoer_wagner_min_cut(
...
+                // Handle errors by converting them into Python exceptions
+                Err(PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(format!(
+                    "An error occurred: {:?}",
+                    e
+                )))
+            }
+        }
+    })
+}
+
```
IvanIsCoding, line 950: My piece of advice is the `rustworkx-core` function looks good but we need to do some work on the Python bindings. I think we can just remove the Python bindings and add it later on

IvanIsCoding:
I think the code is looking better now, the compiler is complaining about some unused variables + trait missing for Python but overall this was a big improvement! Thanks

gluonhiggs:
I have modified the functions in `minimum_cycle_basis.rs` so that they are able to take `weight_fn` that returns `Result<K,E>`. By this, I guess we can deal with more general weights (e.g not only `i32`). The thing is I also want to and the python wrapper for this core, but I don't quite understand how precisely to implement this. It would be wonderful if you provide some keywords or document that I can search to read and learn, I think I can also do this (it will take a while).

IvanIsCoding:
> I have modified the functions in `minimum_cycle_basis.rs` so that they are able to take `weight_fn` that returns `Result<K,E>`. By this, I guess we can deal with more general weights (e.g not only `i32`). The thing is I also want to and the python wrapper for this core, but I don't quite understand how precisely to implement this. It would be wonderful if you provide some keywords or document that I can search to read and learn, I think I can also do this (it will take a while). Thank you!

If you know that you need to access the weights multiple times I recommend following this pattern: https://github.com/Qiskit/rustworkx/blob/e4dae8c495cb04aae276a12b76b2005eafa265df/src/shortest_path/all_pairs_dijkstra.rs#L69.

If you only need to call this once you can wrap this pattern https://github.com/Qiskit/rustworkx/blob/e4dae8c495cb04aae276a12b76b2005eafa265df/src/shortest_path/floyd_warshall.rs#L90 into a function that you pass to the Rust code.

gluonhiggs:
I've added what I think necessary. Thank you a lot!

PR #1172, opened 2024-04-24
Relax numpy upper version cap

mtreinish:
In #1012 we added an upper version cap to numpy to prevent it from installing numpy 2.0 before we confirmed that rustworkx was compatible with it. Now that numpy 2.0.0rc1 has been released we're able to confirm that rustworkx works fine with numpy 2.0. This commit raises the upper bound on the numpy version to < 3 to enable installing numpy 2.0 with rustworkx.


IvanIsCoding (approved):
LGTM, thanks for confirming

mtreinish:
In talking with @jakelishman (who handled the numpy 2.0 compatibility for Qiskit) he pointed out that the `__array__` method has new arguments which we weren't handling in rustworkx. The tests wouldn't cover this because even when running with numpy 2.0.0rc1 we weren't touching that. I updated the method in: https://github.com/Qiskit/rustworkx/pull/1172/commits/fc8cb4bb36fa312073b292e52717134991fe8613 to handle the semantics from the migration guide.

src/iterators.rs
```diff
@@ -601,11 +602,27 @@ macro_rules! custom_vec_iter_impl {
...
+                    return Err(PyValueError::new_err(
+                        "A copy is needed to return an array from this object.",
+                    ));
+                }
+                let res = self.$data.convert_to_pyarray(py)?;
+                Ok(match dtype {
+                    Some(dtype) => {
+                        let locals = [("res", res), ("dtype", dtype)].into_py_dict_bound(py);
+                        py.eval_bound(
+                            "__import__('numpy').asarray(res, dtype=dtype)",
```
IvanIsCoding, line 618: I know this method exists but are there any alternatives? `__import__`  is strongly discouraged (https://docs.python.org/3/library/functions.html#import__)

IvanIsCoding (commented):
Can we refactor the dtype handling to use `py.import_bound()` and then handle the NumPy module as a traditional Python object to call the methods?

I'd rather avoid `eval_bound` and `__import__` in the same code

src/iterators.rs
```diff
@@ -601,11 +602,27 @@ macro_rules! custom_vec_iter_impl {
...
+                            None,
+                            Some(&locals),
+                        )?
+                        .into()
```
IvanIsCoding, line 622:
```suggestion
                        let numpy_mod = py.import_bound("numpy")?;
                        let args = (res,);
                        let kwargs = PyDict::new_bound(py);
                        kwargs.set_item("dtype", dtype)?;
                        numpy_mod.call_method("asarray", args, Some(&kwargs))?.into()
```

IvanIsCoding: Ah I need to make a minor tweak

IvanIsCoding, line 622:
```suggestion
                        let numpy_mod = py.import_bound("numpy")?;
                        let args = (res,);
                        let kwargs = [("dtype", dtype)].into_py_dict_bound(py);
                        numpy_mod
                            .call_method("asarray", args, Some(&kwargs))?
                            .into()
```

Issue #1173, opened 2024-04-24
Improve the performance of the Katz Centrality computation.


Paulo-21:
### What is the expected enhancement?
Improve the performance of Katz Centrality.

I saw that for every iteration of all node the beta values was looked from a hashmap index by the node index (usize) exactely like the x vector .
I think the hashmap could be replace with a simple vector, this lead to better performance.
This could be done without breaking the api.

PR #1174, opened 2024-04-24
Remove the Hashmap from the katz centrality computation

Paulo-21:
Proposed fix for issue #1173


IvanIsCoding (approved):
LGTM, that was a detail that we did not catch on the review.

In general the bottleneck will still be iterating over the edges of the graph, but this shouldn't hurt the performance

Issue #1166, opened 2024-04-19
Move `collect_bicolor_runs()` functionality to rustworkx-core


mtreinish:
The `collect_bicolor_runs()` function is only exposed via a Python interface currently:

https://github.com/Qiskit/rustworkx/blob/ab93329be6d8aaccb6c090f4b3ab6f9540ee8114/src/dag_algo/mod.rs#L627-L754

We should port it to rustworkx-core, so that rust users can leverage the function.

One tweak that probably makes sense for the rustworkx-core version is that instead of returning a `Vec` of `Vec`s of node weights we should have it return of an iterator of `Vec`s node ids. This would be more flexible and performant for rust space users and for the python side of rustworkx we can just collect the iterator to a `Vec` (for backwards compatibility).

ElePT:
I'm interested!

raynelfss:
This seems to have been completed. Could someone close the issue?

PR #1186, opened 2024-05-08
Move `collect_bicolor_runs()` to rustworkx-core

ElePT:
This is a first attempt at solving #1166, the PR introduces a new `collect_bicolor_runs()` function in `rustworx-core` and modifies the python-facing `collect_bicolor_runs` to use the core function under the hood (no user-facing changes).

The current function returns a `<Vec<Vec<G::NodeId>>`, I could not find a way to implement the iterator mentioned in the original issue because I kept getting "the size cannot be known at compilation time", but I am open to implementation suggestions.


rustworkx-core/src/collect_bicolor_runs.rs
```diff
@@ -0,0 +1,203 @@
...
+///     the edge is ignored.
+///
+/// :returns: a list of groups with exactly two edge colors, where each group
+///     is a list of node data payload/weight for the nodes in the bicolor run
+/// :rtype: list
+pub fn collect_bicolor_runs<G, F, C, B, E>(
+    graph: G,
+    filter_fn: F,
+    color_fn: C,
+) -> Result<Vec<Vec<&<G as Data>::NodeWeight>>, CollectBicolorSimpleError<E>> //OG type: PyResult<Vec<Vec<PyObject>>>
```
mtreinish, line 91: Instead of returning `NodeWeight` here I think it would be simpler to just return something like:

```
impl Iterator<Item=Vec<G::NodeId>>
```
(or something like that I might have made a mistake in the exact syntax) Or alternatively a `Vec<Vec<G::NodeId>>`

The idea is to return an iterator of node indices for each run. This isn't exactly what the python api was returning, but retunring the node ids will let us easily iterate and map that to the weights for python, but for other rust consumer (especially Qiskit in the future) this will be lighter weight to work with because it's a simple vec get to go from an index to a weight. 

Then using an iterator would be a more rust native interface to return an iterator of the runs

rustworkx-core/src/lib.rs
```diff
@@ -73,6 +73,7 @@ pub type Result<T, E = Infallible> = core::result::Result<T, E>;
 pub mod bipartite_coloring;
 /// Module for centrality algorithms.
 pub mod centrality;
+pub mod collect_bicolor_runs;
```
ElePT, line 76: This was moved from under `pub mod connectivitiy` when I ran `cargo fmt`, I guess that they must be in alphabetical order (I will move it back to coloring algorithms).

rustworkx-core/src/dag_algo.rs
```diff
@@ -313,6 +316,170 @@ where
...
+    }
+}
+
+fn fmt_dag_has_cycle(f: &mut Formatter<'_>) -> std::fmt::Result {
+    write!(f, "Sort encountered a cycle")
+}
+
+fn fmt_callable_error<E: Error>(f: &mut Formatter<'_>, inner: &E) -> std::fmt::Result {
+    write!(f, "The function failed with: {:?}", inner)
+}
```
ElePT, line 343: I am not 100% sure that defining these custom error classes is the best way to go, maybe it's too cumbersome? It took me a while to figure out how to convert to and from `CollectBicolorError::CallableError` and`PyErr`, and I ended up using [this function](https://github.com/Qiskit/rustworkx/pull/1186/files#diff-8d6abcabece14baab9fb4c082786ee2a676f21d9fa70162ba5222164622ff69aR610). The custom errors have also made it difficult to come up with a simple example to add to the docs, because I always need to define the type hint for the error, but this might be also lack of experience on my side.

rustworkx-core/src/dag_algo.rs
```diff
@@ -599,3 +766,218 @@ mod test_lexicographical_topological_sort {
         assert_eq!(result, Ok(Some(vec![nodes[7]])));
     }
 }
+
+#[cfg(test)]
+mod test_collect_bicolor_runs {
```
ElePT, line 771: I tried my best migrating the python tests to rust but I skipped `test_color_with_ignored_edge` and `test_two_colors_with_barrier`.

rustworkx-core/src/dag_algo.rs
mtreinish: So I think this pattern is fine, especially if there are multiple sources of potential error. Although in the case there really is two conditions to think about, either the input graph is not a DAG (i.e. has a cycle) or the callback has an error. To simplify the function what I've typically been doing is  using `None` to represent a graph with a cycler and then using a fully generic `E` for the error type from the callback and the functions result. That lets you use `?` to pass through the callback's error directly to the caller and still handle the error case without too much extra complexity. This would basically be changing the function signature to be like:

```rust
pub fn collect_bicolor_runs<G, F, C, B, E>(
    graph: G,
    filter_fn: F,
    color_fn: C,
) -> Result<Option<Vec<Vec<G::NodeId>>>, E>
where
    F: Fn(&<G as Data>::NodeWeight) -> Result<Option<bool>, E>,
    C: Fn(&<G as Data>::EdgeWeight) -> Result<Option<usize>,E>,
    G: NodeIndexable
        + IntoNodeIdentifiers
        + IntoNeighborsDirected
        + IntoEdgesDirected
        + Visitable
        + DataMap,
    <G as GraphBase>::NodeId: Eq + Hash,
{
```
The only wrinkle in this case is because you have 2 callbacks a user needs to ensure `filter_fn` and `color_fn` return the same error type. But that should be ok in practice (and for rustworkx's python crate it will just be a `PyErr`).

rustworkx-core/src/dag_algo.rs
```diff
@@ -313,6 +314,161 @@ where
...
+    color_fn: C,
+) -> Result<Option<Vec<Vec<G::NodeId>>>, E>
+where
+    F: Fn(&<G as Data>::NodeWeight) -> Result<Option<bool>, E>,
+    C: Fn(&<G as Data>::EdgeWeight) -> Result<Option<usize>, E>,
+    G: IntoNodeIdentifiers // Used in toposort
+        + IntoNeighborsDirected // Used in toposort
+        + IntoEdgesDirected // Used for .edges_directed
+        + Visitable // Used in toposort
+        + DataMap, // Used for .node_weight
```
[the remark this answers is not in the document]
ElePT: Hmm I agree that this would make the code more readable, and more general, as the filtering/coloring functions could also just depend on the node/edge id. At the same time, I don't know if many coloring functions would use the ids directly, they would probably access the weights (especially with larger graphs, handling 30 node ids vs, let's say, 4 weights, looks painful), so we would be adding that "extra work" to the users. However, I think accessing the weights from the ids is not too bad or surprising, so we can go the general route. I'll apply the changes and see if I see anything else.

tests/digraph/test_collect_bicolor_runs.py
```diff
@@ -99,6 +99,7 @@ def filter_function(node):
                 return None
 
         def color_function(node):
+            print("node name:", node)
```
mtreinish, line 102:
```suggestion
```

rustworkx-core/src/dag_algo.rs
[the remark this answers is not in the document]
mtreinish: FWIW, the only thing I know using this function is Qiskit, we added it to rustworkx based on an algorithm we were doing in Python previously: https://github.com/Qiskit/rustworkx/issues/265. It's kind of a bespoke thing we wanted to do for Qiskit's dag traversal. The reason we're porting it is to rustworkx-core is so we can call it from Rust in Qiskit :) 

So if using weights here makes the Qiskit use case easier I guess we can leave it as is, but with my rustworkx-core api design hat on I felt using ids felt more natural.

ElePT:
I have tried out the suggestion of using ids instead of weights in c256e27721c9b836899a64cfd93b2e71993765ae, the only impact I noticed on the user side was the need to switch from one shared function for all tests to per-test closures in the unit tests, because a reference to the graph is needed to get the weights from the node/edge ids. I actually think this makes the tests more readable. I had hoped to be able to get rid of the result annotations, but the error type still needs to be determined, shame.

rustworkx-core/src/dag_algo.rs
```diff
@@ -313,6 +314,165 @@ where
...
+/// * Raises an error if found computing the bicolor runs
+///
+/// # Example:
+///
+/// ```rust
+/// use rustworkx_core::dag_algo::collect_bicolor_runs;
+/// use petgraph::graph::{DiGraph, NodeIndex};
+/// use std::convert::Infallible;
+/// use std::error::Error;
+/// let mut graph = DiGraph::new();
```
mtreinish, line 352:
```suggestion
///
/// let mut graph = DiGraph::new();
```

rustworkx-core/src/dag_algo.rs
```diff
@@ -313,6 +314,165 @@ where
...
+/// let n3 = graph.add_node(1);
+/// let n4 = graph.add_node(0);
+/// let n5 = graph.add_node(0);
+/// graph.add_edge(n0, n2, 0);
+/// graph.add_edge(n1, n2, 1);
+/// graph.add_edge(n2, n3, 0);
+/// graph.add_edge(n2, n3, 1);
+/// graph.add_edge(n3, n4, 0);
+/// graph.add_edge(n3, n5, 1);
+/// let filter_fn = |node_id| -> Result<Option<bool>, Infallible> {
```
mtreinish, line 365:
```suggestion
///
/// let filter_fn = |node_id| -> Result<Option<bool>, Infallible> {
```

rustworkx-core/src/dag_algo.rs
```diff
@@ -313,6 +314,165 @@ where
...
+///     Ok(Some(*graph.node_weight(node_id).unwrap() > 0))
+/// };
+/// let color_fn = |edge_id| -> Result<Option<usize>, Infallible> {
```
mtreinish, line 368:
```suggestion
///
/// let color_fn = |edge_id| -> Result<Option<usize>, Infallible> {
```

rustworkx-core/src/dag_algo.rs
```diff
@@ -313,6 +314,165 @@ where
...
+///     Ok(Some(*graph.edge_weight(edge_id).unwrap() as usize))
+/// };
+/// let result = collect_bicolor_runs(&graph, filter_fn, color_fn).unwrap();
```
mtreinish, line 371:
```suggestion
///
/// let result = collect_bicolor_runs(&graph, filter_fn, color_fn).unwrap();
```

rustworkx-core/src/dag_algo.rs
```diff
@@ -313,6 +314,165 @@ where
...
+    filter_fn: F,
+    color_fn: C,
+) -> Result<Option<Vec<Vec<G::NodeId>>>, E>
+where
+    F: Fn(<G as GraphBase>::NodeId) -> Result<Option<bool>, E>,
+    C: Fn(<G as GraphBase>::EdgeId) -> Result<Option<usize>, E>,
+    G: IntoNodeIdentifiers // Used in toposort
+        + IntoNeighborsDirected // Used in toposort
+        + IntoEdgesDirected // Used for .edges_directed
+        + IntoEdgeReferences
```
mtreinish, line 386: Is this used anywhere?

mtreinish (approved):
This LGTM now thanks for the updates. I just left some whitespace nits in the docstring. Also I think we can drop one trait bound.

PR #1194, opened 2024-05-17
Add `layers` function to `rustworkx-core`

### LINKED ISSUE #1167: Move `layers` functionality to rustworkx-core
The `layers()` function is only exposed via a Python interface currently:

https://github.com/Qiskit/rustworkx/blob/ab93329be6d8aaccb6c090f4b3ab6f9540ee8114/src/dag_algo/mod.rs#L227-L345

We should port it to rustworkx-core, so that rust users can leverage the function.

One tweak that probably makes sense for the rustworkx-core version is that instead of returning a `PyList` for a Vec of Vecs of node weights we should have it return of an iterator of Vecs of node ids. This would be more flexible and performant for rust space users and for the python side of rustworkx we can just collect the iterator to a pylist (for backwards compatibility).

raynelfss:
Attempts to resolve #1167 

### Summary
These commits bring the `layers` function down to `rustworkx-core` so that it can be used within the rust interface. This function was only available through a python interface previously.

### Changes
- `rustworkx.rustworkx-core.layers` **`layers()`**: This function returns a list of subgraphs whose nodes are disjointed.
   - The main difference between this and its python counterpart is that this function only returns a list of `NodeId` instances. To get the weights/data stored in these nodes, the user can iterate through the graph accessing the weights based on the instance of `NodeId`.
- `rustworkx.dag_algo` **`layers()`**: This function is similar to the one in `rustworkx-core`, but allows the user to choose between getting node indices and node weights/data.
   - The logic has been updated to use the `rustworkx-core` version, but its behavior is identical to what it was previously.

### Future directions:
- ~~To return an `Iterator` instance from `rustworkx-core`, instead of a `Vec<Vec<G::NodeId>>` to prevent from unnecessary allocations.~~ **Added!**
- Open to suggestions.


mtreinish (commented):
Thanks for doing this, just some quick high level comments from a first skim over the code. 

The other things are can you add a couple of other rust space tests so we have some more dedicated coverage of the rust api (more than just the doctest) and also a release note?

rustworkx-core/src/dag_algo.rs
```diff
@@ -313,6 +316,115 @@ where
...
+///  (2, 3),
+///  (3, 4),
+/// ];
+///
+/// let graph = DiGraph::<u32, u32>::from_edges(&edge_list);
+/// let layers: Vec<Vec<usize>> = layers(&graph, vec![0,1]).unwrap();
+/// let expected_layers = vec![vec![0,1], vec![1,2], vec![2,3], vec![3,4], vec![4]];
+/// assert_eq!(layers, expected_layers)
+/// ```
+pub fn layers<G>(graph: G, first_layer: Vec<usize>) -> Result<Vec<Vec<usize>>, LayersError>
```
ElePT, line 347: I was comparing your PR with #1186 and noticed that even though we both return node ids, in this case you return node ids as a `Vec<Vec<usize>>` instead of a `Vec<Vec<G::NodeId>>`. Was there any particular reason for this choice? I based mine on the `longest_path` function, which also returned a `Vec<Vec<G::NodeId>>`.

raynelfss: In this case, I don't think there is a particular reason, might just be a case of me following the previous code a little bit too tightly.

From what I can see it's just so we can send a regular `Vec` with `usize` values, since it might be a bit easier for the user to deal with, but then again, it might work the same way with `G::NodeId`. I will get back to you on this.

mtreinish (commented):
Just some quick comments from a first skim through the PR.

rustworkx-core/src/dag_algo.rs
```diff
@@ -313,6 +316,119 @@ where
...
+            }
+        }
+        if !next_layer.is_empty() {
+            output_indices.push(next_layer.to_owned());
+        }
+        cur_layer = next_layer;
+        next_layer = Vec::new();
+    }
+    Ok(output_indices)
+}
```
mtreinish, line 430: The iterator comment I referred to in the original issue can actually be done here. It requires a bit of a restructuring of the code, you have to move the tracking variables into a struct and implement the `Iterator` trait for that struct. This ends up looking something like:

```suggestion
    Ok(LayersIterator {
        graph,
        cur_layer,
        output_indices,
        next_layer,
        predecessor_count,
        node_set,
        first_iter: false,
    })
}

struct LayersIterator<G, N> {
    graph: G,
    cur_layer: Vec<N>,
    output_indices: Vec<Vec<N>>,
    next_layer: Vec<N>,
    predecessor_count: HashMap<N, usize>,
    node_set: HashSet<N>,
    first_iter: bool,
}

impl<G, N> Iterator for LayersIterator<G, N>
where
    G: NodeIndexable // Used in from_index and to_index.
        + NodeCount // Used in find_cycle
        + EdgeCount // Used in find_cycle
        + Visitable // Used in find_cycle
        + IntoNodeIdentifiers<NodeId = N> // Used for .node_identifiers
        + IntoNeighborsDirected // Used for .neighbors_directed
        + IntoEdgesDirected // Used for .edged_directed
        + GraphBase<NodeId = N>,
    N: Copy + Eq + Hash,
{
    type Item = Vec<N>;
    fn next(&mut self) -> Option<Self::Item> {
        if !self.first_iter {
            self.first_iter = true;
            Some(self.cur_layer.clone())
        } else if self.cur_layer.is_empty() {
            None
        } else {
            while !self.cur_layer.is_empty() {
                for node in &self.cur_layer {
                    let children = self
                        .graph
                        .neighbors_directed(*node, petgraph::Direction::Outgoing);
                    let mut used_indices: HashSet<N> = HashSet::new();
                    for succ in children {
                        // Skip duplicate successors
                        if used_indices.contains(&succ) {
                            continue;
                        }
                        used_indices.insert(succ);
                        let mut multiplicity: usize = 0;
                        let raw_edges: G::EdgesDirected = self
                            .graph
                            .edges_directed(*node, petgraph::Direction::Outgoing);
                        for edge in raw_edges {
                            if edge.target() == succ {
                                multiplicity += 1;
                            }
                        }
                        self.predecessor_count
                            .entry(succ)
                            .and_modify(|e| *e -= multiplicity)
                            .or_insert(
                                // Get the number of incoming edges to the successor
                                self.graph
                                    .edges_directed(succ, petgraph::Direction::Incoming)
                                    .count()
                                    - multiplicity,
                            );
                        if *self.predecessor_count.get(&succ).unwrap() == 0 {
                            self.next_layer.push(succ);
                            self.predecessor_count.remove(&succ);
                        }
                    }
                }
                if !self.next_layer.is_empty() {
                    return Some(self.next_layer.to_owned());
                }
                std::mem::swap(&mut self.cur_layer, &mut self.next_layer);
                self.next_layer.clear();
            }
            None
        }
    }
}
```

Then you return the return type to 

```rust
impl Iterator<Item=Vec<G::NodeId>>
```

I think having the true iterator here would make sense because we can stop early if someone hits the target, or we can operate on the data in place as we iterate which I think would be more powerful than collecting into a Vec.

Note, I haven't tested the functionality of this suggestion, just that it compiles locally. But there are some warnings so I probably messed up something in the migration. This is more the basic structure to shoot for.

rustworkx-core/src/dag_algo.rs
```diff
@@ -313,6 +316,119 @@ where
...
+        }
+    }
+    for layer_node in &cur_layer {
+        if !node_set.contains(layer_node) {
+            return Err(LayersError(Some(format!(
+                "An index input in 'first_layer' {} is not a valid node index in the graph",
+                graph.to_index(*layer_node)
+            ))));
+        }
+    }
```
mtreinish, line 385: Do we need this in the rust API? I'm wondering if we should just treat this with a panic because it's a straight programming error in rust, not an expected input. We need this python side so we avoid a panic and return an `IndexError` but in rust this condition is the same as doing:

```rust
let test = vec![0, 1]
println!("{}", test[42])
```

so guarding against it feels unnecessary.

mtreinish (changes requested):
Thanks for doing this, it's a good start. I think in this case we should make `layers()` a true iterator because it doesn't need to do a complete traversal to compute the final output. We can and should return this layer by layer. The reason we couldn't do that in python is we can't put references in a pyclass for the iteration, but in a Rust API we don't have that limitation. I left an inline code suggestion to start you down that path.

The other thing is I'm thinking we may not need to return a `Result<>` for this function, the only error condition we returned before was input checking we needed for Python interoperability, but doesn't really feel right in a rust API. You added a cycle check in this case which we could error for, but I'd also be fine just panicking in that case because this function doesn't work if there are cycles. Or alternatively just stop iterating when we reach a cycle. In either case we could just document that behavior and avoid the `Result` and I think either would probably be acceptable because this function is really only for DAGs so you shouldn't be passing a graph with cycles.

rustworkx-core/src/dag_algo.rs
```diff
@@ -313,6 +316,119 @@ where
...
+    let node_set = graph.node_identifiers().collect::<HashSet<G::NodeId>>();
+    // Throw error if a cycle is found at the current node
+    if let Some(node) = &cur_layer.first() {
+        let check_cycle = find_cycle(&graph, Some(**node));
+        if !check_cycle.is_empty() {
+            return Err(LayersError(Some(
+                "The provided graph has a cycle".to_string(),
+            )));
+        }
+    }
```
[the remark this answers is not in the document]
raynelfss: The reason why I had the error in the first place was to communicate to Python about what errors were encountered. I could change it to panic, and then just have Python make the cycle checks beforehand. I would still have to implement a check on the rust side that involves less overhead. I will think about the logic for that.

rustworkx-core/src/dag_algo.rs
raynelfss: This makes sense, I'll have something like this ready soon!

rustworkx-core/src/dag_algo.rs
raynelfss: I'll move it to Python only then

raynelfss: This check might still be needed as it is only performed on that first layer, which is not validated in any way. So I kept it, but a `panic!` is called instead. I wonder if this is not the desired behavior @mtreinish.

rustworkx-core/src/dag_algo.rs
```diff
@@ -314,6 +319,176 @@ where
...
+                "An index input in 'first_layer' {:?} is not a valid node index in the graph",
+                node
+            );
+        }
+    }
+    Ok(LayersIter::new(graph, cur_layer))
+}
+
+#[derive(Debug, Clone)]
+pub struct LayersIter<G, N> {
```
mtreinish, line 395: I would leave this private and have the function just return `impl Iterator`

rustworkx-core/src/dag_algo.rs
```diff
@@ -314,6 +319,176 @@ where
...
+///     vec![2.into(),3.into()],
+///     vec![3.into(),4.into()],
+///     vec![4.into()]
+/// ];
+/// assert_eq!(layers, expected_layers)
+/// ```
+pub fn layers<G>(
+    graph: G,
+    first_layer: Vec<G::NodeId>,
+) -> Result<LayersIter<G, G::NodeId>, LayersError>
```
mtreinish, line 359: If possible I would do this as

```suggestion
) -> Result<impl Iterator<Item=Vec<G, G::NodeId>>, LayersError>
```

It leaves us the flexibility to leave `LayersIter` private and also tweak the implementation in the future if needed. We basically are making the guarantee that the API will just return an iterator of vecs of node ids.

rustworkx-core/src/dag_algo.rs
```diff
@@ -314,6 +319,176 @@ where
...
+    fn new(graph: G, first_layer: Vec<N>) -> Self {
+        Self {
+            graph,
+            cur_layer: first_layer,
+            next_layer: vec![],
+            predecessor_count: HashMap::new(),
+            first_iter: true,
+        }
+    }
+}
```
mtreinish, line 423: Do we need a `new()` here? It feels a bit unnecessary since this is just the iterator return type from this function and not something users should be manually interacting with.

rustworkx-core/src/dag_algo.rs
```diff
@@ -314,6 +319,176 @@ where
...
+
+impl<G, N> Iterator for LayersIter<G, N>
+where
+    G: NodeIndexable // Used in from_index and to_index.
+        + NodeCount // Used in find_cycle
+        + EdgeCount // Used in find_cycle
+        + Visitable // Used in find_cycle
```
mtreinish, line 430: This shouldn't be necessary on this trait impl. We're not running `find_cycle` here.

rustworkx-core/src/dag_algo.rs
```diff
@@ -314,6 +319,176 @@ where
...
+    // Throw error if a cycle is found at the current node
+    if let Some(node) = &cur_layer.first() {
+        let check_cycle = find_cycle(&graph, Some(**node));
+        if !check_cycle.is_empty() {
+            return Err(LayersError(Some(format!(
+                "Found cycle at {:?}",
+                check_cycle
+            ))));
+        }
+    }
```
mtreinish, line 380: I'm still worried about this, you're adding extra overhead for a complete DFS of the graph just for error checking when it didn't exist before. This could be a noticeable performance regression especially for large graphs. If you think we need to do result handling here we should do the checking during the iteration since we're already iterating over the graph.

raynelfss: I was thinking about it and the method you suggested before about having a set to keep track of all the nodes was right, but didn't seem to be working well for me, due to a programming error on my end. In previous iterations of me trying to fix that, I forgot to add the elements of the first layer into that set, which resulted in the traversal continuing infinitely. This should be fixed in a future commit.

rustworkx-core/src/dag_algo.rs
raynelfss: No, it is not necessary, I was just trying to avoid having to initialize all the other empty attributes manually. This will be removed in a future commit.

raynelfss:
Thank you for the reviews @mtreinish, I believe this PR should be in a better position after the latest commtis. I managed to remove some of the overhead of iterating through the entire graph to find cycles, instead I keep track of all the layers being found, if one of them repeats, then a cycle is present. I thought of maybe doing it by index but, in some cases, the indices might repeat without it necessarily being a cycle.

```rust
let edge_list = vec![
  (0, 1),
 (1, 2),
 (2, 3),
 (3, 4),
];

let graph = DiGraph::<u32, u32>::from_edges(&edge_list);
let layers: Vec<Vec<NodeIndex>> = layers(&graph, vec![0.into(),1.into()]).collect();
let expected_layers: Vec<Vec<NodeIndex>> = vec![
    vec![0.into(),1.into()],
    vec![1.into(),2.into()],
    vec![2.into(),3.into()],
    vec![3.into(),4.into()],
    vec![4.into()]
];
assert_eq!(layers, expected_layers)
```
In this case, while the layers show repeated indices, they don't represent a cycle.

Also due to the nature of the graph being that when a node is removed the indices shift, therefore keeping the nodes within a range, if a node is provided outside that range (which can be found using `graph.node_bound() -> usize`), the graph doesn't contain that node. This is the only way I found that allowed me to check if a node is in the graph without having to collect all of the nodes.

As always there is room for improvement. If there's any other inconsitencies that anyone might find feel free to leave it as a review comment. Thanks!

rustworkx-core/src/dag_algo.rs
```diff
@@ -314,6 +316,140 @@ where
...
+            }
+            self.cycle_check.insert(self.cur_layer.to_owned());
+            self.next_layer.clear();
+            if self.cur_layer.is_empty() {
+                None
+            } else {
+                Some(self.cur_layer.to_owned())
+            }
+        }
+    }
```
mtreinish, line 450: I still have some concerns over the cycle checking here. This is much better than calling `find_cycle` but this potentially involves 2 copies which feels like one too many. I know we talked about this offline but I still feel like the cycle checking should be by node, not by layer. If you're getting duplicates here that's pointing to a bug somewhere, likely in the user provided first layer.

rustworkx-core/src/dag_algo.rs
```diff
@@ -314,6 +316,140 @@ where
     Ok(Some((path, path_weight)))
 }
 
+/// Return an iterator of graph layers
+///
+/// A layer is a subgraph whose nodes are disjoint, i.e.,
+/// a layer has depth 1. The layers are constructed using a greedy algorithm.
+///
```
mtreinish, line 323: Can you mention this will panic if there are cycles?

rustworkx-core/src/dag_algo.rs
```diff
@@ -314,6 +316,140 @@ where
...
+                        );
+                    if *self.predecessor_count.get(&succ).unwrap() == 0 {
+                        self.next_layer.push(succ);
+                        self.predecessor_count.remove(&succ);
+                    }
+                }
+            }
+            swap(&mut self.cur_layer, &mut self.next_layer);
+            if self.cycle_check.contains(&self.cur_layer) {
+                panic!("The layers algorithm doesn't support graphs with cycles.")
```
mtreinish, line 440: I'm starting to think this should be a result now actually; I apologize this was my earlier mistake. I was thinking about how we document this on the python side, and the answer is we need to raise an exception to python, and that's tricky to do if we panic. I just really don't like iterators with `Result<T>` items. But at the end of the day doing:
```
let layers: Result<Vec<_>> = layers().map().collect();
```
isn't terrible I guess.

releasenotes/notes/add-layers-rustworkx-core-1caaccb1ca292ca8.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Add a function :meth:`~rustworkx-core.dag_algo.layers` in rustworkx-core to 
+    get the layers of a directed acyclic graph.  This is equivalent to the
+    :func:`.layers` function that existed in the Python API but now exposes it
+    for Rust users too.
```
mtreinish, line 7: We should have a fix note documenting the behavior change around calling the python function with a graph with cycles.

src/dag_algo/mod.rs
```diff
@@ -295,99 +297,38 @@ pub fn layers(
...
-                    layer_node.index()
-                )));
-            }
+    let node_set = dag.graph.node_indices().collect::<HashSet<NodeIndex>>();
+    for layer_node in &first_layer {
+        if !node_set.contains(&NodeIndex::new(*layer_node)) {
+            return Err(InvalidNode::new_err(format!(
+                "An index input in 'first_layer' {} is not a valid node index in the graph",
+                layer_node
+            )));
```
mtreinish, line 306: This ends up iterating over the graph twice for an error check. Instead you can do it like this:

```suggestion
    for layer_node in &first_layer {
        if !dag.contains_node(NodeIndex::new(*layer_node)) {
            return Err(InvalidNode::new_err(format!(
                "An index input in 'first_layer' {} is not a valid node index in the graph",
                layer_node
            )));
```

rustworkx-core/src/dag_algo.rs
raynelfss: Yes I will be updating it to be by nodes instead, I just haven't modified the code to make it work that way. I haven't played with this since we spoke offline 😅. I am actively thinking of which ways to modify this method to make it work as intented. I will keep you posted on this.

rustworkx-core/src/dag_algo.rs
raynelfss: Sounds good, I will continue to evaluate this in that case.

releasenotes/notes/add-layers-rustworkx-core-1caaccb1ca292ca8.yaml
raynelfss: Sure thing!

rustworkx-core/src/dag_algo.rs
raynelfss: Not anymore since we do result handling.

rustworkx-core/src/dag_algo.rs
```diff
@@ -333,6 +337,154 @@ where
...
+    N: Debug + Copy + Eq + Hash,
+{
+    type Item = Result<Vec<N>, LayersError>;
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.first_iter {
+            self.first_iter = false;
+            for node in &self.cur_layer {
+                if self.graph.to_index(*node) >= self.graph.node_bound() {
+                    panic!("Node {:#?} is not present in the graph.", node);
+                }
```
mtreinish, line 422: Do we need this explicitly, won't we just get a panic if we try to access `node` in `neighbors_directed()` below?

rustworkx-core/src/dag_algo.rs
```diff
@@ -333,6 +337,154 @@ where
...
+                if self.cycle_check.contains(node) {
+                    return Some(Err(LayersError(
+                        "An invalid first layer was provided.".to_owned(),
+                    )));
```
mtreinish, line 426: Is it possible to have a cycle here? Or are you protecting against duplicate entries in `first_layer`. If it's duplicates we should put that in the error message.

rustworkx-core/src/dag_algo.rs
```diff
@@ -333,6 +337,154 @@ where
...
+                self.cycle_check.insert(*node);
+            }
+            Some(Ok(self.cur_layer.to_owned()))
+        } else if self.cur_layer.is_empty() {
+            None
+        } else {
+            for node in &self.cur_layer {
+                if self.graph.to_index(*node) >= self.graph.node_bound() {
+                    panic!("Node {:#?} is not present in the graph.", node);
+                }
```
mtreinish, line 437: Same question here

rustworkx-core/src/dag_algo.rs
```diff
@@ -333,6 +337,154 @@ where
...
+                        self.predecessor_count.remove(&succ);
+                    }
+                }
+            }
+            swap(&mut self.cur_layer, &mut self.next_layer);
+            self.next_layer.clear();
+            if self.cur_layer.is_empty() {
+                None
+            } else {
+                Some(Ok(self.cur_layer.to_owned()))
```
mtreinish, line 482: I think this a bit more explicit:

```suggestion
                Some(Ok(self.cur_layer.clone()))
```

mtreinish (approved):
This LGTM now thanks for the updates and thoroughness on this PR. I just have a few minor nits and questions inline. But otherwise I think this is ready to merge.

rustworkx-core/src/dag_algo.rs
raynelfss: Not really. From the [docs](https://docs.rs/petgraph/latest/petgraph/graph/struct.Graph.html#method.neighbors_directed):

> Produces an empty iterator if the node doesn’t exist.
Iterator element type is NodeIndex<Ix>.

It only produces an empty iterator, therefore I had to introduce a way to make it panic if the index does not belong.

PR #1196, opened 2024-05-18
Add hyperbolic random graph model generator

SILIZ4:
This adds a generator for the hyperbolic random graph model. There are two issues I wasn't able to solve:

1. The model is only defined for undirected graphs. Since any graph structure works with the generator, I used the `Data` trait and return an error (runtime) when the graph is directed. It's clear to me that it should be a compile error, but I'm not familiar enough with traits and `petgraph`.
2. I wasn't able to add LaTeX in the rustworkx-core documentation and so it feels incomplete.

Related to #150. This is my first contribution and I'm very open to suggestions.


releasenotes/notes/hyperbolic-random-graph-d85c115930d8ac08.yaml
```diff
@@ -0,0 +1,11 @@
...
+  - |
+    Adds new random graph generator function, :func:`.undirected_hyperbolic_random_graph`
+    to generate a random graph using the H² random graph model.
+  - |
+    Adds new function to the rustworkx-core module ``rustworkx_core::generators``
+    ``hyperbolic_random_graph()`` that generates a H^2 hyperbolic random graph.
+
+issues:
+  - |
+    Related to issue `#150 <https://github.com/Qiskit/rustworkx/issues/150>`.
```
IvanIsCoding, line 11: The `issues` release note is related to known issues at release time, not Github issues. You can ommit this, the `features` section is all you need

rustworkx-core/src/generators/random_graph.rs
```diff
@@ -619,14 +620,141 @@ where
     Ok(graph)
 }
 
+/// Generate a hyperbolic random undirected graph.
```
IvanIsCoding, line 623: Maybe we should also note that they can be called "Hyperbolic geometric graph"

src/random_graph.rs
```diff
@@ -380,6 +380,64 @@ pub fn random_geometric_graph(
...
+/// :param float beta: Sigmoid sharpness (nonnegative) of the connection probability.
+/// :param float r: Distance at which the connection probability is 0.5 for the probabilistic model.
+///     Threshold when `beta` is `None`.
+/// :param int seed: An optional seed to use for the random number generator.
+///
+/// :return: A PyGraph object
+/// :rtype: PyGraph
+#[pyfunction]
+#[pyo3(text_signature = "(radii, angles, r, beta, /, seed=None)")]
+pub fn undirected_hyperbolic_random_graph(
```
IvanIsCoding, line 418: If literature always assumes a Hyperbolic Random Graph is undirected, you can shorten the name of this function to `undirected_hyperbolic_random_graph`

rustworkx-core/src/generators/random_graph.rs
```diff
@@ -619,14 +620,141 @@ where
...
+    let num_nodes = radii.len();
+    if num_nodes == 0 || radii.len() != angles.len() {
+        return Err(InvalidInputError {});
+    }
+    if radii.iter().any(|x| x.is_infinite() || x.is_nan()) {
+        return Err(InvalidInputError {});
+    }
+    if angles.iter().any(|x| x.is_infinite() || x.is_nan()) {
+        return Err(InvalidInputError {});
+    }
```
IvanIsCoding, line 688: You could `.chain` these iterators to perform this check

rustworkx-core/src/generators/random_graph.rs
```diff
@@ -619,14 +620,141 @@ where
...
+    if radii.iter().fold(0_f64, |a, &b| a.min(b)) < 0. {
```
IvanIsCoding, line 689: f64 doesn't implement `Ord` but you can use the workaround from https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.min to simplify

rustworkx-core/src/generators/random_graph.rs
```diff
@@ -619,14 +620,141 @@ where
...
+        return Err(InvalidInputError {});
+    }
+    if angles.iter().map(|x| x.abs()).fold(0_f64, |a, b| a.max(b)) > PI {
```
IvanIsCoding, line 692: Same comment as above, I'd replace the fold with the `reduce` and `f64::max`

rustworkx-core/src/generators/random_graph.rs
```diff
@@ -619,14 +620,141 @@ where
...
+                );
+            }
+        }
+    }
+    Ok(graph)
+}
+
+#[inline]
+fn hyperbolic_distance(r1: &f64, theta1: &f64, r2: &f64, theta2: &f64) -> f64 {
+    (r1.cosh() * r2.cosh() - r1.sinh() * r2.sinh() * (theta1 - theta2).cos()).acosh()
```
IvanIsCoding, line 746: Genuinely by biggest concern about this code is numerical stability, we should probably rewrite this to avoid multiplications

rustworkx-core/src/generators/random_graph.rs
```diff
@@ -619,14 +620,141 @@ where
...
+        .enumerate()
+        .take(num_nodes - 1)
+    {
+        for (w, (r2, theta2)) in radii.iter().zip(angles.iter()).enumerate().skip(v + 1) {
+            let dist = hyperbolic_distance(r1, theta1, r2, theta2);
+            let is_edge = match beta {
+                Some(b) => {
+                    let prob = 1. / ((b / 2. * (dist - r)).exp() + 1.);
+                    let u: f64 = between.sample(&mut rng);
+                    u < prob
```
IvanIsCoding, line 728: Higly recommend us to expand this comparison instead of caculating the reciprocal of a large f64

IvanIsCoding (commented):
Overall this is looking pretty good! I wouldn't worry to much about the `petgraph`'s traits, this seems to make sense to me.

I don't have much to comment on Hyperbolic Random Graphs as it is not our core area of expertise, but I did add some comments about floating-point arithmetic:
* Intsead of multiplying `cosh(a)*cosh(b)`, I'd look for `(cosh(a + b) - cos(a - b))/2` as this can might be able to boost the accuracy. Same applies for `sinh`. Check with https://herbie.uwplse.org/ if you want to be sure
* I'd also be careful with comparisons involving floats

Lastly, our pure Rust documentation doesn't support LaTex as we'd need to add KaTeX with something like https://docs.rs/rustdoc-katex-demo/0.1.5/rustdoc_katex_demo/. So you can leave it as is and we'll figure it out later

rustworkx-core/src/generators/random_graph.rs
SILIZ4: Good catch, this is fixed.

SILIZ4:
I believe I have fixed everything that was talked about. I don't see how to rewrite this version of the hyperbolic distance in a numerical friendly way. That said, I generalized the code to allow for any valid dimension of the hyperbolic space. To do so, it was easier to change the coordinate system and as a bonus, the hyperbolic distance has a more numerical friendly expression.

SILIZ4:
I have no idea why Coverall now fails. Do you know why that is?

IvanIsCoding:
> I have no idea why Coverall now fails. Do you know why that is?

I think an update broke our CI, I'll try to fix it

IvanIsCoding (approved):
I think this is good to go

Issue #1165, opened 2024-04-19
Move `lexicographical_topological_sort()` functionality to rustworkx-core


mtreinish:
The `lexicographical_topological_sort()` function is only exposed via a Python interface currently:

https://github.com/Qiskit/rustworkx/blob/ab93329be6d8aaccb6c090f4b3ab6f9540ee8114/src/dag_algo/mod.rs#L347-L485

We should port it to rustworkx-core, so that rust users can leverage the function.

One tweak that probably makes sense for the rustworkx-core version is that instead of returning a `PyList` (realistically a `Vec<N>`) of node weights we should have it return of an iterator of node ids. This would be more flexible and performant for rust space users and for the python side of rustworkx we can just consume the iterator to build the pylist return there (for api compatibility).

PR #1197, opened 2024-05-20
Add lexicographical_topological_sort function to rustworkx-core

mtreinish:
This commit adds a rust generic implementation of the lexicographical topological sort function that was previously only exposed via the python interface to rustworkx-core. This new function is used internally by rustworkx and replaces the python specific implementation that used to be there. But it now exposes an interface to use it to rust.

Fixes: #1165


IvanIsCoding (approved):
LGTM

PR #1198, opened 2024-05-21
Fix `coveralls` by changing the extension to `.lcov`

IvanIsCoding:
I have no idea why but our Coveralls integration started complaining about `lcov` files with `.info`, so I switched to `.lcov`


mtreinish (approved):
Heh, I'm surprised the action cares about the file extension now, but this has fixed the job so let's go with it.

PR #1200, opened 2024-05-22
Add stochastic block model generator

SILIZ4:
This naive O(n^2) algorithm might not be the fastest possible algorithm since there exists a O(m) algorithm for G(n, p), but I didn't see how to generalize it for non square "sub-matrices". I'm not convinced that networkx's implementation with iterators is actually more efficient.

I was unable to add hyperlinks to the G(n, p) generator in rustworkx and rustworkx-core. Is there a way to do so?


src/random_graph.rs
```diff
@@ -273,6 +273,114 @@ pub fn undirected_gnm_random_graph(
...
+/// nodes ``u`` and ``v`` depends on their block (or community) and is given by
+/// ``probabilities[blocks[u]][blocks[v]]``. The number of nodes and the number of
+/// blocks are inferred from ``blocks``.
+///
+/// This algorithm has a time complexity of :math:`O(n^2)` for :math:`n` nodes.
+///
+/// Arguments:
+///
+/// :param list[int] blocks: Block membership (between 0 and B-1) of each node.
+/// :param list[list[float]] probabilities: B x B matrix that contains the
```
IvanIsCoding, line 289: This can probably be a 2d numpy.array

rustworkx-core/src/generators/random_graph.rs
```diff
@@ -305,6 +305,133 @@ where
...
+///     || (),
+///     || (),
+/// )
+/// .unwrap();
+/// assert_eq!(g.node_count(), 3);
+/// assert_eq!(g.edge_count(), 6);
+/// ```
+pub fn sbm_random_graph<G, T, F, H, M>(
+    blocks: &[usize],
+    probabilities: &[Vec<f64>],
```
IvanIsCoding, line 347: You can import the ndarray crate and have this a 2d array

releasenotes/notes/sbm-random-graph-bf7ccd8e938f4218.yaml
```diff
@@ -0,0 +1,9 @@
+features:
+  - |
+    Adds new random graph generator in rustworkx for the stochastic block model.
+    There is a generator for directed :func:`.directed_sbm_random_graph` and
+    undirected graphs :func:`.undirected_sbm_random_graph`.
+  - |
+    Adds new function ``sbm_random_graph`` to the rustworkx-core module
+    ``rustworkx_core::generators`` that samples a graph from the stochastic
+    block model.
```
IvanIsCoding, line 9: You can add drawings of some example graphs if you want, check https://github.com/Qiskit/rustworkx/blob/955d9890ac7073c3eb8878e4fdcfeedfeef3e73d/releasenotes/notes/0.14/add-albert-graph-5a7d393e1fe18e9d.yaml#L5 for an example

IvanIsCoding (commented):
I haven't had time to give a complete review but:
* Complexity seems fine, if you think there are improvements you can open an issue after we merge this one
* We should probably accept numpy arrays as an input, I have a feeling that is the datatype people would use for a matrix

rustworkx-core/src/generators/random_graph.rs
SILIZ4: Is rustworkx-core intended to work with ndarray? I feel like I should avoid adding a dependency if possible. I agree that for the Python API it makes sense to directly support a numpy ndarray.

IvanIsCoding: Eventually we will have to add ndarray because of some of the centrality functions/Floyd-Warshall. So you can be the first one to add it

IvanIsCoding: cc @mtreinish for asking a second opinion

mtreinish: I'm fine adding an ndarray dependency on rustworkx core. I think eventually we'll want to start using features and optional dependencies. But for right now I don't see a problem with doing this for core, especially because we're using it in the python side already.

IvanIsCoding:
> This naive O(n^2) algorithm might not be the fastest possible algorithm since there exists a O(m) algorithm for G(n, p), but I didn't see how to generalize it for non square "sub-matrices". I'm not convinced that networkx's implementation with iterators is actually more efficient.
> 
> I was unable to add hyperlinks to the G(n, p) generator in rustworkx and rustworkx-core. Is there a way to do so?

By the way, for links you want:
* rustworkx you use something like this (https://github.com/Qiskit/rustworkx/blob/955d9890ac7073c3eb8878e4fdcfeedfeef3e73d/releasenotes/notes/fix-type-search-6abecbc064e1dbf0.yaml#L4),
* For rustworkx-core follow the rustdoc guide https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

SILIZ4:
I wasn't sure how dependencies versions should be handled so I added ndarray directly with cargo. Docs hyperlinks are fixed. I also changed the first argument for the community sizes instead of membership. It makes more sense and is what networkx uses.

IvanIsCoding:
> I wasn't sure how dependencies versions should be handled so I added ndarray directly with cargo. Docs hyperlinks are fixed. I also changed the first argument for the community sizes instead of membership. It makes more sense and is what networkx uses.

I will tweak ndarray to use [Cargo workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html)

IvanIsCoding (approved):

PR #1204, opened 2024-05-31
Update mpl_draw() to fix multigraph plots

### LINKED ISSUE #774: `mpl_draw()` does not work for multigraphs
### Information

- **Python version**: 3.9
- **Rust version**: I do not know, but not relevant for this issue.
- **Operating system**: macOs

When plotting a multigraph with parallel edges using `rustworkx.visualization.mpl_draw()`, two issues occur:
1. The arrows for multiple edges are printed on top of each other.
2. Only one label is drawn for every pair of nodes. (The labels are currently saved in a dict, which uses node pair as the key.)

I see three possible solutions:
- The documentation could warn that `mpl_draw()` does not work for multigraphs.
- `mpl_draw()` could raise a warning when a multigraph is passed. ([networkx seems to do that for at least the edge labels](https://github.com/networkx/networkx/blob/8334e96bd794e56940e5b554c6773d62617141e3/networkx/drawing/nx_pylab.py#L1156))
- Ideally, `mpl_draw()` would be modified to show parallel edges. As far as I can see, this would include three steps, similar to what has been outlined in https://stackoverflow.com/a/70245742:
  1. Modifying `draw_edges()` so it bends the arrows of parallel edges to prevent overlap.
  2. Modifying `draw_edge_labels()` so it plots the edge labels at the positions of the arrows and not simply on the line connecting the nodes.
  3. Modifying `draw_graph()` so `kwds["edge_labels"]` contains one entry per edge and not one per node-pair. (The key of the dict should be the edge id and not the node pair.) 

1. Create a graph with two parallel edges:
```Python
graph = rustworkx.PyDiGraph()
graph.add_node('A')
graph.add_node('B')
graph.add_edge(0, 1, 1)
graph.add_edge(0, 1, 2)
mpl_draw(graph, with_labels=True, labels=str, edge_labels=str, alpha=0.5)
```
2. Save the image as a vector graphic and inspect it.

The image will contain two arrows on top of each other and the label for edge 2. The label for edge 1 is not drawn.

maxwell04-wq:
Fixes #774. 
Updates:
- `draw_edges()`: set `rad=0.25` for looped nodes. 
- `draw_edge_labels()`: Modify label offsets for looped nodes.


maxwell04-wq:
@mtreinish one major hiccup in the implementation is that the position of the edges is calculated on the plot in the `draw_edges()` function and the position of the labels can only be evaluated relative to the position of the nodes and not the edges in the `draw_edge_labels()` function. While the current parameters for the curved edges of multigraphs work for plotting the labels, maybe there's a better way to evaluate the position of the labels.

rustworkx/visualization/matplotlib.py
```diff
@@ -763,7 +773,7 @@ def _connectionstyle(posA, posB, *args, **kwargs):
             mutation_scale=mutation_scale,
             color=arrow_color,
             linewidth=line_width,
-            connectionstyle=_connectionstyle,
+            connectionstyle=connectionstyle + f", rad = {rad}",
```
IvanIsCoding, line 776:
```suggestion
            connectionstyle=f"{connectionstyle}, rad = {rad}",
```

rustworkx/visualization/matplotlib.py
```diff
@@ -754,6 +754,16 @@ def _connectionstyle(posA, posB, *args, **kwargs):
         else:
             line_width = width
 
+        # radius of edges
+
+        reverse_edge = np.concatenate(([dst], [src]))
+        for edge in edge_pos:  # the loop can be optimized
+            if bool(np.sum(np.all(np.equal(edge, reverse_edge)))):
```
IvanIsCoding, line 761: I think there are easier and simpler ways to write this. I don't quite understand why we are using NumPy for this check either

IvanIsCoding (commented):
I think we need to work on the loop condition. You'll also need to add a release note (https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md#release-notes)

maxwell04-wq:
@IvanIsCoding I have not tried to align the labels vertically as, following the logic of `rad` in `connectionstyle=arc3`, simply offsetting the `y` coordinates of labels to `rad` generates a figure wherein the labels are perfectly placed on the edges. 
test_fig
Would you suggest that I work on aligning the edge labels?

rustworkx/visualization/matplotlib.py
IvanIsCoding: Reopening as the last commit deleted

rustworkx/visualization/matplotlib.py
```diff
@@ -754,6 +754,15 @@ def _connectionstyle(posA, posB, *args, **kwargs):
         else:
             line_width = width
 
+        # radius of edges
+        reverse_edge = [dst, src]
+        if (
+            len(np.where(np.all(edge_pos == reverse_edge, axis=(1, 2)))[0]) != 0
```
IvanIsCoding, line 760: So this scales quadratically on the number of edges. `O(E^2)`  is not ideal, if we plot a dense graph this is going to be a problem. You need to rewrite this condition using a `set` or something similar instead of naively looking through all the edges

releasenotes/notes/fix-mpl-draw-digraph-plots-aecf86738ab9b0db.yaml
```diff
@@ -0,0 +1,10 @@
+---
+fixes:
+  - |
+    Fixes the plots of multigraphs using the `mpl_draw()` function. Refer to
+    `#12345 <https://github.com/Qiskit/rustworkx/issues/774>` for more
+    details.
+other:
+  - |
+     The radius of the edges of self-loops in multigraphs is set to `0.25`. 
+     The edge labels are offset accordingly.
```
IvanIsCoding, line 10: You don't need the other section

releasenotes/notes/fix-mpl-draw-digraph-plots-aecf86738ab9b0db.yaml
```diff
@@ -0,0 +1,10 @@
+---
+fixes:
+  - |
+    Fixes the plots of multigraphs using the `mpl_draw()` function. Refer to
+    `#12345 <https://github.com/Qiskit/rustworkx/issues/774>` for more
+    details.
```
IvanIsCoding, line 6: The release notes would be better if you actually drew a multigraph with `mpl_draw` as an example. You can use https://github.com/Qiskit/rustworkx/blob/53545127b2235b8f5eb517fefe0502029fdb69d7/releasenotes/notes/0.14/add-albert-graph-5a7d393e1fe18e9d.yaml#L16 as an example, it becomes this: https://www.rustworkx.org/release_notes.html#new-features

IvanIsCoding:
release_notes_2_0
It seems to be working, I will merge the PR

IvanIsCoding (approved):

PR #1207, opened 2024-06-01
Add periodic, with_positions to hexagonal_lattice_graph

### LINKED ISSUE #803: Add support to `hexagonal_lattice_graph` for periodcity and position coordinates
networkx's [`hexagonal_lattice_graph`](https://networkx.org/documentation/stable/reference/generated/networkx.generators.lattice.hexagonal_lattice_graph.html) function provides support for periodicity and setting position coordinates, and it would be great if rustworkx's `hexagonal_lattice_graph` did the same.

Definition of done:
- a `periodic` bool argument added to `hexagonal_lattice_graph`
- a `with_positions` bool argument added to `hexagonal_lattice_graph`

cc @mrossinek

WingCode:
closes #803


rustworkx-core/src/generators/hexagonal_lattice_graph.rs
```diff
@@ -78,16 +78,24 @@ pub fn hexagonal_lattice_graph<G, T, F, H, M>(
     mut default_node_weight: F,
     mut default_edge_weight: H,
     bidirectional: bool,
+    periodic: bool,
+    with_positions: Option<F>,
```
IvanIsCoding, line 82: This is a breaking API change on the Rust side. On the Python side, you can use optional arguments but on Rust there is not much we can do.

IvanIsCoding (commented):
Thanks for the PR! I know your code does not compile yet, but I think we can get started on dicussing how to tackle the breaking change on the Rust side

jpacold:
Hi @WingCode -- I noticed that this PR seems to only add `with_positions`, so I opened a separate draft PR just on the `periodic` part. Please let me know if you are planning to continue looking at this issue (I don't want to get in the way or duplicate your work).

WingCode:
Hi @jpacold , Thank you for asking. Yes, you can go right ahead with issue. I am not sure when I will be able to get back to this issue.

IvanIsCoding:
Closing this as obsolete after #1213. Nevertheless, thanks for getting the work started!

PR #1208, opened 2024-06-01
Move ancestors and descendants to rustworkx-core

mtreinish:
This commit adds an implementation of the ancestors and descendants functions to the rustworkx-core crate exposing the functions to rust users. The existing implementation in the rustworkx crate is removed and it is updated to call the rustworkx-core functions. These new functions will be more efficient as they're not using dijkstra's algorithm to find a path from nodes now and instead are just doing a BFS. The rustwork-core functions also return an iterator of nodes.


IvanIsCoding (approved):
LGTM, this seems pretty straightforward

PR #1209, opened 2024-06-02
Move bfs_successors and bfs_predecessors to rustworkx-core

mtreinish:
This commit adds an implementation of the bfs_successors and bfs_predecessors functions to the rustworkx-core crate exposing the functions to rust users. The existing implementation in the rustworkx crate is removed and it is updated to call the rustworkx-core functions. The rustwork-core functions return an iterator of nodes indices and the Python crate side consumes the iterator to build the bfs view these functions previously returned.


IvanIsCoding (approved):
LGTM

PR #1210, opened 2024-06-07
Move `collect_runs()` to rustworkx-core

### LINKED ISSUE #1199: Move `collect_runs()` functionality to rustworkx-core

The `collect_runs()` function is only exposed via a Python interface currently:

https://github.com/Qiskit/rustworkx/blob/0ec113b077fc27943947496cbef4f88daed558bf/src/dag_algo/mod.rs#L607-L670

We should port it to rustworkx-core, so that rust users can leverage the function.

One tweak that probably makes sense for the rustworkx-core version is that instead of returning a Vec of Vecs of node weights we should have it return of an iterator of Vecs node ids. This would be more flexible and performant for rust space users and for the python side of rustworkx we can just collect the iterator to a Vec (for backwards compatibility).

eliarbel:
This PR moves `collect_runs()` functionality to rustworkx-core into `rustworkx-core/src/dag_algo.rs`. This will be used by `collect_runs()` in `src/dag_algo/mod.rs`. 

Closes #1199 

**Tasks:** 
- [x] Read the CONTRIBUTING document
- [x] Implement in core, including error handling
- [x] Add tests in core 
- [x] Convert core implementation to return iterator over runs
- [x] Use core implementation in rustworkx
- [x] Update documentation 
- [x] Format code


eliarbel:
@mtreinish this is ready for review now, however CI `clippy` complains on parts in the code not included in this PR (with `error: usage of a legacy numeric constant`). Any idea?

mtreinish:
My guess is that the release of rust 1.79 last week is causing new clippy rules to be run and that's causing the failure. Although I thought we should have everything pinned to the MSRV in CI to avoid this kind of situation.

eliarbel:
> My guess is that the release of rust 1.79 last week is causing new clippy rules to be run and that's causing the failure. Although I thought we should have everything pinned to the MSRV in CI to avoid this kind of situation.

Yes, I was on 1.77. Updating to 1.79 I see those warnings locally as well. I'll wait for it to be resolved.

rustworkx-core/src/dag_algo.rs
```diff
@@ -475,6 +475,150 @@ where
...
+///
+/// let positive_payload = |n| -> Result<bool, ()> {Ok(*graph.node_weight(n).expect("i32") > 0)};
+/// let mut runs = collect_runs(&graph, positive_payload).expect("Some");
+///
+/// assert_eq!(runs.next(), Some(Ok(vec![n3])));
+/// assert_eq!(runs.next(), Some(Ok(vec![n2])));
+/// assert_eq!(runs.next(), None);
+/// ```
+///
+pub fn collect_runs<G, F, E>(graph: G, include_node_fn: F) -> Option<Runs<G, F, E>>
```
mtreinish, line 516: I think it would be a bit cleaner to have this be:
```suggestion
pub fn collect_runs<G, F, E>(graph: G, include_node_fn: F) -> Option<impl Iterator<Item = Result<Vec<G::NodeId>, E>>>
```

The signature is a bit uglier but it preserves the flexibility for us to change things up a bit more in the future as all we're committing to is returning an iterator. It also lets us leave `Runs` as a private.

src/dag_algo/mod.rs
```diff
@@ -561,47 +562,28 @@ pub fn collect_runs(
...
-        if !filter_node(&graph.graph[node])? || seen.contains(&node) {
-            continue;
-        }
-        seen.insert(node);
-        let mut group: Vec<PyObject> = vec![graph.graph[node].clone_ref(py)];
-        let mut successors: Vec<NodeIndex> = graph
-            .graph
-            .neighbors_directed(node, petgraph::Direction::Outgoing)
+
+    let mut result: Vec<Vec<PyObject>> = Vec::new();
```
mtreinish, line 575: Can we use `size_hint` (https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint) along with `Vec::with_capacity()` here to save some allocations?

rustworkx-core/src/dag_algo.rs
```diff
@@ -475,6 +475,150 @@ where
...
+    Some(runs)
+}
+
+/// Auxiliary struct to make the output of [`collect_runs`] iteratable
+///
+/// If the filtering function passed to [`collect_runs`] returns an error, it is propagated
+/// through `next` as `Err`. In this case the run in which the error occured will be skipped
+/// but the iterator can be used further until consumed.
+///
+pub struct Runs<G, F, E>
```
mtreinish, line 549: With my above suggestion I think we can leave this private.

```suggestion
struct Runs<G, F, E>
```

src/dag_algo/mod.rs
```diff
@@ -29,10 +30,10 @@ use pyo3::types::PyList;
 use pyo3::Python;
 
 use petgraph::algo;
-use petgraph::graph::NodeIndex;
+// use petgraph::graph::NodeIndex;
```
mtreinish, line 33:
```suggestion
```

src/dag_algo/mod.rs
```diff
@@ -29,10 +30,10 @@ use pyo3::types::PyList;
...
 use petgraph::prelude::*;
 use petgraph::stable_graph::EdgeReference;
-use petgraph::visit::NodeCount;
+// use petgraph::visit::NodeCount;
```
mtreinish, line 36:
```suggestion
```

mtreinish (commented):
Besides the compiler errors in the rust tests this is looking good. Just a few small inline comments.

src/dag_algo/mod.rs
```diff
@@ -561,47 +560,28 @@ pub fn collect_runs(
...
-        if !filter_node(&graph.graph[node])? || seen.contains(&node) {
-            continue;
-        }
-        seen.insert(node);
-        let mut group: Vec<PyObject> = vec![graph.graph[node].clone_ref(py)];
-        let mut successors: Vec<NodeIndex> = graph
-            .graph
-            .neighbors_directed(node, petgraph::Direction::Outgoing)
+
+    let mut result: Vec<Vec<PyObject>> = Vec::with_capacity(core_runs.size_hint().1.unwrap_or(0));
```
mtreinish, line 573: This will result in some overallocation, but I don't think it's a big deal, it's also only temporary as we convert this to a pylist almost immediately so this is fine.

mtreinish (approved):
LGTM now, thanks for doing this.

PR #1212, opened 2024-06-08
Hyperbolic generator improvements + fix

SILIZ4:
The hyperbolic distance returned NaN for very close vertices. This meant the hyperbolic geometric graph generator wouldn't create edges that should exist.

I realized it didn't make sense to accept infinite coordinates (hyperbolic distance is undefined in most cases) and return an error instead.

I swapped the order of the parameters in the rustworkx-core api to match the rustworkx api.

Finally, it's inconvenient to specify the time coordinates since it can be computed using the other coordinates. Users only need to specify the useful coordinates.


releasenotes/notes/hyperbolic-random-graph-fix-2677e7679bc9b18b.yaml
```diff
@@ -0,0 +1,8 @@
+fixes:
```
IvanIsCoding, line 1: We never released the code so you don’t have to include release notes this time

rustworkx-core/src/generators/random_graph.rs
```diff
@@ -12,6 +12,7 @@
 
 #![allow(clippy::float_cmp)]
 
+use core::f64;
```
IvanIsCoding, line 15: I think you can just use f64 out of the box? We always include `std`, I think we never really use core

rustworkx-core/src/generators/random_graph.rs
```diff
@@ -856,17 +860,23 @@ where
...
+    let mut sum_squared_x = 0.;
+    let mut sum_squared_y = 0.;
+    let mut inner_product = 0.;
+    for (x_i, y_i) in x.iter().zip(y.iter()) {
+        if x_i.is_infinite() || y_i.is_infinite() || x_i.is_nan() || y_i.is_nan() {
+            return f64::NAN;
+        }
+        sum_squared_x += x_i * x_i;
+        sum_squared_y += y_i * y_i;
+        inner_product += x_i * y_i;
```
IvanIsCoding, line 873:
```suggestion
        sum_squared_x = x_i::mul_add(x_i, x_i, sum_squared_x);
        sum_squared_y = x_i::mul_add(y_i, y_i, sum_squared_y);
        inner_product = x_i::mul_add(x_i, y_i, inner_product);
```
I think we might as well use fused add-multiply after we confirmed the floats are valid

SILIZ4: I wasn't aware of that function, good to know!

IvanIsCoding (approved):

PR #1213, opened 2024-06-10
Add `periodic` to `hexagonal_lattice_graph`

### LINKED ISSUE #803: Add support to `hexagonal_lattice_graph` for periodcity and position coordinates

jpacold:
Implements one of the features needed to close #803. I think this is complementary to PR #1207, which looks like it starts implementing the other part (`with_positions`).

Rust changes:
* Moved calls to `graph.add_edge` into a closure to make the code more readable
* Added `periodic` argument. The comments on #803 suggest implementing this by following the logic in networkx, which constructs a non-periodic graph and then contracts pairs of nodes on the lattice boundaries. I chose to instead create a smaller number of nodes up front.
* Added tests.

Python changes:
* Updated stub file.
* Added tests, including a check for isomorphism with the periodic lattices generated by networkx.


rustworkx-core/src/generators/hexagonal_lattice_graph.rs
```diff
@@ -88,95 +93,88 @@ where
...
+    };
+
+    // Number of vertical chains
+    let collen = if periodic { cols } else { cols + 1 };
+
+    let num_nodes = if periodic {
+        rowlen * collen
+    } else {
+        rowlen * collen - 2
+    };
```
IvanIsCoding, line 117: You can rewrite all the variables with a single match on that condition using destructuring e.g. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ef7020b44cc19135583d87a1dcf6d1b2

rustworkx-core/src/generators/hexagonal_lattice_graph.rs
```diff
@@ -317,8 +315,119 @@ mod tests {
...
+        let g: petgraph::graph::UnGraph<(), ()> =
+            hexagonal_lattice_graph(2, 2, || (), || (), false, true).unwrap();
+        assert_eq!(g.node_count(), 8);
+        assert_eq!(g.edge_count(), expected_edges.len());
+        assert_eq!(
+            expected_edges,
+            g.edge_references()
+                .map(|edge| (edge.source().index(), edge.target().index()))
+                .collect::<Vec<(usize, usize)>>(),
+        );
```
IvanIsCoding, line 363: I wouldn't assert the order of the edges in the generated list because that is something we reserve the right to change. You might want to compare the equality with a `HashSet` or a sorted list though.

I am glad the hard-coded edge order didn't affect your PR though.

rustworkx-core/src/generators/hexagonal_lattice_graph.rs
```diff
@@ -317,8 +315,119 @@ mod tests {
...
+        let g: petgraph::graph::DiGraph<(), ()> =
+            hexagonal_lattice_graph(2, 2, || (), || (), false, true).unwrap();
+        assert_eq!(g.node_count(), 8);
+        assert_eq!(g.edge_count(), expected_edges.len());
+        assert_eq!(
+            expected_edges,
+            g.edge_references()
+                .map(|edge| (edge.source().index(), edge.target().index()))
+                .collect::<Vec<(usize, usize)>>(),
+        );
```
IvanIsCoding, line 391: If you have multiple assertions you come make a macro or function names `asserted_unordered_eq!`

rustworkx-core/src/generators/hexagonal_lattice_graph.rs
```diff
@@ -317,8 +315,119 @@ mod tests {
...
+        let g: petgraph::graph::DiGraph<(), ()> =
+            hexagonal_lattice_graph(2, 2, || (), || (), true, true).unwrap();
+        assert_eq!(g.node_count(), 8);
+        assert_eq!(g.edge_count(), expected_edges.len());
+        assert_eq!(
+            expected_edges,
+            g.edge_references()
+                .map(|edge| (edge.source().index(), edge.target().index()))
+                .collect::<Vec<(usize, usize)>>(),
+        );
```
IvanIsCoding, line 431: Same

IvanIsCoding (commented):
I will let people that made the feature request review the content, but overall the Rust code is looking pretty good! I left some minor comments

rustworkx-core/src/generators/hexagonal_lattice_graph.rs
jpacold: Thanks for the feedback -- I ended up adding two functions to do this for directed and undirected graphs.

mtreinish:
Can you leave a comment on #803 so we can assign the issue to you for unitaryhack?

jpacold:
Hi @mtreinish -- I can do that, though I'm not sure about closing the issue since the `with_positions` part of this is still not done. It's fine with me if that ends up taking longer than the unitaryhack deadline.

rustworkx-core/src/generators/hexagonal_lattice_graph.rs
jpacold: Forgot to mention that I did this as well in 6497d44.

jpacold:
I tried adding the `with_positions` argument, and got something that works. However, I think I need some help here since it seems like there should be a better way to do this.

In Rust we started with this function
```
pub fn hexagonal_lattice_graph<G, T, F, H, M>(
    rows: usize,
    cols: usize,
    default_node_weight: F,
    default_edge_weight: H,
    bidirectional: bool,
    periodic: bool,
) -> Result<G, InvalidInputError>
where
    G: Build + Create + Data<NodeWeight = T, EdgeWeight = M> + NodeIndexable,
    F: FnMut() -> T,
    H: FnMut() -> M,
    G::NodeId: Eq + Hash,
```

Coming from C++, I really wanted to change the argument `default_node_weight` to something analogous to `std::variant<FNoArgs, FTwoArgs>`, where `FNoArgs` would be constrained to `FnMut() -> T` and `FTwoArgs` would be constrained to `FnMut(usize, usize) -> T`. However, I couldn't find a way to do this with the Rust traits system.

I ended up adding a second function
```
pub fn hexagonal_lattice_graph_weighted<G, T, F, H, M>(
    rows: usize,
    cols: usize,
    node_weight: F,
    default_edge_weight: H,
    bidirectional: bool,
    periodic: bool,
) -> Result<G, InvalidInputError>
where
    G: Build + Create + Data<NodeWeight = T, EdgeWeight = M> + NodeIndexable,
    F: FnMut(usize, usize) -> T,
    H: FnMut() -> M,
    G::NodeId: Eq + Hash,
```
and factoring out as much of the shared logic as possible into a builder object. This makes it possible to match the functionality in networkx, but seems like an awkward design and leads to some really messy code in `generators.rs`.

rustworkx-core/src/generators/hexagonal_lattice_graph.rs
```diff
@@ -17,19 +17,200 @@ use petgraph::visit::{Data, NodeIndexable};
 
 use super::InvalidInputError;
 
+mod utils {
```
IvanIsCoding, line 20: I don't really think we need this module, it could be include with the rest of the code. But that is a minor detail for now

tests/generators/test_hexagonal.py
```diff
@@ -415,3 +416,162 @@ def test_hexagonal_graph_0_0(self):
...
+        graph2cols = rustworkx.generators.hexagonal_lattice_graph(2, 2, periodic=True)
+        subGraph = graph2cols.subgraph(hexagons[0])
+        self.assertFalse(rustworkx.is_isomorphic(subGraph, C6))
+        self.assertEqual(len(subGraph.edges()), 7)
+
+    def test_hexagonal_graph_periodic_degree(self):
+        """In a periodic hexagonal lattice, all nodes must have
+        degree 3 (idea copied from the networkx test suite)."""
+        for nRows in range(2, 8):
+            for nCols in range(2, 8, 2):
```
IvanIsCoding, line 440: These should be subtests but it's something we can fix later

tests/generators/test_hexagonal.py
```diff
@@ -415,3 +416,162 @@ def test_hexagonal_graph_0_0(self):
...
+        self.assertEqual(len(graph), 12)
+        self.assertEqual(len(graph.edges()), len(expected_edges))
+        self.assertEqual(set(graph.edge_list()), set(expected_edges))
+
+    def test_hexagonal_graph_periodic_networkx_equivalent(self):
+        """Networkx uses different logic to construct a periodic hexagonal
+        lattice graph, so here we check that the results from rustworkx and
+        networkx are isomorphic for a few cases."""
+        for nRows in range(2, 8):
+            for nCols in range(2, 8, 2):
```
IvanIsCoding, line 566: These should be subtests but it's something we can fix later

IvanIsCoding (approved):
The test cases look pretty strong and give me confidence that this code is correct.

There are some minor details like using the `utils` module, leaving `ToDo`s, we don't have a release note, etc. But in the grand scheme of things, this is a great addition I want to be in the 0.15 release

jpacold:
Thanks for approving the PR -- should I open another issue for the remaining ToDos or just open a second PR once I get to them?

IvanIsCoding:
> Thanks for approving the PR -- should I open another issue for the remaining ToDos or just open a second PR once I get to them?

Sure, you can make a second PR if you want to clean up the code

PR #1215, opened 2024-06-10
Add `is_semi_connected()` to `rustworkx`

### LINKED ISSUE #446: Add `is_semiconnected` function.
A graph is semiconnected if, and only if, for any pair of nodes, either one is reachable from the other, or they are mutually reachable. 

For reference, see https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.components.is_semiconnected.html#networkx.algorithms.components.is_semiconnected

For the implementation, we can consider adapting `petgraph::condesation` function https://docs.rs/petgraph/0.6.0/petgraph/algo/fn.condensation.html. We can't use it directly since it only works with `petgraph::Graph` but we have a `petgraph::StableGraph` as our underlying graph data structure.

Procatv:
Implemented the is_semi_connected function, I don't believe it's fully ready to go so I would appreciate feedback regarding my changes, and any more tests I can add, and etc. Followed implementation from definition: [https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.components.is_semiconnected.html#networkx.algorithms.components.is_semiconnected ](url)

Fixes #446


src/connectivity/mod.rs
```diff
@@ -266,6 +269,51 @@ pub fn is_weakly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
+        temp_graph.add_edge(node_map[source.index()], node_map[target.index()], ());
+    }
+
+    let condensed = condensation(temp_graph, false);
+    let condensed_digraph = DiGraph::from(condensed);
+
+    let topo_sort = toposort(&condensed_digraph, None)
+        .map_err(|_| PyValueError::new_err("Graph has cycles"))?;
+
+    for window in topo_sort.windows(2) {
```
IvanIsCoding, line 305: We should parallelize this loop with Rayon using `par_iter`/`into_par_iter` and use a `RwLock` to prevent calling `petgraph::algo::has_path_connecting` when we know the answer is `false`

IvanIsCoding (commented):
Rust code is looking good BUT I think we can still take advantage of multi-threading to boost our speedup compared to sequential libraries. Let me know if you have more questions.

The tests are good but I think we should add some larger inputs. Our generators can help with that

src/connectivity/mod.rs
```diff
@@ -266,6 +270,63 @@ pub fn is_weakly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
+                nodes.push(node);
+            }
+            nodes.reverse();
+            nodes
+        },
+    };
+
+    let pairs: Vec<_> = topo_sort.windows(2).map(|window| (window[0], window[1])).collect();
+    pairs.into_par_iter().for_each(|(u,v)| {
+        if !(petgraph::algo::has_path_connecting(&condensed_digraph, u, v, None)){
```
IvanIsCoding, line 320: Thinking more deeply about this we don't need `has_path_connecting`, I think just checking if the edge exists is good enough

IvanIsCoding (commented):
LGTM, just leaving one final comment about removing `has_path_connecting` for something simpler

src/connectivity/mod.rs
```diff
@@ -266,6 +270,65 @@ pub fn is_weakly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
+            nodes.reverse();
+            nodes
+        },
+    };
+
+    // petgraph::algo::has_path_connecting(&condensed_digraph, u, v, None
+
+    let pairs: Vec<_> = topo_sort.windows(2).map(|window| (window[0], window[1])).collect();
+    pairs.into_par_iter().for_each(|(u,v)| {
+        if !(graph.has_edge(u.index(), v.index())) && !(graph.has_edge(v.index(), u.index())){
```
IvanIsCoding, line 322:
```suggestion
        if !graph.has_edge(u.index(), v.index()){
```

tests/digraph/test_is_semi_connected.py
```diff
@@ -0,0 +1,67 @@
...
+
+        self.assertTrue(rustworkx.is_semi_connected(graph))
+
+    def test_is_semi_connected_reverse(self):
+        graph = rustworkx.PyDiGraph()
+        graph.add_nodes_from(list(range(3)))
+        graph.add_edge(0, 1, None)
+        graph.add_edge(2, 1, None)
+
+        self.assertTrue(rustworkx.is_semi_connected(graph))
```
IvanIsCoding, line 34:
```suggestion
        self.assertFalse(rustworkx.is_semi_connected(graph))
```

IvanIsCoding (commented):
I left a comment on a test case that is wrong, the condition is simpler, we check only one direction not both

src/connectivity/mod.rs
```diff
@@ -266,6 +270,65 @@ pub fn is_weakly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
+    let topo_sort = match toposort(&condensed_digraph, None){
+        Ok(topo_sort) => topo_sort,
+        Err(_) => {
+            let mut dfs = DfsPostOrder::new(&condensed_digraph, NodeIndex::new(0));
+            let mut nodes = Vec::new();
+            while let Some(node) = dfs.next(&condensed_digraph){
+                nodes.push(node);
+            }
+            nodes.reverse();
+            nodes
```
IvanIsCoding, line 314: In theory this should never happen if we pass `true` to condensation. So we could just return an error

src/connectivity/mod.rs
```diff
@@ -266,6 +270,65 @@ pub fn is_weakly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
+    for _node in graph.graph.node_indices() {
+        node_map.push(temp_graph.add_node(()));
+    }
+
+    for edge in graph.graph.edge_indices() {
+        let (source, target) = graph.graph.edge_endpoints(edge).unwrap();
+        temp_graph.add_edge(node_map[source.index()], node_map[target.index()], ());
+    }
+
+    let condensed = condensation(temp_graph, false);
```
IvanIsCoding, line 300:
```suggestion
    let condensed = condensation(temp_graph, true);
```
I think we shouldn't let self-loops impact and multi-edges impact the topological sort

IvanIsCoding:
I fixed clippy due to the Rust update to 1.79, sorry for the inconvenience. Let's hope CI passes now.

IvanIsCoding:
So some test cases were not correct, I pushed the test cases I believe are correct (you can double check with NetworkX, every path graph should be semi connected). So let's focus on finding the bug and making the test cases pass.

Procatv:
> So some test cases were not correct, I pushed the test cases I believe are correct (you can double check with NetworkX, every path graph should be semi connected). So let's focus on finding the bug and making the test cases pass.

- I believe `test_is_semi_connected_disconnected_graph` should stay false given that there are no edges in the graph.
- I made a mistake with `test_is_semi_connected_false` I was copying the test provided in the [is_semiconnected](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.components.is_semiconnected.html#networkx.algorithms.components.is_semiconnected)() page and realized that in `networkx` adding nodes using `path_graph` starts from 0 to n-1, while `add_nodes_from` in `rustworkx` is from 1 to n, so the implementation was working correctly, but the test itself was wrong if I wanted to assert false (3 should be 4)
- Also I believe that we should still check edges in both directions, one direction checking only works for all other graphs except any `directed_path_graph` which makes sense given that we only check one direction with has_edge -> when I implement the above changes all tests pass given that I change `test_is_semi_connected_disconnected_graph` to false

IvanIsCoding:
> > So some test cases were not correct, I pushed the test cases I believe are correct (you can double check with NetworkX, every path graph should be semi connected). So let's focus on finding the bug and making the test cases pass.
> 
> * I believe `test_is_semi_connected_disconnected_graph` should stay false given that there are no edges in the graph.
> * I made a mistake with `test_is_semi_connected_false` I was copying the test provided in the [is_semiconnected](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.components.is_semiconnected.html#networkx.algorithms.components.is_semiconnected)() page and realized that in `networkx` adding nodes using `path_graph` starts from 0 to n-1, while `add_nodes_from` in `rustworkx` is from 1 to n, so the implementation was working correctly, but the test itself was wrong if I wanted to assert false (3 should be 4)
> * Also I believe that we should still check edges in both directions, one direction checking only works for all other graphs except any `directed_path_graph` which makes sense given that we only check one direction with has_edge -> when I implement the above (bidirectional edge check and `test_is_semi_connected_disconnected_graph` to false) changes all tests pass

The disconnected graph was my mistake, but graphs like `A -> B <- C`  are not semi-connected but with the wrong check the code mistakenly reported them as so. 

We can use the `floyd_warshall` functions with the definition in the tests to estabilish a "ground truth" and work from there. For `A -> B <- C` there is no path from `A` to `C` so the algorithm should report it as not connected.

IvanIsCoding (approved):
After thinking about the problem more thoroughly, I am jumping in to implement `is_semi_connected` in a way that reuses almost 100% of our existing code.

I apologize for not thinking about it earlier, but the condition of `u->v`  for each vertex of the NetworkX implementation is simply checking that there is a path covering all vertices in the DAG. It turns out that `longest_path` that we recently moved to be pure Rust code in #1192 does exactly that! So the NetworkX check is equivalent to seeing if there is a longest path with `N-1` nodes in a DAG of `N` nodes.

Thanks a lot of for writing the original code! I know it ended up not being the original idea but this process led us to implement the least amount of code possible, which will be simpler to maintain in the future.

PR #1216, opened 2024-06-14
Update core's `lexicographical_topological_sort` to work with non-string keys.

kevinhartman:
Also improves error handling.

(We need this in Qiskit, for example, to support use from the new DAGCircuit Rust implementation, which does not use string keys.)


mtreinish (commented):
This lgtm I like making the return from the callback generic and not tied to a string. I think CI is catching a bug in the error handling via python here though. Once that's fixed I think this is ready to merge.

mtreinish (approved):
LGTM thanks for the quick update.

PR #1217, opened 2024-06-18
Fix Clippy warnings due to 1.79

IvanIsCoding:
Currently CI is broken because Clippy introduced new lints in the 1.79 update. This PR fixes that

### DIFF
--- a/src/iterators.rs
+++ b/src/iterators.rs
@@ -577,7 +577,7 @@ macro_rules! custom_vec_iter_impl {
                             Err(PyIndexError::new_err(format!("Invalid index, {}", idx)))
                         } else if idx < 0 {
                             let len = self.$data.len();
-                            Ok(self.$data[len - idx.abs() as usize].clone().into_py(py))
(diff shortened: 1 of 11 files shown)


src/iterators.rs
```diff
@@ -577,7 +577,7 @@ macro_rules! custom_vec_iter_impl {
                             Err(PyIndexError::new_err(format!("Invalid index, {}", idx)))
                         } else if idx < 0 {
                             let len = self.$data.len();
-                            Ok(self.$data[len - idx.abs() as usize].clone().into_py(py))
+                            Ok(self.$data[len - idx.unsigned_abs()].clone().into_py(py))
```
mtreinish, line 580: I seem to remember some odd behavior around this specifically when we encountered this clippy check in the past. It might just have been an msrv thing though. It's passing all the tests so it's probably be fine.

mtreinish (approved):

src/iterators.rs
IvanIsCoding: I think it was MSRV https://github.com/Qiskit/rustworkx/pull/609. We should have it by now I think

PR #1221, opened 2024-06-20
Remove SciPy as a development dependency

IvanIsCoding:
This PR removes SciPy as a development dependency. Instead of relying on NetworkX's new implementation based on SciPy, this introduced forked code from NetworkX's old pure Python implementation.

I added NetworkX's BSD-3 license at the top of the files. We have already done similar for the visualization module in the past.

Partially motivated by #1220. But it is nice to test the `hits` and `pagerank` methods in architectures that previously skipped the test before


tests/digraph/test_hits.py
```diff
@@ -19,16 +55,56 @@
...
+        # normalize vector
+        s = 1.0 / max(a.values())
+        for n in a:
+            a[n] *= s
+        # check convergence, l1 norm
+        err = sum(abs(h[n] - hlast[n]) for n in h)
+        if err < tol:
+            break
+    else:
+        raise nx.PowerIterationFailedConvergence(max_iter)
```
mtreinish, line 96: Do we need to raise a networkx error here? Since this is our fork of the function it feels like we should be able to just raise `ValueError` or `RuntimeError` as it's only for testing and not something we're expecting to be user facing at all.

tests/digraph/test_pagerank.py
```diff
@@ -19,16 +55,67 @@
...
+            # this matrix multiply looks odd because it is
+            # doing a left multiply x^T=xlast^T*W
+            for _, nbr, wt in W.edges(n, data=weight):
+                x[nbr] += alpha * xlast[n] * wt
+            x[n] += danglesum * dangling_weights.get(n, 0) + (1.0 - alpha) * p.get(n, 0)
+        # check convergence, l1 norm
+        err = sum(abs(x[n] - xlast[n]) for n in x)
+        if err < N * tol:
+            return x
+    raise nx.PowerIterationFailedConvergence(max_iter)
```
mtreinish, line 115: Same comment as on hits about removing the the networkx exception and usage from the tests.

mtreinish (commented):
This looks fine to me, I like removing scipy as a test dependency, just a comment inline about the exceptions.

tests/digraph/test_hits.py
IvanIsCoding: Done

tests/digraph/test_pagerank.py
IvanIsCoding: Done

mtreinish (approved):
Thanks for the quick update

PR #1223, opened 2024-06-22
Subtle name change for `with_capacity` hints

IvanIsCoding:
This is an opinionated follow up for #975. It is mostly about renaming the arguments and making it accept `None` as argument.  The reasoning:
* Despite the docstring of `initial_node_count` making it explictly that it does not initialize with nodes, I have a bad feeling eventually we will get an issue complaining about it. I think the name with `hint` is clearer that you are giving rustworkx a heads-up to pre-allocate memory
* By using the prefix `_hint` I feel that we can have much more freedom to make optimizations later. If it's a hint we can pre-allocate another value e.g. `2^ceil(log2(hint))`
* `None` seems like a more sensible default value for the argument because it means we didn't specify anything. I think if in Rust we had to do `with_capacity(0)` we'd find it weird, but doing `with_capacity(capacity.unwrap_or_default())` seems like "ok we want to initialize with capacity but it could be unspecified`

Lastly, we completely forgot about the `__init__` method for `PyGraph`! So I added that as well.


src/digraph.rs
```diff
@@ -317,10 +320,10 @@ impl PyDiGraph {
...
         (
             (self.check_cycle, self.multigraph, self.attrs.clone_ref(py)).into_py(py),
-            [
-                ("initial_node_count", self.graph.node_bound()),
-                ("initial_edge_count", self.graph.edge_bound()),
-            ]
+            indexmap::indexmap! {
+                "node_count_hint" => self.graph.node_bound(),
+                "edge_count_hint" => self.graph.edge_bound(),
+            }
```
mtreinish, line 326: Do we need to make the intermediate `IndexMap` here? This is basically doing two allocations now, one for the indexmap and one for the pydict while before it was just a single one for the pydict.

src/digraph.rs
```diff
@@ -295,17 +295,20 @@ impl PyDiGraph {
...
+    #[pyo3(signature=(/, check_cycle=false, multigraph=true, attrs=None, *, node_count_hint=None, edge_count_hint=None))]
     fn new(
         py: Python,
         check_cycle: bool,
         multigraph: bool,
         attrs: Option<PyObject>,
-        initial_node_count: usize,
-        initial_edge_count: usize,
+        node_count_hint: Option<usize>,
+        edge_count_hint: Option<usize>,
```
mtreinish, line 305: TBH, I'm not the biggest fan of this change. There doesn't seem to be a difference between `Some(0)` and `None`  and I feel it makes it simpler both for python and rust to just default to `0`. I'd personally prefer if we just left the type as a `usize` instead of an `Option<usize>`. That being said it's super minor and won't effect the runtime performance or anything so I don't feel strongly enough to block on this.

mtreinish (commented):
Overall I'm fine with making this change it's as you said mostly just opinion and I'm fine with the suggestions. I just left an inline comment with my thoughts on the new default, but I'm also fine with your proposal here. The only thing I really wonder about is the indexmap creation for pickle serialization, I don't think it's necessary in this case.

src/digraph.rs
IvanIsCoding: I reverted the change, I just really wish there was a macro to make Python dictionaries. Creating a list of tuples doesn't tell the reader that it's a dictionary in my opinion.

mtreinish (approved):

src/digraph.rs
IvanIsCoding: I agree but my main argument for `None` is that maybe one day we could want to change. Something more along the lines of small vector optimizations. With `None` I know we can override it to whatever default it sets. With `0`, I think it would be poor taste to not allocate a vector with zero capacity.

PR #1203, opened 2024-05-30
Potential fix to the tooltip with special characters  bug

anushkrishnav:
Fixes #750 

Updated the fn attr_map_to_string function 
I can add a separate function to handle the escape values if needed 
```
fn escape_graphviz_value(value: &str) -> String {
    value
        .replace("\\", "\\\\")  // Escape backslashes first
        .replace("\"", "\\\"")  // Escape double quotes
        .replace("\n", "\\n")   // Escape newlines
        .replace(":", "\\:")    // Escape colons
}
```

Will finish up docs and tests once the solution is satisfactory 
- [ ] I ran rustfmt locally
- [ ] I have added the tests to cover my changes.
- [ ] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.

PR #1229, opened 2024-06-28
Fix dot file generation

mtreinish:
This commit fixes a regression introduced in #1203. That PR was attempting to fix missing character escaping in some string fields but it was doing so too eagerly. This would result in invalid dot files being produced for users upgrading from rustworkx < 0.15.0 that were wrapping strings in quotes as needed previously. For example if you were setting `'color="#aaaaaa"'` previously before this would become `color="\"#aaaaaa\""` after #1203. In order to quickly release a 0.15.1 this commit reverts the dot generation component of #1203 but updates the code to wrap tooltip in addition to label which was what the original bug reported. In 0.16.0 we should investigate adding a flag to control the escaping behavior of the function to either decide to wrap values in quotes or not.

### DIFF
--- a/releasenotes/notes/fix-graphviz-escaping-39a2d1cebb586eca.yaml
+++ b/releasenotes/notes/fix-graphviz-escaping-39a2d1cebb586eca.yaml
@@ -13,6 +13,7 @@ fixes:
             return {
                 "color": '"#422952"'
             }
+
     to set the color attribute in the output dot file with the string
     `"#422952"` (with the quotes) this was incorrectly being converted to
     `\"#422952\"`. This no longer occurs, in rustworkx 0.16.0 there will likely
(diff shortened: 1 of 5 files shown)


IvanIsCoding (approved):
We should have added a test case for that in hindsight

releasenotes/notes/fix-graphviz-escaping-39a2d1cebb586eca.yaml
```diff
@@ -0,0 +1,21 @@
...
+    placed quotes in a string for an attr callback get that to pass through to 
+    the output dot file this was incorrectly being converted in rustworkx 0.15.0
+    to duplicate the quotes and escape them. For example, if you defined a
+    callback like::
+
+        def color_node(_node):
+            return {
+                "color": '"#422952"'
+            }
+    to set the color attribute in the output dot file with the string
```
mtreinish, line 16:
```suggestion

    to set the color attribute in the output dot file with the string
```

PR #1232, opened 2024-06-28
Hexagonal lattice followup

jpacold:
Follow-up to #1213.

* Addressed comments on the previous PR
    * Split tests with a lot of sub-cases into subtests
    * Removed unnecessary module

* We now pre-compute the number of edges in each graph, so that we can call `G::with_capacity(self.num_nodes, self.num_edges)` to create the graph where before we had `G::with_capacity(self.num_nodes, self.num_nodes)`
 
* Fixed two bugs in the position calculation
   * In my initial implementation I missed a factor of two, so the hexagons were not actually regular
   * For lattices with an odd number of columns, the node index arithmetic had an error that led to configurations like this:
hexlattice_bug
Probably the reason I missed this is that I mostly wrote tests for cases with an even number of columns (required when `periodic = True`). We now get the expected result:
hexlattice_fixed

### DIFF
--- a/releasenotes/notes/hexagonal-lattice-graph-fixes-1cd4aa88f1591701.yaml
+++ b/releasenotes/notes/hexagonal-lattice-graph-fixes-1cd4aa88f1591701.yaml
@@ -1,0 +1,10 @@
+---
+fixes:
+  - |
+    Fixed two bugs in the node position calculation done by the generator functions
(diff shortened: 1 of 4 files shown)


mtreinish:
Could you add a release note to document the bugfixes? I'm planning to push a 0.15.1 release shortly and we can include this in it if you think it's warranted.

mtreinish (approved):
This LGTM, thanks for catching and fixing this.

PR #1235, opened 2024-07-01
Port VF2 to `rustworkx-core`

kevinhartman:
Ports Rustworkx's modified VF2 algorithm to `rustworkx-core`, supporting generic `petgraph` types like `StableGraph`.

### To-do
- [ ] Fix edge matcher regression (I'm aware of the issue, working on fix).
- [ ] Release note.
- [ ] Add more detailed PR description 🙂.


kevinhartman:
It looks like I may need to put the release note I've just added into a more-specific (version?) folder.

What's the right place for it?

src/isomorphism/vf2_mapping.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    fn eq(
+        &mut self,
+        g0: &StablePyGraph<Ty>,
+        g1: &StablePyGraph<Ty>,
+        n0: NodeIndex,
+        n1: NodeIndex,
+    ) -> Result<bool, Self::Error> {
+        if let (Some(a), Some(b)) = (g0.node_weight(n0), g1.node_weight(n1)) {
+            unsafe {
```
IvanIsCoding, line 51: Just because PyO3 lets you do this does mean we should take it. I think it is a big amount of risk for a refactor

IvanIsCoding (changes requested):
Georgios would be the best person to review this but I will have to step in. Overall the `rustworkx-core` looks like a clean port! 10/10, I think we might get more Rust users for `rustworkx-core` as Isomorphism is one of the methods that differentiate our library.

However, I do not want to use `Python::assume_gil_acquired()`. So far we have always let `PyO3` handle all the uncertainities in the Python FFI, we have zero unsafe code in  Rustworkx directly. Can't we use ` Python::with_gil` or some type of closure? There has to be a better way

kevinhartman:
Thanks for the review, @IvanIsCoding!

Regarding the Python GIL acquisition in the `PyMatcher` implementations of `NodeMatcher` and `EdgeMatcher`, I think assuming the GIL is acquired is the best thing we can do.

The main issue with threading the actual GIL into those callbacks is that the callbacks get moved into the `(Di|)GraphVF2Mapping` pyclass structs, which means they can't have any lifetime parameters (pyclass structs cannot have lifetime parameters in PyO3) and thus can't hold a `Python<'py>` token. In the original Python-only code, this was worked around by having the `next` method on `VF2Mapping` take a `Python<'py>` token, which was then threaded down to the callbacks when pumping the iterator. But this isn't appropriate in `rustworkx-core`, of course.

We certainly _could_ use `Python::with_gil` instead of `Python::assume_gil_acquired`, but my expectation is that this would introduce a performance regression, since PyO3 would be checking for the GIL for every single comparison, be it a node or an edge.

Given that the matchers are only invoked while we're pumping the iterator from Python, we should be safe to assume the GIL is acquired, as long as we aren't using Rayon (or other multithreading) to perform matching on other threads.

IvanIsCoding:
I think `unsafe` as a last-resource is OK and with Matthew's +1 we can merge it as is.

However, before we do that: can you try using a PyFunction similar to how we use it in here: https://github.com/Qiskit/rustworkx/blob/2734eee56f246da0d1618f0302268d5f66d38bbd/src/lib.rs#L312?

We pass PyFunction's to Dijkstra. Couldn't we do the same here? Then the calls would be `fn.call1((a, b))`  instead of requiring the `py` argument

kevinhartman:
Thanks for the additional review @IvanIsCoding!

> We pass PyFunction's to Dijkstra. Couldn't we do the same here? Then the calls would be fn.call1((a, b)) instead of requiring the py argument

I'm not sure I'm finding / understanding exactly what you're describing here. It looks like `CostFn` is an enum that can be either a default `f64` value or a `PyObject` callable from the user. In the [`impl` for `CostFn`](https://github.com/Qiskit/rustworkx/blob/2734eee56f246da0d1618f0302268d5f66d38bbd/src/lib.rs#L345), the `call` method appears to require a `py` token.

I think the only way to really get a Python token into the matchers would be to remove them from the `VF2Algorithm` struct and pass them in to the call to `next` every time, i.e. inject the Python token into `node_matcher` and `edge_matcher` closures and use a custom `next` method that takes them as parameters. But, that feels like a bad API design to me, since it allows the user to change the definition of what constitutes a match with every pump of the "iterator" (in quotes because we'd also no longer be able to implement `Iterator` since we'd need a custom signature for `next`).

Let me know if I'm misunderstanding the example you mentioned. It'd be neat to get things working without `unsafe`, but I'm not optimistic it is possible without making serious concessions to the integrity of the Rust API for this. And, I'm not worried about what we're doing here with `unsafe` because it is tightly controlled and entirely encapsulated from our users.

PR #1241, opened 2024-07-02
Cleaned up some control flow and optimized some parts.

Larmbs:
This is my second time doing this, forgot to format final result.

I just made some changes to control flow to optimize some parts. The code does the exact same thing but is just shorter and more readable.


IvanIsCoding (commented):
Thanks for doing this! I am surprised Clippy didn’t find some of these. Some of them are more trivial than others so I still need to re-review some of the if -> match conditions. Test cases should catch regressions, but just to be safe

Larmbs:
Yeah, I believe the control flow should be identical but it's best not to risk it.

IvanIsCoding (approved):

IvanIsCoding:
Code coverage is 100% and I checked the matches, I think it is good to go

Issue #832, opened 2023-03-14
PyGraph not typed correctly in VSCode, disallowing autocomplete


thomasaarholt:
### Information

- **rustworkx version**: 0.12.1
- **Python version**: 3.10.9
- **Rust version**: Not installed, installed via [wheel](https://pypi.org/project/rustworkx/#files)

- **Operating system**:
Debian Buster

### What is the current behavior?
VSCode is unable to parse the type of PyGraph, see below.

<img width="404" alt="Screenshot 2023-03-14 at 07 49 48" src="https://user-images.githubusercontent.com/2721423/224918597-ebc8eb66-aa3c-4c32-a13f-aa1c7bdfe81c.png">

<img width="298" alt="image" src="https://user-images.githubusercontent.com/2721423/224918117-28268df4-cba4-4e5e-b0ef-34268f85d33e.png">

This makes it very difficult to write code, especially while getting used to a new API.

Autocompletion works in iPython, and assumedly, the Jupyter Notebook (both use Jedi).

### What is the expected behavior?
VSCode should identify the type of the graph

### Steps to reproduce the problem
In VSCode: 
```python
import rustworkx

graph = rustworkx.PyGraph()

a = graph.add_node("A")
```

IvanIsCoding:
We just merged #401 which should fix that. It has not been released yet.

Would you mind installing from the main branch and sharing what you see?

thomasaarholt:
Wahey! It works great!

The typing here is very cool. Took me a moment to figure out, but its quite useful.
<img width="928" alt="image" src="https://user-images.githubusercontent.com/2721423/225579878-9a5b40f9-41cb-4141-ae4c-9667fcc265b9.png">

Two suggestions:
- Rename the TypeVars `S` and `T` to `NodeType` and `PayloadType`
- Rename the `edge` argument to `payload`.

thomasaarholt:
Hilariously, the above code doesn't actually run 😅 
```python
g: rx.PyGraph[tuple[int, int], float] = rx.PyGraph() # this typing is very cool

n1 = g.add_node((1,2))
n2 = g.add_node((3,4))

g.add_edge(n1, n2, 10)

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[52], line 1
----> 1 g: rx.PyGraph[tuple[int, int], float] = rx.PyGraph() # this typing is very cool
      3 n1 = g.add_node((1,2))
[...]

PR #1246, opened 2024-07-11
Specify `Any` for `PyGraph` and `PyDiGraph` type annotations for pyright compatibility

### LINKED ISSUE #1242: Problem type checking `ancestors()` with pyright
### Information

- **Python version**: 3.12.4
- **Rust version**: 1.79.0
- **Operating system**: Void Linux x86_64

Pyright is unable to fully type check the `rustworkx.ancestors()` function in strict mode.

Pyright should be able to determine the type of `ancestors()`, including the parameterized type of the `graph` parameter.

```python
# pyright: strict

import rustworkx as rx

graph: rx.PyDiGraph[int, float] = rx.PyDiGraph()
node = graph.add_node(7)
rx.ancestors(graph, node)
```
```bash
$ pyright .
./main.py
  ./main.py:7:1 - error: Type of "ancestors" is partially unknown
    Type of "ancestors" is "(graph: PyDiGraph[Unknown, Unknown], node: int, /) -> set[int]" (reportUnknownMemberType)
1 error, 0 warnings, 0 informations
```
I looked at the stub file
https://github.com/Qiskit/rustworkx/blob/553bff1823a30293c82fb811f4457b094700a728/rustworkx/rustworkx.pyi#L985-L990
and `graph` doesn't have `PyDiGraph[_S, _T]` like in some of the other functions. I also ran into this issue with the `descendants()` function.

IvanIsCoding:
Closes #1242 

This is a pretty annoying PR but it avoids `pyright` from infering the types as `Unknown`. Although we will only keep `mypy` in the CI, `pyright` is the base for `pylance` which ships with VSCode. And we got VSCode issues in the past (#832), so I think it's worth supporting.


rustworkx/__init__.pyi
```diff
@@ -272,13 +272,13 @@ _DijkstraVisitor = TypeVar("_DijkstraVisitor", bound=visit.DijkstraVisitor)
 class PyDAG(Generic[_S, _T], PyDiGraph[_S, _T]): ...
 
 def distance_matrix(
-    graph: PyGraph | PyDiGraph,
+    graph: PyGraph[Any, Any] | PyDiGraph[Any, Any],
```
mtreinish, line 275: Instead of `Any` should we be using `_T` and `_S` here? I'm not familiar with pyright but I just always figured we'd use the type var for this.

mtreinish (commented):
Just one question inline about using the generic type var instead `Any`. But if that doesn't work I'm fine with this.

rustworkx/__init__.pyi
IvanIsCoding: Distance matrix calculates the unweighted distance matrix so the node and edge types don’t matter. We can use _S and _T to narrow it down but that would be more restrictive than we are in practice

rustworkx/rustworkx.pyi
```diff
@@ -552,39 +554,39 @@ def directed_gnm_random_graph(
     num_edges: int,
     /,
     seed: int | None = ...,
-) -> PyDiGraph: ...
```
mtreinish, line 555: So based on the earlier comment is this actually `[Any, Any]`. The weights from the generator functions are a fixed type depending on the arguments. Like for this gnm random graph function the nodes and edges are always `None`.

IvanIsCoding:
Now that we are doing a major release, I might use the suggestion from https://github.com/Qiskit/rustworkx/issues/1242#issuecomment-2231841772 to fix it in a more proper way using `TypeVar`

rustworkx/rustworkx.pyi
IvanIsCoding: I think now if someone specifies `graph: PyDiGraph[None, None] = directed_gnm_random_graph(10, 10)`, MyPy will not complain which should

IvanIsCoding:
I tested with `pyright`, I don't want to add it to the Nox file as of now because of the Node.js aspect but it worked. Now we get:
```
```

mtreinish (approved):
LGTM now, thanks for sticking with this.

PR #1247, opened 2024-07-11
Fix `node_link_json` type annotation

### LINKED ISSUE #1243: Incorrect Python stubs typing for node/edge attrs parameters in json serializers
### Information

- **Python version**: 3.11.5
- **Rust version**: Rustworkx installed via pip
- **Operating system**: macOS Sonoma 14.5

Both `graph` and `digraph` `_node_link_json` serializers show expected `callable` output as `str` type.
https://github.com/Qiskit/rustworkx/blob/553bff1823a30293c82fb811f4457b094700a728/rustworkx/rustworkx.pyi#L632-L647

Meanwhile, the docs state _'is expected to return a dictionary of string keys to string values representing the data payload'_ (same as for the `graph_attrs` parameter, for which the typing corresponds with the documentation).

From the rust function, the `dictionary` is also expected
https://github.com/Qiskit/rustworkx/blob/553bff1823a30293c82fb811f4457b094700a728/src/json/node_link_data.rs#L125-L128

On inserting a `callable` return a `str` datatype, this throws an error:
`E       TypeError: 'str' object cannot be converted to 'PyDict'`
No error is thrown on return a `dict`

```
def digraph_node_link_json(
    graph: PyDiGraph[_S, _T],
    /,
    path: str | None = ...,
    graph_attrs: Callable[[Any], dict[str, str]] | None = ...,
    node_attrs: Callable[[_S], dict[str, str]] | None = ...,
    edge_attrs: Callable[[_T], dict[str, str]] | None = ...,
) -> str | None: ...

def graph_node_link_json(
    graph: PyGraph[_S, _T],
    /,
    path: str | None = ...,
    graph_attrs: Callable[[Any], dict[str, str]] | None = ...,
    node_attrs: Callable[[_S], dict[str, str]] | None = ...,
    edge_attrs: Callable[[_T], dict[str, str]] | None = ...,
) -> str | None: ...
```

Inserting `node_attrs=lambda x: 'a'`, which corresponds to the `Callable[[_S], str] | None` type

IvanIsCoding:
Closes #1243 

We let this one slip through by accident.


mtreinish (approved):
LGTM, I'm a bit surprised this got through the tests but this is clearly correct.

IvanIsCoding:
I think our merge queue is stuck, do we need to tweak the CI configs? https://github.com/Qiskit/rustworkx/queue/main

PR #1249, opened 2024-07-12
add -x to graphviz to prune isolated nodes and peninsulas

1ucian0:
On many nodes, `neato` seems to fold on its own.

image

With the parameter `-x`, the problem seems fixed:

image

### DIFF
--- a/rustworkx/visualization/graphviz.py
+++ b/rustworkx/visualization/graphviz.py
@@ -196,9 +196,13 @@ def graphviz_draw(
             )
         prog = method
 
+    extra_args = []
+    if prog == "neato":
+        extra_args.append("-x")
+
     if not filename:
         dot_result = subprocess.run(
-            [prog, "-x", "-T", output_format],
+            [prog, "-T", output_format] + extra_args,
             input=dot_str.encode("utf-8"),
             capture_output=True,
             encoding=None,
@@ -210,7 +214,7 @@ def graphviz_draw(
         return image
     else:
         subprocess.run(
-            [prog, "-x", "-T", output_format, "-o", filename],
+            [prog, "-x", "-T", output_format, "-o", filename] + extra_args,
             input=dot_str,
             check=True,
             encoding="utf8",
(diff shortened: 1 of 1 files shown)


rustworkx/visualization/graphviz.py
```diff
@@ -210,7 +214,7 @@ def node_attr(node):
         return image
     else:
         subprocess.run(
-            [prog, "-T", output_format, "-o", filename],
+            [prog, "-x", "-T", output_format, "-o", filename] + extra_args,
```
mtreinish, line 217:
```suggestion
            [prog, "-T", output_format, "-o", filename] + extra_args,
```

mtreinish (changes requested):
We should really add an extra kwarg option for this, or a better choice would be generic way to set extra flags when calling graphviz. We definitely don't want to always do this with neato because it removes nodes from the graph. If you look at the examples from the pr summary the after example is missing a fair number of nodes.

That all being said does neato not expose this option via graph attributes or something a user can already set today?

mtreinish:
As an alternative approach, from qiskit's usage of `graphviz_draw` may give setting `overlap` in the graph attributes a try: https://graphviz.org/docs/attrs/overlap/ I'm not sure how well that will work, but it sounds like what you're looking for instead of pruning nodes.

IvanIsCoding:
> We should really add an extra kwarg option for this, or a better choice would be generic way to set extra flags when calling graphviz. We definitely don't want to always do this with neato because it removes nodes from the graph. If you look at the examples from the pr summary the after example is missing a fair number of nodes.
> 
> That all being said does neato not expose this option via graph attributes or something a user can already set today?

I agree with Matthew that we should have `kwargs`. We can have default `kwargs` that are sane (like what you provided), but graphviz has a big surface of settings so forwarding `kwargs` to the CLI seems like the best path forward

1ucian0:
Closing because https://github.com/Qiskit/qiskit/pull/12770 is the right way to do it.

PR #1254, opened 2024-07-23
Add `PyDiGraph.neighbors_undirected`

Cryoris:
### Summary

This PR exposes `petgraph`s existing `neighbors_undirected` method on the directed graph efficiently get the undirected neighbors.

### Details

Getting the undirected neighbors in a directed graph is very useful e.g. when investigating coupling maps of quantum devices (for example: physical neighboring qubits affect on another, independent of the gate directionality). Currently this is already possible, as 
```python
digraph.to_undirected().neighbors(node)
```
but this involves a copy.


src/digraph.rs
```diff
@@ -1720,6 +1720,25 @@ impl PyDiGraph {
         }
     }
 
+    /// Get the direction-agnostic neighbors (i.e. successors and predecessors) of a node.
+    ///
```
IvanIsCoding, line 1724: Can you add an example use on the docstring? I was thinking something along the lines of calling this method and comparing to calling `to_undirected` in the whole graph and then calling `neighbors`

tests/digraph/test_neighbors.py
```diff
@@ -57,3 +57,14 @@ def test_no_neighbor(self):
         dag = rustworkx.PyDAG()
         node_a = dag.add_node("a")
         self.assertEqual([], dag.neighbors(node_a))
+
+    def test_undirected_neighbors(self):
```
IvanIsCoding, line 61: Can you add a test case using a digraph generator, calling `to_unditected` in the graph and then comparing this function call to the `neighbors` call for all nodes? I was thinking a directed cycle graph with 10 nodes

IvanIsCoding (approved):
LGTM, just asking for one more example in the doc string & a test to cover this behaviour compared to `to_undirected`

src/digraph.rs
Cryoris: Sounds good! Added in 2824fe4 🙂

tests/digraph/test_neighbors.py
Cryoris: Added in 2824fe4 👍🏻

src/digraph.rs
```diff
@@ -1720,6 +1720,38 @@ impl PyDiGraph {
...
+    ///     import rustworkx
+    ///
+    ///     dag = rustworkx.generators.directed_cycle_graph(num_nodes=10, bidirectional=False)
+    ///
+    ///     node = 3
+    ///     neighbors = dag.neighbors_undirected(node)
+    ///     same_neighbors = dag.to_undirected().neighbors(node)
+    ///
+    ///     assert neighbors == same_neighbors
```
IvanIsCoding, line 1736:
```suggestion
    ///     assert sorted(neighbors) == sorted(same_neighbors)
```

tests/digraph/test_neighbors.py
```diff
@@ -57,3 +58,24 @@ def test_no_neighbor(self):
...
+
+    def test_undirected_neighbors_cycle(self):
+        num_nodes = 10
+        dag = rustworkx.generators.directed_cycle_graph(num_nodes, bidirectional=False)
+        undirected_dag = dag.to_undirected()
+
+        for node in dag.node_indices():
+            undirected_neighbors = dag.neighbors_undirected(node)
+            expected_neighbors = undirected_dag.neighbors(node)
+            self.assertEqual(undirected_neighbors, expected_neighbors)
```
IvanIsCoding, line 81:
```suggestion
            self.assertEqual(sorted(undirected_neighbors), sorted(expected_neighbors))
```

IvanIsCoding (approved):
Just some minor stuff because we don't guarantee ordering when returning nodes (it is a big coincidence)

Cryoris:
Great, thanks for the fast review! 🙂

PR #1256, opened 2024-07-28
Fix Clippy warnings on stable/0.15 branch

IvanIsCoding:
This is to unblock CI for the future backports of #1246 and #1247. The lints were super simple this time, we only changed docstrings not code.


mtreinish (approved):

IvanIsCoding:
@mtreinish I think this PR got stuck waiting for steps that don’t exist. Is there a way to disable those?

mtreinish:
Ah yeah, I see that now. I'll update the branch protection rules

PR #1257, opened 2024-07-29
Added general form of transitive closure

Procatv:
Bulids on PR #707 to implement the general form of the transitive closure of a graph. Implemented function by following structure from @IvanIsCoding.

- Condense the graph into strongly connect components (which we already implement)
- Apply the DAG algorithm and get the transitive closure for the intermediate graph
- Map the answer from the intermediate graph to the original graph


IvanIsCoding:
CI should be unblocked now that #1258 is merged

Issue #1141, opened 2024-03-14
Implementation of Louvain Algorithm


jamesdbaker:
The [Louvain algorithm](https://en.wikipedia.org/wiki/Louvain_method) is commonly used to identify communities in graphs, and is present in other libraries/tools such as [NetworkX](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.community.louvain.louvain_communities.html) and [Neo4J](https://neo4j.com/docs/graph-data-science/current/algorithms/louvain/).

It would be great if an implementation was available in RustworkX too.

IvanIsCoding:
Indeed we will probably map them to isize https://github.com/Qiskit/rustworkx/blob/af0d36caa3d3b3488bdb0297b2014e5f3154f994/src/graphml.rs#L129 likes we already do with `int`

jamestwebber:
A general question: is the goal of `rustworkx` include algorithms like this out-of-the-box, or would it be better to support third-party packages for things like this? I see that networkx includes Louvain now, but originally it was a separate package. Similarly `leidenalg` implements community-detection on `igraph` graphs.

One thing I don't know: how complicated is it to write a package that works with `rustworkx` graphs in a performant way (_i.e._ not going through python when possible)? One item on my ever-growing list of projects is to try to implement Leiden (or similar) in Rust in a way that's compatible with `rustworkx`, potentially using the underlying graph structures directly.

IvanIsCoding:
One last thought: for `leiden` specifically, I think if you build upon `petgraph` and `rustworkx-core` it would make life simpler to port your own version to be included in rustworkx. And we would gladly merge & distribute it if you made a PR submitting it!

jpacold:
I'm interested in taking this on if nobody has started on it.

traviscross:
@jpacold: It seems the right place to start is to first make a PR upstream to [petgraph](https://github.com/petgraph/petgraph) with the new algorithm.

jpacold:
So far all I have done is implement the [modularity](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.community.quality.modularity.html#networkx.algorithms.community.quality.modularity) function (which the Louvain algorithm tries to maximize) in [this branch](https://github.com/jpacold/rustworkx/tree/louvain).

Before going any further I wanted to check whether the architecture makes sense. The idea is to eventually put the full algorithm in, e.g., `rustworkx-core/src/community/louvain.rs`.

[...]

PR #1277, opened 2024-09-05
[WIP] Louvain algorithm for undirected graphs

jpacold:
I finally had some time to work on #1141, so far only on the Rust side and only for undirected graphs. Opening this for feedback on how the Rust code is organized, and maybe to restart the discussion on whether this should go here or in petgraph.

The big additions here are: 

* `metrics.rs`
    * `Modularity`: a trait for graphs for which we can compute [modularity](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.community.quality.modularity.html#networkx.algorithms.community.quality.modularity) (by extension this means we can apply the Louvain method)
    * `modularity`: computes the modularity, given a graph and a set of subsets.
    * `Partition`: a struct holding a graph partition. This mainly exists to keep the implementation of `modularity` organized.

* `louvain.rs`
    * `InnerGraph`: at each level of the algorithm we work with an aggregated graph, where each node of this graph corresponds to one of the communities that were identified in the previous level). `InnerGraph` keeps track of this correspondence.
    * `LouvainAlgo`: helper functions for updating the `Partition` at each level of the algorithm.
    * `one_level_undirected`: most of the business logic is in here since it constructs the new communities at each level, roughly equivalent to `_one_level` in the [networkx implementation](https://networkx.org/documentation/stable/_modules/networkx/algorithms/community/louvain.html).
    * `louvain_communities`: repeatedly calls `one_level_undirected`, stopping either after a fixed number of iterations or if the modularity improvement falls below the given threshold.

I realize this is a lot to review at once and can go into more detail about each piece if necessary.


rustworkx-core/src/community/louvain.rs
```diff
@@ -0,0 +1,432 @@
...
+            let two_m_sq = 2.0 * m * m;
+
+            degrees[best_com] -= deg;
+
+            let delta = if let Some(&w) = neighbor_weights.get(&best_com) {
+                w
+            } else {
+                0.0
+            };
+            let remove_cost = -delta / m + resolution * (s_tot[best_com] * deg) / two_m_sq;
```
IvanIsCoding, line 215: You might want to make this expression a closure. Just to keep it consistent inside and outside the loop

rustworkx-core/src/community/louvain.rs
```diff
@@ -0,0 +1,432 @@
...
+            s_tot[best_com] += deg;
+
+            if best_com != init_com {
+                performed_move = true;
+                total_gain += best_gain;
+                node_to_community[node] = best_com;
+            }
+        }
+        if !performed_move {
+            break;
```
IvanIsCoding, line 234: I think we should loop on `while performed_move` or something like that to make the intention clearer.

rustworkx-core/src/community/louvain.rs
```diff
@@ -0,0 +1,432 @@
...
+            // For a barbell graph, we expect the Louvain algorithm to identify
+            // the two complete subgraphs as the final communities
+            assert_eq!(result.len(), 2);
+            assert_eq!(result[0].len(), n);
+            assert_eq!(result[1].len(), n);
+        }
+    }
+
+    #[test]
+    fn test_louvain_karate_club_graph() {
```
IvanIsCoding, line 328: Friendly reminder to myself that we should add a Karate club graph generator. It could be a separate PR

rustworkx-core/src/community/louvain.rs
```diff
@@ -0,0 +1,432 @@
...
+            (32, 33, 5.0),
+        ];
+        let graph: UnGraph<(), f64> = UnGraph::from_edges(edges.iter());
+        let communities = louvain_communities(&graph, 1.0, 0.01, None, Some(7));
+
+        // The result is very sensitive to the random seed. For this seed we
+        // happen to get the same result as:
+        //      import networkx as nx
+        //      g = nx.karate_club_graph()
+        //      communities = nx.community.louvain_communities(g, weight='weight', seed=12)
```
IvanIsCoding, line 416: I think this is a brittle test (https://testing.googleblog.com/2024/04/how-i-learned-to-stop-writing-brittle.html) but leave a TODO there and we can revisit it later. I am glad it passes though

rustworkx-core/src/community/louvain.rs
```diff
@@ -0,0 +1,432 @@
...
+    // Compute the resulting new partition of the input graph
+    let input_graph = &partition.graph;
+    let mut final_index = HashMap::new();
+    let mut next_com = 0;
+    let mut updated_partition: Vec<usize> = vec![0; input_graph.node_count()];
+
+    for n in input_graph.node_identifiers() {
+        let prev_com = partition.subset_idx(n);
+        let inner_com = node_to_community[prev_com];
+        let new_com = if let Some(&c) = final_index.get(&inner_com) {
```
IvanIsCoding, line 251: Nitpicking but a `match` statement would be clearer in this case

rustworkx-core/src/community/metrics.rs
```diff
@@ -0,0 +1,233 @@
...
+                .iter()
+                .zip(outgoing_weights.iter())
+                .map(|(&x, &y)| x * y)
+                .sum()
+        } else {
+            outgoing_weights.iter().map(|&x| x * x).sum::<f64>() / 4.0
+        };
+
+        let m: f64 = total_edge_weight(self.graph);
+        sigma_internal / m - resolution * sigma_total_squared / (m * m)
```
IvanIsCoding, line 137: We might want to reorganize this expression to avoid floating point erros, specially because of the common factor `1/m`

IvanIsCoding (commented):
Overall this is a great start! I would call it more than a draft given it already has good tests.

I left some Rust comments, I need to study the algorithm more myself to give more "algorithmic" advice

rustworkx-core/src/community/louvain.rs
IvanIsCoding: Once #1280 gets merged this test will hopefully be much shorter

rustworkx-core/src/community/louvain.rs
jpacold: Done in 12ffc33.

rustworkx-core/src/community/louvain.rs
jpacold: Done in 234659e.

rustworkx-core/src/community/louvain.rs
jpacold: I'm not sure yet what to do with this. I can adjust the resolution parameter so that we consistently get two output communities. However, there is still a lot of dependence on the seed (in networkx I get at least 5 different partitions with the same parameters)

rustworkx-core/src/community/louvain.rs
jpacold: Done in 11a74fd.

rustworkx-core/src/community/metrics.rs
jpacold: I started doing this in 58857fd. Other than the common factor what are the possible issues? I think `sigma_internal`, `sigma_total_squared / m`, and `m` are all about the same order of magnitude but could be missing something.

jpacold:
* Fixed two mistakes in the modularity gain calculation (I am actually a little surprised that the tests passed earlier and need to come up with some more)
* Addressed some of the comments

It looks as if the CI is failing because I used [associated type bounds](https://blog.rust-lang.org/2024/06/13/Rust-1.79.0.html#bounds-in-associated-type-position).

IvanIsCoding:
> * Fixed two mistakes in the modularity gain calculation (I am actually a little surprised that the tests passed earlier and need to come up with some more)
> * Addressed some of the comments
> 
> It looks as if the CI is failing because I used [associated type bounds](https://blog.rust-lang.org/2024/06/13/Rust-1.79.0.html#bounds-in-associated-type-position).

Indeed our current MSRV is Rust 1.70 from June 2023. Rust 1.79 is from June 2024. In the past we used to follow Debian stable's Rustc which believe it or not is in [1.63 still](https://packages.debian.org/trixie/rustc).

I'd try rewriting it before we discuss bumping MSR, 1.79 might break a couple people that install from source

PR #1280, opened 2024-09-09
Add Zachary's Karate Club

IvanIsCoding:
Adds the famous [Zachary's karate club](https://en.wikipedia.org/wiki/Zachary%27s_karate_club), a network that is used in many examples.

Makes #1277 easier and I assume it will be used in lots of example for community detection algorithms


rustworkx-core/src/generators/karate_club.rs
```diff
@@ -0,0 +1,89 @@
...
+// under the License.
+
+use std::hash::Hash;
+
+use petgraph::data::{Build, Create};
+use petgraph::visit::{Data, NodeIndexable};
+
+use super::InvalidInputError;
+
+// Adapted from NetworkX, licensend under the MIT license
```
alexanderivrii, line 20:
```suggestion
// Adapted from NetworkX, licensed under the MIT license
```

rustworkx-core/src/generators/karate_club.rs
```diff
@@ -0,0 +1,96 @@
...
+    mut default_edge_weight: H,
+) -> Result<G, InvalidInputError>
+where
+    G: Build + Create + Data<NodeWeight = T, EdgeWeight = M> + NodeIndexable,
+    F: FnMut(bool) -> T,
+    H: FnMut(usize) -> M,
+    G::NodeId: Eq + Hash,
+{
+    let mut graph = G::with_capacity(0, 0);
+    let membership: std::collections::HashSet<usize> = MR_HI_MEMBERS.into_iter().collect();
```
mtreinish, line 75: We've typically been using `hashbrown` instead of the stdlib `HashSet`.

rustworkx-core/src/generators/karate_club.rs
```diff
@@ -0,0 +1,96 @@
...
+use std::hash::Hash;
+
+use petgraph::data::{Build, Create};
+use petgraph::visit::{Data, NodeIndexable};
+
+use super::InvalidInputError;
+
+// Adapted from NetworkX, licensed under the MIT license
+// https://github.com/networkx/networkx/blob/409979eff35f02eff54f4eea3731736bd431dc2e/networkx/generators/social.py#L49
+const ZACHARY: &str = "\
```
mtreinish, line 22: Instead of a const string like this, would it make more sense as a static array? Something like: 
```suggestion
static ZACHARY: [[u8; 34]; 34] = [
```

rustworkx-core/src/generators/karate_club.rs
```diff
@@ -0,0 +1,96 @@
...
+0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 4 0 0 0 2\n\
+0 0 2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 4 3 0 0 0 0 0 0 0 0 4\n\
+0 0 2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 2 0 2\n\
+0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 3 0 0 4 0 0 0 0 0 3 2\n\
+0 2 0 0 0 0 0 0 3 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 3 3\n\
+2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 2 7 0 0 2 0 0 0 4 4\n\
+0 0 2 0 0 0 0 0 3 0 0 0 0 0 3 3 0 0 1 0 3 0 2 5 0 0 0 0 0 4 3 4 0 5\n\
+0 0 0 0 0 0 0 0 4 2 0 0 0 3 2 4 0 0 2 1 1 0 3 4 0 0 2 4 2 2 3 4 5 0";
+
+const MR_HI_MEMBERS: [usize; 17] = [0, 1, 2, 3, 4, 5, 6, 7, 8, 10, 11, 12, 13, 16, 17, 19, 21];
```
mtreinish, line 58: Is there a reason we need to use a `usize` for this? I'd be inclined to use `u8` instead since nothing will exceed 255.

rustworkx-core/src/generators/karate_club.rs
```diff
@@ -0,0 +1,96 @@
...
+    mut default_node_weight: F,
+    mut default_edge_weight: H,
+) -> Result<G, InvalidInputError>
+where
+    G: Build + Create + Data<NodeWeight = T, EdgeWeight = M> + NodeIndexable,
+    F: FnMut(bool) -> T,
+    H: FnMut(usize) -> M,
+    G::NodeId: Eq + Hash,
+{
+    let mut graph = G::with_capacity(0, 0);
```
mtreinish, line 74: This graph has a fixed size right? Shouldn't we allocate it appropriately as:

```suggestion
    let mut graph = G::with_capacity(34, 78);
```

rustworkx-core/src/generators/karate_club.rs
```diff
@@ -0,0 +1,96 @@
...
+    let membership: std::collections::HashSet<usize> = MR_HI_MEMBERS.into_iter().collect();
+    let mut node_indices = Vec::new();
```
mtreinish, line 76: Should this be?:
```suggestion
    let mut node_indices = Vec::with_capacity(34);
```

rustworkx-core/src/generators/karate_club.rs
```diff
@@ -0,0 +1,96 @@
...
+    for (row, line) in ZACHARY.split('\n').enumerate() {
+        let node_id = graph.add_node(default_node_weight(membership.contains(&row)));
+        node_indices.push(node_id);
+        let this_row: Vec<usize> = line
+            .split_whitespace()
+            .map(|b| b.parse::<usize>().unwrap())
+            .collect();
```
mtreinish, line 83: I feel like if we make this a static array as I suggested above this gets a lot simpler, because we just are iterating over the array instead of having to do string manipulation.

rustworkx-core/src/generators/karate_club.rs
```diff
@@ -0,0 +1,96 @@
...
+2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 2 7 0 0 2 0 0 0 4 4\n\
+0 0 2 0 0 0 0 0 3 0 0 0 0 0 3 3 0 0 1 0 3 0 2 5 0 0 0 0 0 4 3 4 0 5\n\
+0 0 0 0 0 0 0 0 4 2 0 0 0 3 2 4 0 0 2 1 1 0 3 4 0 0 2 4 2 2 3 4 5 0";
+
+const MR_HI_MEMBERS: [usize; 17] = [0, 1, 2, 3, 4, 5, 6, 7, 8, 10, 11, 12, 13, 16, 17, 19, 21];
+
+/// Generates Zachary's Karate Club graph.
+///
+/// Zachary's Karate Club graph is a well-known social network that represents
+/// the relations between 34 members of a karate club.
```
mtreinish, line 63: I feel like the docstring here needs a bit more detail, some details on the 2 arguments would be useful for rust users. Also a citation to the source of the algorithm like in Python would be good. Also we've typically been having a doc example for everything in rustworkx-core so there is at least a doctest of the public function.

rustworkx-core/src/generators/karate_club.rs
```diff
@@ -0,0 +1,96 @@
...
+pub fn karate_club_graph<G, T, F, H, M>(
+    mut default_node_weight: F,
+    mut default_edge_weight: H,
+) -> Result<G, InvalidInputError>
```
mtreinish, line 67: Can this ever return an `InvalidInputError`? I didn't see anything that returns an error although I might have missed it.

tests/graph/test_karate.py
```diff
@@ -0,0 +1,42 @@
...
+# under the License.
+
+import unittest
+
+import rustworkx as rx
+import networkx as nx
+
+
+class TestKarate(unittest.TestCase):
+    def test_isomorphic_to_networkx(self):
```
mtreinish, line 20: I'm not sure what else we'd test besides this, but I'm not thrilled about relying on networkx for the correct behavior here. If there was a bug in a release of networkx in their `karate_club_graph` function we'll be effected by that. However, in practice that seems pretty unlikely so I'm fine with this. But maybe add a test of the multigraph flag just so we have coverage of it.

src/generators.rs
```diff
@@ -1715,6 +1715,62 @@ pub fn dorogovtsev_goltsev_mendes_graph(py: Python, n: usize) -> PyResult<graph:
...
 
+/// Generates Zachary's Karate Club graph.
+///
+/// Zachary's Karate Club graph is a well-known social network that represents
+/// the relations between 34 members of a karate club, as described in the
+/// paper by Zachary [1]_.
+///
+/// The graph is undirected and contains 34 nodes (members) and 78 edges
+/// (interactions). Each node represents a club member, and each edge represents
+/// a relationship between two members.
```
mtreinish, line 1726: Maybe we should mention what the node and edge data payloads are since they're fixed in the output.

mtreinish (commented):
Overall this LGTM thanks for adding this generator. I just have a few inline comments that will slightly improve the performance and/or simplify the code.

rustworkx-core/src/generators/karate_club.rs
IvanIsCoding: I tried to keep the implementation as close as possible to NetworkX. I might swap to a list of tuples though

tests/graph/test_karate.py
IvanIsCoding: I will export NetworkX’s graph to XML and read the file, that way the test won’t depend on NetworkX directly

IvanIsCoding: Done.

rustworkx-core/src/generators/karate_club.rs
IvanIsCoding: I switched to u8

rustworkx-core/src/generators/karate_club.rs
IvanIsCoding: Done

rustworkx-core/src/generators/karate_club.rs
IvanIsCoding: It is infallible now

src/generators.rs
IvanIsCoding: Done. Hopefully it makes sense, I wanted to give callers the freedom to choose their types but the original paper is pretty fixed about what the edge weights are.

tests/graph/test_karate.py
```diff
@@ -0,0 +1,402 @@
...
+
+        with tempfile.NamedTemporaryFile("wt") as fd:
+            fd.write(karate_xml)
+            fd.flush()
+            nx_graph = rx.read_graphml(fd.name)[0]
+
+        graph = rx.generators.karate_club_graph()
+
+        self.assertTrue(
+            rx.is_isomorphic(graph, nx_graph, node_matcher=node_matcher, edge_matcher=edge_matcher)
```
mtreinish, line 51:
```suggestion
            expected = rx.read_graphml(fd.name)[0]

        graph = rx.generators.karate_club_graph()

        self.assertTrue(
            rx.is_isomorphic(graph, expected, node_matcher=node_matcher, edge_matcher=edge_matcher)
```

mtreinish (approved):

PR #1284, opened 2024-09-20
Create simple_paths_generator_with_score.rs

ranjana-mishra:
It uses Dijkstra Algo to generate all possible paths , by deleting one edge at a time, and gives the paths with their score.

### DIFF
--- a/rustworkx-core/src/simple_paths_generator_with_score.rs
+++ b/rustworkx-core/src/simple_paths_generator_with_score.rs
@@ -1,0 +1,280 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License. You may obtain
+// a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations
+// under the License.
+
+use crate::petgraph::graph::{DiGraph, NodeIndex};
+use crate::petgraph::algo::Measure;
+use crate::petgraph::visit::{EdgeRef, IntoEdges, VisitMap, Visitable};
+
+use std::collections::{HashMap, BinaryHeap};
+use std::{f32,thread};
+use std::collections::hash_map::Entry::{Occupied, Vacant};
+use std::hash::Hash;
+use std::cmp::Ordering;
+use std::fmt::Debug;
+// use rand::Rng;
+
+
+
+#[derive(Debug,Clone)]
+pub struct  SimplePath {
+    Score: f32,
+    Path : Vec<NodeIndex>,
+}
+
+// The code provides the shortest distance cost of all Nodes from the start Node 
+#[derive(Copy, Clone, Debug)] struct MinScored<K, T>(pub K, pub T);
+
+impl<K: PartialOrd, T> PartialEq for MinScored<K, T> {
+    #[inline]
+    fn eq(&self, other: &MinScored<K, T>) -> bool {
+        self.cmp(other) == Ordering::Equal
+    }
+}
+
+impl<K: PartialOrd, T> Eq for MinScored<K, T> {}
+
+impl<K: PartialOrd, T> PartialOrd for MinScored<K, T> {
+    #[inline]
+    fn partial_cmp(&self, other: &MinScored<K, T>) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl<K: PartialOrd, T> Ord for MinScored<K, T> {
+    #[inline]
+    fn cmp(&self, other: &MinScored<K, T>) -> Ordering {
+        let a = &self.0;
+        let b = &other.0;
+        if a == b {
+            Ordering::Equal
+        } else if a < b {
+            Ordering::Greater
+        } else if a > b {
+            Ordering::Less
+        } else if a.ne(a) && b.ne(b) {
+            // these are the NaN cases
+            Ordering::Equal
+        } else if a.ne(a) {
+            // Order NaN less, so that it is last in the MinScore order
+            Ordering::Less
+        } else {
+            Ordering::Greater
+        }
+    }
+}
+
+// This is mutation of petgraph dijkastra to get full path between source to target
+fn dijkstra<G, F, K>(
+   graph: G,
+   start: G::NodeId,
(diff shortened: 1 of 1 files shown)


ranjana-mishra:
@mtreinish  can you review the PR and suggest how we move forward. Few identified points are -
1) Make dijkastra Algo work with sub graphs to have faster results
2) Generate the multiple graphs through iterator to consume memory for one at a time.

rustworkx-core/src/generators/simple_shortest_paths.rs
```diff
@@ -0,0 +1,368 @@
...
+        paths.push(s.target);
+        let mut node = &s.target;
+        loop { 
+            let pre_node = path.get(node).expect("Error");
+            paths.push(*pre_node);
+            if *pre_node == s.source { 
+                break;
+            }
+            node = pre_node;
+            }
```
mtreinish, line 270: In general it'd be better to rework this as a while loop so the exit conditions are a bit more clear.

rustworkx-core/src/generators/simple_shortest_paths.rs
```diff
@@ -0,0 +1,368 @@
...
+// index: used for iterating over edged to delete one and calculate path once again
+// Source: to store the start point
+// Target: to store goal of path
+// graph: store the path to be used
+// unique_path: stores all unique_paths to verify that we are not returning same path again after random edge removal
+// switch : used for switching to next shortest path in case for one all possible paths are generated.
+/////////////
+
+#[derive(Debug, Clone)]
+pub struct SimplePath {
```
mtreinish, line 38: Typically the expectation of this kind of functionality would be to have a a public function to serve as the main interface for this functionality that takes in a reference to a graph object and returns an iterator structure like this.

mtreinish (changes requested):
Thanks for your first contribution to rustworkx, I did a quick first pass through the code and left some high level inline comments. The biggest things missing right now is that this code isn't actually included in the library, the module isn't included anywhere that gets linked in from the root library. This means while the source file is in tree nothing is building it as part of rustworkx-core. Besides not including the source as part of the library this is preventing things like the formatter and linter from actually running on it. I think you should move your new file out of `rustworkx-core/src/generators/` and into `rustworkx-core/src/shortest_path`, then you will need to modify https://github.com/Qiskit/rustworkx/blob/main/rustworkx-core/src/shortest_path/mod.rs to include your new module and re-export the public interface like is done for the other shortest path functions. Once that is done then the code will be included in the library at build time. 

Following that I expect there will be lint and formatting things to fix as I observed a couple of places that needed to be fixed. The other thing is I left a comment inline about tests we should be sure we're exercising the code as part of the PR to make sure we're adequately testing everything. I would suggest taking a look at the project's contributing guide: https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md which covers these aspects in a bit more detail.

I haven't gone through the bulk of the logic of the new function yet, I'll review the algorithm and the function in more depth along with a deeper code review when these higher level issues are fixed. There is also the question of introducing a python interface too, but we can discuss that after we have a functional rustworkx-core function.

rustworkx-core/src/generators/simple_shortest_paths.rs
ranjana-mishra: it actually back tracks the path till source, if the next node is source, then it exits.

> In general it'd be better to rework this as a while loop so the exit conditions are a bit more clear.

it actually back tracks the path till source, if the next node is source, then it exits.

rustworkx-core/src/generators/simple_shortest_paths.rs
```diff
@@ -0,0 +1,368 @@
...
+            // Order NaN less, so that it is last in the MinScore order
+            Ordering::Less
+        } else {
+            Ordering::Greater
+        }
+    }
+}
+
+// This is mutation of petgraph dijkastra to get full path between source to target
+fn dijkstra<G, F, K>(
```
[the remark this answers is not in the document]
ranjana-mishra: I have not review this specific Dijkastra, but petgraph dijkastra returns only score of each node between source & target. but how to find the path from source -> target is not clear from that score. thats why i have updated the code to return the previous node as, that it followed to reach to the destination.
like - if the path for lowest score was A->B-->C->D, it will return the score, also Node (C), which is the previous node, similarly for C , it will return score & B as previous node, and for B it will return A as previous node, which we can use to reach till A from D.

Let me know if I am missing something on existing Dijkastra functioning.

rustworkx-core/src/generators/simple_shortest_paths.rs
```diff
@@ -0,0 +1,368 @@
...
+            // these are the NaN cases
+            Ordering::Equal
+        } else if a.ne(a) {
+            // Order NaN less, so that it is last in the MinScore order
+            Ordering::Less
+        } else {
+            Ordering::Greater
+        }
+    }
+}
```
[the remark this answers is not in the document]
ranjana-mishra: yes the mic_scored already exists, as it is private section for dijkastra, I have to include it again here.

rustworkx-core/src/generators/simple_shortest_paths.rs
```diff
@@ -0,0 +1,368 @@
...
+/////////////
+
+#[derive(Debug, Clone)]
+pub struct SimplePath {
+    pub Score: f32,
+    pub Path: Vec<NodeIndex>,
+    index: usize,
+    source: NodeIndex,
+    target: NodeIndex,
+    graph: DiGraph<(), f32>,
```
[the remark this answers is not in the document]
ranjana-mishra: yes this is the main problem to the logic. I am trying to save reference , then when I do next, it needs to reference all objects of previous struct, and then it gives lifetime errors, saying 1 should exists after 2. when I try to copy the reference from self to next().

I will try this to use only reference.

rustworkx-core/src/generators/simple_shortest_paths.rs
ranjana-mishra: I will try that, as yield is not there , I am still looking how to do this, i want to calculate only 1 path at a time, want to go for next path only if .next() or something similar is called.

rustworkx-core/src/generators/simple_shortest_paths.rs
ranjana-mishra: see the current status, even though its same with comment one will get to know the exit state.

ranjana-mishra:
Please review the PR -
- it is fix over k-shortest paths ( same cost paths ), as that is not considering already visited points and fails on test cases.
- Provide function to get n paths in increasing cost order.

ranjana-mishra:
New PR for this https://github.com/Qiskit/rustworkx/pull/1363

PR #1290, opened 2024-10-03
Remove support for Python 3.8

IvanIsCoding:
This is our yearly Python version maintenance. 3.8 reached EOL. I will add the newly released 3.13 in a separate PR to test the wheel release etc.

I profited to update `black`, `ruff` and `mypy` as well. `mypy` passes without any changes, but there were some minor differences with `black` and `ruff`.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,9 +1,6 @@
 queue_rules:
   - name: automerge
     conditions:
-      - check-success=python3.8-x64 windows-latest
-      - check-success=python3.8-x64 ubuntu-latest
-      - check-success=python3.8-x64 macOS-latest
       - check-success=python3.9-x64 windows-latest
       - check-success=python3.9-x64 ubuntu-latest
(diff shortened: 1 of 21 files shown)


.mergify.yml
```diff
@@ -1,9 +1,6 @@
 queue_rules:
   - name: automerge
     conditions:
-      - check-success=python3.8-x64 windows-latest
```
mtreinish, line 4: I need to push a PR to update this config with GitHub merge queues we don't need this, and we can only use mergify for automated backporting.

mtreinish (approved):
This lgtm and  it looks like is ci passing with the correct python versions. Thanks for doing this.

.mergify.yml
IvanIsCoding: Ditto. I actually quite like the GitHub queues, we could move some more expensive checks to the final merge instead of running them each time

PR #1291, opened 2024-10-04
Add Python 3.13 to CI and to release

IvanIsCoding:
This adds Python 3.13 to CI and bumps `cibuildwheel` to the latest version.

I will test the wheel build on my personal repository before merging.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -24,7 +24,7 @@ stubs_deps = [
 
 def install_rustworkx(session):
     session.install(*deps)
-    session.install(".", "-c", "constraints.txt")
+    session.install(".[all]", "-c", "constraints.txt")
 
 # We define a common base such that -e test triggers a test with the current
 # Python version of the interpreter and -e test_with_version launches
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,7 +58,7 @@ jobs:
     strategy:
       matrix:
         rust: [stable]
-        python-version: [3.9, "3.10", "3.11", "3.12"]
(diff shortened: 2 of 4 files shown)


noxfile.py
```diff
@@ -24,7 +24,7 @@
 
 def install_rustworkx(session):
     session.install(*deps)
-    session.install(".", "-c", "constraints.txt")
+    session.install(".[all]", "-c", "constraints.txt")
```
IvanIsCoding, line 27: I removed this by accident in #1290

IvanIsCoding:
Putting this on hold until October 7

IvanIsCoding:
And we are affected by https://github.com/python-pillow/Pillow/issues/6676, MyPy is getting smarter although not when I wanted

mtreinish (commented):
This LGTM, but before merging we probably should update the trove classifiers in the setup.py: https://github.com/Qiskit/rustworkx/blob/main/setup.py#L72

IvanIsCoding:
> This LGTM, but before merging we probably should update the trove classifiers in the setup.py: https://github.com/Qiskit/rustworkx/blob/main/setup.py#L72

I added 3.13 to that list. I also just removed `win32` from the wheels being built because of #1293. I will test the wheels build in my repository before merging.

IvanIsCoding:
The test build passed, all the failures are permission issues because we omitted the PyPI tokens. This is good to go

IvanIsCoding:
After the NumPy bump in #1305 we no longer need to drop the Windows 32-bit build. All builds are working now

mtreinish (approved):

PR #1293, opened 2024-10-10
Update PyO3 to 0.22

IvanIsCoding:
Closes #1250

This depends on:
* `rust-numpy` releasing 0.22, which will happen soon. For the time being, we build from their Git main branch
* `py-clone` feature is enabled, if it is disabled too many things fail
* ~The update broke `iterators.rs`, I haven't had the time to fix so for this draft I just deleted support for slicing. Of course the final version will need to reintroduce it~ Slicing is now fixed
* ~We have 68 warnings! But hey it builds~  Warnings are fixed

I also profited from this PR to update `ndarray` and `ndarray-stats`.

For `py-clone` I think we should handle it in a future PR

PR #1294, opened 2024-10-13
Swap `ahash` with `foldhash`

IvanIsCoding:
This matches `hashbrown`'s 0.15 release that switches the default hasher. This is more to keep consistency with the internals.

I will put this on hold until #1293 is merged, we are still stuck with an older version of `hashbrown` until we update PyO3 anyway

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,7 +9,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "getrandom",
  "once_cell",
  "version_check",
  "zerocopy",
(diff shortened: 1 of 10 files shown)


IvanIsCoding:
This fails MSRV, I guess hashbrown will be stuck as well

jamestwebber:
I used to use `hashbrown` for my own code but stopped when I noticed this on their github repo:

> Since Rust 1.36, this is now the HashMap implementation for the Rust standard library. However you may still want to use this crate instead since it works in environments without std, such as embedded systems and kernels.

Maybe there have been additional features in `hashbrown` that make it worth using now but I'm curious is the difference is measurable.

IvanIsCoding:
> I used to use `hashbrown` for my own code but stopped when I noticed this on their github repo:
> 
> > Since Rust 1.36, this is now the HashMap implementation for the Rust standard library. However you may still want to use this crate instead since it works in environments without std, such as embedded systems and kernels.
> 
> Maybe there have been additional features in `hashbrown` that make it worth using now but I'm curious is the difference is measurable.

That is a fair point. It is not strictly necessary, but I feel `hashbrown` is always trying to push a little bit above the standard library in terms of API and performance.

Also, it is a transitive dependency of `indexmap` so the crate is already in the dependencies anyway. We might as well benefit from some performance improvements like the one from `foldhash`

mtreinish:
> Maybe there have been additional features in hashbrown that make it worth using now but I'm curious is the difference is measurable.

The key reasons we use hashbrown on rustworkx is the default hasher is higher performance, it actually is surprising how much a difference using ahash made here when I did the benchmarking for it many years ago (I hadn't heard of foldhash before this PR though so I can't comment on that). We could have gotten this by directly using ahash on the stdlib hashmap though (or ahash's mapping type). The other key reason though was rayon support, we don't use it extensively but if we need a parallel iterator over a hashmap or hashset hashbrown has a rayon feature we can use.

IvanIsCoding:
I am closing this as indexmap is now pinned to a version with hashbrown that still uses ahash

PR #1299, opened 2024-10-20
Remove unused feature from `sprs` dependency

IvanIsCoding:
In short, this disables the `alga` feature of the `sprs` crate. It is enabled by default and we don't use those traits, so the code still compiles and our tests still pass.

It also solves the annoying issue of having a older `num-complex` in Cargo.lock, because sometimes `cargo update` swaps the version we use which causes issues with PyO3 and ndarray/NumPy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,30 +16,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "alga"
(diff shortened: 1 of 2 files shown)


mtreinish (approved):
LGTM, thanks for cleaning this up.

PR #1307, opened 2024-11-03
Remove The Hashmap from Shorted Path for Centrality Computation

Paulo-21:
Hello, 
I removed the hashmap for the shorted path for centrality.

This may improve performance.

Tell me what do u think about :)

### DIFF
--- a/rustworkx-core/src/centrality.rs
+++ b/rustworkx-core/src/centrality.rs
@@ -353,35 +353,50 @@ where
 {
     let mut verts_sorted_by_distance: Vec<G::NodeId> = Vec::new(); // a stack
     let c = graph.node_count();
-    let mut predecessors = HashMap::<G::NodeId, Vec<G::NodeId>>::with_capacity(c);
-    let mut sigma = HashMap::<G::NodeId, f64>::with_capacity(c);
-    let mut distance = HashMap::<G::NodeId, i64>::with_capacity(c);
+    let mut predecessors: Vec<Vec<G::NodeId>> = Vec::with_capacity(c);
+    let mut sigma: Vec<f64> = Vec::with_capacity(c);
+    let mut distance: Vec<i64> = Vec::with_capacity(c);
     #[allow(non_snake_case)]
     let mut Q: VecDeque<G::NodeId> = VecDeque::with_capacity(c);
-
     for node in graph.node_identifiers() {
-        predecessors.insert(node, Vec::new());
-        sigma.insert(node, 0.0);
-        distance.insert(node, -1);
+        let node_index = graph.to_index(node);
+        predecessors[node_index] = Vec::new();
+        sigma[node_index] = 0.0;
+        distance[node_index] = -1;
     }
-    sigma.insert(*node_s, 1.0);
-    distance.insert(*node_s, 0);
+    let node_s_index = graph.to_index(*node_s);
+    sigma[node_s_index] = 1.0;
+    distance[node_s_index] = 0;
     Q.push_back(*node_s);
     while let Some(v) = Q.pop_front() {
         verts_sorted_by_distance.push(v);
-        let distance_v = distance[&v];
+        let v_idx = graph.to_index(v);
+        let distance_v = distance[v_idx];
         for w in graph.neighbors(v) {
-            if distance[&w] < 0 {
+            let w_idx = graph.to_index(w);
+            if distance[w_idx] < 0 {
                 Q.push_back(w);
-                distance.insert(w, distance_v + 1);
+                distance[w_idx] = distance_v + 1;
             }
-            if distance[&w] == distance_v + 1 {
-                sigma.insert(w, sigma[&w] + sigma[&v]);
-                let e_p = predecessors.get_mut(&w).unwrap();
-                e_p.push(v);
+            if distance[w_idx] == distance_v + 1 {
+                sigma[w_idx] = sigma[w_idx] + sigma[v_idx];
+                predecessors[w_idx].push(v);
             }
         }
     }
+    let mut sigma_h = HashMap::with_capacity(c);
+    for (idx, s) in sigma.iter().enumerate() {
+        sigma_h.insert(graph.from_index(idx), *s);
+    }
+    let mut pred_h = HashMap::with_capacity(c);
+    let mut idx = predecessors.len() - 1;
+
+    while let Some(p) = predecessors.pop() {
+        pred_h.insert(graph.from_index(idx), p);
+        idx -= 1;
+    }
+    let sigma = sigma_h;
+    let predecessors = pred_h;
     verts_sorted_by_distance.reverse(); // will be effectively popping from the stack
     ShortestPathData {
         verts_sorted_by_distance,


rustworkx-core/src/centrality.rs
```diff
@@ -342,43 +342,39 @@ where
...
-    let mut verts_sorted_by_distance: Vec<G::NodeId> = Vec::new(); // a stack
     let c = graph.node_count();
-    let mut predecessors = HashMap::<G::NodeId, Vec<G::NodeId>>::with_capacity(c);
-    let mut sigma = HashMap::<G::NodeId, f64>::with_capacity(c);
-    let mut distance = HashMap::<G::NodeId, i64>::with_capacity(c);
+    let max_index = graph.node_bound();
+    let mut verts_sorted_by_distance: Vec<G::NodeId> = Vec::with_capacity(c); // a stack
+    let mut predecessors: Vec<Vec<usize>> = vec![Vec::new(); max_index];
+    let mut sigma: Vec<f64> = vec![0.; max_index];
+    let mut distance: Vec<i64> = vec![-1; max_index];
```
IvanIsCoding, line 358: A more appropriate type here is `Option<i64>` if you want to represent missing paths. I'd even say `Option<usize>`

rustworkx-core/src/centrality.rs
```diff
@@ -280,11 +280,11 @@ fn _accumulate_vertices<G>(
...
         let iw = graph.to_index(*w);
-        let coeff = (1.0 + delta[iw]) / path_calc.sigma[w];
-        let p_w = path_calc.predecessors.get(w).unwrap();
-        for v in p_w {
-            let iv = graph.to_index(*v);
-            delta[iv] += path_calc.sigma[v] * coeff;
+        let coeff = (1.0 + delta[iw]) / path_calc.sigma[iw];
+        let p_w = path_calc.predecessors.get(iw).unwrap();
+        for iv in p_w {
+            //let iv = graph.to_index(*v);
```
IvanIsCoding, line 286: Remove this comment

IvanIsCoding (commented):
I think minimizing the number of hashing operations that happen during the shortest path is a good idea in general. But I am not convinced this works, we need to benchmark this more carefully.

I have a feeling this method will be slower for directed graphs, where some nodes are not able to reach all other nodes in the graph. In those cases, having a small hashmap with the nodes that can be reached is much faster than having a large vector with mostly non-visited entries.

rustworkx-core/src/centrality.rs
```diff
@@ -342,43 +342,39 @@ where
...
+    let mut sigma: Vec<f64> = vec![0.; max_index];
+    let mut distance: Vec<i64> = vec![-1; max_index];
     #[allow(non_snake_case)]
     let mut Q: VecDeque<G::NodeId> = VecDeque::with_capacity(c);
-
-    for node in graph.node_identifiers() {
-        predecessors.insert(node, Vec::new());
-        sigma.insert(node, 0.0);
-        distance.insert(node, -1);
-    }
```
Paulo-21, line 366: Se how the hashmap was full filled with value for all node of the graph, so replacing with a vec of size of node bound will  no be a problem for cache efficiency, because hashmap was already full when the algorithm started

Paulo-21:
> I think minimizing the number of hashing operations that happen during the shortest path is a good idea in general. But I am not convinced this works, we need to benchmark this more carefully.
> 
> I have a feeling this method will be slower for directed graphs, where some nodes are not able to reach all other nodes in the graph. In those cases, having a small hashmap with the nodes that can be reached is much faster than having a large vector with mostly non-visited entries.

I heard your argument and i agree that we should benchmark to be sure that it will be faster when the hashmap was not full filled
But in this particular case the hashmap was full filled before the algorithm started so i don't think it will cause any different.
I think we can replace every hashmap indexed by NodeId type and that is full filled before algorithm start.

Paulo-21:
Hello, i have finaly run some benchmark
This is a screenshot from the perf tool on linux with the version with hashmap.
We can see that there a non -negligable time spend with the hashmap.
image

I have run some benchmark and in single thread : 
with hashmap 1789 micro second 
without hasmap 1050 micro second .

Do you want more benchmark ?

rustworkx-core/src/centrality.rs
```diff
@@ -470,41 +465,37 @@ where
         + NodeCount
         + GraphBase
         + IntoEdges,
-    G::NodeId: Eq + Hash,
-    G::EdgeId: Eq + Hash,
+    G::NodeId: Eq,
+    G::EdgeId: Eq,
 {
     let mut verts_sorted_by_distance: Vec<G::NodeId> = Vec::new(); // a stack
     let c = graph.node_count();
```
IvanIsCoding, line 472: I believe we'd need to change this to `node_bound()`. It is definetely worth adding a test case where some nodes are removed. I need to check if we have any

rustworkx-core/src/centrality.rs
IvanIsCoding: So I was not involved in the original review of #799. Your version is definetely better than what was submitted in #799. With that being said, I'd like to compare it to a more optimized version like in https://github.com/IvanIsCoding/rustworkx/blob/f9de1db7fd5f9efafe4d6f5d9012ae2c75364081/rustworkx-core/src/centrality.rs#L1137. But that can come in a follow up PR.

IvanIsCoding (commented):
I will stand corrected that indeed this is better than the code from #799. I was not involved in the review of #799, but to be honest we can do much better than hashing every single node in the graph (multiple) times.

Your numbers are nice but not very reproducible at the moment, I suggest you test both edge_betweness_centrality and betweness_centrality with two best/worst case scenarios from Python:
* https://www.rustworkx.org/apiref/rustworkx.generators.directed_path_graph.html -> worst case scenario, very sparse braph
* https://www.rustworkx.org/apiref/rustworkx.generators.complete_graph.html -> best case scenario for the optimization

With the benchmark code, we can reuse it to testperformance improvements. I think at the end, the status of thee code will be:
* Undirected graphs always call the vector version (this PR)
* Directed graphs call an optimized version with hash map (future PR)

IvanIsCoding:
Also, last but not least add a test to test edge betweness centrality with a deleted node: https://github.com/Qiskit/rustworkx/blob/30897c5bc46efcb17bc81c33946a0a4260e2d652/tests/graph/test_centrality.py#L62

Paulo-21:
Hello,
come back from benchmark

I have run some benchmark with the graphs you mentionned before.
The supposed worse case and best case senarios.
### For the Directed graph of 10000 nodes :

With hashmap : 2.5117154121398926 secondes
without : 0.8792691230773926 secondes


### For Complete graph of 1000 nodes : 
With hashmap : 4.231908559799194 secondes
Without : 2.379117965698242 secondes

if you want to test by your self
`pip install --force-reinstall git+https://github.com/Paulo-21/rustworkx`

```
from rustworkx import betweenness_centrality
import rustworkx.generators
from time import time
#graph = rustworkx.generators.directed_path_graph(100000)
graph = rustworkx.generators.complete_graph(1000)
t = time()
betweenness_centrality(graph)
print(time() - t)
```

Paulo-21:
> Also, last but not least add a test to test edge betweness centrality with a deleted node:
> 
> https://github.com/Qiskit/rustworkx/blob/30897c5bc46efcb17bc81c33946a0a4260e2d652/tests/graph/test_centrality.py#L62

Okay i will do it later

IvanIsCoding:
There are some Clippy failures, but they are from the update to Rust 1.89. They are not your fault.

you can leave the PR as is, I’ll tweak it and merge it during 0.18. Right now, reviews for 0.17 are a bit stuck

Paulo-21:
I ran "cargo test -all" with the code up to date and i get this error 
<details>
<summary>Test error trace</summary>

```
failures:

---- centrality::test_newman_weighted_closeness_centrality::test_the_same_as_closeness_centrality_when_weights_are_1_improved_digraph stdout ----

thread 'centrality::test_newman_weighted_closeness_centrality::test_the_same_as_closeness_centrality_when_weights_are_1_improved_digraph' panicked at rustworkx-core/src/centrality.rs:1394:13:
assertion `left == right` failed
  left: [Some(inf), Some(6.0), Some(2.0), Some(1.0), Some(0.6), Some(0.4000000000000001), Some(0.2857142857142857)]
 right: [Some(0.0), Some(0.16666666666666666), Some(0.2222222222222222), Some(0.25), Some(0.26666666666666666), Some(0.27777777777777773), Some(0.2857142857142857)]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- centrality::test_newman_weighted_closeness_centrality::test_the_same_as_closeness_centrality_when_weights_are_1_not_improved_digraph stdout ----

thread 'centrality::test_newman_weighted_closeness_centrality::test_the_same_as_closeness_centrality_when_weights_are_1_not_improved_digraph' panicked at rustworkx-core/src/centrality.rs:1373:13:
assertion `left == right` failed
  left: [Some(inf), Some(6.0), Some(2.0), Some(1.0), Some(0.6), Some(0.4), Some(0.2857142857142857)]
 right: [Some(0.0), Some(1.0), Some(0.6666666666666666), Some(0.5), Some(0.4), Some(0.3333333333333333), Some(0.2857142857142857)]

---- centrality::test_newman_weighted_closeness_centrality::test_weighted_closeness_graph stdout ----

thread 'centrality::test_newman_weighted_closeness_centrality::test_weighted_closeness_graph' panicked at rustworkx-core/src/centrality.rs:1348:13:
assertion `left == right` failed
  left: [Some(0.2608695652173913), Some(0.3333333333333333), Some(0.4), Some(0.42857142857142855), Some(0.4), Some(0.3333333333333333), Some(0.2608695652173913)]
 right: [Some(0.2857142857142857), Some(0.375), Some(0.46153846153846156), Some(0.5), Some(0.46153846153846156), Some(0.375), Some(0.2857142857142857)]


failures:
    centrality::test_newman_weighted_closeness_centrality::test_the_same_as_closeness_centrality_when_weights_are_1_improved_digraph
    centrality::test_newman_weighted_closeness_centrality::test_the_same_as_closeness_centrality_when_weights_are_1_not_improved_digraph
    centrality::test_newman_weighted_closeness_centrality::test_weighted_closeness_graph

test result: FAILED. 308 passed; 3 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.70s

error: test failed, to rerun pass `-p rustworkx-core --lib`
```

</details>

rustworkx-core/src/centrality.rs

rustworkx-core/src/centrality.rs

rustworkx-core/src/centrality.rs
IvanIsCoding: Done. This was an actual bug, for instance.

PR #1312, opened 2024-11-11
Fix bug where edge colors are randomly shuffled in `mpl_draw`

### LINKED ISSUE #1308: `rx.visualization.mpl_draw()` does not apply an `edge_color` list to the correct edges
### Information

- **Python version**: 3.11.9
- **Rust version**: N/A
- **Operating system**: Ubuntu 20.04

[`rustworkx.visualization.matplotlib.draw_edges()`](https://github.com/Qiskit/rustworkx/blob/312a70bf60437d0d7ad61d42e0b886f601cb8076/rustworkx/visualization/matplotlib.py#L485) does not correctly apply `edge_color` of type `list`. Edge colors are not being applied to the edges of `graph.edge_list()` in the correct order. Issue sourced from: https://github.com/qBraid/qBraid/issues/802

The entries of a given `edge_color` list should be mapped to the correct edges, corresponding to the order of `graph.edge_list()`.

The code below plots 3 nodes and 3 edges. Edge `(c, a)` should be colored `blue`. But if you run the code multiple times, you will find that the blue edge changes, and is not always `(c, a)`. 

```python
import rustworkx as rx
from rustworkx.visualization import mpl_draw

graph = rx.PyDiGraph()

a = graph.add_node("A")
b = graph.add_node("B")
c = graph.add_node("C")

graph.add_edges_from([(a, b, 0.1), (b, c, 1.2), (c, a, 2.0)])

edge_colors_dict = {(c, a): "blue"}
edge_colors = [edge_colors_dict.get((u, v), "grey") for u, v in graph.edge_list()]

mpl_draw(graph, edge_color=edge_colors, with_labels=True, labels=str)

plt.show()
```

If you `pip install rustworkx<0.15.0`, and run the code again, it works fine.

Therefore, I was able to trace this bug back the following commit: https://github.com/Qiskit/rustworkx/commit/646057ae16505293f48fcc260cfa384bcad78fbc

Using the `draw_edges()` function from before that commit (i.e. `rustworkx` version 0.14.2) resolves the issue. But the latest update seems to form an `arrow_color` list that is not consistent with the indexing of the `edge_color` list.

IvanIsCoding:
Closes #1308. A bit embarrasing but we let that slip through in #1204.

The solution is a bit hacky but reusing Python dictionaries was the easiest way to get a set-like object that preservers iteration order.

### DIFF
--- a/releasenotes/notes/correct-edge-colors-e082e1761e1c060b.yaml
+++ b/releasenotes/notes/correct-edge-colors-e082e1761e1c060b.yaml
@@ -1,0 +1,8 @@
+---
+fixes:
+  - |
+    Fixed a bug introduced in version 0.15 where the edge colors specified as
+    a list in calls to :func:`~rustworkx.visualization.mpl_draw` were not
+    having their order respected. Instead, the order of the colors was
+    being shuffled. This has been restored and now the behavior should
+    match that of 0.14.
--- a/rustworkx/visualization/matplotlib.py
+++ b/rustworkx/visualization/matplotlib.py
@@ -639,7 +639,7 @@ def draw_edges(
     edge_pos_keys = dict()
     for e in edge_list:
         edge_pos_keys[(tuple(pos[e[0]]), tuple(pos[e[1]]))] = None
-    edge_pos = edge_pos.keys()
+    edge_pos = edge_pos_keys.keys()
 
     # Check if edge_color is an array of floats and map to edge_cmap.
     # This is the only case handled differently from matplotlib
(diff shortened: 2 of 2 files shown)


mtreinish (approved):
It's a little weird as a pattern, but this makes sense to have set like mechanics while retaining insertion order. This LGTM, before merging though was there a test we wanted to add for this to catch a potential future regression?

IvanIsCoding:
I was checking https://github.com/Qiskit/rustworkx/blob/main/tests/visualization/test_mpl.py. Currently we have no tests asserting any attributes about the images. We just assert they can be saved.

I will create another issue to add Matplotlib that actually catch regressions

PR #1314, opened 2024-11-14
Impl `__repr__` for custom vector

zen-xu:
I found that rustworkx's custom vectors do not provide `__repr__` method, which I have to use `print(indices)`


IvanIsCoding:
This seems straightforward to me, I think we should have it

IvanIsCoding (approved):

PR #1315, opened 2024-11-15
Read gzipped graphml files

fabmazz:
Create python function to load a graphml file that has been compressed with gzip without necessarily unzipping it before.
This makes reading large graphs in graphml much faster, and they can use less space

- [x] I ran rustfmt locally
- [x] I ran `cargo clippy` and it suggests no changes 


IvanIsCoding:
Thanks for submitting this. I think this is a very welcoming addition.

However, I don’t think we should have a new mehod to support compressed files. I’d refactor the existing method to try to decompress the file based on the file extension. And perhaps an optional argument like “force_decompression” to decompress anyway even if the file extension is not what we expect

fabmazz:
Ok I have now included everything in a single function, with the optional argument `compression` to force.

src/graphml.rs
```diff
@@ -703,9 +732,9 @@ impl GraphML {
 /// :rtype: list[Union[PyGraph, PyDiGraph]]
 /// :raises RuntimeError: when an error is encountered while parsing the GraphML file.
 #[pyfunction]
-#[pyo3(text_signature = "(path, /)")]
-pub fn read_graphml(py: Python, path: &str) -> PyResult<Vec<PyObject>> {
-    let graphml = GraphML::from_file(path)?;
+#[pyo3(signature=(path, compression=""),text_signature = "(path, compression=\"\", /)")]
+pub fn read_graphml(py: Python, path: &str, compression: &str) -> PyResult<Vec<PyObject>> {
```
IvanIsCoding, line 736: Prefer `Option<&str>` over an empty string. For Python the default would become `None` which gets converted to `None` in Rust as well

IvanIsCoding (commented):
Awesome this is looking good. I left a minor comment. 

Lastly, check CONTRIBUTING.md for how to update the type stubs now that you added a new argument to an existing function. Our CI should fail you in the “Stubs” section until you update

src/graphml.rs
fabmazz: Yeah I thought about that too, but apparently it's being phased out: 
https://pyo3.rs/main/function/signature#trailing-optional-arguments

IvanIsCoding: > Yeah I thought about that too, but apparently it's being phased out: https://pyo3.rs/main/function/signature#trailing-optional-arguments
You just need to specify the None in the signature like you specified the empty string. We still use that in lots of places. What is deprecated is None being inserted by default without declaration

IvanIsCoding (approved):
I am looking forward to releasing this in 0.16!

PR #1323, opened 2024-11-19
Add immediate_dominators function

airwoodix:
If I didn't miss it, dominance-related functionality is not yet exposed. Apologies if I missed it.

This patch adds a `rustworkx.immediate_dominators` function that has the same functionality as [`networkx.immediate_dominators`](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.dominance.immediate_dominators.html). The core functionality is directly provided by the `petgraph` crate, such that only a thin Python binding was necessary.

The tests are directly ported from [those in networkx](https://github.com/networkx/networkx/blob/9c5ca54b7e5310a21568bb2e0104f8c87bf74ff7/networkx/algorithms/tests/test_dominance.py#L6).

An equivalent to `networkx.dominance_frontiers` could be added in a later patch.


src/dominance.rs
```diff
@@ -0,0 +1,60 @@
...
+/// :rtype: Mapping[int, int]
+///
+/// :raises NullGraph: the passed graph is empty
+/// :raises InvalidNode: the start node is not in the graph
+#[pyfunction]
+#[pyo3(text_signature = "(graph, start_node, /)")]
+pub fn digraph_immediate_dominators(
+    graph: &digraph::PyDiGraph,
+    start_node: usize,
+) -> PyResult<HashMap<usize, usize>> {
```
IvanIsCoding, line 40: We have`DictMap` instead of HashMap to preserve the output order. Most Python users get confused by the randomized iteration order of HashMap

rustworkx/rustworkx.pyi
```diff
@@ -1048,6 +1048,12 @@ def graph_union(
     merge_edges: bool = ...,
 ) -> PyGraph[_S, _T]: ...
 
+# Dominance
+
+def digraph_immediate_dominators(
+    graph: PyDiGraph[_S, _T], start_node: int, /
+) -> Mapping[int, int]: ...
```
IvanIsCoding, line 1055: I think putting the return type here as `dict[int,int]` is more appropriate. The method is not generic PyO3 always cast the HashMap to a dictionary

IvanIsCoding (commented):
Thanks for submitting this! It is a welcome addition.

I just left some minor comments but overall this is very close to getting merged and release in 0.16

tests/digraph/test_dominance.py
```diff
@@ -0,0 +1,140 @@
...
+        with self.assertRaises(rx.InvalidNode):
+            rx.digraph_immediate_dominators(graph, 1)
+
+    def test_singleton(self):
+        """
+        Edge cases: single node, optionally cyclic.
+        """
+        graph = rx.PyDiGraph()
+        graph.add_node(0)
+        assert rx.digraph_immediate_dominators(graph, 0) == {0: 0}
```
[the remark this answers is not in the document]
IvanIsCoding: And for all other assert calls in general

src/lib.rs
```diff
@@ -464,6 +466,7 @@ fn rustworkx(py: Python<'_>, m: &Bound<PyModule>) -> PyResult<()> {
     m.add_wrapped(wrap_pyfunction!(graph_vf2_mapping))?;
     m.add_wrapped(wrap_pyfunction!(digraph_union))?;
     m.add_wrapped(wrap_pyfunction!(graph_union))?;
+    m.add_wrapped(wrap_pyfunction!(digraph_immediate_dominators))?;
```
IvanIsCoding, line 469: Because this algorithm only applies to directed graphs, I think calling it `immediate_dominators` is appropriate

rustworkx/rustworkx.pyi
airwoodix: Done. Also changed the return type in the docstring.

src/dominance.rs
airwoodix: Good to know, thanks for the hint!

tests/digraph/test_dominance.py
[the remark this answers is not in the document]
airwoodix: Thanks! Two functions and the pytest habits went back...

Changed to using `unittest` methods everywhere.

IvanIsCoding (approved):
Fantastic. This should come out when 0.16 is released 🚀

PR #1329, opened 2024-11-23
Add dominance_frontiers function

airwoodix:
As follow-up to #1323, this patch adds the `rustworkx.dominance_frontiers` function, with the same functionality as [networkx.dominance_frontiers](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.dominance.dominance_frontiers.html).

The algorithm is described in Figure 5 of [this paper](https://hdl.handle.net/1911/96345). The tests are adapted from the [networkx ones](https://github.com/networkx/networkx/blob/f21f284c7d985726878cd38195716b44e7d054b1/networkx/algorithms/tests/test_dominance.py#L4).


IvanIsCoding (approved):
LGTM, very nice follow up of #1323

Issue #1322, opened 2024-11-19
Wrong type annotation for `find_node_by_weight` and `node_link_json`


BarakKatzir:
### Information

- **rustworkx version**: 0.15.1
- **Python version**: 3.12

### What is the current behavior?
Wrong annotation in `rustworkx/rustworkx/rustworkx.pyi` 
```python
def find_node_by_weight(
        self,
        obj: Callable[[_S], bool],
        /,
    ) -> int | None: ...
```

### What is the expected behavior?
Correct to
```python
def find_node_by_weight(self, obj: _S, /) -> int | None: ...
```

IvanIsCoding:
All out methods are manually annotated so you found a bug. For the JSON function it is a known issue (#1243 ) and the fix will is already merged in the main branch.

find_node_by_weight is new though, we should send a PR to fix it

PR #1228, opened 2024-06-26
Prepare 0.15.0 release

mtreinish:
This commit prepares the 0.15.0 release. It moves the release notes for the release into a separate directory and adds a prelude to the release notes. This should be the last commit to merge for the 0.15.0 release and after it merges that commit will be what is should be tagged as the 0.15.0 release.

As part of this there are few small changes other changes being made. The release notes are update to have consistent formatting and wording.

TODO:

- [x] Wait for last milestone PRs to merge

PR #1330, opened 2024-11-23
Fix typos detected by the typos tool

airwoodix:
This is a proposal to add the [typos](https://github.com/crate-ci/typos) tool to the ``lint`` session.

The patch adds the tool to the Nox ``lint`` session and fixes the reported issues. Some issues require overriding the tools decision. To limit these cases, db4c4ef does some minor renaming, which hopefully also improves the code's readability.

Since file names in `releasenotes/notes/**/*.yaml` are automatically generated, they may contain false positives. This scenario is dealt with by suppressing filename spell-checking for files under `releasenotes/`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: "3.10"
-      - run: pip install -U ruff==0.6.8 black~=24.8 typos~=1.27
+      - run: pip install -U ruff==0.6.8 black~=24.8 typos~=1.28
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: rustfmt
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: "3.10"
-      - run: pip install -U ruff==0.6.8 black~=24.8
+      - run: pip install -U ruff==0.6.8 black~=24.8 typos~=1.27
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: rustfmt
@@ -42,6 +42,8 @@ jobs:
         run: ruff check rustworkx retworkx setup.py tests
       - name: Check stray release notes
         run: python tools/find_stray_release_notes.py
+      - name: Spell check
+        run: typos --exclude releasenotes/ && typos --no-check-filenames releasenotes/
       - name: rustworkx-core Rust Tests
         run: pushd rustworkx-core && cargo test && popd
       - name: rustworkx-core Docs
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,9 +27,9 @@ jobs:
         with:
           python-version: "3.10"
       - run: pip install -U ruff==0.6.8 black~=24.8 typos~=1.28
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@1.82
         with:
-          components: rustfmt
+          components: rustfmt, clippy
       - name: Test Build
         run: cargo build
       - name: Rust Format
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,7 +27,7 @@ jobs:
         with:
           python-version: "3.10"
       - run: pip install -U ruff==0.6.8 black~=24.8
-      - uses: dtolnay/rust-toolchain@1.82
+      - uses: dtolnay/rust-toolchain@stable
         with:
           components: rustfmt, clippy
       - name: Test Build
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: "3.10"
-      - run: pip install -U ruff==0.6.8 black~=24.8 typos~=1.28
+      - run: pip install -U ruff==0.6.8 black~=24.8
       - uses: dtolnay/rust-toolchain@1.82
         with:
           components: rustfmt, clippy
@@ -42,8 +42,6 @@ jobs:
         run: ruff check rustworkx retworkx setup.py tests
       - name: Check stray release notes
         run: python tools/find_stray_release_notes.py
-      - name: Spell check
-        run: typos --exclude releasenotes/ && typos --no-check-filenames releasenotes/
       - name: rustworkx-core Rust Tests
         run: pushd rustworkx-core && cargo test && popd
       - name: rustworkx-core Docs
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ it just as it would if there was a prebuilt binary available.
 > [!NOTE]  
 > To build from source you will need to ensure you have pip >=19.0.0
 installed, which supports PEP-517, or that you have manually installed
-`setuptools-rust` prior to running `pip install rustworkx`. If you recieve an
+`setuptools-rust` prior to running `pip install rustworkx`. If you receive an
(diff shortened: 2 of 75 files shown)


.github/workflows/main.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: "3.10"
-      - run: pip install -U ruff==0.6.8 black~=24.8
+      - run: pip install -U ruff==0.6.8 black~=24.8 typos~=1.27
```
IvanIsCoding, line 29: Let's remove this from CI for now and try to add it as a warning

README.md
```diff
@@ -77,7 +77,7 @@ it just as it would if there was a prebuilt binary available.
 > [!NOTE]  
 > To build from source you will need to ensure you have pip >=19.0.0
 installed, which supports PEP-517, or that you have manually installed
-`setuptools-rust` prior to running `pip install rustworkx`. If you recieve an
+`setuptools-rust` prior to running `pip install rustworkx`. If you receive an
```
IvanIsCoding, line 80: This is an example of high-value contribution because README.md typos are read by everyone and a bit embarrasing

docs/source/release_notes.rst
```diff
@@ -266,7 +266,7 @@ New Features
 
 - A new method, :meth:`~rustworkx.PyDiGraph.remove_node_retain_edges`, has been
   added to the :class:`~rustworkx.PyDiGraph` class. This method can be used to
-  remove a node and add edges from its predecesors to its successors.
+  remove a node and add edges from its predecessors to its successors.
```
IvanIsCoding, line 269: I am not even sure if we process this file anymore but it is a high-value find

noxfile.py
```diff
@@ -49,6 +50,8 @@ def lint(session):
     session.run("ruff", "check", "rustworkx", "retworkx", "setup.py")
     session.run("cargo", "fmt", "--all", "--", "--check", external=True)
     session.run("python", "tools/find_stray_release_notes.py")
+    session.run("typos", "--exclude", "releasenotes")
+    session.run("typos", "--no-check-filenames", "releasenotes")
```
IvanIsCoding, line 54: Ironically, releasenotes/ is one of the places I'd like to check for typos the most

rustworkx-core/src/max_weight_matching.rs
```diff
@@ -371,7 +371,7 @@ fn expand_blossom<E>(
         // base.
         assert!(label_ends[blossom].is_some());
         let entry_child = in_blossoms[endpoints[label_ends[blossom].unwrap() ^ 1]];
-        // Decied in which direction we will go around the blossom.
+        // Decide in which direction we will go around the blossom.
```
IvanIsCoding, line 374: This is an example of a find with low-value,, I don't want to block people when they make a typo on a code comment. Except if that code comment is a docstring of course

tests/graph/test_bellman_ford.py
```diff
@@ -175,7 +175,7 @@ def test_raises_negative_cycle_bellman_ford_paths(self):
         with self.assertRaises(rustworkx.NegativeCycle):
             rustworkx.bellman_ford_shortest_paths(graph, 0, weight_fn=float)
 
-    def test_raises_negative_cycle_bellman_ford_path_lenghts(self):
+    def test_raises_negative_cycle_bellman_ford_path_lengths(self):
```
IvanIsCoding, line 178: I made this typo, it is silly but this is one of the cases where the tool blocking CI would not add much value. The test is still correct.

IvanIsCoding (commented):
Firstly, thanks for getting this started. I'd focus for now on submitting the typo fixes and not trying to get the tool on CI. I will explain why.

Typos should be a warning not a CI failure. It adds friction for new contributors and unfortunately we already have:
* cargo clippy (fails CI)
* ruff (fails CI)
* cargo fmt (fails CI)
* black (fails CI)
* type annotations (fails CI)
* release notes (fails only if not in the folder, if missing the CI passes)
* documentation (fails only if formatted incorrectly, if missing the CI passes)

Because the development is asynchronous, that can delay how fast others can contribute. Some things like `ruff` add a lot of value, they find real bugs from time to time. Code formatters like `black` and `cargo fmt` are not strictly necessary but they help the codebase be more consistent. The type annotation checks are to prevent backsliding to a level before #401 (but it still allows bugs like #1322).

Ideally, we should have a bot messasing us with the new typos found in the PR. Something similar to the Code Coverage bot: if someone sends a PR and I see the bot commenting the coverage is low, I can immediately take action and ask for more tests. Something similar could happen for typos.

IvanIsCoding:
With that being said, for the next release we should run the release PR (like in #1228) through the tool if we don't have a bot by then. Many code review comments were humans trying to find typos.

noxfile.py
airwoodix: `releasenotes/` is checked, but not the filenames, because they the hashes from `reno` produce many false positives.

It may be possible to exclude file name checking for all YAML files or even YAML files within `releasenotes/` by adjusting `typos.toml`. The split command looked less magic to me.

typos.toml
```diff
@@ -0,0 +1,4 @@
+[default]
+extend-ignore-words-re = [
+  "[Ss]toer",
```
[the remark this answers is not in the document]
airwoodix: Yes, `store` is the suggested correction.

These cases become dangerous when using `typos -w` (edit in-place). With just a linter report, the problem is quickly spotted.

airwoodix:
Thanks for the feedback!

I understand your concerns and removed the CI check as you suggested. Would it be ok to keep it in the Nox `lint` session?

[airwoodix/gh-typos-comment](https://github.com/airwoodix/gh-typos-comment) contains a prototype of an automatic PR comment with the spell checking report (see example in airwoodix/gh-typos-comment#3 - the comment is removed if the spell checker doesn't find anything). I'm not extremely familiar with Github Actions, but it seems that the [recommended way](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) of doing this is to have dependent workflows. However, these [only trigger if the workflow file is in the `main` branch](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run), so an equivalent of [pr_comment.yml](https://github.com/airwoodix/gh-typos-comment/blob/136a4803c6c5468ac6b09d7c2af35357f0e04671/.github/workflows/pr_comment.yml) would need to be contributed separately to get the trigger in this PR. Would you consider reviewing such a patch?

typos.toml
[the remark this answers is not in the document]
IvanIsCoding: Looks good to me. The only change I will make is moving this from its own file to `pyproject.toml` as it is better to have all configurations centralized in a single file

noxfile.py
```diff
@@ -49,6 +50,8 @@ def lint(session):
     session.run("ruff", "check", "rustworkx", "retworkx", "setup.py")
     session.run("cargo", "fmt", "--all", "--", "--check", external=True)
     session.run("python", "tools/find_stray_release_notes.py")
+    session.run("typos", "--exclude", "releasenotes")
```
IvanIsCoding, line 53: I will keep this is `nox -etypos` for two reasons:
* it doesn't need rustworkx to be compiled (just like Black)
* It might make our life easier if we want to have the separate CI pass

.github/workflows/main.yml
```diff
@@ -27,9 +27,9 @@ jobs:
         with:
           python-version: "3.10"
       - run: pip install -U ruff==0.6.8 black~=24.8
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@1.82
```
IvanIsCoding, line 30: I will revert to use stable, we always want to build on the latest version

IvanIsCoding (approved):

PR #1337, opened 2024-12-08
Implement condensation graph generation

kazuki0824:
This pull request adds functionality to generate condensation graphs. The Rust implementation is copied and slightly modified from https://github.com/petgraph/petgraph because the original `petgraph` implementation cannot be applied as it is. Condensation graphs represent strongly connected components of a directed graph as single nodes. The update includes a new implementation and its test to support this feature. 

- [x] I ran rustfmt locally


kazuki0824:
Docs, usages

kazuki0824:
@IvanIsCoding can anyone review this code?

IvanIsCoding:
> @IvanIsCoding can anyone review this code?

Sorry I haven’t acknowledged the PR. I haven’t had time to review it.

I will do the review after Christmas, it might still make the cut for rustworkx 0.16.

kazuki0824:
> > @IvanIsCoding can anyone review this code?
> 
> 
> 
> Sorry I haven’t acknowledged the PR. I haven’t had time to review it.
> 
> 
> 
> I will do the review after Christmas, it might still make the cut for rustworkx 0.16.

Thank you for your quick response.

src/connectivity/mod.rs
```diff
@@ -114,6 +114,79 @@ pub fn strongly_connected_components(graph: &digraph::PyDiGraph) -> Vec<Vec<usiz
         .collect()
 }
 
+fn condensation_inner<N, E, Ty, Ix>(
+    py: &Python,
+    g: Graph<N, E, Ty, Ix>,
+    make_acyclic: bool,
+    sccs: Option<Vec<Vec<usize>>>,
+) -> StableGraph<PyObject, PyObject, Ty, Ix>
```
alexanderivrii, line 122: I am wondering if it would make sense to also return a map between the nodes of the condensed graph and the (sets of) nodes of the original graph, so that information computed for the condensed graph could be lifted to the original graph as well?

tests/digraph/test_strongly_connected.py
```diff
@@ -65,3 +65,55 @@ def test_number_strongly_connected_big(self):
...
+
+        self.graph.add_edge(self.node_b, self.node_e, "b->e")
+
+        self.graph.add_edge(self.node_e, self.node_f, "e->f")
+        self.graph.add_edge(self.node_f, self.node_g, "f->g")
+        self.graph.add_edge(self.node_g, self.node_h, "g->h")
+        self.graph.add_edge(self.node_h, self.node_e, "h->e")  # サイクル: e -> f -> g -> h -> e
+
+    def test_condensation(self):
+        # condensation関数を呼び出し
```
alexanderivrii, line 97: It would be nice to translate all comments to English.

alexanderivrii (commented):
This is not a review, but I was wondering: would it make sense to move this functionality to rustworkx-core? would it make sense to implement this both for directed and undirected graphs? would it make sense to return additional information relating the nodes of the original and the condensed graphs (see in-place comment).

src/connectivity/mod.rs
```diff
@@ -114,6 +114,79 @@ pub fn strongly_connected_components(graph: &digraph::PyDiGraph) -> Vec<Vec<usiz
...
+
+    let condensed = condensation_inner(&py, g.into(), true, sccs);
+
+    digraph::PyDiGraph {
+        graph: condensed,
+        cycle_state: algo::DfsSpace::default(),
+        check_cycle: false,
+        node_removed: false,
+        multigraph: true,
+        attrs: py.None(),
```
IvanIsCoding, line 186: Either the `attrs` should contain the node mapping like in NetworkX. Or we need to update the signature to return a mapping

src/connectivity/mod.rs
```diff
@@ -114,6 +114,79 @@ pub fn strongly_connected_components(graph: &digraph::PyDiGraph) -> Vec<Vec<usiz
...
+    condensed.map(|_, w| w.to_object(*py), |_, w| w.to_object(*py))
+}
+
+#[pyfunction]
+#[pyo3(text_signature = "(graph, /, sccs=None)", signature=(graph, sccs=None))]
+pub fn condensation(
+    py: Python,
+    graph: &digraph::PyDiGraph,
+    sccs: Option<Vec<Vec<usize>>>,
+) -> digraph::PyDiGraph {
```
IvanIsCoding, line 175: Sibling comment to either update the signature or return the node mapping somewhere else

src/connectivity/mod.rs
IvanIsCoding, line 122:
```suggestion
) -> StablePyGraph<Directed>
```
Stick to the shared type alias for consistency with the other methods.

src/connectivity/mod.rs
```diff
@@ -114,6 +114,79 @@ pub fn strongly_connected_components(graph: &digraph::PyDiGraph) -> Vec<Vec<usiz
         .collect()
 }
 
+fn condensation_inner<N, E, Ty, Ix>(
```
IvanIsCoding, line 117: This function is called only once in the code, so I don't think you need to make it generic. It is fine to accept only one kind of type for each argument.

n.b: the feedback would be different if this wasn't a private function

rustworkx/rustworkx.pyi
```diff
@@ -192,6 +192,7 @@ def number_connected_components(graph: PyGraph, /) -> int: ...
 def number_weakly_connected_components(graph: PyDiGraph, /) -> bool: ...
 def node_connected_component(graph: PyGraph, node: int, /) -> set[int]: ...
 def strongly_connected_components(graph: PyDiGraph, /) -> list[list[int]]: ...
+def condensation(graph: PyDiGraph, /, sccs=None) -> PyDiGraph: ...
```
IvanIsCoding, line 195:
```suggestion
def condensation(graph: PyDiGraph, /, sccs: list[int] | None=...) -> PyDiGraph: ...
```
You don't specify the value here, just that the argument is optional because there is already a default

src/connectivity/mod.rs
```diff
@@ -114,6 +114,79 @@ pub fn strongly_connected_components(graph: &digraph::PyDiGraph) -> Vec<Vec<usiz
...
+    py: &Python,
+    g: Graph<N, E, Ty, Ix>,
+    make_acyclic: bool,
```
IvanIsCoding, line 120: This argument is always true, I'd rather remove the argument & the logic for when it is false

src/connectivity/mod.rs
```diff
@@ -114,6 +114,79 @@ pub fn strongly_connected_components(graph: &digraph::PyDiGraph) -> Vec<Vec<usiz
...
+
+    // Consume nodes and edges of the old graph and insert them into the new one.
+    let (nodes, edges) = g.into_nodes_edges();
+    for (nix, node) in nodes.into_iter().enumerate() {
+        condensed[node_map[nix]].push(node.weight);
+    }
+    for edge in edges {
+        let source = node_map[edge.source().index()];
+        let target = node_map[edge.target().index()];
+        if make_acyclic {
```
IvanIsCoding, line 158: This is always true, you can remove the if

src/connectivity/mod.rs
```diff
@@ -114,6 +114,79 @@ pub fn strongly_connected_components(graph: &digraph::PyDiGraph) -> Vec<Vec<usiz
...
+            if source != target {
+                condensed.update_edge(source, target, edge.weight);
+            }
+        } else {
```
IvanIsCoding, line 162: This is always false, remove this statmenet

tests/digraph/test_strongly_connected.py
```diff
@@ -65,3 +65,55 @@ def test_number_strongly_connected_big(self):
             node = G.add_node(i)
             G.add_child(node, str(i), {})
         self.assertEqual(len(rustworkx.strongly_connected_components(G)), 200000)
+
+
+class TestCondensation(unittest.TestCase):
+    def setUp(self):
+        # グラフをセットアップ
```
IvanIsCoding, line 72: I appreciate the code comments but before submitting we'll either need to translate them to English

tests/digraph/test_strongly_connected.py
```diff
@@ -65,3 +65,55 @@ def test_number_strongly_connected_big(self):
...
+        self.node_e = self.graph.add_node("e")
+        self.node_f = self.graph.add_node("f")
+        self.node_g = self.graph.add_node("g")
+        self.node_h = self.graph.add_node("h")
+
+        # エッジを追加
+        self.graph.add_edge(self.node_a, self.node_b, "a->b")
+        self.graph.add_edge(self.node_b, self.node_c, "b->c")
+        self.graph.add_edge(self.node_c, self.node_d, "c->d")
+        self.graph.add_edge(self.node_d, self.node_a, "d->a")  # サイクル: a -> b -> c -> d -> a
```
IvanIsCoding, line 87:
```suggestion
        self.graph.add_edge(self.node_d, self.node_a, "d->a")  # cycle : a -> b -> c -> d -> a
```

tests/digraph/test_strongly_connected.py
```diff
@@ -65,3 +65,55 @@ def test_number_strongly_connected_big(self):
...
+
+        self.graph.add_edge(self.node_b, self.node_e, "b->e")
+
+        self.graph.add_edge(self.node_e, self.node_f, "e->f")
+        self.graph.add_edge(self.node_f, self.node_g, "f->g")
+        self.graph.add_edge(self.node_g, self.node_h, "g->h")
+        self.graph.add_edge(self.node_h, self.node_e, "h->e")  # サイクル: e -> f -> g -> h -> e
```
IvanIsCoding, line 94:
```suggestion
        self.graph.add_edge(self.node_h, self.node_e, "h->e")  # cycle: e -> f -> g -> h -> e
```

tests/digraph/test_strongly_connected.py
```diff
@@ -65,3 +65,55 @@ def test_number_strongly_connected_big(self):
             node = G.add_node(i)
             G.add_child(node, str(i), {})
         self.assertEqual(len(rustworkx.strongly_connected_components(G)), 200000)
+
+
+class TestCondensation(unittest.TestCase):
```
IvanIsCoding, line 70: The existing test case is excellent! But we need to add one test covering the case where we pass a list to `sccs` as an argument

IvanIsCoding (commented):
Overall the condensation logic looks good to me. I left comments that are mostly to simplify the existing code and make it more maintainable in the long run. I don't think you'll need to change much for the algorithm

I think the only debate left is how to return the node mappings. I personally like the short return type signature, so for me a graph attribute with the mapping seems like a good compromise.

IvanIsCoding:
Some notes to myself:
* Add a release note for this once the code is ready (https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md#release-notes)
* Add a docstring to the method once the code is ready

Feel free to do those on your own too. But I don't mind doing the documentation for you, it is fine to focus on the code.

Again, thanks for your repeated contributions to rustworkx

kazuki0824:
@IvanIsCoding Now ready to review. I also added Rust doc and release notes. Please check them too.

IvanIsCoding:
I plan to take a look at this soon, sorry for the delay

kazuki0824:
> I plan to take a look at this soon, sorry for the delay 

Thank you for your response.

src/connectivity/mod.rs
```diff
@@ -192,6 +194,161 @@ pub fn is_strongly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
+        },
+    );
+    Ok((mapped, node_map))
+}
+
+#[pyfunction]
+#[pyo3(text_signature = "(graph, /, sccs=None)", signature=(graph, sccs=None))]
+pub fn condensation(
+    py: Python,
+    graph: PyObject,
```
IvanIsCoding, line 312: We generally have the split at `__init__.py` with a Python function with `@_rustworkx_dispatch` and then two Rust functions with `digraph_*` and `graph_*`: https://github.com/Qiskit/rustworkx/blob/217d58ded7b457c30d409f594b720c5be637778e/rustworkx/__init__.py#L148

I'd rather keep the existing pattern, of course this one works but consistency in the library is important

src/connectivity/mod.rs
```diff
@@ -192,6 +194,161 @@ pub fn is_strongly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
+        let mut result = digraph::PyDiGraph {
+            graph: condensed,
+            cycle_state: algo::DfsSpace::default(),
+            check_cycle: false,
+            node_removed: false,
+            multigraph: true,
+            attrs: PyDict::new(py).into(),
+        };
+        let node_map_py = node_map.into_pyobject(py)?;
+        let attrs = PyDict::new(py);
```
IvanIsCoding, line 327: I believe you can use a `DictMap` instead of a `PyDict` and just call `into_pyobject()` at the end. In general, working with Rust primitives might be easier. In general the Python APIs force you to handle errors that don't happen very often. For Rust, putting an item in the hashmap would "just work"

src/connectivity/mod.rs
```diff
@@ -192,6 +194,161 @@ pub fn is_strongly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
+        attrs.set_item("node_map", node_map_py)?;
+        result.attrs = attrs.into();
+        Ok(result.into_pyobject(py)?.into())
```
IvanIsCoding, line 330: If you split the Rust methods in `digraph_condensation` and `graph_condensation`, you'll be able to just return the `PyDiGraph` or `PyGraph` object without needing this conversion

src/connectivity/mod.rs
```diff
@@ -192,6 +194,161 @@ pub fn is_strongly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
+    let components_usize: Vec<Vec<NodeIndex<usize>>> = components
+        .iter()
+        .map(|comp| comp.iter().map(|ix| NodeIndex::new(ix.index())).collect())
+        .collect();
+
+    let mut condensed: StableGraph<Vec<N>, E, Ty, u32> =
+        StableGraph::with_capacity(components_usize.len(), g.edge_count());
+
+    // Build a map from old indices to new ones.
+    let mut node_map = vec![usize::MAX; g.node_count()];
```
IvanIsCoding, line 248: Prefer using `None` instead of `usize::MAX` to represent missing objects

src/connectivity/mod.rs
```diff
@@ -192,6 +194,161 @@ pub fn is_strongly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
+        }
+    }
+
+    // Consume nodes and edges of the old graph and insert them into the new one.
+    let (nodes, edges) = g.into_nodes_edges();
+    for (nix, node) in nodes.into_iter().enumerate() {
+        let idx = node_map.get(nix).copied().unwrap_or(usize::MAX);
+        if idx != usize::MAX {
+            condensed[NodeIndex::new(idx)].push(node.weight);
+        }
```
IvanIsCoding, line 262: This becomes
```suggestion
        if let Some(idx) = node_map.get(nix).copied()  {
            condensed[NodeIndex::new(idx)].push(node.weight);
        }
```

src/connectivity/mod.rs
```diff
@@ -192,6 +194,161 @@ pub fn is_strongly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
+    }
+    for edge in edges {
+        let source = node_map
+            .get(edge.source().index())
+            .copied()
+            .unwrap_or(usize::MAX);
```
IvanIsCoding, line 268: Same comment about prefering `Option<usize>` instead of using `usize::MAX` as a replacement for `None`

src/connectivity/mod.rs
```diff
@@ -192,6 +194,161 @@ pub fn is_strongly_connected(graph: &digraph::PyDiGraph) -> PyResult<bool> {
...
+        if make_acyclic && Ty::is_directed() {
+            if source != target {
+                condensed.update_edge(source, target, edge.weight);
+            }
+        } else {
+            condensed.add_edge(source, target, edge.weight);
+        }
+    }
+
+    let mapped = condensed.map(
```
IvanIsCoding, line 287: I'd say focus on returning Rust structs in this function, leave dealing with Python to `digraph_condensation` and `graph_condensation`. It's ok to loop over results and focus on building the `PyGraph`/`PyDiGraph` to return there.

IvanIsCoding (commented):
Some comments. Overall this works but there are some things that can be improved.

The tests look good though. You can keep them as they are.

src/connectivity/mod.rs
kazuki0824: > We generally have the split at `__init__.py` with a Python function with `@_rustworkx_dispatch` and then two Rust functions with `digraph_*` and `graph_*`:
> 
> https://github.com/Qiskit/rustworkx/blob/217d58ded7b457c30d409f594b720c5be637778e/rustworkx/__init__.py#L148
> 
> I'd rather keep the existing pattern, of course this one works but consistency in the library is important

@IvanIsCoding I got it, and where do we place two Rust functions with `digraph_*` and `graph_*`?

IvanIsCoding: The  Python function goes here: https://github.com/Qiskit/rustworkx/blob/30f29079eeaea6aefa1eb1485d496230173cb1b8/rustworkx/__init__.py#L852

It’s in `__init__.py`, there’s not a lot of code there just the doc string essentially.

The Rust code can stay where it is right now

kazuki0824: I just tried to address this issue following other code.

IvanIsCoding (approved):
LGTM, thanks for addressing the comments

PR #1348, opened 2024-12-30
Allow PyGraph and PyDiGraph to be annotated as generic classes at runtime

### LINKED ISSUE #1345: Add `__class_getitem__` to generic classes
I noticed that despite the `PyGraph`, `PyDiGraph` and `PyDAG` classes being typed as generic over node payload and edge payload type the classes cannot be typed in runtime because the do not implement the [`__class_getitem__`](https://docs.python.org/3/reference/datamodel.html#object.__class_getitem__) class method.

This means that using them in annotations requires one of the following: (1)  use forward reference (i.e. put them quotation marks); (2) be enclosed in an `if TYPE_CHECKING:` block; (3) have `from __future__ import annotations` used in the module; (4) used only in `.pyi` files. (5) use the `type` statement in python>=3.12 instead of `TypeAlias` (which does lazy evaluation).

These are a bit of a hassle but not too troubling, but this has another implication: these type annotations cannot be inspected at runtime.
Runtime inspection is something that packages like `pydantic`, `attrs` do (even the standard library `dataclasses` module does a bit of it).

I suggest adding a `__class_getitem__` classmethod for these classes. In python its easy to do (although I'm not suggesting implementing this in python):

```python
from types import GenericAlias

class PyGraph:
    @classmethod
    def __class_getitem__(cls, key, /):
        return GenericAlias(cls, key)
    ...
```

This should probably be implemented in rust via PyO3. The `GenericAlias` class has a python-stable ABI counterpart [`Py_GenericAlias`](https://docs.python.org/3/c-api/typehints.html) and I assume that it's supported by PyO3 (but haven't checked).

For reference you can look at `numpy` C-implementation, where they do minimal check that the `key` is either a tuple of  length 1 or 2 or any other python object (the `Py_GenericAlias` converts it to a one-tuple in that case): https://github.com/numpy/numpy/blob/ebbd9442e9ccd2508979d58354d16e3eeca96d9c/numpy/_core/src/multiarray/methods.c#L2805

If you do add the `__class_getitem__` method, then its probably also best to add its signature to the `pyi` files:

```python
class PyGraph(Generic[_S, _T]):
    def __class_getitem__(cls, item: Any, /) -> GenericAlias: ...
    ...
```

IvanIsCoding:
Fixes #1345. It should address the core aspects for `PyGraph` and `PyDiGraph`. However, custom return types will come in a follow up PR

This makes the following work without requiring `from __future__ import annotations`:

```python
import rustworkx as rx

graph: rx.PyDiGraph[int, int] = rx.PyDiGraph()
```

~~Also, this is worth highlighting: I refrained from adding `__class_getitem__` to the type stubs as in the stubs that method is inherited from `Generic[_S, _T]`~~ I ended up adding the method to the stubs as `mypy` stub tests disagreed with my reasoning


releasenotes/notes/improve-pep560-support-66f059006db8e3b2.yaml
```diff
@@ -0,0 +1,21 @@
+---
+features:
+  - |
+    The :class:`PyDiGraph() <rustworkx.PyDiGraph>` and :class:`PyGraph() <rustworkx.PyGraph>`
+    classes now have better support for
+    `PEP 560 <https://peps.python.org/pep-0560/>`__. Building off of the previous
+    releases which introduced type annotations, the following code snippet is now valid:
+
+    .. jupyter-execute::
```
mtreinish, line 9: Does this need to be executed? It seems like a static code-block since we just need syntax highlighting

mtreinish (approved):
One quick question about the release note. But otherwise this lgtm it's a pretty straightforward feature and one I could see people using in typing contexts.

releasenotes/notes/improve-pep560-support-66f059006db8e3b2.yaml
IvanIsCoding: I have no preference either is fine. But I will say the advantage of executing is that it tests the code path.

 If you execute that with 0.15.1 you’ll get a runtime error!

mtreinish: I'm fine leaving it. We're not super worried about our docs build time right now. If we were though this is the kind of thing we can change.

PR #1349, opened 2025-01-01
Allow some custom return types to be annotated as generic classes at runtime 

IvanIsCoding:
Related to #1345 
Follow up of #1348 

True diff: https://github.com/IvanIsCoding/rustworkx/compare/class-get-item...IvanIsCoding:rustworkx:pep563-custom-return-types

After #1348 is merged this should be easier to review


src/iterators.rs
```diff
@@ -618,6 +621,24 @@ macro_rules! custom_vec_iter_impl {
...
             }
 
+            #[classmethod]
+            #[pyo3(signature = (key, /))]
+            pub fn __class_getitem__(
+                cls: &Bound<'_, pyo3::types::PyType>,
+                key: &Bound<'_, PyAny>,
+            ) -> PyResult<PyObject> {
+                let class_string = stringify!($name);
+                const ALLOWED_GENERIC_CLASSES: [&str; 3] =
```
IvanIsCoding, line 631: Out of all changes this is the one I am least confident about because it feels like a hack

IvanIsCoding:
Putting this on-hold again as it could profit from the `multiple-pymethods` feature from PyO3

PR #1353, opened 2025-01-05
Fix issues with `rustworkx.visit` annotations

### LINKED ISSUE #1352: Typing: `visit` not re-exported and partial/missing annotations
### Information

I found two issues will using `rustworkx.visit`:

1) re-export of `visit` modules currently doesn't work for mypy for some reason
2) there are several partial & missing  type annotations remaining in rustworkx.

More info:

1. The following python code doesn't pass mypy:

    ```python
    from __future__ import annotations
    import rustworkx as rx
    from typing import override

    rx.visit  # error: Module "rustworkx" does not explicitly export attribute "visit"  [attr-defined]

    # error: Class cannot subclass "BFSVisitor" (has type "Any")  [misc]
    # error: Name "rx.visit.BFSVisitor" is not defined  [name-defined]
    class MyVistor(rx.visit.BFSVisitor):
        # error: Method "discover_vertex" is marked as an override, but no base method was found with this name  [misc]
        @override
        def discover_vertex(self, v: int) -> None:
            return

    ```

    To solve this I suggest replacing in `rustworkx/__init__.pyi` the line `import rustworkx.visit as visit` with `from . import visit as visit`. This seems to fix the re-export locally for me. After the fix I only get the expected  `error: Missing type parameters for generic type "BFSVisitor"  [type-arg]` from running mypy on the script above (this type error is unavoidable until https://github.com/Qiskit/rustworkx/pull/1349 is addressed).

2. Looking at `visit.pyi` I noticed that most of the methods do not have return type (should be `-> None`). This raised a flag for me since I expected `rustworkx`'s tests to catch partial typing. It seems that `stubtest` does not test for partial type hints - I tried running it and it did complain about several `__all__` attributes missing in stubs but nothing else (`rustworkx.generators.__all__`, `rustworkx.rustworkx.__all__`, `rustworkx.visualization.graphviz.__all__` and `rustworkx.visualization.matplotlib.__all__`). 

    However, if I run `mypy --strict -p rustworkx` I get a bunch of missing generic's parameters and some other type annotations. Either this or `pyright --verifytypes rustworkx` catches also a lot of errors.  I suggest adding one of these or both as a test(s) to `rustworkx` (and fixing the errors they emit).

- **Python version**: 3.13
- **Operating system**: ubuntu 24

1) re-export of `visit` modules currently doesn't work for mypy for some reason
2) there are several partial & missing type annotations remaining in rustworkx.

See above.

---
If my suggestions above sound good to you, I'm willing to open a PR and try implementing them.

---
PS, I really am using `BFSVisitor`, so this isn't just a continuation of my previous issue regarding generics.
Thanks for responding and addressing, all my type-related issues so far. I know it might be petty of me since it doesn't change the runtime behavior by much, but I love using rustworkx and I love having my code type-checked.

IvanIsCoding:
Closes #1352


barakatzir:
Why not combine the `visit.pyi` into `visit.py`?

IvanIsCoding:
> Why not combine the `visit.pyi` into `visit.py`?

`visit.py` gets ignored in my experience. There is another issue where someone reported the visit annotations as missing when it was only  `visit.py`.

mtreinish (approved):

PR #1358, opened 2025-01-08
Make type annotations for NumPy arrays more specific

IvanIsCoding:
Related to #1352

This is trying to address some complaints from `pyright` infering the NumPy array as `Unknown`.


barakatzir:
I don't know if you want yo add it to this PR, but pyright also complains about `nunpy.dtype` needing parameter annotations.

barakatzir:
> > I don't know if you want yo add it to this PR, but pyright also complains about `numpy.dtype` needing parameter annotations.
> 
> I annotated it with `np.dtype[np.generic]` but to be honest I find that par counter-intuitive.

You can set it to `np.dtype[Any]` which matches the new return type `npt.NDArray[Any]`. It sounds more intuitive to me.

rustworkx/rustworkx.pyi
```diff
@@ -1079,7 +1080,9 @@ class _RustworkxCustomVecIter(Generic[_T_co], Sequence[_T_co], ABC):
     def __len__(self) -> int: ...
     def __ne__(self, other: object) -> bool: ...
     def __setstate__(self, state: Sequence[_T_co]) -> None: ...
-    def __array__(self, dtype: np.dtype | None = ..., copy: bool | None = ...) -> np.ndarray: ...
+    def __array__(
+        self, dtype: np.dtype[Any] | None = ..., copy: bool | None = ...
+    ) -> npt.NDArray[Any]: ...
```
mtreinish, line 1085: Do we want to use a generic/typevar here? Since the `Any` for the dtype argument needs to match the dtype inside the returned array?

mtreinish (commented):
Overall this LGTM, just one small inline question about the hint for `__array__`

rustworkx/rustworkx.pyi
IvanIsCoding: For the record we ignore the `dtype` at runtime. However, I cannot annotate the ndarray with `_T` as that would be incorrect. There is a mapping from `_T` to numpy types that we’d need to manually add and that is a lot of work.

Also, this method exists to support `np.array` calls no one should call it directly. I think NumPy type stubs will mask this call anyway

mtreinish (approved):

PR #1359, opened 2025-01-09
Silence rust clippy lint enabled after the 1.84 release

IvanIsCoding:
[In short the new clippy lint suggests us to replace `map_or` with `is_none_or`.](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_map_or)

But `is_none_or` is only available for Rust 1.82 or newer (https://doc.rust-lang.org/std/option/enum.Option.html#method.is_none_or). So we can revisit this but at the time it is best to ignore

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,6 +71,12 @@ version = "^0.11"
 default-features = false
 features = ["multi_thread"]
 
+[features]
+default = []
+# TODO: remove this once PyO3 is updated to 0.23. Currently, this is a bug in PyO3 0.22
+# that leaks some of their features in a public macro. However, this was removed in 0.23.
+gil-refs = ["pyo3/gil-refs"]
+
 [profile.release]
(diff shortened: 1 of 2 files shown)


IvanIsCoding:
There are still some failures I need to investigate

Cargo.toml
```diff
@@ -71,6 +71,12 @@ version = "^0.11"
 default-features = false
 features = ["multi_thread"]
 
+[features]
```
IvanIsCoding, line 74: This is ugly but it is the only way to make PyO3 0.22 work with Clippy. I will not bother opening a bug as the latest version already has a fix

mtreinish: Yeah I think this is fine as a temporary fix I had to do the same thing on qiskit recently too: https://github.com/Qiskit/qiskit/pull/13645

Issue #1361, opened 2025-01-13
python BFS and DFS search documentation improvements


barakatzir:
### What is the expected enhancement?

I find the documentations for the python visit functions (`dfs_search` etc.) confusing, and I think all users can benefit from some additions to them.
It might be that my sparse prior knowledge in graph algorithms, but I found the documentation confusing with all the colors (WHITE, GRAY, BLACK) mentioned without context to the BFSVisitor events, or the term "forward or cross edge" which I didn't know.
In the end, I read the rust implementation to understand the algorithm. The code is clearly well thought out and it would be an injustice to keep not let it shine in the documentation as well.

I suggest the following changes:
1.  In docstring, mentioning the visitor events in the pseudo-code in the docs, so it's clear which event occurs when. See suggestions below.
2.  Fix `rustworkx.dfs_search` docstring: mention that the `graph` can be a `PyDiGraph` as well.
3. Changing the docstring's reported type-annotation for `source` from `List[int]` to `Sequence[int] | None` with default `None`.
4. Changing the type annotations (in `rustworkx.pyi`) for the different visit functions so that `visitor=None` is not accepted by static type checkers. This can be achieved with overloads:
    ```python
    @overload
    def digraph_bfs_search(
        graph: PyDiGraph,
        *,
        visitor: _BFSVisitor,
    ) -> None: ...
    @overload
    def digraph_bfs_search(
        graph: PyDiGraph,
        source: Sequence[int] | None,
        visitor: _BFSVisitor,
    ) -> None: ...
    ```
6. In docstring, replace the "Graph can not be mutated while traversing." to "Trying to mutate the graph while traversing raises a `RuntimeError`.". This way the user can know which exception type he should catch in such cases.
7. Documenting that raising `PruneSearch` in a `finish_vertex` method raises an exception. I checked and this behavior is documented in `rustworkx-core` where the function call panics in such cases.
8. (This is the only actual suggested change to runtime behavior) I suggest not raising a `PanicException` if `finish_vertex` raises `PrunceSearch`. This exception isn't caught by `except Exception` and user cannot easily narrow it with `except PanicException` since `PanicException` is in `pyo3_runtime` module, which is not site-packages usually and cannot be easily imported.

I only looked at the `bfs_search` and `dfs_search` which I use, but I suspect `dijkstra_search` also could benefit from similar changes.
If you agree, I can open a PR with the agreed upon changes and also look into the Dijkstra case.

[...]

PR #1362, opened 2025-01-15
Improvements for rustworkx.visit annotations

IvanIsCoding:
Related to #1352 and #1361 

This is kind of a mix of using `TypeVar`'s default argument ~~with some feedback to mitigate the issue introduced in #787 that made us realize that the order of the arguments in the signature was wrong~~

### DIFF
--- a/rustworkx/__init__.py
+++ b/rustworkx/__init__.py
@@ -1490,7 +1490,7 @@ def cartesian_product(
 
 
 @_rustworkx_dispatch
-def bfs_search(graph, source, visitor):
+def bfs_search(graph, source=None, visitor=None):
     """Breadth-first traversal of a directed/undirected graph.
 
     The pseudo-code for the BFS algorithm is listed below, with the annotated
@@ -1564,7 +1564,7 @@ def bfs_search(graph, source, visitor):
 
 
 @_rustworkx_dispatch
-def dfs_search(graph, source, visitor):
+def dfs_search(graph, source=None, visitor=None):
     """Depth-first traversal of a directed/undirected graph.
 
     The pseudo-code for the DFS algorithm is listed below, with the annotated
@@ -1634,7 +1634,7 @@ def dfs_search(graph, source, visitor):
 
 
 @_rustworkx_dispatch
-def dijkstra_search(graph, source, weight_fn, visitor):
+def dijkstra_search(graph, source=None, weight_fn=None, visitor=None):
     """Dijkstra traversal of a graph.
 
     The pseudo-code for the Dijkstra algorithm is listed below, with the annotated
--- a/rustworkx/__init__.py
+++ b/rustworkx/__init__.py
@@ -1490,7 +1490,7 @@ def cartesian_product(
 
 
 @_rustworkx_dispatch
-def bfs_search(graph, source=None, visitor=None):
+def bfs_search(graph, source, visitor):
     """Breadth-first traversal of a directed/undirected graph.
 
     The pseudo-code for the BFS algorithm is listed below, with the annotated
@@ -1564,7 +1564,7 @@ def bfs_search(graph, source=None, visitor=None):
 
 
 @_rustworkx_dispatch
-def dfs_search(graph, source=None, visitor=None):
+def dfs_search(graph, source, visitor):
     """Depth-first traversal of a directed/undirected graph.
 
     The pseudo-code for the DFS algorithm is listed below, with the annotated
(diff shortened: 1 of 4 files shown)


rustworkx/rustworkx.pyi
```diff
@@ -42,6 +39,11 @@ if sys.version_info >= (3, 13):
 else:
     from typing_extensions import TypeVar
 
+if sys.version_info >= (3, 11):
```
IvanIsCoding, line 42: This is not strictly necessary but I noticed we could refactor this to avoid depending on `typing_extension` for 3.13

rustworkx/rustworkx.pyi
```diff
@@ -974,41 +976,41 @@ def graph_transitivity(graph: PyGraph, /) -> float: ...
...
     graph: PyDiGraph,
     source: Sequence[int] | None = ...,
-    visitor: _BFSVisitor | None = ...,
+    visitor: _BFSVisitor = ...,
 ) -> None: ...
 def graph_bfs_search(
     graph: PyGraph,
     source: Sequence[int] | None = ...,
-    visitor: _BFSVisitor | None = ...,
+    visitor: _BFSVisitor = ...,
```
IvanIsCoding, line 991: This is consistent with the behavior for the universal functions: https://github.com/Qiskit/rustworkx/blob/f182dc232cf972526e381cdb17536e411ad27616/rustworkx/__init__.pyi#L587

This was an artifact of #787 that we translated literally into the typing stubs. In practice this is not a breaking change because `None` errors at runtime anyway.

rustworkx/visit.pyi
```diff
@@ -11,10 +11,17 @@
...
+if sys.version_info >= (3, 13):
+    from typing import TypeVar
+else:
+    from typing_extensions import TypeVar
+
 class StopSearch(Exception): ...
 class PruneSearch(Exception): ...
 
-_T = TypeVar("_T")
+_T = TypeVar("_T", default=Any)
```
IvanIsCoding, line 24: This silences pyright in strict mode like reported in #1352

rustworkx/rustworkx.pyi
```diff
@@ -974,41 +976,99 @@ def graph_transitivity(graph: PyGraph, /) -> float: ...
...
-_DijkstraVisitor = TypeVar("_DijkstraVisitor", bound=DijkstraVisitor)
+_BFSVisitor = TypeVar("_BFSVisitor", bound=BFSVisitor, default=BFSVisitor)
+_DFSVisitor = TypeVar("_DFSVisitor", bound=DFSVisitor, default=DFSVisitor)
+_DijkstraVisitor = TypeVar("_DijkstraVisitor", bound=DijkstraVisitor, default=DijkstraVisitor)
 
+@overload
 def digraph_bfs_search(
     graph: PyDiGraph,
-    source: Sequence[int] | None = ...,
-    visitor: _BFSVisitor | None = ...,
```
IvanIsCoding, line 984: This is consistent with the behavior for the universal functions: https://github.com/Qiskit/rustworkx/blob/f182dc232cf972526e381cdb17536e411ad27616/rustworkx/__init__.pyi#L587

This was an artifact of #787 that we translated literally into the typing stubs. In practice this is not a breaking change because `None` errors at runtime anyway.

rustworkx/rustworkx.pyi
```diff
@@ -974,41 +976,99 @@ def graph_transitivity(graph: PyGraph, /) -> float: ...
...
+    source: Sequence[int] | None,
+    visitor: _BFSVisitor,
 ) -> None: ...
+@overload
```
IvanIsCoding, line 989: This overload is to represent the keyword only case

rustworkx/rustworkx.pyi
```diff
@@ -974,41 +976,99 @@ def graph_transitivity(graph: PyGraph, /) -> float: ...
...
+    source: Sequence[int] | None,
+    visitor: _DFSVisitor,
 ) -> None: ...
+@overload
+def graph_dfs_search(
+    graph: PyGraph,
+    *,
+    visitor: _DFSVisitor,
+) -> None: ...
+@overload
```
IvanIsCoding, line 1031: For Dijkstra we have the unfortunate case of three overloads

rustworkx/__init__.py
```diff
@@ -1490,7 +1490,7 @@ def cartesian_product(
 
 
 @_rustworkx_dispatch
-def bfs_search(graph, source, visitor):
+def bfs_search(graph, source=None, visitor=None):
```
IvanIsCoding, line 1493: Because of the dispatch I think this has no impact but it looks like we had `source` as a required argument in the universal functions by accident?

IvanIsCoding:
Overloading is harder than I thought. https://github.com/Qiskit/rustworkx/actions/runs/12780463862/job/35626988746 had quite a lot of errors.

Let's try to merge a smaller PR first with some obvious, correct improvements. I will study more on the overloads side to try to fix the issues in a separate PR.

mtreinish (approved):

Issue #1365, opened 2025-01-17
Add Local Complementation


S-Erik:
### What is the expected enhancement?
Add a function which performs the local complementation of a node applied to a graph. Local complementation is often used when working with graph states where each node corresponds to a qubit and each edge between two qubits represents a CZ gate ([wikipedia page](https://en.wikipedia.org/wiki/Graph_state)).

Figure 1 and the text around equation (2) in [this paper](https://arxiv.org/abs/1910.03969) describes the local complementation of a node applied to a graph.
Adapted from the above mentioned paper:
Local complementation of a node n applied to a graph acts to complement the neighbourhood of the node n. That is, in the neighbourhood of n, it removes edges if they are present, and adds any edges that are missing.

Local complementation is only defined on indirected graphs. It is not defined on graphs that have self-loops or have multiple edges between nodes.

I will immediately add a pull request for this feature.

PR #1366, opened 2025-01-17
Add Local Complementation

S-Erik:
See #1365.

Any comments on the tests, documentation and the error handling are welcome.


tests/graph/test_local_complement.py
```diff
@@ -0,0 +1,71 @@
...
+import unittest
+
+import rustworkx
+
+
+class TestLocalComplement(unittest.TestCase):
+    def test_clique(self):
+        N = 5
+        graph = rustworkx.PyGraph(multigraph=False)
+        graph.extend_from_edge_list([(i, j) for i in range(N) for j in range(N) if i < j])
```
IvanIsCoding, line 22: We actually have a complete graph generator! https://www.rustworkx.org/apiref/rustworkx.generators.complete_graph.html

tests/graph/test_local_complement.py
```diff
@@ -0,0 +1,71 @@
...
+            self.assertTrue(
+                rustworkx.is_isomorphic(
+                    expected_graph,
+                    complement_graph,
+                )
+            )
+
+    def test_empty(self):
+        N = 5
+        graph = rustworkx.PyGraph(multigraph=False)
```
IvanIsCoding, line 39: Likewise, we have an empty graph generator https://www.rustworkx.org/apiref/rustworkx.generators.complete_graph.html

tests/graph/test_local_complement.py
```diff
@@ -0,0 +1,71 @@
...
+        graph.add_nodes_from([i for i in range(N)])
+
+        expected_graph = rustworkx.PyGraph(multigraph=False)
```
IvanIsCoding, line 42: ditto

src/connectivity/mod.rs
```diff
@@ -561,6 +561,86 @@ pub fn digraph_complement(py: Python, graph: &digraph::PyDiGraph) -> PyResult<di
...
+    }
+
+    // Local complementation
+    let node_neighbors: HashSet<NodeIndex> = graph.graph.neighbors(node).collect();
+    let node_neighbors_vec: Vec<&NodeIndex> = node_neighbors.iter().collect();
+    for (i, neighbor_i) in node_neighbors_vec.iter().enumerate() {
+        // Ensure checking pairs of neighbors is only done once
+        let (_nodes_head, nodes_tail) = node_neighbors_vec.split_at(i + 1);
+        for neighbor_j in nodes_tail.iter() {
+            if complement_graph.has_edge(neighbor_i.index(), neighbor_j.index()) {
```
IvanIsCoding, line 607: Use a `match` statement with the result of `remove_edge`. For the `Some(_)` case you'd ignore the value and for the `None` case you have the code of your current `else` branch.

src/connectivity/mod.rs
```diff
@@ -561,6 +561,86 @@ pub fn digraph_complement(py: Python, graph: &digraph::PyDiGraph) -> PyResult<di
...
+            "The input index for 'node' is not a valid node index",
+        ));
+    }
+
+    // Local complementation
+    let node_neighbors: HashSet<NodeIndex> = graph.graph.neighbors(node).collect();
+    let node_neighbors_vec: Vec<&NodeIndex> = node_neighbors.iter().collect();
+    for (i, neighbor_i) in node_neighbors_vec.iter().enumerate() {
+        // Ensure checking pairs of neighbors is only done once
+        let (_nodes_head, nodes_tail) = node_neighbors_vec.split_at(i + 1);
```
IvanIsCoding, line 605: If you are not going to use the first variable you can just assign it to `_`

tests/graph/test_local_complement.py
```diff
@@ -0,0 +1,71 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import unittest
+
+import rustworkx
+
+
+class TestLocalComplement(unittest.TestCase):
```
IvanIsCoding, line 18: Add a test case with a multigraph with `assertRaises` to  addcoverage for the error section

src/connectivity/mod.rs
```diff
@@ -561,6 +561,86 @@ pub fn digraph_complement(py: Python, graph: &digraph::PyDiGraph) -> PyResult<di
...
+
+    let node = NodeIndex::new(node);
+    if !graph.graph.contains_node(node) {
+        return Err(InvalidNode::new_err(
+            "The input index for 'node' is not a valid node index",
+        ));
+    }
+
+    // Local complementation
+    let node_neighbors: HashSet<NodeIndex> = graph.graph.neighbors(node).collect();
```
IvanIsCoding, line 601: This is one of the rare cases where I'd recommend using https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html. It is slightly slower but it will make your graph indices deterministic. Looping over a `HashSet` yields a randomized order so the edge indices would change on every run

src/lib.rs
```diff
@@ -585,6 +585,8 @@ fn rustworkx(py: Python<'_>, m: &Bound<PyModule>) -> PyResult<()> {
     m.add_wrapped(wrap_pyfunction!(digraph_core_number))?;
     m.add_wrapped(wrap_pyfunction!(graph_complement))?;
     m.add_wrapped(wrap_pyfunction!(digraph_complement))?;
+    m.add_wrapped(wrap_pyfunction!(graph_local_complement))?;
+    m.add_wrapped(wrap_pyfunction!(digraph_local_complement))?;
```
IvanIsCoding, line 589: Idem, you'll need to remove this

IvanIsCoding (commented):
Thanks for making your first contribution! It is very welcome, this seems to be a more niche function and that is fine. `rustworkx` has plenty of methods that are more Quantum Computing centric let's say.

I did leave comments and there are still some TODOs:
* Adding type annotations, the CI is failing because they are missing https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md#type-annotations
* Sign the CLA

Lastly, I can do this for you when I merge the code but feel free to do it yourself if you want:
* Adding release notes https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md#release-notes  
*  Add your method to the documentation somewhere in https://github.com/Qiskit/rustworkx/tree/main/docs/source/api

src/connectivity/mod.rs
S-Erik: Why would a deterministic loop be beneficial here?

src/connectivity/mod.rs
S-Erik: Very nice idea. The removes the call to `has_edge`.

tests/graph/test_local_complement.py
S-Erik: I took inspiration from `tests/graph/test_complement.py` where this is also used. I will change my code to use the generator. One might then also change the tests in test_complement.py to use generators. Should I include this in this pull request?

src/connectivity/mod.rs
IvanIsCoding: This is an old issue but read #347. People get frustrated when you run the same code twice and the output is different

tests/graph/test_local_complement.py
IvanIsCoding: Keep it separate, I think when we wrote the complement function we didn’t have the generator.

IvanIsCoding (approved):
Fantastic, this is ready to be merged!

PR #1369, opened 2025-01-21
Add in and out edge indices functions

ThisuraGallage:
This PR introduces a new method, in_edge_indices, to the rustworkx.PyDiGraph class. The method returns the indices of incoming edges for a specified node.
https://github.com/Qiskit/rustworkx/issues/1360

- [x] I ran rustfmt locally

I'll add the documentation once the function signature is finalized.


rustworkx/rustworkx.pyi
```diff
@@ -1423,6 +1423,7 @@ class PyDiGraph(Generic[_S, _T]):
     def in_edges(self, node: int, /) -> WeightedEdgeList[_T]: ...
     def incident_edge_index_map(self, node: int, /, all_edges: bool = ...) -> EdgeIndexMap: ...
     def incident_edges(self, node: int, /, all_edges: bool = ...) -> EdgeIndices: ...
+    def in_edge_indices(self, node: int, /) -> EdgeIndices: ...
```
IvanIsCoding, line 1426: Please add this method to `PyGraph` as well. It is dumb BUT it doesn't hurt to keep the API consistent.

It will just return all edges for the `PyGraph` case and you should note that in the docstring

src/digraph.rs
```diff
@@ -1842,6 +1842,29 @@ impl PyDiGraph {
...
+    ///
+    /// :returns: A list of the incoming edge indices to a node in the graph
+    /// :rtype: EdgeIndices
+    #[pyo3(text_signature = "(self, node, /)")]
+    pub fn in_edge_indices(&self, node: usize) -> EdgeIndices {
+        let node_index = NodeIndex::new(node);
+        EdgeIndices {
+            edges: self
+                .graph
+                .edges_directed(node_index, petgraph::Direction::Incoming)
```
IvanIsCoding, line 1862:
```suggestion
                .edges_directed(node_index, dir)
```

src/digraph.rs
```diff
@@ -1842,6 +1842,29 @@ impl PyDiGraph {
...
+    ///
+    /// :param int node: The node index to get incoming edges from. If
+    ///     this node index is not present in the graph this method will
+    ///     return an empty list and not error.
+    ///
+    /// :returns: A list of the incoming edge indices to a node in the graph
+    /// :rtype: EdgeIndices
+    #[pyo3(text_signature = "(self, node, /)")]
+    pub fn in_edge_indices(&self, node: usize) -> EdgeIndices {
+        let node_index = NodeIndex::new(node);
```
IvanIsCoding, line 1858:
```suggestion
        let node_index = NodeIndex::new(node);
        let dir = petgraph::Direction::Incoming;
```

This is a nit the other functions just use this style. I didn't write them but I am ok sticking with it

src/digraph.rs
```diff
@@ -1842,6 +1842,29 @@ impl PyDiGraph {
...
+    /// ``node``.
+    ///
+    /// :param int node: The node index to get incoming edges from. If
+    ///     this node index is not present in the graph this method will
+    ///     return an empty list and not error.
+    ///
+    /// :returns: A list of the incoming edge indices to a node in the graph
+    /// :rtype: EdgeIndices
+    #[pyo3(text_signature = "(self, node, /)")]
+    pub fn in_edge_indices(&self, node: usize) -> EdgeIndices {
```
IvanIsCoding, line 1857: Another suggestion is adding `out_edge_indices` which is the same function but with `dir = petgraph::Direction::Outgoing`

tests/digraph/test_edges.py
```diff
@@ -813,6 +813,23 @@ def test_incident_edges_all_edges(self):
         res = graph.incident_edges(node_d, all_edges=True)
         self.assertEqual([2, 1], res)
 
+    def test_in_edge_indices(self):
```
IvanIsCoding, line 816: I know you copied this test case from above but in hindsight we can improve it

tests/digraph/test_edges.py
```diff
@@ -813,6 +813,23 @@ def test_incident_edges_all_edges(self):
...
+        graph = rustworkx.PyDiGraph()
+        node_a = graph.add_node(0)
+        node_b = graph.add_node(1)
+        node_c = graph.add_node("c")
+        node_d = graph.add_node("d")
+        graph.add_edge(node_a, node_c, "edge a")
+        graph.add_edge(node_b, node_d, "edge b")
+        graph.add_edge(node_d, node_c, "edge c")
+        res = graph.in_edge_indices(node_c)
+        self.assertEqual([2, 0], res)
```
IvanIsCoding, line 826:
```suggestion
        edge_ac = graph.add_edge(node_a, node_c, "edge a")
        graph.add_edge(node_b, node_d, "edge b")
        edge_dc = graph.add_edge(node_d, node_c, "edge c")
        res = graph.in_edge_indices(node_c)
        self.assertEqual([edge_dc, edge_ac], res)
```

IvanIsCoding (commented):
The current implementation looks good to me. I left some minor nitpick comments but they are not blocking.

What is going to be more work is adding even more methods...
* `PyGraph` should have the same methods even if the implementation is just calling `incident_edges` for consistency
* `PyDiGraph` and `PyGraph` could use a `out_edge_indices`

So that is 3 more methods which were in the original request but are useful to maintain some coherence accross the API

IvanIsCoding (approved):
LGTM but I will run this through @mtreinish before adding four similar methods for a second opinion

IvanIsCoding:
Also, Matthew kindly reminded me:

Congratulations on snatching commit 1000 🥳 🥳

PR #1371, opened 2025-01-24
Use native aarch64 GitHub Action workers

### LINKED ISSUE #1367: Use Linux arm64 runners for the release process

They are now available for us: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/. They will make the release faster, especially now that we have to use QEMU to build `musl` versions as well.

In short it is https://github.com/Qiskit/qiskit/pull/13682. Mind you, I'd like to do it after the 0.16 because I have tested that the wheels build and "if it ain't broke, don't fix it"

IvanIsCoding:
Closes #1367 

In hindsight we should have used this before the release as the QEMU jobs are really unstable. I might need to backport this to the 0.16 branch and trigger a workflow again.

This is a copy of https://github.com/Qiskit/qiskit/pull/13682. I will test it in my branch and check it works.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -47,7 +47,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-13, macos-14]
+        os: [ubuntu-latest, windows-latest, macos-13, macos-14, ubuntu-24.04-arm]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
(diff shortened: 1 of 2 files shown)


IvanIsCoding:
Test run: https://github.com/IvanIsCoding/rustworkx/actions/runs/12943314687

IvanIsCoding:
> > `WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested`
> 
> Hmm, that makes me think we have a configuration issue that's explicitly setting the image for aarch64 to x86_64. But that's a bit odd, because I think we only set an image explicitly on x86_64.

This will go away when I refactor this PR.

Right now, the aarch64 build is faster than the default build because i686 spends a lot of time compiling NumPy. So there could be a window of time where there could be no source/pre-compiled wheels for new platforms which is chaos.

I will merge `musl` and `manylinux` for aarch64 and put them with the main infrastructure now.

IvanIsCoding:
Simplified test build is running: https://github.com/IvanIsCoding/rustworkx/actions/runs/12971477288

 I think we'll be able to just add aarch64 to the OS matrix. That would simplify our setup quite a lot

mtreinish (commented):
This lgtm, but do you want to promote it to tier 1 and add a test job to our ci test matrix?

IvanIsCoding:
> This lgtm, but do you want to promote it to tier 1 and add a test job to our ci test matrix?

Let's merge this first because the release is kind of broken right now.

I will defer the tiers for later, but I think we need to some cleanups:
* win32 got removed from Qiskit and I think it should be removed from rustworkx
* aarch64 manylinux could be promoted to tier 1
* aarch64 musllinux could be promoted to tier 2 to include tests
* i686 was the slowest step in the release last time, I am inclined not to downgrade the tier as NumPy does not release binaries for it anymore

mtreinish (approved):

mtreinish:
Fwiw in qiskit 2.0 we're planning on dropping all 32 bit platform support. I'm not sure we need to necessarily follow what qiskit does here. We can check the download statistics for different platforms and make the call based on that.

IvanIsCoding:
I think that is fair. Let’s check download statistics.

My main concern is 32-bit builds breaking the build because of NumPy not compiling. Maybe we could downgrade i686/win32 to Tier 4

IvanIsCoding:
@Mergifyio backport stable/0.16

PR #1375, opened 2025-01-29
Enhance traversal docs

barakatzir:
This PR fixes https://github.com/Qiskit/rustworkx/issues/1361.
It add some elaboration to the traversal BFS and DFS functions, mainly with a more concrete pseudo-code.
I also fixed small doc errors in betweenness function's docstrings.
In the issue I also suggested doing a face-lift to the dijkstra_search. I can add this in to this PR in the style of the other docstrings changes I'm proposing If you'd like.

- [v] I ran the `nox -e docs`  and `cargo doc` to see that the result is good.
- [v] I have read the CONTRIBUTING document.

### DIFF
--- a/rustworkx/__init__.py
+++ b/rustworkx/__init__.py
@@ -1503,7 +1503,7 @@ def bfs_search(graph, source, visitor):
         def BFSIterator(G, I, F):
           # color each vertex in WHITE, GRAY, BLACK for undiscovered,
           # discovered and finished, respectively, to track its status
-          color = {}                          
+          color = {}
           for u in G:                         # u is a vertex in G
             color[u] = WHITE                  # color all as undiscovered
           for s in I:                         # s is a vertex in I
@@ -1534,7 +1534,7 @@ def bfs_search(graph, source, visitor):
             color[u] = BLACK
             F.Finish(u)
 
-    If an exception is raised inside the callback method of the 
+    If an exception is raised inside the callback method of the
     :class:`~rustworkx.visit.BFSVisitor` instance, the graph traversal
     will be stopped immediately. You can exploit this to exit early by raising a
     :class:`~rustworkx.visit.StopSearch` exception, in which case the search function
@@ -1586,7 +1586,7 @@ def bfs_search(graph, source, visitor):
         graph.add_edges_from_no_data(
             [(school, home), (school, market), (market, home)]
         )
-        
+
         class DistanceHomeFinder(BFSVisitor):
 
             def __init__(self):
@@ -1636,7 +1636,7 @@ def dfs_search(graph, source, visitor):
     """Iterative depth-first traversal of a directed/undirected graph.
 
     Pseudo-code for the iterative depth-first search algorithm is
-    listed below, including the event points, for which the given 
+    listed below, including the event points, for which the given
     :class:`rustworkx.visit.DFSVisitor` visitor object will be called with
     the appropriate method.
 
@@ -1649,8 +1649,8 @@ def dfs_search(graph, source, visitor):
           time = -1
           for s in I:                         # s is a vertex in I
             time = DFS(G, s, F, color, time)
-        
-        
+
+
         def DFS(G, s, F, color, time):
           if color[u] != WHITE:
             return time
--- a/rustworkx/__init__.py
+++ b/rustworkx/__init__.py
@@ -1500,39 +1500,39 @@ def bfs_search(graph, source, visitor):
 
     ::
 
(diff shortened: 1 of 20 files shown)


IvanIsCoding:
Thanks for submitting this! I will try to review it when I can, but again this is a welcome contribution.

rustworkx/__init__.py
```diff
@@ -1491,40 +1491,63 @@ def cartesian_product(
...
-                      if (color[v] = GRAY)         (u,v) has a gray target
-                          ...
-                      else if (color[v] = BLACK)   (u,v) has a black target
-                          ...
-              end for
-              color[u] := BLACK                    finish vertex u
-          end while
-
-    If an exception is raised inside the callback function, the graph traversal
+        def BFSIterator(G, I, F):
```
IvanIsCoding, line 1503: For Python I think it makes more sense to talk about `BFS` than `BFSIterator`. I'd rewrite it to keep the old signature ` BFS(G, s)`

rustworkx/__init__.py
```diff
@@ -1565,38 +1633,68 @@ def tree_edge(self, edge):
 
 @_rustworkx_dispatch
 def dfs_search(graph, source, visitor):
-    """Depth-first traversal of a directed/undirected graph.
+    """Iterative depth-first traversal of a directed/undirected graph.
```
IvanIsCoding, line 1636: We could have a recursive DFS with something like https://crates.io/crates/recursive these days, so I am not opposed to saying this.

With that in mind, DFS order is implementation dependent and even the order we iterate edges is not promised to be stable so it is what it is.

rustworkx/__init__.py
```diff
@@ -1565,38 +1633,68 @@ def tree_edge(self, edge):
...
-                  all DFS-VISIT(G, v)
-              else if (color[v] = GRAY)         (u,v) is a back edge
-              ...
-             else if (color[v] = BLACK)         (u,v) is a cross or forward edge
-             ...
-          end for
-          color[u] := BLACK                     finish vertex u
-
-    If an exception is raised inside the callback function, the graph traversal
+        def DFSIterator(G, I, F):
```
IvanIsCoding, line 1645: Same comment from BFS

rustworkx/__init__.py
```diff
@@ -1565,38 +1633,68 @@ def tree_edge(self, edge):
...
+            time = DFS(G, s, F, color, time)
+
+
+        def DFS(G, s, F, color, time):
+          if color[u] != WHITE:
+            return time
+          color[s] = GRAY
+          time += 1
+          F.Discover(s, time)
+          Q = [s]                             # LIFO vertex queue
```
IvanIsCoding, line 1660: You can call it a stack

IvanIsCoding (approved):
This looks good to me with the caveat that I think for Python docstrings we should focus on the methods that highlight `BFS`/`DFS` only. I'd drop the iterator details

rustworkx/__init__.py
barakatzir: I copied the `Iterative` prefix from the `rustworkx-core` since it made sense to me to differentiate this method from the DFS algorithm which has a single source vertex.
I didn't know about DFS (the CLRS ref you sent me was helpful!) but my friends who do know (both the algo and of CLRS) were confused a bit by how several source vertices are incorporated.
If the implementation detail of how several source vertices are incorporated into DFS is not guaranteed, the docs should probably mention that as well.

IvanIsCoding: I will double check Georgios’ implementation in the source code and check.

I myself forgot the specifics, either the user passes it as an input or we pick an arbitrary node I think

IvanIsCoding: I think if you rewrite the pseudo code to be similar to the implementation, users will like it. Go for it

barakatzir: I'll try to get to it tomorrow.

barakatzir:
I reverted to the pseudo-code styling of the algorithms. I couldn't simplify the algorithms while staying similar to the implementation, so I thought going with pseudo-code conveys better that this does not follow the implementation.

I also redid the Dijkstra search as it was similarly unclear (to me) and added an example.

I fixed a minor type annotation issue in the universal `dijkstra_search`.

IvanIsCoding (approved):
Thanks for the documentation improvements!

I will merge #1380 with this and then lets check the docs. I revised but there are multiple repetitions of the same docstring (unfortunately, that is how the docs work). So let's see the end result in the website and judge. I have a feeling it will be better

PR #1377, opened 2025-02-02
Bump MSRV to 1.79

IvanIsCoding:
This bumps our MSRV to a newer Rust version and aligns the MSRV of rustworkx with Qiskit. https://github.com/Qiskit/qiskit/pull/13709 was the bump in Qiskit's side.

There is nothing tying us to 1.70 and now that we released a new major version I think it is fair to use newer a `rustc` as the next release doesn't have a specified date yet.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,7 +69,7 @@ jobs:
         ]
         include:
           # Test minimal supported Rust version
-          - rust: 1.70.0
+          - rust: 1.79.0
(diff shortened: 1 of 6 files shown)


IvanIsCoding:
I also profited to update some dependencies, I don't think any of them were tied to 1.70 specifically but I was curious about the MSRV-aware Cargo update introduced in 1.79

mtreinish (approved):

PR #1380, opened 2025-02-03
Bump GH Actions version to fix documentation publishing

IvanIsCoding:
We ignored a warning and now the docs publishing job is broken: https://github.com/Qiskit/rustworkx/actions/runs/13105099930/job/36558547792

I think just bumping the version of the actions will fix


IvanIsCoding:
Abandoning in favor of #1375 with the changes

PR #1388, opened 2025-02-16
Fix `bfs_search` and other search methods panicking with invalid sources

### LINKED ISSUE #1386: Panic in traversal

Currently, setting source to non present indices in `bfs_search` will panic.

Running
```python
import rustworkx as rx
from rustworkx.visit import BFSVisitor

graph = rx.PyGraph()
rx.bfs_search(graph, [100], BFSVisitor())
```

raises:

```console
pyo3_runtime.PanicException: put at index 100 exceeds fixedbitset size 0
```

I suggest:
1. raising a IndexError instead.
2. In `rustworkx-core` either returning an Error or documenting the panic behavior
3. If needed do the same for other traversals (dfs, dijkstra).

IvanIsCoding:
Closes #1386 

I opted to check at the Python level because it seemed simpler to handle the failures before calling `rustworkx-core`.

### DIFF
--- a/releasenotes/notes/fix-search-panic-fd54b704f7ee2101.yaml
+++ b/releasenotes/notes/fix-search-panic-fd54b704f7ee2101.yaml
@@ -1,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixed a panic when passing invalid source nodes to search methods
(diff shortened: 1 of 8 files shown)


mtreinish (approved):
LGTM this is a straightforward fix, thanks for doing this.

PR #1385, opened 2025-02-11
Generalizing `Closeness` centrality to weighted networks using Newman method

FedericoBruzzone:
Close #1384

- [x] I ran rustfmt locally
- [x] I have added the tests to cover my changes.
- [x] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.

PR #593, opened 2022-04-23
Implementation of closeness_centrality

derbuihan:
This PR implements closeness centrality for graphs and digraphs.
It supports boolean option `wf_improved` with the same names and meanings as their networkx counterparts.

https://github.com/Qiskit/retworkx/issues/441

Issue #1387, opened 2025-02-14
Add Python tests for `newman_weighted_closeness_centrality`


IvanIsCoding:
### What is the expected enhancement?

In #1385, we skipped the Python tests for `newman_weighted_closeness_centrality`. Before the 0.17 release, we need to test the function.

Notice that the PR has Rust tests, so basing the tests from the version should be fine. We also need to test the behaviour when `weight_fn` is None.

boonware:
@IvanIsCoding Can you clarify what you mean by "basing the tests from the version should be fine."? I am interesting in taking this issue.

PR #1392, opened 2025-02-28
Add parallelism to `cloness_centrality`

IvanIsCoding:
This PR makes `closeness_centrality` and `newman_weighted_closeness_centrality` parallel for better performance. This is possible because the task is embarrassingly parallel.

We probably missed this in #593 but #1385 reminded me of it.

The parallelism is toggled via the `parallel_threshold` just like other centrality methods.


IvanIsCoding:
Because #1387, I added a test to boost the coverage in the report. I am confident the Rust tests guarantee the correctness so for now that is a stop gap

rustworkx-core/src/centrality.rs
```diff
@@ -1095,42 +1095,50 @@ mod test_katz_centrality {
...
 ///     (0, 4), (1, 2), (2, 3), (3, 4), (1, 4)
 /// ]);
-/// let output = closeness_centrality(&dg, true);
+/// let output = closeness_centrality(&dg, true, 200);
 /// assert_eq!(
 ///     vec![Some(0.), Some(0.), Some(1./4.), Some(1./3.), Some(4./5.)],
 ///     output
 /// );
 /// ```
-pub fn closeness_centrality<G>(graph: G, wf_improved: bool) -> Vec<Option<f64>>
```
mtreinish, line 1104: Do you want to mention in the docstring that this function runs in parallel now?

rustworkx-core/src/centrality.rs
```diff
@@ -1095,42 +1095,50 @@ mod test_katz_centrality {
...
+        + IntoEdgesDirected
+        + std::marker::Sync,
+    G::NodeId: Eq + Hash + Send,
+    G::EdgeId: Eq + Hash + Send,
 {
     let max_index = graph.node_bound();
-    let mut closeness: Vec<Option<f64>> = vec![None; max_index];
-    for node_s in graph.node_identifiers() {
+    let node_indices: Vec<G::NodeId> = graph.node_identifiers().collect();
+    let closeness: Mutex<Vec<Option<f64>>> = Mutex::new(vec![None; max_index]);
```
mtreinish, line 1123: Instead of doing this as an vec wrapped in a mutex could you do this as a collection over the parallel iterator? It looks like each iteration is the node index so I think you should be able to do this with `collect()` and not need the mutex.

mtreinish (commented):
This LGTM, it makes a lot of sense to run this with rayon. Thanks for doing this. I just left a couple of comments inline the main one is I think we can do this without a mutex using a rayon iterator collection into the vec which should be faster..

rustworkx-core/src/centrality.rs
IvanIsCoding: I will try, I don't remeber explicitly if `collect()` on a parallel iterator from `par_iter()` preserves order. So I will need to check. That is one of the only advantages of the the mutex approach.

An `enumerate()` could also work here but then I will need to fight with Rust's type system I think

mtreinish: `par_iter()` does preserve order for `collect()`, so I don't think it should be too difficult.

IvanIsCoding: Indeed: https://github.com/rayon-rs/rayon/issues/669

It is only `par_bridge` that doesn't preserve the order

rustworkx-core/src/centrality.rs
IvanIsCoding: Done

rustworkx-core/src/centrality.rs

releasenotes/notes/parallel-closeness-centrality-5165f62e370181eb.yaml
```diff
@@ -0,0 +1,10 @@
+---
+features:
+  - |
+    The :func:`~.closeness_centrality` method is now parallel for better performance.
+    The parallelism is toggled via the new `parallel_threshold` optional parameter.
+other:
```
mtreinish, line 6: I'd put this in the upgrade section because that's where we typically put breaking api changes. It's something the user may need to update when they upgrade to the next release.

```suggestion
upgrade:
```

releasenotes/notes/parallel-closeness-centrality-5165f62e370181eb.yaml
```diff
@@ -0,0 +1,10 @@
+---
+features:
+  - |
+    The :func:`~.closeness_centrality` method is now parallel for better performance.
```
mtreinish, line 4: Do you want to mention `newman_weighted_closeness_centrality` here too since it's a separate function?

rustworkx/__init__.py
```diff
@@ -1171,6 +1171,10 @@ def closeness_centrality(graph, wf_improved=True):
         :class:`~rustworkx.PyGraph` or :class:`~rustworkx.PyDiGraph`.
     :param bool wf_improved: This is optional; the default is True. If True,
         scale by the fraction of nodes reachable.
+    :param int parallel_threshold: The number of nodes to calculate the
+        the closeness centrality in parallel at if the number of nodes in
+        the graph is less than this value it will run in a single thread. The
+        default value is 50.
```
mtreinish, line 1177: Do you want to mention `RAYON_NUM_THREADS` to control the size of the thread pool here too? We do this in the other parallel functions typically.

mtreinish (commented):
Overall this looks much better to me now, thanks for making the change. I just had a couple of small inline comments about the release notes and the docstring updates. But code-wise I think this is good to go.

rustworkx/__init__.py
IvanIsCoding: Done.

releasenotes/notes/parallel-closeness-centrality-5165f62e370181eb.yaml
IvanIsCoding: Done.

releasenotes/notes/parallel-closeness-centrality-5165f62e370181eb.yaml
IvanIsCoding: In theory `newman_weighted_closeness_centrality` is a function that has not been released publicly yet, it will be first available in 0.17. It already has its own release note, it is like if we implemented it in parallel from the start.

mtreinish (commented):
I think this looks fine now, besides the lint failure. The `RAYON_NUM_THREADS` doesn't look like it's in the Python docs though, we normally document it there too because it's an env variable and works from python too.

mtreinish (approved):

PR #1398, opened 2025-03-20
Fix some typos in docs and test function names

eumiro:
These are just some typos fixes in docs and test function names.


IvanIsCoding (approved):
And this is after we ran an automated typo checker on the codebase. Thanks!

PR #1397, opened 2025-03-15
Consolidate docstrings in generators.rs

eumiro:
Another bunch of docstrings reviewed.

Issue #1390, opened 2025-02-23
Advanced testing for `rustworkx-core`


IvanIsCoding:
### What is the expected enhancement?

Currently, we test `rustworkx` with unit tests in Python and we test `rustworkx-core` with unit tests and doctests in Rust.

Because `rustworkx-core` is written completely in Rust and has a simpler testing setup (no need for tools that understand two languages), we could run some more advanced tests for it:
* [Miri](https://github.com/rust-lang/miri): find undefined behaviour in Rust
* [Mutation Testing](https://github.com/sourcefrog/cargo-mutants): use `cargo-mutants` to find lines of code that are covered by tests but do not cause test failures
* [Fuzzing](https://rust-fuzz.github.io/book/cargo-fuzz.html): use `cargo-fuzz`, most likely to find panics

IvanIsCoding:
Also, Miri is known to cause issues to some dependencies we have in `Cargo.lock` so check https://github.com/Qiskit/qiskit/blob/e4048320db76dcd738a36f5376628855d20e97b3/.github/workflows/miri.yml#L28 to see some patches Qiskit uses.

Krishn1412:
Hey @IvanIsCoding , can I pick this one up and add the tests?

IvanIsCoding:
Congratulations, after #1407 was merged we have our first fuzz target.

Some action items:
* add more fuzzers for popular methods
* add a GithubActions workflows that installed a pinned rust nightly version, C++ toolchain and builds the fuzzer
* Join https://github.com/google/oss-fuzz maybe?

Krishn1412:
should I add more fuzz targets (and work on the other action items you mentioned) or go ahead with building a basic structure for the other types of tests?

IvanIsCoding:
You can focus on adding more targets. I think that is the harder part. I will figure out how to run it

Krishn1412:
Got it!

Krishn1412:
Hey @IvanIsCoding, now that basic fuzzing is done, should now focus on adding other kinds of tests?

IvanIsCoding:
@Krishn1412 I think the next best thing would be property based tests with `quickcheck`.

I will try to find some examples later from other packages.

IvanIsCoding:
[...]

Issue #1316, opened 2024-11-16
Add better Matplotlib tests


IvanIsCoding:
### What is the expected enhancement?

Motivated by #1308 

Currently our tests only catch serious regressions like image generation failing. We don’t assert anything about the image themselves.

I have little idea about the state of testing Matplotlib, but I think we can improve the current setup

Rachel-Eva:
Hi! I’d love to work on this. Could you please suggest if you have a preferred approach for asserting image correctness? Thanks!

IvanIsCoding:
@Rachel-Eva I don’t have a preferred approach. I think part of the task is exploring how other libraries do this.

There’s a balance between a test that is too strict and breaks with each matplotlib update. And a test that catches regressions like we had , we were unfortunately shuffling the colours on the plot.

I assigned the issue to you.

Rachel-Eva:
@IvanIsCoding Hi! I'll start by checking how libraries like NetworkX and Seaborn, test their Matplotlib plots, and based on that, I'll try adding some tests for the rustworkx plotting functions.

Does that sound good?

IvanIsCoding:
Sounds good to me. Also, feel free to make this test Linux only. I am expecting some differences across operating systems. Let’s keep it simple, we just want to catch some obvious errors

Rachel-Eva:
Okay, I'm currently using Windows, so I'll set up a VM for Linux and run the tests there.

PR #1399, opened 2025-03-21
Extend docstrings for some graph connectivity methods

eumiro:
This should help new users to discover related methods and understand the method signature with simple examples. I believe there's enough space on the API docs website for more material.

I didn't find the following methods: `number_strongly_connected_components`, `is_strongly_connected`, `number_semi_connected_components`, and `semi_connected_components`. Otherwise I'd like to cross-reference them in the related docstrings.

Also I found it a bit confusing that `connected_components` and `weakly_connected_components` return a list of sets, whereas `strongly_connected_components` returns a list of lists, although the concept is similar.

OT: Did you actually consider using pytest for tests?

### DIFF
--- a/src/connectivity/mod.rs
+++ b/src/connectivity/mod.rs
@@ -100,15 +100,28 @@ pub fn simple_cycles(
     johnson_simple_cycles::PySimpleCycleIter::new(py, graph)
 }
 
-/// Compute the strongly connected components for a directed graph
+/// Find the strongly connected components in a directed graph
 ///
-/// This function is implemented using Kosaraju's algorithm
+/// A strongly connected component (SCC) is a maximal subset of vertices
+/// such that every vertex is reachable from every other vertex
+/// within that subset.
 ///
-/// :param PyDiGraph graph: The input graph to find the strongly connected
-///     components for.
+/// This function is implemented using Kosaraju's algorithm.
 ///
-/// :return: A list of list of node ids for strongly connected components
-/// :rtype: list
+///     >>> G = rx.PyDiGraph()
+///     >>> G.extend_from_edge_list([(0, 1), (1, 2), (2, 0), (3, 4)])
+///     >>> rx.strongly_connected_components(G)
+///     [[4], [3], [0, 1, 2]]
+///
+/// See also [weakly_connected_components].
+///
+/// For undirected graphs, see [connected_components].
+///
+/// :param PyDiGraph graph: The directed graph to find the strongly connected
+///     components in
+///
+/// :return: A list of lists of node indices of strongly connected components
+/// :rtype: list[list[int]]
 #[pyfunction]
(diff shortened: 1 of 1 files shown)


IvanIsCoding (approved):
LGTM. Thanks for adding the examples on the docstrings. I think our documentation suffers from lacks of examples.

It is mostly that when contributing new code, we already need to add Rust code + Python bindings + tests + type annotations + a release note. So by the time authors get to the docstring, they are tired. I like this approach of refining docstrings after the fact.

For missing methods: again, similar fate from #1397. @eumiro if you could send a new PR editing this file docs/source/api/algorithm_functions/connectivity_and_cycles.rst, that would fix the issue. I can't believe we had methods undocumented for so long!

IvanIsCoding:
Re: `pytest`. Matthew has his own test runner that is [stestr](https://github.com/mtreinish/stestr). I myself like the built-in `unittest` module and at work I mostly use [abseil-py](https://github.com/abseil/abseil-py/tree/main) which has a test module built on top of the `unittest` module.

I don't foresee us migrating to `pytest`. I am quite happy with the existing Python test setup. If you want to spend time contributing tests (which is welcome!), check #587 and #1390, they are mostly about Rust. #1316 is a Python-only test issue if you want to try, currently our `matplotlib` tests are close to non-existant.

IvanIsCoding:
Last but not least: API inconsitencies. The methods were added by different people at different times. I don't think when Matthew wrote the earlier versions, he thought we'd have as big of a user base.

We have also a policy of avoiding breaking changes at all costs. For example, we have purposefully kept some API mistakes like having a sub-optimal order of Python arguments because changing the order would break users.

We might fix those things for 1.0 one day, so maybe comment on #1000 some improvements you can think of. Or just open a new issue.

eumiro:
Thanks for the feedback, @IvanIsCoding . I find Rustworkx very useful for some of my use cases and while I approximately know, what I expect from my graphs, I do not always know the names of the methods/concepts, so I have to search around docs a lot (and learn stuff at the same time!). This is how I got the motivation to improve the docs for the `*_connected_*` methods and write some examples.

Am I searching in the wrong place? I cannot find the four missing methods from my top post at all. The only place where I find `number_strongly_connected_components` is in `tests/digraph/test_strongly_connected.py` and even that one uses the verbose `strongly_connected_components` to match against a list. Is it possible they do not exist at all?

I first thought these methods were not commonly used in the graph theory, but while networkx does not have the two `*semi_connected*` methods, it has both `*strongly_connected*` methods. Poor man's solution would be to use `strongly_connected_components` to count the `number_strongly_connected_components` and bool it to `is_strongly_connected`, but I believe there's some smarter shortcut.

And I understand your preference for `unittest`. I find `pytest` code more straightforward and easier to read (and therefore well suitable even for docs examples), but that's just personal taste and no reason to change working system.

PR #1400, opened 2025-03-22
Add is_strongly_connected and number_strongly_connected_components methods

eumiro:
In #1399 I documented the connection methods and realized I could not find the `is_strongly_connected` and `number_strong_connected_components` methods anywhere in the code.

Since `is_weakly_connected` method looks like just checking the result of `weakly_connected_components`, I tried my luck and cloned it for `is_strongly_connected` for code, docs, and tests. The same for `number_strongly_connected_components`.


rustworkx/rustworkx.pyi
```diff
@@ -207,9 +207,11 @@ def graph_bipartite_edge_color(graph: PyGraph, /) -> dict[int, int]: ...
 
 def connected_components(graph: PyGraph, /) -> list[set[int]]: ...
 def is_connected(graph: PyGraph, /) -> bool: ...
+def is_strongly_connected(graph: PyDiGraph, /) -> bool: ...
 def is_weakly_connected(graph: PyDiGraph, /) -> bool: ...
 def is_semi_connected(graph: PyDiGraph, /) -> bool: ...
 def number_connected_components(graph: PyGraph, /) -> int: ...
+def number_strongly_connected_components(graph: PyDiGraph, /) -> bool: ...
```
IvanIsCoding, line 214:
```suggestion
def number_strongly_connected_components(graph: PyDiGraph, /) -> int: ...
```

docs/source/sources.txt
```diff
@@ -143,6 +143,7 @@ rustworkx.is_isomorphic.html
 rustworkx.is_isomorphic_node_match.html
 rustworkx.is_matching.html
 rustworkx.is_maximal_matching.html
+rustworkx.is_strongly_connected.html
```
IvanIsCoding, line 146: Hmmm, I don't think this file does anything. You can notice by the fact we haven't updated it in nearly two years.

It must be from our old doc setup. I will investigate if we can remove it.

IvanIsCoding: Indeed, when we migrate the URL of our docs Matthew added a redirect step: https://github.com/Qiskit/rustworkx/blob/5f10121b0e4e9a8ee03b52bd69393164403fc0b9/docs/source/conf.py#L125

It doesn't hurt to be there, but it is mostly for old methods.

rustworkx/rustworkx.pyi
eumiro: Oh, I cloned the line below (`number_weakly_connected_components`) and it had also `bool`. Will fix both. Thanks.

docs/source/sources.txt
eumiro: If the file is not needed, then just remove it from the repo?

IvanIsCoding: Let's do it in a separate PR. It is no longer needed but there might be one or two places reading it. If we just remove it, we'll need to update the script

rustworkx/rustworkx.pyi
```diff
@@ -207,10 +207,12 @@ def graph_bipartite_edge_color(graph: PyGraph, /) -> dict[int, int]: ...
 
 def connected_components(graph: PyGraph, /) -> list[set[int]]: ...
 def is_connected(graph: PyGraph, /) -> bool: ...
+def is_strongly_connected(graph: PyDiGraph, /) -> bool: ...
 def is_weakly_connected(graph: PyDiGraph, /) -> bool: ...
 def is_semi_connected(graph: PyDiGraph, /) -> bool: ...
 def number_connected_components(graph: PyGraph, /) -> int: ...
-def number_weakly_connected_components(graph: PyDiGraph, /) -> bool: ...
+def number_strongly_connected_components(graph: PyDiGraph, /) -> int: ...
+def number_weakly_connected_components(graph: PyDiGraph, /) -> int: ...
```
IvanIsCoding, line 215: Thanks for fixing the old signature as well

IvanIsCoding (approved):

PR #1401, opened 2025-03-25
Enhance docs for adj/pred/succ methods in PyDiGraph and add find_successor_node_by_edge method

eumiro:
- add examples to all adj/pred/succ methods in PyDiGraph
- add method `find_successor_edge_by_node` that is just just a minimally modified version of `find_predecessor_edge_by_node`.

--- 

- [x] I ran rustfmt locally


src/digraph.rs
```diff
@@ -632,11 +632,29 @@ impl PyDiGraph {
...
+    ///
+    ///     >>> G = rx.PyDiGraph()
+    ///     >>> G.add_nodes_from(["A", "B", "C", "D", "E"])
+    ///     NodeIndices[0, 1, 2, 3, 4]
+    ///     >>> G.extend_from_edge_list([(0, 1), (1, 2), (1, 3), (1, 4)])
+    ///     >>> G.successors(1)  # successors of the 'B' node
+    ///     ['E', 'D', 'C']
+    ///
+    /// To filter the successors by the attributes of the connecting edge,
+    /// see [find_successors_by_edge].
```
IvanIsCoding, line 649: See https://www.rustworkx.org/dev/apiref/rustworkx.strongly_connected_components.html#rustworkx.strongly_connected_components, I don’t think we have been using hyperlinks right. Apologies for noticing it so late.

I think the right syntax starts with ~ like in the release notes. Try copying examples from there

IvanIsCoding (commented):
I will review this later today but I left a comment about hyperlinks

src/digraph.rs
eumiro: Thanks for the hint. I updated the link syntax in this PR. The rest will have to wait.

IvanIsCoding (approved):

IvanIsCoding:
Whatever we did here has the right syntax for hyperlinks: https://www.rustworkx.org/dev/apiref/rustworkx.PyDiGraph.find_successor_node_by_edge.html#rustworkx.PyDiGraph.find_successor_node_by_edge has the right links

PR #1402, opened 2025-03-25
Consolidate string formatting and converting

eumiro:
This is just a small refactoring that consolidates string formatting (f-strings).


IvanIsCoding (approved):
This should be easy to merge. Thanks!

PR #1403, opened 2025-03-28
Advanced Tests for rust repo

Krishn1412:
Starting off with the implementation of cargo-fuzz which is used to find panics.

Implemented the "test_is_connected" test in the fuzz. Working on more.


IvanIsCoding:
I will review this more carefully, but in short:
* https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md should have instructions on how to run the fuzzing
* We should build fuzz targets on CI
* However, we shouldn't run fuzz targets on CI as that can be expensive

Cargo.lock
```diff
@@ -415,29 +459,29 @@ dependencies = [
...
-checksum = "1cd3927b5a78757a0d71aa9dff669f903b1eb64b54142a9bd9f757f8fde65fd7"
+checksum = "94f6cbe86ef3bf18998d9df6e0f3fc1050a8c5efa409bf712e661a4366e010fb"
 dependencies = [
  "once_cell",
  "target-lexicon",
 ]
 
 [[package]]
 name = "pyo3-ffi"
-version = "0.23.4"
```
IvanIsCoding, line 428: It looks like you ran an update while adding the dependencies. Revert this file and  just run something along the lines of `cargo add crate --features feature_foo` instead.

Let's try to keep this PR as small as possible in terms of dependencies added.

Krishn1412:
Hey @IvanIsCoding, I have reverted the changes in dependencies. I could not find instructions on how to run the fuzzing in the Contributing file. Could you please help me with that.

IvanIsCoding:
@Krishn1412 apologies for the confusion. There is no fuzzing section (yet). I am asking you to write a brand new subsection, under the rustworkx-core section, explaining how to run the fuzzing.

rustworkx-core/fuzz/fuzz_targets/is_connected_fuzz.rs
```diff
@@ -0,0 +1,58 @@
...
+    let start_node = nodes[0];
+
+    let mut visit_map = graph.visit_map();
+    let component: HashSet<usize> =
+        connectivity::bfs_undirected(&graph, start_node, &mut visit_map)
+            .into_iter()
+            .map(|x| x.index())
+            .collect();
+
+    assert_eq!(
```
IvanIsCoding, line 53: I am not sure this assertion is correct. At best, it assumes you only generate connected graphs which I am not sure is always the case.

A more reasonable assertion would be that the visit nodes from a DFS are equal to a BFS. We could start with that

CONTRIBUTING.md
```diff
@@ -208,6 +208,44 @@ cargo test
...
+cargo fuzz build
+```
+#### To run a fuzz test (e.g., is_connected_fuzz):
+
+```sh
+cargo fuzz run is_connected_fuzz
+```
+For nightly toolchain:
+```sh
+cargo +nightly fuzz run is_connected_fuzz
```
IvanIsCoding, line 232: Is there a way to run all the fuzz targets? If not, we should point users to Cargo.toml with the list of targets.

IvanIsCoding (commented):
Overall code is looking good, but I left a comment about the graph correctness. Because this is the first fuzz target, maybe let's start with asserting that running a BFS yields the same set of nodes as running a DFS

Krishn1412:
Hey @IvanIsCoding, should I create different PRs for the other tests or build in this one?

IvanIsCoding:
> Hey @IvanIsCoding, should I create different PRs for the other tests or build in this one?

Make different PRs. I will let you know that right now Clippy will complain about our code at head, so my plan is:
* Send a PR to fix clippy
* Merge your PR
* Review your future PRs

Let's keep it at one PR per test. Again, thanks for the support. I hope I can merge this soon.

Krishn1412:
Got it, thank you!

IvanIsCoding:
I will try to resolve the Cargo.lock conflicts sometime tomorrow

IvanIsCoding:
Ok, some updates:
* I verified that `cargo publish --dry-run` doesn't publish the fuzz code to crates.io. That is how it should work. Nice
* I confirmed `cargo build` doesn't build the fuzz target, which is intended because we don't want people installing rustworkx to compile the fuzz unecessarily


Things that are probably ok but I need to setup on my machine. I am getting this error following the instructions:
```
error: linking with `cc` failed: exit status: 1
  |
  = note:  "cc" "-m64" "/tmp/rustcuNTlld/symbols.o" "-Wl,-Bstatic" "-Wl,--whole-archive" "-Wl,<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/lib/librustc-nightly_rt.asan.a" "-Wl,--no-whole-archive" "<1 object files omitted>" "-Wl,--as-needed" "/tmp/rustcuNTlld/{liblibfuzzer_sys-2fb9a77a5fde2004.rlib}.rlib" "<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/lib/{libcompiler_builtins-*}.rlib" "-Wl,-Bdynamic" "-lc" "-lm" "-lrt" "-lpthread" "-lstdc++" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-L" "/tmp/rustcuNTlld/raw-dylibs" "-B<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/bin/gcc-ld" "-fuse-ld=lld" "-Wl,-znostart-stop-gc" "-Wl,--eh-frame-hdr" "-Wl,-z,noexecstack" "-L" "/home/ivan/Projects/rustworkx-dev/target/x86_64-unknown-linux-gnu/release/build/libfuzzer-sys-b91be800a92a6008/out" "-L" "<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "/home/ivan/Projects/rustworkx-dev/target/x86_64-unknown-linux-gnu/release/deps/test_traversal_node_coverage-40ae2202594db3b6" "-pie" "-Wl,-z,relro,-z,now" "-Wl,-O1" "-nodefaultlibs"
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: rust-lld: error: undefined symbol: __sancov_gen_.573
```

It looks like we'll need some sort of C++ toolchain to include this on CI. So let's keep it away from CI right now.

I will update the original issue, but later we'll need to include a new workflow that runs from time to time to try to build the fuzzer. It will need to be separate, as it will use both rustc nightly and C++ which are way more likely to break than our existing setup.

IvanIsCoding (approved):

PR #1404, opened 2025-03-28
docs: add a note that subgraph recreates indexes

alessio-locatelli:
The inability to find the desired index in the subgraph was unexpected. This behavior, if it is a feature rather than a bug, needs to be documented.

### DIFF
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -3003,7 +3003,9 @@ impl PyDiGraph {
     ///     It is worth noting that node and edge weight/data payloads are
     ///     passed by reference so if you update (not replace) an object used
     ///     as the weight in graph or the subgraph it will also be updated in
-    ///     the other.
+    ///     the other. Because the indexes will be recreated, the subgraph will
+    ///     have different indexes, meaning you cannot access nodes using the
+    ///     original graph's indexes.
     /// :rtype: PyGraph
     ///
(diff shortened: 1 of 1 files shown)


IvanIsCoding:
@mtreinish wrote this function originally. I haven't asked him but I can explain a line of reasoning for this behavior.

`StableGraph` behind the scenes if backed by two `Vec` entries in `petgraph` a couple of layers down. This means that for us to preserve the indexes of a graph with `N` nodes, we'd need roughly `O(N)` space.

For subgraphs, if you create many subgraphs of size `k` with `k < N`, this would be a waste of memory and performance. Therefore, Matthew implemented this function rewriting the indexes. That way, the space is `O(k)`.

Let's document that the indices are not preserved, but the behavior is intended.

src/digraph.rs
```diff
@@ -3003,7 +3003,9 @@ impl PyDiGraph {
     ///     It is worth noting that node and edge weight/data payloads are
     ///     passed by reference so if you update (not replace) an object used
     ///     as the weight in graph or the subgraph it will also be updated in
-    ///     the other.
+    ///     the other. Because the indexes will be recreated, the subgraph will
+    ///     have different indexes, meaning you cannot access nodes using the
+    ///     original graph's indexes.
```
IvanIsCoding, line 3008:
```suggestion
    ///     the other. Node and edge the indices will be recreated for the subgraph for compactness.
    ///     Therefore, do not access data using the original graph's indices.
```

IvanIsCoding (approved):

PR #1405, opened 2025-03-28
Enhance docs and tests for pred/succ/anc/desc

eumiro:
I realized the `ancestors` and `descendants` raise an `IndexError` if the provided node is not present in the directed graph. On the other hand `predecessors` and `successors`, which are quite similar methods, just return an empty list.

I do not want to introduce a breaking change, so I just added a test pair to make sure this persists.

Also for the formatting, I discovered the docs website was using [Furo](https://pradyunsg.me/furo/reference/admonitions/) which offers some nice details to distinguish different parts of a method documentation. Before we have lots of docs, we'll have to find some useful and decent style.

### DIFF
--- a/src/digraph.rs
+++ b/src/digraph.rs
@@ -21,6 +21,7 @@ use std::io::prelude::*;
 use std::io::{BufReader, BufWriter};
 use std::str;
 
+use hashbrown::hash_set::Entry;
 use hashbrown::{HashMap, HashSet};
 
 use rustworkx_core::dictmap::*;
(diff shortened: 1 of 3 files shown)


src/digraph.rs
```diff
@@ -705,9 +715,8 @@ impl PyDiGraph {
         let mut predec: Vec<&PyObject> = Vec::new();
         let mut used_indices: HashSet<NodeIndex> = HashSet::new();
         for pred in parents {
-            if !used_indices.contains(&pred) {
+            if used_indices.insert(pred) {
```
IvanIsCoding, line 718: Actually, let's rewrite this to use this API with a `match` statement: https://docs.rs/hashbrown/latest/hashbrown/struct.HashSet.html#method.entry

That way the behavior is clearer. Your code and the existing codes are equivalent, your code saves a lookup but it is not obvious.

IvanIsCoding (approved):
LGTM, thanks for documenting the current behaviors.

I am curious about what `..seealso` is going to look on the docs, but it is a Sphinx directive and it is pretty appropriate for our situation. Our theme is patched but it should support it.

I left a minor Rust comment

src/digraph.rs
eumiro: Changed both predecessors/successors syntax.

Since the usage of this syntax is not consistent across the whole project, I'm going to consolidate it elsewhere.

IvanIsCoding (approved):

eumiro:
> I am curious about what `..seealso` is going to look on the docs, but it is a Sphinx directive and it is pretty appropriate for our situation. Our theme is patched but it should support it.

It looks good, but the link is not rendered, so it's not useful.

PR #1407, opened 2025-03-29
Replace lambdas with direct functions or operator module

eumiro:
This just replaces `lambda`s with direct functions (e.g. `lambda x: float(x)` → `float`) or with `operator` module (e.g. `lambda x, y: x == y` → `op.eq`). Makes a lot of lines actually fit into a single line and hopefully improved the readability.

### DIFF
--- a/rustworkx/__init__.py
+++ b/rustworkx/__init__.py
@@ -236,7 +236,7 @@ def adjacency_matrix(graph, weight_fn=None, default_weight=1.0, null_value=0.0):
 
         to return a weight of 1 for all edges. Also::
 
-            adjacency_matrix(graph, weight_fn: lambda x: float(x))
+            adjacency_matrix(graph, weight_fn: float)
 
         to cast the edge object as a float as the weight. If this is not
         specified a default value (either ``default_weight`` or 1) will be used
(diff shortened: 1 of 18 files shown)


IvanIsCoding:
I will review this later, but it is one fo the things that I find on the edge of performance and being Pythonic.

First thing first, using `float` instead of `lambda x: x` is faster and we should probably document that to users. With that being said, reading `floyd_warshall_numpy(graph, weight_fn=float)`  is not the most readable. It works, `float` is a type and its constructor acts as a function, ok. But we still need to teach users the callback can be anything e.g. `lambda x: x["time
"]` or even a fully fledged function with `def`.

For tests, I'd say in general avoid changing as I think we are happy with the existing coverage. The exception would be isomorphism tests, where I think `op.eq` is actually quite idiomatic.

rustworkx/__init__.py
```diff
@@ -236,7 +236,7 @@ def adjacency_matrix(graph, weight_fn=None, default_weight=1.0, null_value=0.0):
 
         to return a weight of 1 for all edges. Also::
 
-            adjacency_matrix(graph, weight_fn: lambda x: float(x))
+            adjacency_matrix(graph, weight_fn: float)
```
IvanIsCoding, line 239: Both the old example and the new one are wrong by the way, it would be `adjacency_matrix(graph, weight_fn: lambda x:=float(x))`

IvanIsCoding (commented):
I will add to my original comment, maybe we should keep things as is and use [Sphinx hints](https://docutils.sourceforge.io/docs/ref/rst/directives.html#hint) to say: by the way, this is faster

eumiro:
I agree with you. At the end it really does not add any value. I'll close this PR.

PR #1408, opened 2025-03-29
Enhance docs and unpack tuples in tree.rs

eumiro:
Enhance the documentation and add example for the methods `minimum_spanning_edges` and `minimum_spanning_tree`. Also refactor the code to unpack tuples in loops and use verbose variable names instead of accessing numerated items subsequently.


IvanIsCoding (approved):
LGTM

IvanIsCoding:
Our code coverage tool is down, I will merge when it is back up

PR #1409, opened 2025-03-31
Enhance docs in toposort

eumiro:
Format cross-references and make the example code more explicit.

BTW, do you prefer using jupyter-execute or `>>>`-indented docstring-like code?

### DIFF
--- a/src/toposort.rs
+++ b/src/toposort.rs
@@ -34,11 +34,20 @@ enum NodeState {
 
 /// Provides functionality to topologically sort a directed graph.
 ///
+/// A topological sorter is used to arrange the nodes of a directed acyclic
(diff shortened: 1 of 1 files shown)


IvanIsCoding (approved):
With regards to `jupyter-execute`. For code that has plots (i.e. generators, graph layouts) we always want to use the Jupyter option to generate visual output for the user.

For code that needs to run but outputs text, it is 50/50. I talked with Matthew, too many would slow down our documentation. But right now we have too few.

So in general if the method is non-trivial and the outptu is useful,  `jupyter-execute`. But either is fine, I appreciate just the plain markdown in places where we used to have no documentation

PR #1413, opened 2025-04-04
Enhance docstrings for topological sort methods

eumiro:
I realized the `lexicographical_topological_sort` method does not raise `DAGHasCycle` if it gets a graph with cycles. It just ignores the participating nodes and returns the rest.

For instance:

```
G = rx.PyDiGraph()
G.add_nodes_from(["A", "B"])
G.add_edges_from_no_data([(0, 1), (1, 0)])
rx.lexicographical_topological_sort(G, key=str)
```

returns an empty list.

I tried to add a check into the code, but did not succeed. The test in would be:

```
tests/digraph/test_nodes.py (line 611)

    def test_lexicographical_topological_sort_with_cycle(self):
        G = rustworkx.PyDiGraph()
        G.add_nodes_from(("A", "B"))
        G.add_edges_from_no_data(((0, 1), (1, 0)))
        self.assertRaises(rustworkx.DAGHasCycle, rustworkx.lexicographical_topological_sort, G, str)
```


IvanIsCoding (commented):
This is blocked until I merge a fix for clippy lints, which hopefully is soon

src/dag_algo/mod.rs
```diff
@@ -343,41 +343,47 @@ pub fn layers(
...
+/// ordering.  If `reverse` is set to `False`, the edges are treated as if they
+/// point in the opposite direction.
+///
+/// Unlike :func:`~rustworkx.topological_sort`, this function resolves ties
+/// between nodes using the string returned by the `key` argument. The `reverse`
+/// argument only affects the direction of the edges, not the ordering of keys.
+///
+///   >>> G = rx.PyDiGraph()
+///   >>> G.add_nodes_from(["A", "B", "C", "D", "E", "F", "G"])
+///   >>> G.add_edges_from_no_data([(0, 6), (1, 6), (2, 6), (3, 6), (4, 6), (5, 6)])
```
IvanIsCoding, line 363: Nitpicking, but you can unpack this as:
```
a, b, c, d, e, f, g = .add_nodes_from(["A", "B", "C", "D", "E", "F", "G"])
```

And then do e.g. `edges_from_no_data[(a, g)...]`

src/dag_algo/mod.rs
```diff
@@ -413,19 +419,28 @@ pub fn lexicographical_topological_sort(
...
-/// be in the earliest possible generation that they can belong to.
+/// A topological generation is a collection of nodes where all ancestors of a
+/// node are guaranteed to be in a previous generation, and all descendants of a
+/// node are guaranteed to be in a subsequent generation. Nodes are placed in
+/// the earliest possible generation they can belong to.
 ///
-/// :param PyDiGraph graph: The DAG to get the topological generations from
+///   >>> G = rx.PyDiGraph()
+///   >>> G.add_nodes_from([0, 1, 2, 3, 4])
+///   >>> G.add_edges_from_no_data([(0, 1), (0, 2), (1, 3), (2, 3), (3, 4)])
```
IvanIsCoding, line 431: ditto

src/dag_algo/mod.rs
```diff
@@ -468,14 +483,29 @@ pub fn topological_generations(dag: &digraph::PyDiGraph) -> PyResult<Vec<NodeInd
...
-/// :param PyDiGraph graph: The DAG to get the topological sort on
+/// Computes a topological ordering of the nodes in the given directed graph,
+/// ensuring that for every directed edge from node :math:`u` to node :math:`v`,
+/// node :math:`u` appears before node :math:`v` in the resulting sequence. This
+/// is particularly useful in scenarios such as task scheduling and dependency
+/// resolution, where certain tasks must be completed before others.
 ///
+///   >>> G = rx.PyDiGraph()
+///   >>> G.add_nodes_from(["A", "B", "C", "D", "E", "F", "G"])
+///   >>> G.add_edges_from_no_data([(0, 1),(1, 2), (2, 3), (3, 4), (5, 2), (6, 3)])
```
IvanIsCoding, line 497: ditto

IvanIsCoding (commented):
LGTM with a minor caveat

IvanIsCoding:
And about raising an error: that would probably be the correct behaviour. It should be pretty easy to check after the fact as well, if the rustworxk-core lexicographical topological sort did not return all the nodes we'd raise.

src/dag_algo/mod.rs
eumiro: This is true for many of other examples and I'll be happy to update them later.

However, since this method returns a numerical list of indices `NodeIndices[6, 5, 0, 1, 2, 3, 4]`, I find it easier for the reader to match these numbers and the arguments to `add_edges_from_no_data`.

src/dag_algo/mod.rs
eumiro: The same as below. Here even more: perfectly matching numbers only.

src/dag_algo/mod.rs
eumiro: In this case it seems to be useful. Updated, thanks.

eumiro:
> And about raising an error: that would probably be the correct behaviour. It should be pretty easy to check after the fact as well, if the rustworxk-core lexicographical topological sort did not return all the nodes we'd raise.

Actually, it is unfortunately correct, as it is. Because of the optional `initial` argument, we cannot expect to have all nodes of the graph returned. And with `initial`, it will “only include those nodes and any nodes that are dominated by them”, so the rest should not be tested for cycles at all.

IvanIsCoding:
Just FYI, I think the error from the CI is actionable:
```
/home/runner/work/rustworkx/rustworkx/.nox/docs-3/lib/python3.10/site-packages/rustworkx/__init__.py:docstring of rustworkx.topological_generations:16:Unknown interpreted text role "cls".
```

If the fix is trivial, I can try patching it after I merge #1403

IvanIsCoding:
Once https://status.coveralls.io/ is back we should be able to merge this

IvanIsCoding (approved):

PR #1414, opened 2025-04-04
Fix clippy warnings for Rust 1.86

IvanIsCoding:
Mostly about https://rust-lang.github.io/rust-clippy/master/index.html#doc_overindented_list_items

There was also https://rust-lang.github.io/rust-clippy/master/index.html#manual_div_ceil


mtreinish (approved):
Thanks for fixing this. I actually preferred the explicit form of `div_ceil()` in the heavy hex graph generator because I think it's closer to what's in the paper. But at the end of the day that doesn't matter too much.

PR #1417, opened 2025-04-05
Fix `cargo clippy` not covering `rustworkx-core` tests

IvanIsCoding:
Right now, if one runs `cargo clippy --all-targets` inside the `rustwork-core` directory, it will find lints from clippy. Sure, they are for tests, but they are still lints.

To address that, in this PR I:
* Fixed the findings from the lints
* Transitioned to running the lint with `cargo clippy --workspace` which does include the tests
* Updated `CONTRIBUTIGNG.md` and CI to stop having to run commands inside the `rustworkx-core` directory. For most cases, `cargo` provides commands that work with all workspace crates at the root

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,9 +43,9 @@ jobs:
       - name: Check stray release notes
         run: python tools/find_stray_release_notes.py
       - name: rustworkx-core Rust Tests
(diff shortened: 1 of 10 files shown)


mtreinish (approved):
LGTM, this is a nice simplification of both the steps to run things in CI and also the test code cleanups look good.

PR #1418, opened 2025-04-05
Swap `ahash` with `foldhash`

IvanIsCoding:
In short, this is #1294 but works because we raised our MSRV and foldhash lowered their MSRV.

I think using the same hasher between `indexmap` and `hashbrown` is a plus.


mtreinish:
Do you have any performance numbers to compare here? I'm fine with this in principle of keeping us in sync with hashbrown, that's why we started using ahash originally. But, I'm curious what impact this has in practice on our performance here.

IvanIsCoding:
I will try to get a benchmark sometime this week. I will focus mostly on Python, but I will try Rust as well if I can.

I am expecting the change to be neutral, to be honest. As long as we don't have a visible regression there it should be fine.

IvanIsCoding:
Ok, I finally had an idea to benchmark this: https://gist.github.com/IvanIsCoding/58157618e75a0c6aaf699dc5882d7829

In short, I used our Dijkstra's implementation on a path graph and run it multiple times. This way the overhead of priority queues and Python is kind of  fixed.

For a path graph of 100k nodes, our latest released version with ahash took `0.6737 ± 0.0607  seconds`. The version with foldhash took `0.6434 ± 0.0024  seconds`. Mind you, I ran this on a MacBook Air with M2 so it's not scientific.

I think overall the performance is pretty comparable and the bottleneck is on the Python side as always. It should be safe to merge.

mtreinish (approved):
I do wonder if the benchmarks are isolated to just this change or other things we've merged since the 0.16.0 release. But since with 0.17 there is no regression and in aggregate things are slightly faster I'm good with this. The other benefit is it makes our build faster by removing a bunch of dependencies.

IvanIsCoding:
> I do wonder if the benchmarks are isolated to just this change or other things we've merged since the 0.16.0 release. But since with 0.17 there is no regression and in aggregate things are slightly faster I'm good with this. The other benefit is it makes our build faster by removing a bunch of dependencies.

I will try re-running tonight, compiling with the same rustc but with ahash.

I think there will not be a big difference, but you are correct that there could be some deltas since 0.16. Nevertheless at least it is on par with what we already have.

Issue #1416, opened 2025-04-05
Explore using `uv` for documentation and tests


IvanIsCoding:
### What is the expected enhancement?

We have been using ruff since #1006.

[`uv`](https://docs.astral.sh/uv/getting-started/installation/) is a tool by the same developers that manages virtual environments. The most interesting feature for us are [its lock files](https://docs.astral.sh/uv/concepts/projects/sync/).

I think our documentation workflow would benefit from a lockfile, as currently it is our process with the most dependencies.  `nox` has a [`uv` backend](https://nox.thea.codes/en/stable/usage.html#changing-the-sessions-default-backend) that we could integrate to make generating the docs more reliable.

For our tests, in most instances I am satisfied with `nox`. Our unit tests have very little dependencies, so I don't think that workflow would benefit a lot for it. The same applies for our lint workflow, which currently is very simple.

Perhaps we could use `uv` for stub tests as currently they run on a pinned `mypy` version. A lockfile could benefit that use case.

IvanIsCoding:
Also, our docs are not building with Python 3.13. Fortunately, we have 3.10 pinned in CI

kazuki0824:
@IvanIsCoding 
Looks like adding `nox.options.default_venv_backend = "uv|virtualenv"` to noxfile.py might be of help.
https://github.com/astral-sh/uv/issues/6579

kazuki0824:
For example, 
```shellscript
uvx uvx black --check retworkx rustworkx retworkx tests
uvx ruff check

# Adding nox.options.default_venv_backend = "uv|virtualenv"
uvx nox -estubs
uvx nox -etest
```

IvanIsCoding:
> For example,
> 
> ```
> uvx uvx black --check retworkx rustworkx retworkx tests
> uvx ruff check
> 
> # Adding nox.options.default_venv_backend = "uv|virtualenv"
> uvx nox -estubs
> uvx nox -etest
> ```

I think being able to run `uvx nox -etest` would be cool, but overall my biggest concern from `uv` for drafting #1420 is that we need to think more thoroughly about triggering rebuilds (i.e. using `pip install -e` behaviour doesn't work). Sure that is doable, but right now our setup is so mindless it doesn't make a difference. The only pains I have are with the docs workflow.

[...]

PR #1419, opened 2025-04-06
Move metadata from `setup.py` to `pyproject.toml` and build `retworkx` in a separate `setup.py`

IvanIsCoding:
~~I am removing `retworkx`, it has been more than one year since it was deprecated in #1004. This totals 2 years since the project changed names.~~

I am also profiting to embrace `pyproject.toml`  more. This is motivated by #1416, in general the prospect for most projects is to consolidate most data in `pyproject.toml`.

edit: we set this release as the last one depending on `rustworkx`, I had to use a separate `setup.py` file for this. In short, the `retworkx` code now is nested in a folder. We have `setup.py` that reads from our own `pyproject.toml` in the root. The code is now in `retworkx/retworkx` given the new folder setup.


IvanIsCoding:
I confirmed that the generated metadata in the wheel is correct: https://gist.github.com/IvanIsCoding/135b258198742f49111736ab537b2729

I just need to investigate why the coverage test failed and then this should be good to go.

IvanIsCoding:
https://github.com/Qiskit/qiskit/pull/10958, Qiskit did this a while ago. Although for the Rust part we are moving it to `pyproject.toml`, which Qiskit hasn't done yet.

README.md
```diff
@@ -157,17 +157,14 @@ changes reflected in your python environment.
 ### Develop Mode
 
 If you'd like to build rustworkx in debug mode and use an interactive debugger
-while working on a change you can use `python setup.py develop` to build
-and install rustworkx in develop mode. This will build rustworkx without
-optimizations and include debuginfo which can be handy for debugging. Do note
-that installing rustworkx this way will be significantly slower then using
-`pip install` and should only be used for debugging/development.
+while working on a change you can set `SETUPTOOLS_RUST_CARGO_PROFILE="dev"`
```
mtreinish, line 160: Do we want to set a minimum version for setuptools-rust if we document this. Historically this option didn't exist when rustworkx started. Looking at the changelog it was less recent than I thought but still relatively recent in 1.8.0 from Oct. 2023.

IvanIsCoding:
I might hold on the `retworkx` removal and have something along these lines then:
```
if not retworkx_build:
    setup()
else:
    setup(name="retworkx", ...)
```

Thay way, we can give a final notice to `retworkx` users

README.md
IvanIsCoding: I will update the README to document that.

In my view, any software newer than our `rustc` should be fair to ask for people building from source. The same applies for `setuptools` and `pip`.

IvanIsCoding:
I have an idea on how to make this work. We'll need to move `retworkx` to its own folder but it will be trivial

IvanIsCoding:
Metadata for the retworkx crate: https://gist.github.com/IvanIsCoding/0143766ca7237dc68a6ebeae6f3a0448

It seems to be working. I also tested manually the import is working, we are packaging the files right

README.md
IvanIsCoding: Done, let's say the version from February 2024 is the minimum one.

I also fixed a place where we list an incorrect rustc version.

IvanIsCoding:
Minor update, but I read https://www.reddit.com/r/Python/comments/1jwbymm/psa_you_should_remove_wheel_from_your/ so I discovered:
* we can require newer versions of `setuptools` and `setuptools-rust` that support what we document in the README via PEP 517. It turns out the requirements supports version specifiers
* we no longer need `wheel`

.github/workflows/main.yml
```diff
@@ -37,9 +37,9 @@ jobs:
       - name: Clippy
         run: cargo clippy --workspace --all-targets -- -D warnings
       - name: Black Codestyle Format
-        run: black --check --diff retworkx rustworkx retworkx tests
+        run: black --check --diff rustworkx tests
```
mtreinish, line 40:
```suggestion
        run: black --check --diff rustworkx tests retworkx
```

.github/workflows/main.yml
```diff
@@ -37,9 +37,9 @@ jobs:
...
       - name: Python Lint
-        run: ruff check rustworkx retworkx setup.py tests
+        run: ruff check rustworkx setup.py tests
```
mtreinish, line 42:
```suggestion
        run: ruff check rustworkx setup.py tests retworkx
```

.github/workflows/wheels.yml
```diff
@@ -202,12 +202,10 @@ jobs:
       - name: Install deps
         run: pip install -U setuptools-rust wheel build
       - name: Build sdist
```
mtreinish, line 204: We're actually not building an sdist here not that it matters. Apparently we're only uploading a wheel for retworkx.

noxfile.py
```diff
@@ -48,14 +35,14 @@ def lint(session):
     black(session)
     typos(session)
     session.install(*lint_deps)
-    session.run("ruff", "check", "rustworkx", "retworkx", "setup.py")
+    session.run("ruff", "check", "rustworkx", "setup.py")
```
mtreinish, line 38:
```suggestion
    session.run("ruff", "check", "rustworkx", "retworkx", "setup.py")
```

noxfile.py
```diff
@@ -69,7 +56,7 @@ def docs_clean(session):
 @nox.session(python=["3"])
 def black(session):
     session.install(*[d for d in lint_deps if "black" in d])
-    session.run("black", "rustworkx", "tests", "retworkx", *session.posargs)
+    session.run("black", "rustworkx", "tests", *session.posargs)
```
mtreinish, line 59:
```suggestion
    session.run("black", "rustworkx", "tests", "retworkx", *session.posargs)
```

releasenotes/notes/farewell-retworkx-be18546ac0b0cbc2.yaml
```diff
@@ -0,0 +1,15 @@
+---
+deprecations:
+  - |
+    We are saying goodbye to the legacy ``retworkx`` package that operates as a
+    backwards compatibility alias for ``rustworkx``. It was originally marked as
+    deprecated and this will be its last release. The ``rustworkx`` name has now existed
```
mtreinish, line 6:
```suggestion
    deprecated in the 0.12.0 release and this will be its last release it is supported. The ``rustworkx`` name has now existed
```

releasenotes/notes/farewell-retworkx-be18546ac0b0cbc2.yaml
```diff
@@ -0,0 +1,15 @@
...
+    longer than the ``retworkx`` name and the ``retworkx`` package is no longer needed.
```
mtreinish, line 7: It's almost that long right but not quite? It looks like we released rustworkx in 0.12.0 which was tagged Oct 7, 2022. The first retworkx release was 0.0.1 in May, 2019. So we're still a few  months short of that crossover point.

Although I'd agree that this is probably arguably true because those super early releases were unusable and you couldn't really start to use the package until about 0.3.0 which was released on Feb 20, 2020. So I'm fine keeping this. More was just excited to start discussing history :)

tox.ini
```diff
@@ -37,8 +37,8 @@ deps =
   ruff
 allowlist_externals=cargo
 commands =
-  black --check --diff {posargs} '../rustworkx' '../tests' '../retworkx'
-  ruff check ../rustworkx ../retworkx . ../setup.py
+  black --check --diff {posargs} '../rustworkx' '../tests'
+  ruff check ../rustworkx . ../setup.py
```
mtreinish, line 41:
```suggestion
  black --check --diff {posargs} '../rustworkx' '../tests' '../retworkx'
  ruff check ../rustworkx ../retworkx . ../setup.py
```

tox.ini
```diff
@@ -77,7 +77,7 @@ skip_install = true
 deps =
   black~=24.8
 commands =
-  black {posargs} '../rustworkx' '../tests' '../retworkx'
+  black {posargs} '../rustworkx' '../tests'
```
mtreinish, line 80:
```suggestion
  black {posargs} '../rustworkx' '../tests' '../retworkx'
```

mtreinish (commented):
Thanks for the updates, overall this is looking quite good now. I left some inline suggestions to add the linting for retworkx back. Until we release 0.17.0 we should still be running those in the remote chance there is a change made to the code there. I think it's good to merge now. 

The only open question I have is did you validate the wheel and release jobs still function correctly? I don't expect any hiccups but for a change like this it's good to validate we're still building the release artifacts without issues.

IvanIsCoding:
I will address the comments but regarding:
> The only open question I have is did you validate the wheel and release jobs still function correctly? I don't expect any hiccups but for a change like this it's good to validate we're still building the release artifacts without issues.

I built the wheel locally. Overall, the tests and specially the coverage test validate the wheel is valid. My biggest fear was the METADATA, but the links I posted above for the gist seem to identicate that gets included as well.

For wheels build, after I address all comments I will try to kick cibuildwheel. From inspecting how cibuildwheel, apart from retworkx, nothing should change. For `retworkx`, I ran the command locally to test so there's that.

IvanIsCoding:
Test run: https://github.com/IvanIsCoding/rustworkx/actions/runs/14553292059

I excluded macOS, PowerPC64, and s390x as those are more expensive for me to test on my personal project. But if Linux builds I think it should be all good. Windows is a bonus.

IvanIsCoding:
@mtreinish there's a spurious failure from x86_64 I am retrying (rustup failed to install, unrelated) but we got some wheels [in the first attempt](https://github.com/IvanIsCoding/rustworkx/actions/runs/14553292059/attempts/1). They look correct to me.

I think this is good to go.

releasenotes/notes/farewell-retworkx-be18546ac0b0cbc2.yaml
IvanIsCoding: There's also a delay on when we release 0.17, so I think by them it will be pretty much even! We can always add a `nearly as long` if you think that is necessary

mtreinish (approved):

PR #1420, opened 2025-04-06
New uv branch

### LINKED ISSUE #1416: Explore using `uv` for documentation and tests

IvanIsCoding:
Closes #1416 

This PR introduces [uv](https://docs.astral.sh/uv/guides/integration/github/) into our documentation workflow. The real diff of this PR will be much smaller once #1419 is merged. See: TODO

Overall, I think this accomplishes the goal of having a more reliable docs workflow. Currently, we pin some dependencies but it is still a workflow that has some quirks. Also, `sphinx` and `sphinx-jupyter` pull a lot of packages so I think using a lock file is a good idea.

I am not sold on using `uv` for other workflows, as currently they are fairly straightforward. We are also not a Python-only library, having to deal with `uv` trying to cache might be annoying. But with `reuse_venv=False`, I am somewhat confident we will not run into that issue for the docs.


IvanIsCoding:
Also, I am changing the documentation build to use `SETUPTOOLS_RUST_CARGO_PROFILE` so the binary is in debug mode. Because we discard the virtual environment, I think building faster is a plus.

IvanIsCoding:
I will put this on hold until we merge #1440

IvanIsCoding:
@mtreinish this is now ready to review

pyproject.toml
```diff
@@ -94,11 +95,12 @@ docs = [
     "pydot",
     "pillow>=4.2.1",
     "reno>=3.4.0",
-    "qiskit-sphinx-theme~=1.14.0rc1",
+    "qiskit-sphinx-theme==1.16.1",
     "matplotlib>=3.4",
     "sphinx-reredirects",
     "sphinxemoji",
     "ipykernel",
+    "lxml_html_clean",
```
mtreinish, line 103: What is this used for?

mtreinish (commented):
I like the idea of using a lockfile for docs it should hopefully make things more reliable longer term. I just had one inline question.

I also didn't hand check the uv lock file, I assume it's all correct since you just machine generated it.

pyproject.toml
IvanIsCoding: Currently if you run `nox -edocs` with Python 3.13 it fails. I had to add this dependency to make it work.

It has to do with Python 3.13 resolving a newer lxml or something of sorts.

IvanIsCoding:
> I like the idea of using a lockfile for docs it should hopefully make things more reliable longer term. I just had one inline question.
> 
> I also didn't hand check the uv lock file, I assume it's all correct since you just machine generated it.

I can confirm I have regenerated the `uv.lock` file very recently.

Also, `uv.lock` can be exported  to the new lock format defined in PEP 751. But that format is brand new and very few tools support it yet.

PR #1421, opened 2025-04-06
Use `PyGenericAlias` from PyO3 instead of importing `types.GenericAlias`

IvanIsCoding:
Follow up of #1348 

Since #1415, we have PyO3 0.24 that contains https://github.com/PyO3/pyo3/pull/4917 which will make `__clas_geitem__` slightly simpler and faster


mtreinish (approved):
Cool, it's cool to see the interface you added for this got included in the new release and we're able to use it here so neatly.

PR #1422, opened 2025-04-07
Add docstrings to exceptions and sort their list alphabetically

eumiro:
The [exceptions API](https://www.rustworkx.org/api/exceptions.html) was quite empty and since PyO3 [allows](https://github.com/PyO3/pyo3/pull/2027) adding docstrings to `create_exception`, I moved the comments to docstrings and also completed and sorted alphabetically their list in the API docs.

### DIFF
--- a/docs/source/api/exceptions.rst
+++ b/docs/source/api/exceptions.rst
@@ -4,16 +4,18 @@ Exceptions
 .. autosummary::
    :toctree: ../apiref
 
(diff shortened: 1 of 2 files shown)


IvanIsCoding:
I need to fix a conflict before merging, but thanks for documenting (again). We try to have obvious names, but clearly descriptions add a lot of value

IvanIsCoding (approved):

PR #1426, opened 2025-04-10
Use BFS to compute shortest paths in Closeness Centrality

IvanIsCoding:
Follow up of #1392

This is nothing too complex, but just like in #593 we missed the parallelization, we also missed that the shortest path function could be a faster one.

### DIFF
--- a/rustworkx-core/src/centrality.rs
+++ b/rustworkx-core/src/centrality.rs
@@ -17,7 +17,7 @@ use std::sync::RwLock;
 use hashbrown::HashMap;
 use petgraph::algo::dijkstra;
 use petgraph::visit::{
-    EdgeCount, EdgeIndexable, EdgeRef, GraphBase, GraphProp, IntoEdgeReferences, IntoEdges,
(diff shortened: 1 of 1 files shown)


rustworkx-core/src/centrality.rs
```diff
@@ -1133,11 +1133,26 @@ where
...
     });
+
+    let unweighted_shortest_path = |g: Reversed<&G>, s: G::NodeId| -> HashMap<G::NodeId, usize> {
+        let mut distances = HashMap::new();
+        let mut bfs = Bfs::new(g, s);
+        distances.insert(s, 0);
+        while let Some(node) = bfs.next(g) {
+            let distance = distances[&node];
+            for edge in g.edges(node) {
+                let target = edge.target();
```
mtreinish, line 1144: I was thinking we could do `for target in g.neighbors_directed(node, Outgoing)` here, but that would require we add the trait bound `IntoNeighborsDirected` on `G` and that seems unnecessary so this is fine.

mtreinish (approved):

PR #1427, opened 2025-04-10
Adding Bellman Ford Fuzz targets

Krishn1412:
Added a check to test the principle that for undirected non-negative weights, Bellman Ford and Dijkstra's algorithm must produce the same shortest distance.

Next, will start on adding fuzz targets for connectivity and generator modules.


IvanIsCoding (approved):
LGTM

PR #1429, opened 2025-04-11
Adding fuzz targets for special graphs

Krishn1412:
Added Fuzz targets for heavy-hex graph and lollipop graph to test their robustness for various inputs.

### DIFF
--- a/rustworkx-core/fuzz/Cargo.toml
+++ b/rustworkx-core/fuzz/Cargo.toml
@@ -34,3 +34,17 @@ path = "fuzz_targets/test_fuzz_bellman_ford.rs"
 test = false
 doc = false
 bench = false
+
+[[bin]]
+name = "test_fuzz_heavy_hex_graph"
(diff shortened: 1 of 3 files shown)


IvanIsCoding:
I will review #1430 later, but my opinion for fuzzing is:
* we shoud fuzz random generators to verify assumptions about random graphs
* we should not fuzz deterministic graph generators

In short, anything that takes a seed is a good candidate for fuzzing and everything that does not take a seed... well, we should let it pass.

It is mostly because the input space is smaller. We could write a for loop that creates all the lollipop graphs smaller than 512. Fuzzing doesn't hurt, but I think a more traditional test would have place here.

The same is not as true for randomizes functions as fuzzing could, in fact, lead to a different input that we generally would not be able to find.

Krishn1412:
Okay I understand. I was just trying to fit in fuzzing somehow for the special graphs. I'll move ahead with the random generators.

PR #1423, opened 2025-04-08
Adding random regular graph generator for rustworkx-core

singhmeet11:
This pull request adds a random regular graph function in rustworkx-core(this partly resolves - #1391). I will be adding the PyO3 connection and this functionality to rustworkx in a bit. Just wanted to get review on this first. 

- [x] I ran rustfmt locally
- [x] I have added the tests to cover my changes.
- [ ] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.

PR #1430, opened 2025-04-11
Adding Fuzz targets for random graph generator

Krishn1412:
Added a fuzz target for random graph generator with max edges.

### DIFF
--- a/rustworkx-core/fuzz/fuzz_targets/test_fuzz_random_graph.rs
+++ b/rustworkx-core/fuzz/fuzz_targets/test_fuzz_random_graph.rs
@@ -53,10 +53,12 @@ fn fuzz_gnm_random_graph(input: GnmInput) {
                 assert_eq!(graph1.node_count(), graph2.node_count());
                 assert_eq!(graph1.edge_count(), graph2.edge_count());
 
-                let mut edges1: Vec<_> = graph1.edge_references()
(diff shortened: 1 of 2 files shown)


IvanIsCoding:
I will review this later, but #1423 is a good candidate to be included when it is submited.

And as I said in #1429, I really like this.

rustworkx-core/fuzz/fuzz_targets/test_fuzz_random_graph.rs
```diff
@@ -0,0 +1,44 @@
...
+    let max_m = input.n.saturating_mul(input.n.saturating_sub(1));
+    let capped_m = input.m.min(max_m);
+
+    if let Ok(graph) = gnm_random_graph::<DiGraph<(), ()>, _, _, _, ()>(
+        input.n,
+        capped_m,
+        input.seed,
+        || (),
+        || (),
+    ) {
```
IvanIsCoding, line 35: Some other interesting check that we can have is that calling the method twice with the same seed generates that same graph i.e. the nodes and edges are all the same

Krishn1412: yes that sounds like a good idea, I'll add and push that test too.

IvanIsCoding (approved):

PR #1434, opened 2025-04-22
Adding Fuzz targets for lr planar

Krishn1412:
Added a fuzz target to check the robustness of lr planar function.

### DIFF
--- a/rustworkx-core/fuzz/Cargo.toml
+++ b/rustworkx-core/fuzz/Cargo.toml
@@ -41,3 +41,10 @@ path = "fuzz_targets/test_fuzz_random_graph.rs"
 test = false
 doc = false
 bench = false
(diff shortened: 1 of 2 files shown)


Krishn1412:
Hey @IvanIsCoding, could you take a look at this PR? TIA.

IvanIsCoding (approved):
I'll stand corrected, there are places our code could, in theory, panic: https://github.com/Qiskit/rustworkx/blob/30897c5bc46efcb17bc81c33946a0a4260e2d652/rustworkx-core/src/planar/lr_planar.rs#L393

I think we might be able to make it stronger but as is I think this should be fine.

A possible follow up is trying to generate random planar graphs and augmenting them to be non-planar.

PR #1433, opened 2025-04-20
Update cibuildwheel to 2.23.2 in order to unblock i686 builds

IvanIsCoding:
Closes #1432 

In short, this incorporates https://github.com/pypa/cibuildwheel/issues/2305 which fixes our issue in i686 builds

Test build: https://github.com/IvanIsCoding/rustworkx/actions/runs/14562751655

PR #1436, opened 2025-04-22
Adding fuzz target for centrality

Krishn1412:
Added a fuzz target for closeness centrality with wf_improved = true and false.

Both parallel and sequential modes (parallel_threshold = 1 and 200).

Consistency check between parallel and sequential outputs for each wf_improved mode.

### DIFF
--- a/rustworkx-core/fuzz/fuzz_targets/test_fuzz_centrality.rs
+++ b/rustworkx-core/fuzz/fuzz_targets/test_fuzz_centrality.rs
@@ -1,0 +1,60 @@
+#![no_main]
+
+use arbitrary::{Arbitrary, Unstructured};
+use libfuzzer_sys::fuzz_target;
+use rustworkx_core::centrality::closeness_centrality;
+use rustworkx_core::petgraph::graph::UnGraph;
+
+#[derive(Debug, Arbitrary)]
+struct CentralityFuzzInput {
+    edges: Vec<(usize, usize)>,
+    node_count: usize,
+}
+
+macro_rules! assert_almost_equal {
+    ($x:expr, $y:expr, $d:expr, $i:expr, $wf:expr) => {
+        if ($x - $y).abs() >= $d {
+            panic!(
+                "{} != {} within delta of {} at node {} with wf_improved = {}",
+                $x, $y, $d, $i, $wf
+            );
+        }
+    };
+}
+
+fuzz_target!(|data: &[u8]| {
+    if let Ok(input) = CentralityFuzzInput::arbitrary(&mut Unstructured::new(data)) {
+        fuzz_centrality(input);
+    }
+});
+
+fn fuzz_centrality(input: CentralityFuzzInput) {
+    if input.node_count == 0 || input.node_count > 1000 {
+        return;
+    }
+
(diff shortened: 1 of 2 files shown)


IvanIsCoding (approved):
Sure, it is worth fuzzing parallel vs sequential implementations

rustworkx-core/fuzz/fuzz_targets/test_fuzz_centrality.rs
```diff
@@ -0,0 +1,49 @@
...
+    );
+
+    // Closeness with wf_improved = false
+    let c_seq_false = closeness_centrality(&graph, false, 200);
+    let c_par_false = closeness_centrality(&graph, false, 1);
+    assert_eq!(
+        c_seq_false, c_par_false,
+        "Mismatch in wf_improved=false between sequential and parallel"
+    );
+}
```
IvanIsCoding, line 49: This is more of a nitpick, but I'd loop over an array with `[true, false]` parameters instead of repeating the calls. Not strictly necessary, but it is a nice touch

rustworkx-core/fuzz/fuzz_targets/test_fuzz_centrality.rs
```diff
@@ -0,0 +1,49 @@
...
+    assert_eq!(
+        c_seq_true, c_par_true,
+        "Mismatch in wf_improved=true between sequential and parallel"
+    );
+
+    // Closeness with wf_improved = false
+    let c_seq_false = closeness_centrality(&graph, false, 200);
+    let c_par_false = closeness_centrality(&graph, false, 1);
+    assert_eq!(
+        c_seq_false, c_par_false,
```
IvanIsCoding, line 46: Also, a friendly reminder to use almost equal assertions. We can define a smaller epsilon, but keep this in mind

Example: https://github.com/Qiskit/rustworkx/blob/30f29079eeaea6aefa1eb1485d496230173cb1b8/rustworkx-core/src/centrality.rs#L528

Floating point operations can lead to minor differences, which would cause the fuzzer to fail when in fact there is no bug. Of course, if the `eps` is too big we'll miss bugs so there is a balance

Krishn1412:
hey @IvanIsCoding , I have made the changes you recommended.

IvanIsCoding:
> hey @IvanIsCoding , I have made the changes you recommended.

Ok your changes look good but please synch to `HEAD` (something like `git merge upstream/main` where upstream is something like https://github.com/Qiskit/rustworkx.git. There are some diffs from other PRs

.github/workflows/wheels.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - name: Install cibuildwheel
         run: |
-          python -m pip install cibuildwheel==2.21.3
+          python -m pip install cibuildwheel==2.23.2
```
IvanIsCoding, line 60: This is from #1433, synch to HEAD from https://github.com/Qiskit/rustworkx.git not to include this

Krishn1412:
Sorry about that, I pulled from origin instead of upstream. It is fixed now.

IvanIsCoding (approved):

PR #1439, opened 2025-05-01
Add distance_matrix to rustworkx-core

mtreinish:
This commit moves the distance matrix functionality to rustworkx. This is mostly a straightforward migration as the functionality was written in a generic way already. The only difference is how node holes were handled, the this opted to split the functions into 2, a verison that assumes the graph has compact indices and one that doesn't.


IvanIsCoding:
Let’s defer it to 0.18.0 then.

I tagged it as 0.17.1 because for rustworkx it would not be a change. But for rustworkx-core it is I guess

IvanIsCoding:
I might merge #1491 and use this PR as an excuse to release 0.17.1. There was a problem releasing 0.17.0 with the identifiers.

IvanIsCoding (approved):
This seems straightforward. Maybe we should have merged it for 0.17.0, so I will use it to publish 0.17.1 as 0.17.0 only got published to crates.io

PR #1440, opened 2025-05-06
Use dependency groups in CI

IvanIsCoding:
Follow up of #1419

This leverage the brand new dependency group support introduced in pip 25.1 (https://pip.pypa.io/en/stable/user_guide/#dependency-groups). Notice that other tools we use like `nox` and `uv` already support dependency groups.

The main benefit of this PR is centralizing, again, the depedencies in `pyproject.toml`. Right now, some of the versions are pinned in GitHub Action files. Of course the old way works, but the new way is more aligned with what we do for tests and docs.

note: #1420 is on hold until we merge this, I am making the `uv` PR leverage this.


pyproject.toml
```diff
@@ -63,10 +63,13 @@ name = "Ivan Carvalho"
 email = "ivancarvalho@gatech.edu"
 
 [build-system]
-requires = ["setuptools>=70.1.0", "setuptools-rust>=1.9"]
+requires = ["setuptools>=77.0.0", "setuptools-rust>=1.9"]
```
IvanIsCoding, line 66: This is something I overlooked on #1420: because of https://github.com/pypa/packaging-problems/issues/870, we need a newer version of setuptools.

Nevertheless, `pip` should handle this smoothly.

mtreinish (approved):
This lgtm, I didn't realize they added this feature (but I haven't been following the pypa specs very closely lately) but it seems quite handy.

PR #1441, opened 2025-05-08
Narrow down return type of `PyDAG.copy` with `Self`

zen-xu:


IvanIsCoding (commented):
I think you want to add a method on the child class: https://github.com/Qiskit/rustworkx/blob/f521a50b56560b1c936f3f9150d9811fc31a5c94/rustworkx/__init__.pyi#L297

That also returns `Self` with the same signature. 

I will let you know when I wrote the type annotations originally `Self` was brand new. We ended using `typing_extesions.Self` because one of the custom types needed it! Now we should use it more often. It is more concise.

zen-xu:
Since PyDAG inherits from PyDiGraph, if PyDiGraph can return self, all its subclasses can return self instead of overriding this method in specific subclasses.

IvanIsCoding (approved):
That is correct, I checked on https://mypy-play.net/?mypy=latest&python=3.12 and indeed `Self` narrows down to the type of the child class

With that being said, another reminder to sign the CLA. Otherwise the license-bot won't let me merge your code.

IvanIsCoding:
Actually, #1314 is already submitted so I think this should be fine.

PR #1443, opened 2025-05-08
enhance visualization func type hint

zen-xu:


rustworkx/visualization/graphviz.pyi
```diff
@@ -10,17 +10,68 @@ import typing
 from rustworkx.rustworkx import PyGraph, PyDiGraph
 
 if typing.TYPE_CHECKING:
-    from PIL.Image import Image  # type: ignore
+    from typing_extensions import TypeAlias
+    from PIL.Image import Image
+
+    _Method: TypeAlias = typing.Literal["twopi", "neato", "circo", "fdp", "sfdp", "dot"]
```
IvanIsCoding, line 16: `typing.Literal` can conflict with some users that have `str` from a dictionary or some kind of function. Also, we'd need to stay up to date with graphviz to know what arguments are valid. I'd rather not care.

rustworkx/visualization/graphviz.pyi
```diff
@@ -10,17 +10,68 @@ import typing
...
+    _ImageType: TypeAlias = typing.Literal[
```
IvanIsCoding, line 17: Per #830, this still can fail at runtime so again I'd rather just make the type as a string

IvanIsCoding (commented):
The matplotlib change LGTM, for graphviz I'd rather not step into using `typing.Literal`. The `typing.overload` is fine.

rustworkx/visualization/graphviz.pyi
zen-xu: We can additionally allow image_type and method to support str

IvanIsCoding (approved):

rustworkx/visualization/graphviz.pyi
zen-xu: I inlined the Literal, it shouldn't fail at runtime anymore

IvanIsCoding:
I think you’ll need to run the code formatter. Either `nox -eblack` or just `black rustworkx` if you have it installed

PR #1446, opened 2025-05-10
Minor improvements: prevent future test failures, remove `fixtures`, and more

IvanIsCoding:
This PR is a collection of minor things:
* Removes the `fixtures` test dependency. I don't know when we stopped using it but it seems it can go
* Fixes a warning from NetworkX that I noticed in https://github.com/Qiskit/rustworkx/pull/1440#issuecomment-2858445887.  Our tests would start failing with NetworkX 3.6.
* Fixes a typo that makes `nox -elint` fail
* Pins `ruff`. Right now, we have `ruff~=0.6`. But `ruff` has not achieved a major version yet, so this is essentially running the latest version!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,14 +72,13 @@ testinfra = [
 ]
 test = [
   "setuptools-rust",
-  "fixtures",
   "testtools>=2.5.0",
-  "networkx>=2.5",
+  "networkx>=3.2",
   "stestr>=4.1",
 ]
 lint = [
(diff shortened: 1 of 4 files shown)


IvanIsCoding:
I guess I will need to be a little bit more clever on preventing NetworkX for failing (until 3.9 is dropped)

IvanIsCoding:
Ok, Python 3.9 passes on CI now.

PR #1447, opened 2025-05-10
Make `rustworkx` build and run with pyiodide

IvanIsCoding:
Closes #703 

This adds support for Pyodide. We list Pyodie in a brand new tier that is Tier Experimental. Currently, we build with `pyoide-build==0.30.2` using the following flags:
```
export "RUSTFLAGS="-C target-feature=+atomics,+bulk-memory,+mutable-globals -C link-arg=-sSIDE_MODULE=2 -C link-arg=-sWASM_BIGINT -Z emscripten-wasm-eh"
pyodide build
```

This requires a Rust with `nightly-2025-02-01` toolchain and `emsdk 4.0.8`. I tested with Pyodide 0.28.1a which maps to Python 3.13. Because we depend on so much unstable stuff, this is not ready yet for our CI. I need to figure out a way of adding it.

Related: https://github.com/pyodide/pyodide-recipes/pull/90

PR #1449, opened 2025-05-13
[ON HOLD] Add Pixi workflow to verify Pyodide build works

IvanIsCoding:
Follow up of  #1447

This uses [pixi](https://pixi.sh/) to run the Pyodide build. This is still higly experimental. It is related to #1416 and #1420 in the sense that pixi is to conda what uv is to pypi.

The Github Action workflow I created does the following:
* Installs Python 3.12.7 from Conda via Pixi. Python needs to be pinned
* Installs Emscripten 3.1.58. This is also pinned. Newer versions are required to build with Python 3.13 but that is not on Conda at the moment.
* Installs Node.js
* Installs Rust 1.86. Generally we'd use nightly rust, but Conda does not offer that. So I had to use the `RUSTC_BOOTSTRAP` trick to enable nightly features on a compiler. Of course that is terrible, but the version is pinned so essentially it is like using a pinned nightly compiler
* Installs a pinned version of https://github.com/pyodide/pyodide-build
* The Pyodide CLI installs the tools to build for 0.27.7, which maps to Python 3.12
* Compiles the code with Rust and Emscripten

Tests are included in #1450


CONTRIBUTING.md
```diff
@@ -560,6 +560,59 @@ has the necessary approvals and is tagged as `automerge` unless it has a merge
...
+
+#### Updating `pyodide-build` and dependencies
+
+All the dependencies for the Pyodide build are listed under `[tool.pixi.dependencies]`. To find a set
+of versions that works, visit the [pyodide-cross-build-environments.json](https://github.com/pyodide/pyodide/blob/main/pyodide-cross-build-environments.json) file in the `pyodide` repository.
+
+We'll need to align the [Emscripten](https://anaconda.org/conda-forge/emscripten) version from `conda-forge` with one
+of the public releases. Then, we pick a `pyodide-build` version higher than the required build version and the equivalent Python
+version also specified in the cross build environments. Lastly, update `[tool.pixi.tasks.install_xbuildenv]` to install
+the selected version of Pyodide.
```
mtreinish: I assume this is because pixi is pulling emscripten from conda instead of a system installed package? But it was a bit confusing to read this description as conda isn't mentioned before. Maybe we should have a sentence before this explaining where conda is coming into play. Since this is a very different workflow from the normal rustworkx build process it's best to be a bit more explicit about how all these pieces come together.

CONTRIBUTING.md
```diff
@@ -560,6 +560,59 @@ has the necessary approvals and is tagged as `automerge` unless it has a merge
...
+
+Please refer to the [Pixi](https://pixi.sh/latest/) page for the latest instructions on how
+to install Pixi. Once installed, there's a single command that needs to be run.
+
+#### Building for Pyodide
+
+At the root of the directory, simply run:
+
+```bash
+pixi run build_pyodide
```
mtreinish, line 580: When I run this locally I first got an issue about the `emcc -v` call failing, when I set the env var to skip the version check it errors during the build with:

```
error: linking with `emcc` failed: exit status: 1
  |
  = note:  "emcc" "-s" "EXPORTED_FUNCTIONS=[\"_PyInit_generators\",\"_PyInit_rustworkx\"]" "<1 object files omitted>" "/home/mtreinish/git/qiskit/retworkx/target/wasm32-unknown-emscripten/release/deps/{libcompiler_builtins-ee68dc841fe2a394.rlib}.rlib" "-B<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/bin/gcc-ld" "--target=wasm32-unknown-emscripten" "-sDISABLE_EXCEPTION_CATCHING=0" "-o" "/home/mtreinish/git/qiskit/retworkx/target/wasm32-unknown-emscripten/release/deps/rustworkx.wasm" "-O3" "-g0" "-sSIDE_MODULE=2" "-sWASM_BIGINT" "-sSIDE_MODULE=2" "-sWASM_BIGINT" "-sABORTING_MALLOC=0" "-sWASM_BIGINT"
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: Traceback (most recent call last):
            File "/home/mtreinish/git/qiskit/retworkx/.pixi/envs/default/bin/emcc", line 23, in <module>
              from tools.toolchain_profiler import ToolchainProfiler
          ModuleNotFoundError: No module named 'tools'
```

pyproject.toml
```diff
@@ -154,3 +156,31 @@ repair-wheel-command = "brew install pipx && pipx ensurepath && pipx run --spec
 
 [tool.cibuildwheel.windows]
 repair-wheel-command = "cp {wheel} {dest_dir}/. && pipx run abi3audit==0.0.9 --strict --report {wheel}"
+
+[tool.pixi.workspace]
+channels = ["conda-forge"]
+platforms = ["linux-64", "osx-arm64"]
```
mtreinish, line 159: I thought we weren't supporting pixi on arm64 macos?

mtreinish (commented):
Sorry this took me so long to follow up on. This looks good to me, it exciting to see a workflow for building pyodide in CI. I just had a couple of questions inline. I wasn't able to get pixi working locally and I'm wondering if something went stale while this was waiting on my review.

pyproject.toml
IvanIsCoding: https://github.com/conda-forge/emscripten-feedstock/issues/35 got fixed, now this supports macOS arm as well!

Naturally, emscripten supports all multiple platforms. But I don't want to deal with Windows. As long as Conda Forge provides it, I think it's fine to list it.

CONTRIBUTING.md
IvanIsCoding: I have rewritten the section, take a look.

CONTRIBUTING.md
IvanIsCoding: So I have not been able to reproduce this. I tested on Linux locally, Mac locally, and on CI.

If you tell me the workaround you used, I can add it to `pyproject.toml` or at least mention it in `CONTRIBUTING.md`

IvanIsCoding:
I am putting this on hold after https://peps.python.org/pep-0783/

The correct way of doing this I think is:
1. We use cibuildwheel
2. We test our wheels with cibuildwheel
3. We publish our own wheels

This will need to be revisited as the current workflow is a bit obsolete

PR #1450, opened 2025-05-15
[ON HOLD]  Add tests for Pyodide build with Pixi and uv

IvanIsCoding:
Follow up of  #1449. True diff: https://github.com/IvanIsCoding/rustworkx/compare/pixi-pyodide...IvanIsCoding:rustworkx:test-pyodide

This builds on the preview two PRs by adding a Pixi task that does the following:
* Install uv inside pixi
* Set up a pyodide environment with uv and run `python -m unittest discover .`


IvanIsCoding:
https://peps.python.org/pep-0783/ changes plans, this will be on hold

PR #1451, opened 2025-05-19
[ON HOLD] Add interactive playground for users to try rustworkx

IvanIsCoding:
Follow up combining #1420 and the wheels built on #1449. Edit: the diff is now smaller; we use pre-built wheels from pyodide-recipes

This PR:
* Introduces `jupyterlite-sphinx` and `jupyterlite-pyodide-kernel` as dependencies
* Creates a playground page with `rustworkx`, users can run it in the browser
* The playground uses Pyodide from a CDN, Jupyterlite assets (around 16MB), and a rustworkx wheel that is from the Pyodide index


IvanIsCoding:
Here's  a look at the preview: 
Playground_Preview

We could come with more ideas for the demo to auto-execute. Moreover, we could start adding more interactive snippest throughout the documentation

IvanIsCoding:
I updated the demo to plot the Petersen Graph. It showcases our functionalities a little bit more.

Petersen_Graph_Demo

docs/source/playground.rst
```diff
@@ -0,0 +1,28 @@
+.. _rustworkx_playground:
+
+####################
+Rustworkx Playground
+####################
+
+Welcome to the Rustworkx Playground! This is an interactive environment where you can
+experiment with Rustworkx on your browser.
```
mtreinish: Not sure which is better but in reading it I thought "in" instead of "on" sounded more natural:

```suggestion
experiment with Rustworkx in your browser.
```

docs/source/playground.rst
```diff
@@ -0,0 +1,28 @@
...
+   import rustworkx as rx
+   print(f"Rustworkx Playground with version: {rx.__version__}")
+
+.. note::
+   The `rustworkx` version in the playground is not always the latest. Verify the deployed
+   version with `rustworkx.__version__`.
+
+.. note::
+   The `rustworkx` version in the playground experimental. If you find any issues, please
+   report them at https://github.com/Qiskit/rustworkx/issues.
```
mtreinish, line 28: For this you mean the wasm version of rustworkx is experimental right? Not that 0.17 is experimental?

docs/source/playground.rst
```diff
@@ -0,0 +1,28 @@
...
+
+   import piplite
+   await piplite.install("rustworkx")
+
+   import rustworkx as rx
+   print(f"Rustworkx Playground with version: {rx.__version__}")
+
+.. note::
+   The `rustworkx` version in the playground is not always the latest. Verify the deployed
+   version with `rustworkx.__version__`.
```
mtreinish, line 24: I might have asked this before but is there a path to building the correct package as part of the docs build? It's not like a rustworkx build takes us that long and we already do it a bunch in CI. So we could then have different versions bundled for each docs build so it always matches.

Looking at the recipe you contributed to pyodide: https://github.com/pyodide/pyodide-recipes/blob/main/packages/rustworkx/meta.yaml it doesn't seem too complicated.

mtreinish (commented):
This will be pretty cool to have in the docs. Just a couple questions inline

docs/source/playground.rst
IvanIsCoding: I changed the sentence to:
```
The `rustworkx` WASM version in the playground experimental
```

#1449 changes the platform support to say "rustworkx WASM is experimental". Mostly because Pyodide uses Rust nightly, even between my 0.17.0a3 PR and 0.17.1 PR to Pyodide recipes things collapsed.

docs/source/playground.rst
IvanIsCoding: So, I think the path to have the latest commit from `main` in is to:
1. Merge #1449 
2. Chain `nox -edocs` with the wheel built from `pixi run build_pyodide`

I think the biggest blockers are that #1449 uses Rust nightly (well, not quite but `RUSTC_BOOTSTRAP` is almost that with a pinned rustc). Moreover, setting up `emscripten` is a ritual that I will only accept in CI with some tool like `pixi` or `nixpkgs`. It needs a lockfile in my opinion, too many parts can break

IvanIsCoding:
I am putting this on hold after https://peps.python.org/pep-0783/. This will not make it to 0.18.0

PR #1455, opened 2025-05-29
fix: Expose can_contract_without_cycle as public API and fixes

### LINKED ISSUE #1009: expose can_contract
I would like to potentially contract a set of nodes before I know the operation I want to contract to (there is some computational cost to determining that).  It seems like `can_contract` in digraph -> `contract_nodes` does this which could be useful for this purpose.

luffy-orf:
## Expose can_contract_without_cycle as a public API for node contraction feasibility

**Closes #1009**

### Summary

This PR exposes a function to check if a set of nodes in a directed graph can be contracted without introducing a cycle, as requested in  #1009

### Details

- Renamed the internal `can_contract` function to `can_contract_without_cycle` for clarity and consistency with the suggested API name.
- Made `can_contract_without_cycle` a public function and added a detailed docstring with usage examples.
- Updated all internal references to use the new function name.
- Re-exported `can_contract_without_cycle` in `graph_ext/mod.rs` so it is available as part of the public API (`rustworkx_core::graph_ext::can_contract_without_cycle`).
- Fixed the doctest for `can_contract_without_cycle` to construct the `IndexSet` with `foldhash::fast::RandomState`, matching the function signature.

### DIFF
--- a/rustworkx-core/src/graph_ext/contraction.rs
+++ b/rustworkx-core/src/graph_ext/contraction.rs
@@ -109,7 +109,7 @@ where
         I: IntoIterator<Item = Self::NodeId>,
     {
         let nodes = IndexSet::from_iter(nodes);
-        if check_cycle && !can_contract(self.deref(), &nodes) {
+        if check_cycle && !can_contract_without_cycle(self.deref(), &nodes) {
             return Err(ContractError::DAGWouldCycle);
         }
         Ok(contract_stable(self, nodes, obj, NoCallback::None).unwrap())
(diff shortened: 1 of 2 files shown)


mtreinish (commented):
Thanks for starting this, besides a release note which is missing, you can see details on this here: https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md#release-notes 

I think the other component to close out the issue is to also add the method to the python library. Right now this is just making the rustworkx-core rust interface public which is the first step. But the original issue was about exposing it to the Python API for `PyDiGraph`. It should be pretty straightforward to add a new pyfunction or method on `PyDiGraph` that calls this rustworkx-core function internally.

Issue #1437, opened 2025-04-28
`token_swapper` to return its swaps as `(G::NodeId, G::NodeId)` rather than `NodeIndex`


jakelishman:
### What is the expected enhancement?

This would be a breaking change to the interface, but `token_swapper` currently returns `(NodeIndex, NodeIndex)` as its swaps for a graph `g: impl NodeIndexable` where it's arranged that `g.to_index(n.index())` is the way of converting a `n: NodeIndex` back to the input graph's `NodeId`.

Given that the mapping of the `NodeIndex` back to `G::NodeId` must be possible and known to the `token_swapper`, it would be more convenient to use if `token_swapper` returned its answer in this form.

(side note: there's an internal `HashMap` that tracks this mapping that's unnecessary, given how the inner `StableGraph` is constructed.  (side side note: that mapping needn't be a `HashMap` and the graph needn't be a `StableGraph` with holes - could use a compact `Graph` and a `[G::NodeId]` as the map back))

1ucian0:
👋 This issue is part of [UnitaryHack 2025](https://unitaryhack.dev/projects/rustworkx/). For the duration of the event, it follows a few special rules and considerations:

 * **No need to be assigned** to start working on it, just go ahead! This holds until the end of UnitaryHack. 
 * If multiple PRs are submitted to close this issue, **maintainers may select one** at the end of the event and award the bounty accordingly.
 * If more than one person works on a single PR, please note that the bounty cannot be split.
 * In some cases, **a not-fully-complete but close-enough PR** might receive a partial bounty.
 * If you have any questions or need clarification, feel free to ask in the comments here. Please note that while maintainers do their best to reply, responses might take longer than you expect.

Happy hacking! 🧑‍💻✨

gluonhiggs:
@jakelishman Is this still open. I'm so confused with the PR appears above. If it is, I will try!

gluonhiggs:
@jakelishman I have created the PR as linked above, please have a look! Thank you!

PR #1456, opened 2025-05-29
feat: Make token_swapper return swaps as (G::NodeId, G::NodeId) inste…

luffy-orf:
### Description

This PR addresses #1437 by improving the type returned from the core Rust API `token_swapper`:

* Swaps are now returned as `(G::NodeId, G::NodeId)` instead of `(NodeIndex, NodeIndex)`.
* Internal logic still operates on `NodeIndex` for performance and simplicity.
* Before returning, results are mapped back to `G::NodeId` using a stored mapping.
* Python bindings (`PyGraph`/`PyDiGraph`) remain unchanged: since `NodeId = NodeIndex`, output is still a list of integer pairs.
* The implementation passes `cargo clippy --all-targets -- -D warnings`, with appropriate lint suppressions for type complexity.

### DIFF
--- a/rustworkx-core/src/token_swapper.rs
+++ b/rustworkx-core/src/token_swapper.rs
@@ -99,7 +99,8 @@ where
         }
     }
 
-    fn map(&mut self) -> Result<Vec<Swap>, MapNotPossible> {
+    #[allow(clippy::type_complexity)]
+    fn map(&mut self) -> Result<Vec<(G::NodeId, G::NodeId)>, MapNotPossible> {
         let num_nodes = self.graph.node_bound();
         let num_edges = self.graph.edge_count();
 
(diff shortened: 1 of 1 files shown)


rustworkx-core/src/token_swapper.rs
```diff
@@ -185,7 +186,13 @@ where
...
             Err(e) => Err(*e),
         })
-        .unwrap()
+        .unwrap()?;
+        // Map NodeIndex swaps to NodeId swaps
+        let swaps = result
+            .into_iter()
+            .map(|(a, b)| (self.rev_node_map[&a], self.rev_node_map[&b]))
+            .collect();
+        Ok(swaps)
```
jakelishman, line 195: While this does return the swaps in the `NodeId` form, it adds an extra heap allocation to do it, which should be avoidable - in the parentheses in the original issue, I'd given a couple of suggestions for how (I think - I didn't do it completely) it should be possible to modify the internal tracking structures so that the natural output is `G::NodeId`, rather than needing an explicit remapping.

PR #1459, opened 2025-06-03
Expose can_contract_without_cycle

0xSooki:
This pull request introduces a new method, `can_contract_without_cycle`, to the `PyDiGraph` object and exposes its underlying Rust implementation. The method enables users to check whether contracting a set of nodes in a directed graph would introduce a cycle.

### Details
* Added the `can_contract_without_cycle` method to the `PyDiGraph` class, allowing users to verify if contracting nodes would create a cycle.
* Renamed `can_contract` to `can_contract_without_cycle` function in `rustworkx-core/src/graph_ext/contraction.rs`.
* Replaced all references to `can_contract`.
* Exposed the `can_contract_without_cycle` function in Rust, including documentation and examples for usage.

### Testing
* Added unit tests in Rust and Python to validate the behavior of the `can_contract_without_cycle` method. Tests cover both scenarios where contraction is allowed and where it would introduce a cycle.

### Documentation
* Added release notes describing the new feature and its use cases.


releasenotes/notes/expose-can-contract-without-cycle-bd82ec1e5fd87bc7.yaml
1ucian0, line 1: I'm not a rustworkx maintainer, but based on the release notes in other PRs, I think you can just leave the `features` section and remove the rest.

0xSooki: Thank you, I have updated it according to your suggestions.

rustworkx-core/src/graph_ext/contraction.rs
```diff
@@ -498,8 +498,47 @@ where
...
+///
+/// // Try to contract nodes b and c
+/// let mut nodes = IndexSet::with_hasher(RandomState::default());
+/// nodes.insert(b);
+/// nodes.insert(c);
+///
+/// let can_contract = can_contract_without_cycle(&graph, &nodes);
+/// assert!(can_contract); // true: contracting b and c does not introduce a cycle
+/// ```
+pub fn can_contract_without_cycle<G>(
```
IvanIsCoding: Do not rename this method, it is public. Therefore, existing `rustworkx-core` users could break.

For simplicity, if you find `can_contract` as a confusing name in the `rustworkx` side of the code, use `use contraction::can_contract as can_contract_without_cycle`

0xSooki: Thanks for the clarification! I've reverted the name on `rustworkx` side to `can_contract`

IvanIsCoding (approved):
LGTM. I am happy with the tests and the method in `PyDiGraph` is straightforward. Thanks for making it happen.

Issue #1460, opened 2025-06-03
Question: related to keeping track of subgraph node and edge IDs


JoOkuma:
Hi,

I recently started implementing a higher-level library for graph processing using `rustworkx` as the backbone.

It will rely a lot on computing subgraphs and keeping track of the node and edge ids of the original graph.

I had a bug in my implementation because I was assuming that the `PyDiGraph.subgraph` would keep the order of the `nodes`  indices used to subset, where actually the `nodes` are always sorted --- I was able to fix my bug once I figured this out.

I peeked into `rustworkx` implementation of the subgraph-ing
 https://github.com/Qiskit/rustworkx/blob/7840a2c606dcb3065e960bbd28df7437e59312c7/src/graph.rs#L1928-L1954

and I am a bit puzzled why the `nodes`, which are a list of indices, are not used to access the `<PETGRAPH (self.g?)>.nodes` directly?
Which seems to be a vector, https://github.com/petgraph/petgraph/blob/90a220b2cdb54f5829a87e60ff14322acf7c6f51/src/graph_impl/mod.rs#L390-L394

I'm a newbie when it comes to Rust and a new `rustworkx` user, so there might be something I'm missing out.

One additional thing is a feature request would be to include the [node_map](https://github.com/Qiskit/rustworkx/blob/7840a2c606dcb3065e960bbd28df7437e59312c7/src/graph.rs#L1930) in the subgraph attributes, since it's already being computed.
What do you think?

I'm happy to open a PR if there's an agreement with my comments.

Thanks,

IvanIsCoding:
Given the numerous issues we have had with this unexpected behaviour of `subgraph` (see #1404 that will be in the public docs once we release 0.17), I think a method that returns a `nodemap` is a fair ask.

I see two options:

### Brand new method

`subgraph_with_nodemap` (or a better name): it does the same as `subgraph` but returns a tuple of the new graph  with a NodeMap.

**Pros**: solves the issue, is forward compatible
**Cons**: new method we need to test and maintain

### Reuse `attrs` of graphs in the old method

**Pros**: solves the issue
**Cons**: conflcits with `preserve_attrs`. If `preserve_attrs=True`, we'd need to come up with a solution.

IvanIsCoding:
@mtreinish might have a stronger opinion, for me either is fine but I am more inclided to having a new method. You can start a PR with either option, it shouldn't be too hard to refactor anyway.

IvanIsCoding:
[...]

PR #1461, opened 2025-06-04
New method `subgraph_with_nodemap`

JoOkuma:
This was discussed on #1460 

- [X] I ran rustfmt locally
- [X] I have added the tests to cover my changes.
- [X] I have read the CONTRIBUTING document.

From my understanding the documentation is generated automatically and no update is needed.

I opted to have `subgraph` call `subgraph_with_nodemap` to reduce duplicated code, I can change to a separate implementation to avoid the minor overhead of creating the `nodemap` on the `subgraph` method.


src/graph.rs
```diff
@@ -1945,12 +1955,39 @@ impl PyGraph {
...
             node_removed: false,
             multigraph: self.multigraph,
             attrs,
-        }
+        };
+        (subgraph, node_map)
+    }
+
+    /// Return a new PyGraph object for a subgraph of this graph.
+    ///
```
IvanIsCoding, line 1977: Same comment here about the note

src/digraph.rs
```diff
@@ -3052,14 +3062,41 @@ impl PyDiGraph {
...
             check_cycle: self.check_cycle,
             multigraph: self.multigraph,
             attrs,
-        }
+        };
+        (subgraph, node_map)
+    }
+
+    /// Return a new PyDiGraph object for a subgraph of this graph.
+    ///
```
IvanIsCoding, line 3085: Add a [Sphinx Note](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html#notes) to hint users about `subgraph_with_node`, including a reference that links to the other method.

That way, when users read the documentation they'll choose the new method

IvanIsCoding:
@JoOkuma I think you’ll need to fix the release note and/or do string, we have  CI failure complaining about ambiguous references to the new method. Maybe you need to specify it narrowing for PyGraph/PyDiGraph

JoOkuma:
@IvanIsCoding ok, lets see if this works. I can't build the docs locally, so it's hard to test.

PR #1463, opened 2025-06-09
token_swapper to return its swaps as (G::NodeId, G::NodeId) rather than NodeIndex

### LINKED ISSUE #1437: `token_swapper` to return its swaps as `(G::NodeId, G::NodeId)` rather than `NodeIndex`

gluonhiggs:
Resolve #1437 
The modification is straightforward with some docs  changes

### DIFF
--- a/rustworkx-core/src/token_swapper.rs
+++ b/rustworkx-core/src/token_swapper.rs
@@ -35,6 +35,8 @@ use crate::traversal::dfs_edges;
 
 type Swap = (NodeIndex, NodeIndex);
 type Edge = (NodeIndex, NodeIndex);
+type SwapList<N> = Vec<(N, N)>;
+type SwapResult<N> = Result<SwapList<N>, MapNotPossible>;
 
 /// Error returned by token swapper if the request mapping
 /// is impossible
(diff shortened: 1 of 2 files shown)


1ucian0 (commented):
Not a maintainer here. My understanding of https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md is that you should add tests for this new functionality

gluonhiggs:
> Not a maintainer here. My understanding of https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md is that you should add tests for this new functionality

I ran the existing tests and thought these were enough. Let me check again.

PR #1462, opened 2025-06-06
Add graphml serialization and dot deserialization.

gluonhiggs:
To resolve #840 
I have added a stepping stone for graphml serialization, but I think I need more reviews on this to resolve the first part of the issue.  Because this is a little bit different to other `rustworkx` functions for `graph calculations`, please elaborate what I should do next.

PR #1464, opened 2025-06-09
Fix #840: Add GraphML serializer

### LINKED ISSUE #840: Add reverse direction serialization

thierry-martinez:
This commit adds a GraphML serializer:

```python
def write_graphml(
    graphs: list[PyGraph | PyDiGraph],
    keys: list[tuple[str, Domain, str, Type, Any]],
    path: str,
    /,
    compression: str | None = ...,
) -> None: ...
```

`keys` is a list of tuples: id, domain, name of the key, type, and default value. This commit also introduces the
`read_graphml_with_keys` function, which returns the key definitions in the same format, along with the list of parsed graphs.

The implementation preserves the ids of graphs, nodes, and edges when possible. If some ids conflict, fresh ids are generated in the written GraphML file. The `read_graphml` function has also been updated to store the graph id in the graph attributes, just like node and edge ids are stored in the corresponding attributes.

The `write_graphml` function supports gzip compression, as does `read_graphml`.

Note that the JSON node-link serializer (the other part of #840) was already implemented in #1091.

Compared to #1462:

- Keys are passed explicitly instead of being inferred (which allows to use the types `float` and `int`, and to use default values);

- Attributes for graphs, nodes, and edges are taken from the weight of elements, instead of relying on callbacks. This allows write_graphml to act as a proper reciprocal of read_graphml. Round-trip tests have been added.

- IDs are taken from attributes when possible, instead of being generated from indices.

- Multiple graphs can be written to the same file.

- Gzip compression is supported.

- Tests have been added.

Regarding @IvanIsCoding's
comment (https://github.com/Qiskit/rustworkx/pull/1462#issuecomment-2951935390), about using https://github.com/jonasbb/petgraph-graphml:

- Rustworkx's `graphml.rs` introduces an internal `Graph` data structure, which is used for `read_graphml`. It is natural to have `write_graphml` rely on the same data structure.

- `petgraph-graphml` generates ids from indices, which prevents us from preserving ids accross the `read_graphml`/`write_graphml` round trip.

### DIFF
--- a/rustworkx/rustworkx.pyi
+++ b/rustworkx/rustworkx.pyi
@@ -68,6 +68,20 @@ class ColoringStrategy:
     Saturation: Any
     IndependentSet: Any
 
+@final
+class Domain:
+    Node: Domain
+    Edge: Domain
+    Graph: Domain
+    All: Domain
+
+@final
+class Type:
+    Boolean: Type
+    Int: Type
+    Float: Type
+    Long: Type
+
 # Cartesian product
 
 def digraph_cartesian_product(
@@ -680,11 +694,23 @@ def directed_random_bipartite_graph(
 
 # Read Write
 
+def read_graphml_with_keys(
+    path: str,
+    /,
+    compression: str | None = ...,
+) -> tuple[list[tuple[str, Domain, str, Type, Any]], list[PyGraph | PyDiGraph]]: ...
 def read_graphml(
     path: str,
     /,
     compression: str | None = ...,
 ) -> list[PyGraph | PyDiGraph]: ...
+def write_graphml(
+    graphs: list[PyGraph | PyDiGraph],
+    keys: list[tuple[str, Domain, str, Type, Any]],
+    path: str,
+    /,
+    compression: str | None = ...,
+) -> None: ...
 def digraph_node_link_json(
     graph: PyDiGraph[_S, _T],
     /,
--- a/src/graphml.rs
+++ b/src/graphml.rs
@@ -12,26 +12,29 @@
 
 #![allow(clippy::borrow_as_ptr)]
 
+use std::borrow::{Borrow, Cow};
 use std::convert::From;
 use std::ffi::OsStr;
 use std::fs::File;
-use std::io::{BufRead, BufReader};
+use std::io::{BufRead, BufReader, BufWriter};
 use std::iter::FromIterator;
 use std::num::{ParseFloatError, ParseIntError};
 use std::path::Path;
 use std::str::ParseBoolError;
 
 use flate2::bufread::GzDecoder;
-use hashbrown::HashMap;
+use flate2::write::GzEncoder;
+use flate2::Compression;
+use hashbrown::{HashMap, HashSet};
 use indexmap::IndexMap;
 
-use quick_xml::events::{BytesStart, Event};
+use quick_xml::events::{BytesDecl, BytesStart, BytesText, Event};
 use quick_xml::name::QName;
 use quick_xml::Error as XmlError;
-use quick_xml::Reader;
+use quick_xml::{Reader, Writer};
 
 use petgraph::algo;
-use petgraph::{Directed, Undirected};
+use petgraph::{Directed, EdgeType, Undirected};
 
 use pyo3::exceptions::PyException;
 use pyo3::prelude::*;
@@ -46,6 +49,7 @@ pub enum Error {
     NotFound(String),
     UnSupported(String),
     InvalidDoc(String),
+    IO(String),
 }
 
 impl From<XmlError> for Error {
@@ -76,6 +80,13 @@ impl From<ParseFloatError> for Error {
     }
 }
 
+impl From<std::io::Error> for Error {
+    #[inline]
+    fn from(e: std::io::Error) -> Error {
+        Error::IO(format!("Input/output error: {}", e))
+    }
+}
+
 impl From<Error> for PyErr {
     #[inline]
     fn from(error: Error) -> PyErr {
@@ -84,7 +95,8 @@ impl From<Error> for PyErr {
             | Error::ParseValue(msg)
             | Error::NotFound(msg)
             | Error::UnSupported(msg)
-            | Error::InvalidDoc(msg) => PyException::new_err(msg),
+            | Error::InvalidDoc(msg)
+            | Error::IO(msg) => PyException::new_err(msg),
         }
     }
 }
@@ -112,15 +124,32 @@ fn xml_attribute<'a>(element: &'a BytesStart<'a>, key: &[u8]) -> Result<String, 
         })
 }
 
-#[derive(Clone, Copy)]
-enum Domain {
+#[pyclass(eq)]
+#[derive(Clone, Copy, PartialEq)]
+pub enum Domain {
     Node,
     Edge,
     Graph,
     All,
 }
 
-enum Type {
+impl TryFrom<&[u8]> for Domain {
+    type Error = ();
+
+    fn try_from(value: &[u8]) -> Result<Self, ()> {
+        match value {
+            b"node" => Ok(Domain::Node),
+            b"edge" => Ok(Domain::Edge),
+            b"graph" => Ok(Domain::Graph),
+            b"all" => Ok(Domain::All),
+            _ => Err(()),
+        }
+    }
+}
+
+#[pyclass(eq)]
+#[derive(Clone, Copy, PartialEq)]
+pub enum Type {
     Boolean,
     Int,
     Float,
@@ -129,7 +158,20 @@ enum Type {
     Long,
 }
 
-#[derive(Clone)]
+impl Into<&'static str> for Type {
+    fn into(self) -> &'static str {
+        match self {
+            Type::Boolean => "boolean",
+            Type::Int => "int",
+            Type::Float => "float",
+            Type::Double => "double",
+            Type::String => "string",
+            Type::Long => "long",
+        }
+    }
+}
+
+#[derive(Clone, PartialEq)]
 enum Value {
     Boolean(bool),
     Int(isize),
@@ -138,6 +180,27 @@ enum Value {
     String(String),
     Long(isize),
     UnDefined,
+}
+
+impl Value {
+    fn serialize(&self) -> Option<Cow<str>> {
+        match self {
+            Value::Boolean(val) => Some(Cow::from(val.to_string())),
+            Value::Int(val) => Some(Cow::from(val.to_string())),
+            Value::Float(val) => Some(Cow::from(val.to_string())),
+            Value::Double(val) => Some(Cow::from(val.to_string())),
+            Value::String(val) => Some(Cow::from(val)),
+            Value::Long(val) => Some(Cow::from(val.to_string())),
+            Value::UnDefined => None,
+        }
+    }
+
+    fn to_id(&self) -> PyResult<&str> {
+        match self {
+            Value::String(value_str) => Ok(value_str),
+            _ => Err(PyException::new_err("Expected string value for id")),
+        }
+    }
 }
 
 impl<'py> IntoPyObject<'py> for Value {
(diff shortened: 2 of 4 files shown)


rustworkx/rustworkx.pyi
```diff
@@ -68,6 +68,20 @@ class ColoringStrategy:
     Saturation: Any
     IndependentSet: Any
 
+@final
+class Domain:
```
IvanIsCoding: It will not be obvious for users that don't use GraphML what a domain is. Call this class `XMLDomain` or something along these lines

rustworkx/rustworkx.pyi
```diff
@@ -68,6 +68,20 @@ class ColoringStrategy:
...
+    Node: Domain
+    Edge: Domain
+    Graph: Domain
+    All: Domain
+
+@final
```
IvanIsCoding, line 78: Same comment, maybe call it `XMLType`

src/graphml.rs
```diff
@@ -428,6 +716,24 @@ impl GraphML {
         Ok(())
     }
 
+    fn get_keys(&self, domain: Domain) -> &IndexMap<String, Key> {
```
IvanIsCoding: use `DictMap` instead of `IndexMap`

rustworkx/rustworkx.pyi
IvanIsCoding: Just to add, keep the Rust code as is and rename the pyclass. PyO3 has an option for that

rustworkx/rustworkx.pyi
thierry-martinez: Fixed in c350943, thanks! (I used `GraphML` as prefix.)

src/graphml.rs
thierry-martinez: This function returns the `key_for_*` fields of `struct GraphML`, which are defined as `IndexMap` with the default `std::hash::RandomState`. If I change it here to `DictMap`, I would also need to change the type in the definition of `struct GraphML`, which is not introduced in this PR.

Should I go ahead and change it here, or would it be better to open an issue to update the definition of `struct GraphML` in a separate PR?

rustworkx/rustworkx.pyi
thierry-martinez: Fixed in https://github.com/Qiskit/rustworkx/commit/c350943743f32027c2b463d17d38a6b93690fb86, thanks!

src/graphml.rs
```diff
@@ -112,15 +124,32 @@ fn xml_attribute<'a>(element: &'a BytesStart<'a>, key: &[u8]) -> Result<String,
...
+impl TryFrom<&[u8]> for Domain {
+    type Error = ();
+
+    fn try_from(value: &[u8]) -> Result<Self, ()> {
+        match value {
+            b"node" => Ok(Domain::Node),
+            b"edge" => Ok(Domain::Edge),
+            b"graph" => Ok(Domain::Graph),
+            b"all" => Ok(Domain::All),
+            _ => Err(()),
```
IvanIsCoding, line 148: Do we want to include a return message explaining why a call to this failed though? Right now that is not possible

src/graphml.rs
```diff
@@ -340,6 +444,167 @@ impl<'py> IntoPyObject<'py> for Graph {
...
+}
+
+impl Graph {
+    fn try_from_stable<Ty: EdgeType>(
+        py: Python<'_>,
+        dir: Direction,
+        pygraph: &StablePyGraph<Ty>,
+        attrs: &PyObject,
+    ) -> PyResult<Self> {
+        let mut attrs: Option<std::collections::HashMap<String, Value>> = attrs.extract(py).ok();
```
IvanIsCoding: Use HashMap from https://crates.io/crates/hashbrown or our own `DictMap` (which is just an alias for `IndexMap` with a fast hasher)

src/graphml.rs
IvanIsCoding: DictMap is just IndexMap with a faster hasher. It should be safe to replace

src/graphml.rs
```diff
@@ -756,3 +1172,104 @@ pub fn read_graphml<'py>(
...
+        }
+    }
+}
+
+type GraphMLWithKeys<'py> = PyResult<(Vec<Py<KeySpec>>, Vec<Bound<'py, PyAny>>)>;
+
+/// Read a list of graphs from a file in GraphML format and return the pair containing the list of key definitions and the graph.
+#[pyfunction]
+#[pyo3(signature=(path, compression=None),text_signature = "(path, /, compression=None)")]
+pub fn read_graphml_with_keys<'py>(
```
IvanIsCoding: I'd rather not have this function, let's just keep the `read_graphml` function as is

src/graphml.rs
```diff
@@ -756,3 +1172,104 @@ pub fn read_graphml<'py>(
...
+
+    Ok((keys, out))
+}
+
+/// Write a list of graphs to a file in GraphML format given the list of key definitions.
+#[pyfunction]
+#[pyo3(signature=(graphs, keys, path, compression=None),text_signature = "(graphs, keys, path, /, compression=None)")]
+pub fn write_graphml(
+    py: Python<'_>,
+    graphs: Vec<Py<PyAny>>,
```
IvanIsCoding: Let's make `write_graphml` take only one graph. Split the function to take either a PyGraph and a PyDiGraph. Name it `digraph_write_graphml` and `graph_write_graphml`. Then add an universal wrapper in https://github.com/Qiskit/rustworkx/blob/main/rustworkx/__init__.py.

I'd rather not have to deal with mixed PyGraph/PyDiGraph inputs. Also, I feel that calling `write_graphml(graph)`  leading to a failure is not expected. Let's focus on consistency with the rest of the API.

tests/test_graphml.py
```diff
@@ -10,6 +10,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import math
```
IvanIsCoding: Excellent test coverage!
* Remove the tests for serializing multiple graphs, as currently I am not inclined to maintain that functionality.

src/graphml.rs
```diff
@@ -756,3 +1172,104 @@ pub fn read_graphml<'py>(
...
+    keys: Vec<Py<KeySpec>>,
```
IvanIsCoding: Requiring this argument is very hostile to users, specially because it's not necessarily required. I am sure the specification says that users could specify keys. If you could make it an optional argument with a sensible default, that would be best option.

src/graphml.rs
```diff
@@ -756,3 +1172,104 @@ pub fn read_graphml<'py>(
...
+            name,
+            ty,
+            default,
+        }
+    }
+}
+
+type GraphMLWithKeys<'py> = PyResult<(Vec<Py<KeySpec>>, Vec<Bound<'py, PyAny>>)>;
+
+/// Read a list of graphs from a file in GraphML format and return the pair containing the list of key definitions and the graph.
```
IvanIsCoding: You're going to need to add a docstring explaining each parameter to help users call this function. Do not assume they understand each parameter

IvanIsCoding (commented):
Overall, this is pretty close to being merged but some comments on my opinion of supporting GraphML.

- For `read_graphml`, we don't know which library wrote it so we should do our best to support most things specified in the standard. 
- For `write_graphml`, I think it's ok to have a simplified version. As long as we write  a subset that is valid by the standard and is understood by other libraries, we should be fine. I don't think we should support all the expressiviness of GraphML like serializing multiple graphs.

My vision for this PR is to have code where `rustworkx.write_graphml(graph, "graph.xml")` just works.

IvanIsCoding:
Another comment for `DictMap` vs `IndexMap` vs `HashMap`:
* `HashMap` has a non-deterministic iteration order
* `IndexMap` preservers insertion order when iterating
* `DictMap` is just an alias we have for `IndexMap` with a faster hasher

Why am I mentioning this? For this PR it doesn't matter and it's something we can fix later. But ideally, if we called `rustworkx.write_graphml` twice for the same graph, we'd want to get an identical file. With `HashMap`, that might not be true as the order within the XML could be swapped! Keep that in mind.

src/graphml.rs
thierry-martinez: This function is removed in de6b1a4.

src/graphml.rs
thierry-martinez: I implemented the changes that you described in de6b1a4. Thanks!

tests/test_graphml.py
thierry-martinez: The test `test_write_multiple_graphs` has been removed in de6b1a4.

src/graphml.rs
thierry-martinez: `keys` is now an optional argument: if it is absent, it is inferred from the `graph` (passing `keys` explicitly can still be useful to set attributes' ids and default values).

src/graphml.rs
thierry-martinez: The function `read_graph_with_keys` is now removed, as requested.

src/graphml.rs
thierry-martinez: The intention was for the failure cause to be known and owned by the caller, who can inject the error context that makes sense for their usage. As shown in [this usage](https://github.com/Qiskit/rustworkx/blob/2d51ff352eb6a7911b2000e3c3442b6213f44dd2/src/graphml.rs#L790), the function returns a `Result<..., ()>`, and the responsibility of interpreting or enriching the error lies with the calling code. Typing ensures that this is done properly, since there is no `From<()>` defined for the error types we use.

rustworkx/__init__.py
```diff
@@ -6,6 +6,7 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
+from __future__ import annotations
```
IvanIsCoding:
```suggestion
```

rustworkx/__init__.py
```diff
@@ -2279,3 +2280,17 @@ def single_source_all_shortest_paths(
...
+    path: str,
+    /,
+    keys: list[GraphMLKey] | None = None,
+    compression: str | None = None,
+) -> None:
+    """ """
+    if isinstance(graph, PyGraph):
+        graph_write_graphml(graph, path, keys, compression)
+        return
+    digraph_write_graphml(graph, path, keys, compression)
```
IvanIsCoding:
```suggestion
@_rustworkx_dispatch
def write_graphml(graph, path, /, keys, compression):
    """ """
    raise TypeError(f"Invalid Input Type {type(graph)} for graph")
```

rustworkx/__init__.py
```diff
@@ -2279,3 +2279,9 @@ def single_source_all_shortest_paths(
         For most use cases, consider using `dijkstra_shortest_paths` for a single shortest path, which runs much faster.
     """
     raise TypeError(f"Invalid Input Type {type(graph)} for graph")
+
+
+@_rustworkx_dispatch
+def write_graphml(graph, path, /, keys, compression):
```
IvanIsCoding:
```suggestion
def write_graphml(graph, path, /, keys=None, compression=None):
```

src/graphml.rs
```diff
@@ -756,3 +1260,66 @@ pub fn read_graphml<'py>(
...
+            name,
+            ty,
+            default,
+        }
+    }
+}
+
+/// Write a graph to a file in GraphML format given the list of key definitions.
+#[pyfunction]
+#[pyo3(signature=(graph, path, keys, compression=None),text_signature = "(graph, path, /, keys=None, compression=None)")]
```
IvanIsCoding:
```suggestion
#[pyo3(signature=(graph, path, keys=None, compression=None),text_signature = "(graph, path, /, keys=None, compression=None)")]
```

src/graphml.rs
```diff
@@ -756,3 +1260,66 @@ pub fn read_graphml<'py>(
...
+    let mut graphml = GraphML::default();
+    graphml.graphs.push(Graph::try_from(graph.bind(py))?);
+    graphml.set_or_infer_keys(py, keys)?;
+    graphml.to_file(path, &compression.unwrap_or_default())?;
+    Ok(())
+}
+
+/// Write a digraph to a file in GraphML format given the list of key definitions.
+#[pyfunction]
+#[pyo3(signature=(graph, path, keys, compression=None),text_signature = "(graph, path, /, keys=None, compression=None)")]
```
IvanIsCoding:
```suggestion
#[pyo3(signature=(graph, path, keys=None, compression=None),text_signature = "(graph, path, /, keys=None, compression=None)")]
```

PR #1467, opened 2025-06-21
Adding quickcheck for barbell graphs

Krishn1412:
This QuickCheck test verifies structural properties of barbell_graph.

It ensures the total number of nodes is 2 × mesh_size + path_size.

It checks the number of edges matches the expected sum of two cliques, path edges, and connectors.


IvanIsCoding (approved):
Excellent!

PR #1470, opened 2025-06-25
Adding Quickcheck for binomial tree graph.

Krishn1412:
The quickcheck verifies the structure of binomial tree graph by checking number of node, number of edges and if they are bidirectional.

I plan on adding quickchecks for all the deterministic graph generators and some other algos like bipartite colouring before moving forward.


IvanIsCoding (approved):
I like the plan! Keep up with it

PR #1473, opened 2025-07-01
Adding Quickcheck for cycle graph

Krishn1412:
We are testing the structural property of cycle graph by ensuring:

Node count is exactly n

Edge count is:

n for directed, non-bidirectional
2n for directed, bidirectional
n for undirected

All expected edges are present

In bidirectional mode: for every (u, v) edge, there’s also a (v, u) edge


IvanIsCoding (approved):
LGTM, thanks!

PR #1478, opened 2025-07-06
New function greedy routing

SILIZ4:
This adds the general greedy routing algorithm to rustworkx-core and the version in hyperbolic space in rustworkx. In the process, I also provided other distance functions. I also transferred the hyperbolic distance function from the random graph module to the (new) geometry module.

This algorithm is mainly useful with hyperbolic embeddings, where [it was shown](https://journals.aps.org/pre/abstract/10.1103/PhysRevE.82.036106) that greedy routing paths are often close to the actual shortest paths in the graph. This is why I didn't export other versions of greedy routing to the Python API.

On a side note, to successfully generate the documentation locally I have to change the `earliest_version` in `releasenotes/config.yaml` (e.g. 0.11.0) otherwise I get a ValueError:

> ValueError: earliest-version set to unknown revision '0.8.0'

I haven't checked if that happened to someone else, but I can open an issue if there isn't a quick fix (and it's also not relevant to the pull request).


IvanIsCoding:
I will take a look at this after we release 0.17.0. My plan is to publish this with 0.17.1 or 0.18.0.

With regards to:
```
ValueError: earliest-version set to unknown revision '0.8.0'
```

I think reno uses the git commit history, maybe your fork has a shallow copy? I will investigate it after

PR #1481, opened 2025-07-09
Make Barabasi-Albert's generator return the same graph for the same seed

### LINKED ISSUE #1480: Using barabasi_albert_graph with a random seed
### Information
I'm trying to use the function barabasi_albert_graph with a specific seed in order to reproduce results, but I don't always get identical graphs.

- **rustworkx version**: 0.16.0
- **Python version**: 3.13.1
- **Rust version**: --
- **Operating system**: MacOS

### The current behavior
When passing a seed to the function barabasi_albert_graph (for reproducibility) it generates non-identical graphs.

### The expected behavior
The function barabasi_albert_graph will generate identical graphs when running it multiple times with the same seed.

Run this code a few times and you will see different results
```
from rustworkx import barabasi_albert_graph
sorted(barabasi_albert_graph(7, 2, seed=42).edge_list())
```
e.g. here is a screenshot of two different results (the last edge is not identical)
Image

### Additional note
I'm not sure if this is a bug, but when I run the function with `m=1` I get the error
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
(e.g. when running the line of code `barabasi_albert_graph(7, 1, seed=42)`)

IvanIsCoding:
Closes #1480

The easiest fix was to remove the additional source of randomess that has `HashSet` and replace it with `IndexSet`

### DIFF
--- a/rustworkx-core/src/shortest_path/bellman_ford.rs
+++ b/rustworkx-core/src/shortest_path/bellman_ford.rs
@@ -133,7 +133,7 @@ where
     }
 
     // Build path from predecessors
-    if path.is_some() {
+    if let Some(path) = &mut path {
         for node in graph.node_identifiers() {
             if scores.get_item(node).is_some() {
                 let mut node_path = Vec::<G::NodeId>::new();
@@ -144,7 +144,7 @@ where
                     node_path.push(current_node);
                 }
                 node_path.reverse();
-                path.as_mut().unwrap().insert(node, node_path);
+                path.insert(node, node_path);
             }
         }
     }
--- a/src/graphml.rs
+++ b/src/graphml.rs
@@ -186,7 +186,7 @@ enum Value {
 }
 
 impl Value {
-    fn serialize(&self) -> Option<Cow<str>> {
+    fn serialize(&self) -> Option<Cow<'_, str>> {
         match self {
             Value::Boolean(val) => Some(Cow::from(val.to_string())),
             Value::Int(val) => Some(Cow::from(val.to_string())),


rustworkx-core/src/generators/random_graph.rs
```diff
@@ -1402,6 +1402,27 @@ mod tests {
...
+            barabasi_albert_graph(7, 2, Some(42), None, || (), || ()).unwrap();
+        // Same seeds should yield the same graph
+        let edges1: std::collections::BTreeSet<_> = g1
+            .edge_references()
+            .map(|e| (e.source(), e.target()))
+            .collect();
+        let edges2: std::collections::BTreeSet<_> = g2
+            .edge_references()
+            .map(|e| (e.source(), e.target()))
+            .collect();
```
mtreinish, line 1420: If this test is supposed to check deterministic output with a fixed seed I would expect using a vec here because the btreeset will be iterating based on a sorted order:

```suggestion
        let edges1: Vec<_> = g1
            .edge_references()
            .map(|e| (e.source(), e.target()))
            .collect();
        let edges2: Vec<_> = g2
            .edge_references()
            .map(|e| (e.source(), e.target()))
            .collect();
```

Or is the exact creation order not expected to be deterministic and just the edges created?

mtreinish (commented):
Overall this looks good, the fix is straightforward. Just a quick question about the test and what are expectations around the functions determinism is.

rustworkx-core/src/generators/random_graph.rs
IvanIsCoding: I think it was just that the created edges were deterministic. In the GH issue, we got a report that we are yielding different edges.

mtreinish: Right, that was the reported issue, but I guess I'm asking what our expectations are here for determinism. Do we expect that the edge creation order should be deterministic so a graph that maintains insertion ordering can rely on it? If not I'm fine with this, but if we want to be more explicit here then I'd say we should change this to a vec.

IvanIsCoding: As long as the edges are the same, I think it’s fine! I don’t think we test edge insertion order on any randomized graphs

PR #1483, opened 2025-07-16
Adding quickcheck for karate club

Krishn1412:
Added quick to check the structural property of karate_club_graph, by ensuring the number of nodes, edges, there are no self loops and all nodes referenced are valid.


IvanIsCoding (approved):
LGTM, thanks

PR #1486, opened 2025-07-23
Adding quickcheck for star graph

Krishn1412:
Checking the number of nodes, edges, no self loops and all the edges pass through the central node.


IvanIsCoding (approved):
LGTM

Issue #1487, opened 2025-07-28
Expose multi-target simple paths in Python API


lazyhope:
### What is the expected enhancement?

The underlying Rust implementation for multi-target simple paths is not exposed in the Python API `rx.all_simple_paths`. This feature request proposes exposing this capability in the Python API for improved efficiency when we only need certain pairs of nodes.

**Current behaviour**:
```python
import rustworkx as rx

G = rx.generators.complete_graph(3)
rx.all_simple_paths(G, 0, {1, 2})
```
Output:
```python
File ~/myenv/lib/python3.11/functools.py:909, in singledispatch.<locals>.wrapper(*args, **kw)
    905 if not args:
    906     raise TypeError(f'{funcname} requires at least '
    907                     '1 positional argument')
--> 909 return dispatch(args[0].__class__)(*args, **kw)

TypeError: argument 'to': 'set' object cannot be interpreted as an integer
```

**Expected Output**:
```python
[[0, 1], [0, 1, 2], [0, 2], [0, 2, 1]]
```

Additionally, as #1088 has mentioned, the current document for `rx.all_simple_paths`'s parameter `from_` does not match the underlying parameter name `origin`, causing confusions when calling the function using keyword arguments. 

The `cutoff` parameter also has a mismatch with `networkx`'s implementation, e.g.
```python
import rustworkx as rx
import networkx as nx

rx_paths = {
    tuple(path)
    for path in rx.all_simple_paths(rx.generators.complete_graph(4), 0, 1, cutoff=3)
}
nx_paths_1 = {
    tuple(path) for path in nx.all_simple_paths(nx.complete_graph(4), 0, 1, cutoff=3)
}
nx_paths_2 = {
    tuple(path) for path in nx.all_simple_paths(nx.complete_graph(4), 0, 1, cutoff=2)
}
print(rx_paths == nx_paths_1)  # False
print(rx_paths == nx_paths_2)  # True
```

Thank you for the work!

lazyhope:
OK, I will try to see if I could craft a PR.

IvanIsCoding:
> Sorry I have to close this PR for now, because I noticed a significant performance drop after changing `to`'s type signature from `usize` to `&Bound<PyAny>`

You can probably make an enum that tries to convert first to int and then to a PyList. I’ll send a link to a similar snippet later

PR #1488, opened 2025-07-31
Allow multiple targets for rustworkx.all_simple_paths

lazyhope:
This PR enables multiple target being passed as iterable[int] to `rustworkx.all_simple_paths` as mentioned in #1487, the return type stays the same, similar to `networkx`'s implementation.

To ensure backward capabilities, the implementation for single target handling stays the same.

- [x] I ran rustfmt locally

### DIFF
--- a/src/connectivity/mod.rs
+++ b/src/connectivity/mod.rs
@@ -956,7 +956,7 @@ pub fn local_complement(
 ///
 /// :param PyGraph graph: The graph to find the path in
 /// :param int origin: The node index to find the paths from
-/// :param int to: The node index to find the paths to
+/// :param int | iterable[int] to: The node index(es) to find the paths to
 /// :param int min_depth: The minimum depth of the path to include in the output
 ///     list of paths. By default all paths are included regardless of depth,
 ///     setting to 0 will behave like the default.
@@ -971,7 +971,7 @@ pub fn local_complement(
 pub fn graph_all_simple_paths(
     graph: &graph::PyGraph,
     origin: usize,
-    to: usize,
+    to: &Bound<PyAny>,
     min_depth: Option<usize>,
     cutoff: Option<usize>,
 ) -> PyResult<Vec<Vec<usize>>> {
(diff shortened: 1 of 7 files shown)


src/connectivity/mod.rs
```diff
@@ -971,7 +971,7 @@ pub fn local_complement(
 pub fn graph_all_simple_paths(
     graph: &graph::PyGraph,
     origin: usize,
-    to: usize,
+    to: &Bound<PyAny>,
```
IvanIsCoding: Use the enum I described #1487

This way, handling the cases is a simple `match` statement shared by both methods

src/connectivity/mod.rs
```diff
@@ -981,27 +981,54 @@ pub fn graph_all_simple_paths(
...
+                to_index,
+                min_intermediate_nodes,
+                cutoff_petgraph,
+            )
+            .map(|v: Vec<NodeIndex>| v.into_iter().map(|i| i.index()).collect())
+            .collect();
+        return Ok(result);
+    }
+
+    let mut target_set: HashSet<NodeIndex> = HashSet::new();
```
IvanIsCoding: I think most users will pass in Python lists and expect output to be consistent. Use https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html with `foldhash` as the hasher.

src/connectivity/mod.rs
```diff
@@ -1035,27 +1062,53 @@ pub fn digraph_all_simple_paths(
             "The input index for 'from' is not a valid node index",
         ));
     }
-    let to_index = NodeIndex::new(to);
```
IvanIsCoding, line 1038: Same comments, mirror everything

IvanIsCoding (commented):
Thanks for sending this.

I left some comments on the Rust code. I will take a more careful look at the Python tessts later, but if you didn't break any of the existing test cases that is already a very good sign!

lazyhope:
Thanks for the review! Let me raise an issue to report bug and request multi-target feature in petgraph for now, as the current implementation introduce performance overhead when converting indexmap to flatten vector.

src/connectivity/mod.rs
lazyhope: `all_simple_paths_multiple_targets` is expecting `hashbrown::HashSet<petgraph::prelude::NodeIndex>` so using `IndexSet` will cause a type mismatch, any suggestion to get around this?

IvanIsCoding: Just ignore it then. We can sort the input if someone ever complains about it. As long as the behaviour is consistent with the `all_pairs_*` version, I don’t care

src/connectivity/mod.rs
lazyhope: Fixed in the new commit

IvanIsCoding (approved):
LGTM, thanks! We'll include it in the 0.17 release, I hope that means "very soon"

PR #1489, opened 2025-08-09
from_dot feature

Krishn1412:
Feat: Add from_dot parser support  
- Implement DOT file parsing to load graph structures  
- Support reading DOT content from string input  
- Added grammar using pest
- I ran the rust formatter and clippy
- I have added test coverage for the changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "adler2"
-version = "2.0.1"
+version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
 name = "aho-corasick"
@@ -25,18 +25,24 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "arbitrary"
-version = "1.4.2"
+version = "1.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c3d036a3c4ab069c7b410a2ce876bd74808d2d0888a82667669f8e783a898bf1"
+checksum = "dde20b3d026af13f561bdd0f15edf01fc734f0dafcedbaf42bba506a9517f223"
 dependencies = [
  "derive_arbitrary",
 ]
 
 [[package]]
 name = "autocfg"
-version = "1.5.0"
+version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
+checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
+
+[[package]]
+name = "bitflags"
+version = "2.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c8214115b7bf84099f1309324e63141d4c5d7cc26862f97a0a857dbefe165bd"
 
 [[package]]
 name = "block-buffer"
@@ -49,11 +55,10 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.36"
+version = "1.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5252b3d2648e5eedbc1a6f501e3c795e07025c1e93bbf8bbdd6eef7f447a6d54"
+checksum = "525046617d8376e3db1deffb079e91cef90a89fc3ca5c185bbf8c9ecdd15cd5c"
 dependencies = [
- "find-msvc-tools",
  "jobserver",
  "libc",
  "shlex",
@@ -61,9 +66,9 @@ dependencies = [
 
 [[package]]
 name = "cfg-if"
-version = "1.0.3"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2fd1289c04a9ea8cb22300a459a72a385d7c73d3259e2ed7dcb2af674838cfa9"
+checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "cpufeatures"
@@ -76,9 +81,9 @@ dependencies = [
 
 [[package]]
 name = "crc32fast"
-version = "1.5.0"
+version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9481c1c90cbf2ac953f07c8d4a58aa3945c425b7185c9154d67a65e4230da511"
+checksum = "a97769d94ddab943e4510d138150169a2758b5ef3eb191a9ee688de3e23ef7b3"
 dependencies = [
  "cfg-if",
 ]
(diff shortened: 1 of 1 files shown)


tests/graph/test_dot.py
```diff
@@ -128,3 +128,23 @@ def test_graph_no_args(self):
         graph = rustworkx.undirected_gnp_random_graph(3, 0.95, seed=24)
         dot_str = graph.to_dot()
         self.assertEqual("graph {\n0 ;\n1 ;\n2 ;\n2 -- 0 ;\n2 -- 1 ;\n}\n", dot_str)
+
```
IvanIsCoding, line 131: Can we add more tests, including:
* we go from PyGraph.to_dot and PyDiGraph.to_dot and back, covering multiple combinations of `node_attr`, `edge_attr` and `graph_attr`
*  A test case with an input string generated by https://networkx.org/documentation/stable/reference/generated/networkx.drawing.nx_pydot.write_dot.html

src/dot_parser/dot.pest
```diff
@@ -0,0 +1,43 @@
+WHITESPACE = _{ " " | "\t" | "\r" | "\n" }
```
IvanIsCoding, line 1: Did you write this file yourself or did you copy it from somewhere else? I'd add the LICENSE from the source as a comment

IvanIsCoding (commented):
This is an excellent start!

I will give it a deeper review, specially about graphviz's dot grammar which I need to understand more deeply.

I left some comments about tests and the grammar to get started

src/dot_parser/mod.rs
```diff
@@ -0,0 +1,299 @@
...
+
+                    Rule::assignment => {
+                        let mut parts = stmt.into_inner();
+                        let key = parts.next().map(|p| p.as_str()).unwrap_or("");
+                        let val = parts.next().map(|p| p.as_str()).unwrap_or("");
+                        graph_attrs.set_item(key, val)?;
+                    }
+
+                    Rule::subgraph => {
+                        // TODO: subgraph handling
```
IvanIsCoding: It's ok to return an error for subgraphs now. `dot` might be too expresive. We just want to make sure the output from a rustworkx entry can be re-imported as the original goal.

Once we have that, we can support more advanced features

src/dot_parser/mod.rs
```diff
@@ -0,0 +1,299 @@
...
+                graph,
+                cycle_state: rustworkx_core::petgraph::algo::DfsSpace::default(),
+                check_cycle: false,
+                node_removed: false,
+                multigraph: true,
+                attrs: graph_attrs.clone().into(),
+            });
+        } else {
+            let graph = StablePyGraph::<Undirected>::with_capacity(0, 0);
+            undi_graph = Some(PyGraph {
```
IvanIsCoding: Some thoughts about this handling:
* I'd make a function that is parametrized on StablePyGraph<DirectionTy> e.g. https://github.com/Qiskit/rustworkx/blob/f0240d424f105ada2b0b9af60335094ba69e715b/src/union.rs#L37
* Upon detecting the type, call the parametrization on the if/else

If you need to detect the direction of the graph, call https://docs.rs/petgraph/latest/petgraph/stable_graph/struct.StableGraph.html#method.is_directed within the parametrized function.

I think that choice will simplify your code, especially for handling both `di_graph` and `undi_graph` on every statement

src/dot_parser/mod.rs
```diff
@@ -0,0 +1,299 @@
...
+
+                        // create pairwise edges and add them to the graph
+                        for i in 0..endpoints.len().saturating_sub(1) {
+                            let src = endpoints[i].clone();
+                            let dst = endpoints[i + 1].clone();
+
+                            let src_idx = *node_map.entry(src.clone()).or_insert_with(|| {
+                                let py_node = PyString::new(py, &src).into();
+                                if is_directed {
+                                    di_graph.as_mut().unwrap().graph.add_node(py_node)
```
IvanIsCoding: If you follow the previous recommendation, this becomes `graph.add_node` which arguably is simpler that handling the `Option<PyGraph>` each time

src/dot_parser/dot.pest
Krishn1412: I took inspiration from the pydot parser but I wrote it myself.

src/dot_parser/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+use pest_derive::Parser;
+use pyo3::prelude::*;
+use pyo3::types::{PyDict, PyString};
+
+use crate::digraph::PyDiGraph;
+use crate::graph::PyGraph;
+use crate::StablePyGraph;
+
+use rustworkx_core::petgraph::prelude::{Directed, NodeIndex, Undirected};
+use std::collections::HashMap;
```
IvanIsCoding: Use the hashmap from hashbrown: https://docs.rs/hashbrown/latest/hashbrown/struct.HashMap.html

That's what we use everywhere

src/dot_parser/dot.pest
IvanIsCoding: Awesome!

src/dot_parser/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+    } else {
+        DotGraph::new_undirected()
+    };
+
+    for pair in pairs {
+        if pair.as_rule() != Rule::graph_file {
+            continue;
+        }
+        let mut inner = pair.into_inner();
+        let first = inner.next().unwrap();
```
IvanIsCoding: Let's avoid unwrapping if possible, do `inner.next()?`.

src/dot_parser/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+        for rest in inner {
+            if rest.as_rule() != Rule::stmt_list {
+                continue;
+            }
+
+            for stmt in rest.into_inner() {
+                match stmt.as_rule() {
+                    Rule::node_stmt => {
+                        let mut it = stmt.into_inner();
+                        let nid = it.next().unwrap();
```
IvanIsCoding: Same comment here. If you keep needing a conversion from pest errors to `PyResult`, make a helper closure and do a `.map_err` or something along those lines

src/dot_parser/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+                                        _ => {}
+                                    }
+                                }
+                            }
+                        }
+                    }
+
+                    Rule::assignment => {
+                        let mut parts = stmt.into_inner();
+                        let key = parts.next().map(|p| p.as_str()).unwrap_or("");
```
IvanIsCoding, line 282: This `unwrap_or` is fine. Keep it as is.

IvanIsCoding (commented):
I left some minor comments, but this is very close to getting merged:
* Use the hashmap from `hashbrown` for consistency. It should be slightly faster because of the hasher it uses vs `std` which uses a more conservative hasher
* Avoid panicking, be careful with `unwrap()`. Python users cannot recover easily from panic, they'd rather deal with exceptions.

And last but not least, don't forget to add a release note: https://github.com/Qiskit/rustworkx/blob/main/CONTRIBUTING.md#release-notes. I am planning on promoting this as one of the key new features in the 0.18.0 launch. People will want to read about it, it's handy!

IvanIsCoding (commented):
Also, I think you'll need to downgrade `pest` to compile with our MSRV.

I can also bump the MSRV in a separate PR

Cargo.toml
```diff
@@ -64,6 +64,8 @@ serde_json = "1.0"
 smallvec = { version = "1.0", features = ["union"] }
 rustworkx-core = { path = "rustworkx-core", version = "=0.17.0" }
 flate2 = "1.0.35"
+pest = "2.7"
```
IvanIsCoding: By the way, changes to Cargo.toml should also have a corresponding Cargo.lock change. Did you forget to commit Cargo.lock?

Krishn1412:
Can you take a look @IvanIsCoding

IvanIsCoding:
I will take a look. I also need to fix the Clippy lints with the release of Rust 1.90

IvanIsCoding (approved):
This is excellent, I will modify pest to work with our MSRV but this is good to go. I can take it from here

PR #1491, opened 2025-08-13
Remove invalid classifier and capitalize keywords

IvanIsCoding:
The current pyproject.toml is invalid and cannot be uploaded to pypi, unfortunately

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,9 +22,8 @@ classifiers=[
     "Operating System :: MacOS :: MacOS X",
     "Operating System :: Microsoft :: Windows",
     "Operating System :: POSIX :: Linux",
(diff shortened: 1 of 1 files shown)


mtreinish:
Do you want to upload an sdist to test.pypi.org to test that the metadata is valid. Fwiw we can probably just retag 0.17.0 or do a post release like 0.17.0.1 to fix the broken build without bumping to 0.17.1.

IvanIsCoding:
I opened https://github.com/Qiskit/rustworkx/actions/runs/16924903198 but indeed:
1. Our metadata is valid enough for PyPI to publish it
2. sdist is broken
3. I made too many mistakes, now we have 0.17.1 but only the binaries!

I will send a fix, we should have tested with test.pypi.org but now it's too late.

Issue #1492, opened 2025-08-13
sdist + retworkx publishing are broken (or: 0.17.1 has no source published in PyPI)


IvanIsCoding:
See #1491  and https://github.com/Qiskit/rustworkx/actions/runs/16924903198

PR #1494, opened 2025-08-13
Fix `pyproject.toml` and stop using `setup.py`

IvanIsCoding:
Related to #1492 

I tested this locally and it worked e.g. `python setup.py sdist`  passes.

For `retworkx`, it had more to do with the `dist` being in the wrong path


retworkx/setup.py
```diff
@@ -42,7 +42,7 @@
     long_description_content_type="text/markdown",
     author=pyproject["project"]["authors"][0]["name"],
     author_email=pyproject["project"]["authors"][0]["email"],
-    license=pyproject["project"]["license"],
+    license="Apache-2.0",
```
IvanIsCoding, line 45: In theory this is deprecated, but so is `retworkx` so I think it will be fine

mtreinish (commented):
We probably should move the sdist creation to using `python -m build --sdist` for rustworkx. 
 When I test this locally it's able to build the sdist without errors like the CI job had.
 
The build isolation from `build` probably won't work for retworkx because that's a bit more special but that's ok.

.github/workflows/wheels.yml
```diff
@@ -2,28 +2,15 @@
 name: Wheel Builds
 on:
   push:
-    tags:
-      - '*'
+    branches:
+      - 'stable/0.17'
```
mtreinish: You should undo this change, we don't want to forget about this. After this PR merges we can trigger the wheel build with a throwaway branch that we do as a one shot branched off of stable/0.17's HEAD. We'll also need to relax the release protection to enable triggering from a branch but we can do that in a one off.

IvanIsCoding: I added an `if` condition to check that the date is today, does that work?

IvanIsCoding:
@mtreinish  I updated rustworkx's command to be `python -m build --sdist`.

For the one-off branch: I don't want to create yet another branch. Maybe we can reuse `stable/0.17`? I just added a check for today's date. That way we cannot trigger it by accident on the future.

.github/workflows/wheels.yml
```diff
@@ -205,7 +47,7 @@ jobs:
         run: pushd retworkx && python setup.py bdist_wheel && popd
       - uses: actions/upload-artifact@v4
         with:
-          path: ./retworkx/dist/*
+          path: ./dist/*
```
mtreinish: This isn't actually correct. The wheel gets generated in `./retworkx/dist`

.github/workflows/wheels.yml
mtreinish: I pushed this branch up as a throwaway branch as I suggested and retriggered the sdist and rustworkx job. So the mising sdist and retworkx wheel are live now. We can revert the changes to this file related to manually triggering.

mtreinish (commented):
Overall this is about ready, but in the interest of getting the fix out I pushed up a quick throwaway branch with this branch to manually trigger a build. The fixes are live now on pypi: https://pypi.org/project/retworkx/ and https://pypi.org/project/rustworkx/0.17.1/#files

Do you want to retarget this PR to main and then we can backport it to `stable/0.17`?

IvanIsCoding:
@mergifyio backport stable/0.17

PR #1505, opened 2025-09-12
Fix pickling of graphs with contracted nodes

d-miketa:
Fixes https://github.com/Qiskit/rustworkx/issues/1503. Node contraction removes the contracted nodes but `self.node_removed` wasn't being set. Serialisation works differently depending on the flag. The failure reported in the issue was down to the wrong value of this flag.


IvanIsCoding:
Our code coverage service is down when I need to merge the PR. Don't worry too much about it, I will eventually handle it.

PR #1509, opened 2025-09-21
Remove Python 3.9 support and the legacy `retworkx` import

IvanIsCoding:
In short, this PR:

* Moves the build from 3.9 ABI to 3.10 ABI
* Removes Python 3.9 from GitHub actions
* Bumps Python in CI: 3.10->3.12 to avoid having to do bumps when 3.10 reaches EoL
* Removes the legacy `retworkx` shim
* Run tests with the Python 3.14 RC. I will add a release note for 3.14 once it's formally released and we are closer to the 0.18.0 release


IvanIsCoding:
I bumped `ruff` to lint for Python 3.10 and it found some fixes that are only available for 3.10+.

releasenotes/notes/retworkx-eol-bcbf18d805a73aed.yaml
```diff
@@ -0,0 +1,7 @@
+---
+deprecations:
```
mtreinish: I'd probably make this an `upgrade` note too.

.github/workflows/main.yml
```diff
@@ -70,20 +70,16 @@ jobs:
...
           - rust: 1.79.0
-            python-version: "3.10"
+            python-version: "3.12"
             platform: { os: "ubuntu-latest", python-architecture: "x64", rust-target: "x86_64-unknown-linux-gnu" }
             msrv: "MSRV"
-          # Test ARM + future versions of Rust and Python
+          # Test future versions of Rust and Python
           - rust: beta
-            python-version: "3.13" # upgrade to 3.14-dev when the release candidate is available
+            python-version: "3.14-dev" # upgrade to 3.15-dev when the release candidate is available
```
mtreinish: I know this is due to the delay from me but we should probably do this as:

```suggestion
            python-version: "3.14" # upgrade to 3.15-dev when the release candidate is available
```

now

.github/workflows/main.yml
```diff
@@ -61,7 +61,7 @@ jobs:
     strategy:
       matrix:
         rust: [stable]
-        python-version: [3.9, "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
```
mtreinish, line 64: Do you want to do this here or in a separate PR?
```suggestion
        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
```

releasenotes/notes/retworkx-eol-bcbf18d805a73aed.yaml
IvanIsCoding: Done.

.github/workflows/main.yml
IvanIsCoding: 3.14 will come in a separate PR, for now the Rust Beta + 3.14 will cover some of the testing

Cargo.toml
```diff
@@ -70,7 +70,7 @@ nalgebra-sparse = { version = "=0.10", features = ["io"] }
 
 [dependencies.pyo3]
 version = "0.26"
-features = ["abi3-py39", "extension-module", "hashbrown", "num-bigint", "num-complex", "indexmap", "py-clone"]
+features = ["abi3-py310", "extension-module", "hashbrown", "num-bigint", "num-complex", "indexmap", "py-clone"]
```
mtreinish, line 73: This might actually give us a little speedup, it might be worth benchmarking to see. But it won't be critical either way.

mtreinish (approved):
LGTM thanks for doing this.

PR #1510, opened 2025-09-21
[WIP] Bump MSRV to Rust 1.85, Edition to 2024, and PyO3 to 0.26

IvanIsCoding:
This is still WIP


rustworkx-core/src/centrality.rs
```diff
@@ -520,12 +520,12 @@ where
...
 mod test_edge_betweenness_centrality {
     use crate::centrality::edge_betweenness_centrality;
+    use petgraph::Undirected;
     use petgraph::graph::edge_index;
     use petgraph::prelude::StableGraph;
-    use petgraph::Undirected;
 
     macro_rules! assert_almost_equal {
-        ($x:expr, $y:expr, $d:expr) => {
+        ($x:expr_2021, $y:expr_2021, $d:expr_2021) => {
```
mtreinish: I don't think this change is strictly necessary. `cargo update` is a bit conservative for backwards compatibility, but the change in macro fragment specifiers doesn't really apply here or most of them. I'd suggest reverting this manually and the other places that got flagged so we minimize the diff on this PR and avoid unnecessary churn.

IvanIsCoding:
I will fix the macros but: every single PR conflicts  with edition 2024 and PyO3 0.26. So it was on the convenience of the author vs burden of the reviewer.

hopefully we merge this soon, sorry for the huge delta. But PyO3 renamed their most common struct

rustworkx-core/src/centrality.rs

PR #1515, opened 2025-10-09
Matrix Market support

Krishn1412:
- Added support for conversion of sparse and dense arrays to matrix market and vice versa using nalgrebra-sparse. [Ticket](https://github.com/Qiskit/rustworkx/issues/1513)
- I ran rustfmt locally
- I have added the tests to cover my changes.
- I have updated the documentation accordingly.
-  I have read the CONTRIBUTING document.

### DIFF
--- a/src/matrix_market/mod.rs
+++ b/src/matrix_market/mod.rs
@@ -15,10 +15,7 @@ use rustworkx_core::petgraph::{Directed, EdgeType, Undirected};
 type MatrixMarketData = (usize, usize, Vec<usize>, Vec<usize>, Vec<f64>);
 
 /// Convert any StablePyGraph (directed or undirected) into COO triplets
-fn graph_to_coo<Ty: EdgeType>(
-    _py: Python<'_>,
-    graph: &StablePyGraph<Ty>,
-) -> PyResult<MatrixMarketData> {
+fn graph_to_coo<Ty: EdgeType>(graph: &StablePyGraph<Ty>) -> PyResult<MatrixMarketData> {
     let mut rows = Vec::new();
     let mut cols = Vec::new();
     let mut vals = Vec::new();
@@ -105,7 +102,23 @@ fn coo_to_graph(py: Python<'_>, coo: &CooMatrix<f64>, is_directed: bool) -> PyRe
     }
 }
 
-/// Read Matrix Market from string contents and return the graph (PyGraph or PyDiGraph)
+/// Read a graph from Matrix Market format string contents.
+///
+/// Matrix Market is a human-readable ASCII file format for storing sparse matrices
+/// as coordinate format (row, column, value) triplets. This function parses Matrix
+/// Market formatted string contents and converts it to a graph representation.
+///
+/// For more information about Matrix Market format, see:
+/// https://math.nist.gov/MatrixMarket/formats.html
+///
+/// :param str contents: The Matrix Market formatted string contents to parse.
+///
+/// :returns: A graph object constructed from the Matrix Market data. Returns a
+///          :class:`~rustworkx.PyDiGraph` for directed matrices or a
+///          :class:`~rustworkx.PyGraph` for undirected matrices.
+/// :rtype: PyGraph or PyDiGraph
+///
+/// :raises ValueError: when the Matrix Market string contains invalid data or format.
 #[pyfunction]
 #[pyo3(signature=(contents,),text_signature = "(contents)")]
 pub fn read_matrix_market(py: Python<'_>, contents: &str) -> PyResult<PyObject> {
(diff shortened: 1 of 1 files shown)


rustworkx/rustworkx.pyi
```diff
@@ -754,6 +754,18 @@ def from_node_link_json_file(
 def from_dot(
     dot_str: str,
 ) -> PyDiGraph[_S, _T] | PyGraph[_S, _T]: ...
+def read_matrix_market_data(
```
IvanIsCoding: Name it `read_matrix_market`, make it read from the string contents

rustworkx/rustworkx.pyi
```diff
@@ -754,6 +754,18 @@ def from_node_link_json_file(
...
+    input: str,
+    from_file: bool,
```
IvanIsCoding: Make an equivalent method `read_matrix_market_file` instead. That method takes a path.

rustworkx/rustworkx.pyi
```diff
@@ -754,6 +754,18 @@ def from_node_link_json_file(
...
+) -> tuple[int, int, list[int], list[int], list[float]]: ...
```
IvanIsCoding: In my opinion, this should return a `PyGraph` or `PyDiGraph` object.

rustworkx/rustworkx.pyi
```diff
@@ -754,6 +754,18 @@ def from_node_link_json_file(
...
+def write_matrix_market_data(
+    nrows: int,
```
IvanIsCoding: Likewise, this method should take `PyDiGraph` or `PyGraph` as an input.

src/matrix_market/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+    // load COO matrix from file or string depending on the flag
+    let coo: CooMatrix<f64> = if from_file {
+        load_coo_from_matrix_market_file(input)
+            .map_err(|e| PyErr::new::<pyo3::exceptions::PyIOError, _>(format!("{}", e)))?
+    } else {
+        load_coo_from_matrix_market_str(input)
+            .map_err(|e| PyErr::new::<pyo3::exceptions::PyValueError, _>(format!("{}", e)))?
+    };
+
+    Ok((
```
IvanIsCoding: I'd say from here, try to detect if the graph is directed or undirected. Then, make the graph object and populate it by looping through the COO format.

src/matrix_market/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+    ))
+}
+
+#[pyfunction]
+pub fn write_matrix_market_data(
+    nrows: usize,
+    ncols: usize,
+    rows: Vec<usize>,
+    cols: Vec<usize>,
+    vals: Vec<f64>,
```
IvanIsCoding: As mentioned before, we'll need to take PyGraph or PyDiGraph as arguments. Nevertheless, I'd just make a helper method `graph_to_coo` that returns `nrows, ncols, rows, cols, vals`. Call the helper at the beginning and leave the rest of your code untouched.

tests/test_matrix_market.py
```diff
@@ -0,0 +1,149 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
```
IvanIsCoding: You'll need to update the test because they change the format of the data. You'll also need  to branch the tests into directed/undirected.

With that being said, I am happy with the test cases

tests/test_matrix_market.py
```diff
@@ -0,0 +1,149 @@
...
+# http://www.apache.org/licenses/LICENSE-2.0
+
+import os
```
IvanIsCoding: Prefer https://docs.python.org/3/library/pathlib.html over `os` for what you are doing. But I think you'll be able to avoid it entirely

tests/test_matrix_market.py
```diff
@@ -0,0 +1,149 @@
...
+        os.remove(self.path)
+
+    def test_write_to_string(self):
+        """Write COO data to a Matrix Market string."""
+        nrows, ncols = 3, 3
+        rows = [0, 1]
+        cols = [1, 2]
+        data = [1.0, 1.0]
+
+        mm_str = rustworkx.write_matrix_market_data(nrows, ncols, rows, cols, data, None)
```
IvanIsCoding: Do `with` statements with `tempfile`, using either temporary files or directories. See https://docs.python.org/3/library/tempfile.html#examples for examples

Krishn1412:
Hey @IvanIsCoding, apologies for the delay. I have made the changes as you suggested, I have a couple of doubts.

1) I couldn't find a definitive way to identify if a matrix file is directed or not, currently the logic checks if the graph is symmetric, then make it undirected. Is this alright or should we change this?

2) while passing the graph object in the write_matrix_market function, I couldn't make a python function generic, so I have split it into two. Is this implementation correct?

PR #1517, opened 2025-10-16
Bugfix/json preserves node gaps

d-miketa:
Addresses https://github.com/Qiskit/rustworkx/issues/1516: JSON roundtrip is broken whenever the graph contains deleted (or contracted) nodes.

This leaves just the GraphML format as incapable of a clean round-trip.

### DIFF
--- a/rustworkx/__init__.pyi
+++ b/rustworkx/__init__.pyi
@@ -695,6 +695,7 @@ def bfs_layers(
 ) -> list[list[int]]: ...
 def generate_random_path(
     graph: PyGraph | PyDiGraph,
+    /,
     source: int,
     length: int,
     seed: int | None = None,


IvanIsCoding:
I will take a look at this during the weekend. Thanks for sending the fix for an issue you found!

src/json/node_link_data.rs
```diff
@@ -83,22 +83,70 @@ pub fn parse_node_link_data<Ty: EdgeType>(
...
+        // Replace placeholder nodes with actual data and track which to keep
+        for node in graph.nodes {
+            let payload = match node.data {
+                Some(data) => match node_attrs {
+                    Some(ref callback) => callback.call1(*py, (data,))?,
+                    None => data.into_py_any(*py)?,
+                },
+                None => py.None(),
+            };
+            let node_id = node.id.unwrap_or(0);
```
IvanIsCoding:
```suggestion
            let node_id = node.id?;
```

I don't think we should default to 0 if it is not defined.

src/json/node_link_data.rs
```diff
@@ -83,22 +83,70 @@ pub fn parse_node_link_data<Ty: EdgeType>(
...
+    let node_removed = if preserve_ids {
+        // Find the maximum node ID to determine how many placeholder nodes we need
+        let max_id = graph.nodes.iter().filter_map(|n| n.id).max().unwrap_or(0);
+
+        // Create placeholder nodes up to max_id
+        let mut tmp_nodes: Vec<NodeIndex> = Vec::new();
+        for _ in 0..=max_id {
+            let idx = out_graph.add_node(py.None());
+            tmp_nodes.push(idx);
+        }
```
IvanIsCoding: Maybe a more idiomatic way of writing this:
```
let tmp_nodes: Vec<NodeIndex> = (0..=max_id)
    .map(|_| out_graph.add_node(py.None()))
    .collect();
```

src/json/node_link_data.rs
```diff
@@ -83,22 +83,70 @@ pub fn parse_node_link_data<Ty: EdgeType>(
...
+            let idx = NodeIndex::new(node_id);
+
+            // Replace the placeholder with actual data
+            if let Some(weight) = out_graph.node_weight_mut(idx) {
+                *weight = payload;
+            }
+
+            id_mapping.insert(node_id, idx);
+            // Mark this index as used (remove from tmp_nodes)
+            tmp_nodes.retain(|&n| n != idx);
```
IvanIsCoding: This can lead to quadratic behaviour, we are looping through all `tmp_nodes` to delete the index.

Please use https://docs.rs/hashbrown/latest/hashbrown/struct.HashSet.html instead

src/json/node_link_data.rs
```diff
@@ -83,22 +83,70 @@ pub fn parse_node_link_data<Ty: EdgeType>(
...
+        // Track if we're removing any nodes (indicates gaps in indices)
+        let has_gaps = !tmp_nodes.is_empty();
+
+        // Remove remaining placeholder nodes
+        for tmp_node in tmp_nodes {
+            out_graph.remove_node(tmp_node);
+        }
+
+        has_gaps
+    } else {
```
IvanIsCoding: nit picking: can you swap the if/else order? i.e. use `if ! preserve_ids`.

This is a style choice but I prefer to have the most common case first in if statements.

tests/graph/test_pickle.py
```diff
@@ -39,3 +39,46 @@ def test_weight_graph(self):
...
+        g.add_edge(3, contracted_idx, "3_to_contracted")
+
+        # Test pickle/unpickle
+        gprime = pickle.loads(pickle.dumps(g))
+
+        # Verify complete graph state is preserved
+        self.assertEqual(g.node_indices(), gprime.node_indices())
+        self.assertEqual(g.edge_list(), gprime.edge_list())
+        self.assertEqual(g.nodes(), gprime.nodes())
+        self.assertEqual(dict(g.edge_index_map()), dict(gprime.edge_index_map()))
```
IvanIsCoding: Of course you copied examples that don't use it but: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertDictEqual

This just has a nicer error message, in case someone ever introduces a bug

IvanIsCoding (commented):
Excellent test cases! I have no doubt this works, I just left some comments on what should be improved in Rust.

src/json/node_link_data.rs
IvanIsCoding: This suggestion also works with `HashSet` by the way

PR #1520, opened 2025-10-31
Improve all simple paths multi-target performance

lazyhope:
- [x] I ran rustfmt locally
- [x] I have added the tests to cover my changes. (reusing previous test)
- [x] I have updated the documentation accordingly. (not necessary)

This PR continues the work from https://github.com/Qiskit/rustworkx/pull/1488.

Now that multiple-target support for `all_simple_paths` has been added to `petgraph` (see [all_simple_path_multi](https://docs.rs/petgraph/latest/petgraph/algo/simple_paths/fn.all_simple_paths_multi.html) and https://github.com/petgraph/petgraph/pull/865), this PR simply update the underlying implementation from `rustworkx_core::connectivity::all_simple_paths_multiple_targets` to `petgraph::algo::all_simple_paths_multi`.

This avoids the need to flatten `DictMap` into a `Vec` and results in better performance and cleaner code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ fixedbitset = "0.5.7"
 indexmap = { version = ">=1.9, <3", features = ["rayon"] }
 ndarray = { version = "0.16.1", features = ["rayon"] }
 num-traits = "0.2"
-petgraph = "0.8.3"
+petgraph = "0.8"
 hashbrown = { version = ">=0.13, <0.16", features = ["rayon"] }
 numpy = "0.26"
 rand = "0.9"
--- a/releasenotes/notes/replace-api-for-all-simple-paths-0c44a6c18dbdaaf4.yaml
+++ b/releasenotes/notes/replace-api-for-all-simple-paths-0c44a6c18dbdaaf4.yaml
@@ -1,5 +1,4 @@
-upgrade:
+other:
   - |
-    The underlying implementation for all simple paths with multiple targets is
-    replaced with a new implementation from petgraph to improve performance.
-    This does not affect existing functionality.
+    Avoid unnecessary hashing operations in `rustworkx.all_simple_paths` with
+    multiple targets, improving performance for the use case.
--- a/src/connectivity/mod.rs
+++ b/src/connectivity/mod.rs
@@ -968,7 +968,7 @@ impl<'py> FromPyObject<'py> for TargetNodes {
         if let Ok(int) = ob.extract::<usize>() {
             Ok(Self::Single(NodeIndex::new(int)))
         } else {
-            let mut target_set: HashSet<NodeIndex, foldhash::fast::RandomState> = HashSet::new();
+            let mut target_set: HashSet<NodeIndex> = HashSet::new();
             for target in ob.try_iter()? {
                 let target_index = NodeIndex::new(target?.extract::<usize>()?);
                 target_set.insert(target_index);


Cargo.toml
```diff
@@ -30,7 +30,7 @@ fixedbitset = "0.5.7"
 indexmap = { version = ">=1.9, <3", features = ["rayon"] }
 ndarray = { version = "0.16.1", features = ["rayon"] }
 num-traits = "0.2"
-petgraph = "0.8"
+petgraph = "0.8.3"
```
IvanIsCoding: just keep the `Cargo.lock` change, `petgraph = "0.8"` still can resolve to 0.8.3.

releasenotes/notes/replace-api-for-all-simple-paths-0c44a6c18dbdaaf4.yaml
```diff
@@ -0,0 +1,5 @@
+upgrade:
```
IvanIsCoding: I'd rephrase this release note as `Avoid unnecessary hashing operations in all_simple_paths  with multiple targets`. Users appreciate it, but it's an implementation detail.

src/connectivity/mod.rs
```diff
@@ -968,7 +968,7 @@ impl<'py> FromPyObject<'py> for TargetNodes {
         if let Ok(int) = ob.extract::<usize>() {
             Ok(Self::Single(NodeIndex::new(int)))
         } else {
-            let mut target_set: HashSet<NodeIndex> = HashSet::new();
```
IvanIsCoding, line 971: revert this change as we already use https://docs.rs/hashbrown/latest/hashbrown/hash_set/struct.HashSet.html

`hashbrown` already uses foldhash by defuault AFAIK. and it can change with time, like with `ahash` -> `foldhash`.

releasenotes/notes/replace-api-for-all-simple-paths-0c44a6c18dbdaaf4.yaml
IvanIsCoding: Categorize as `other:`

lazyhope: Thanks for the review! I think the main improvement comes from avoiding flattening the hashset returned by the previous method, as the petgraph's api returns a `Vec`. What do you mean by "unnecessary hashing operations" though?

All other comments look good to me, I will resolve them soon.

IvanIsCoding: Your code is faster, but not for the reason you listed. The flattening of the hashset at the end is not the bottleneck. The biggest benefit comes from not using this line: https://github.com/Qiskit/rustworkx/blob/d1b43b6a2b85a33596f69f0e396e79eb69bc0e30/rustworkx-core/src/connectivity/all_simple_paths.rs#L96

Every node in a path goes through a hashmap look up to find the path. Your new code doesn't perform that lookup.

lazyhope: Thanks for noting that! I will change the release note then.

lazyhope:
> Firstly, thanks for the contribution. It seems you ported #577 to [petgraph/petgraph#865](https://github.com/petgraph/petgraph/pull/865) and made the optimizations.
> 
> It just needs some minor changes, but this should be good to go.

Thanks for the comments and they are now all resolved!

IvanIsCoding (approved):

jakelishman:
Answer: yes - #1552.

PR #1521, opened 2025-11-02
Officially support Python 3.14

IvanIsCoding:
Follow up of #1509. Until #1509 is merged, it contains the diffs of it. True diff: https://github.com/IvanIsCoding/rustworkx/compare/remove-39-retworkx...IvanIsCoding:rustworkx:upgrade-314

Python 3.14 already works, this PR just lists it as supported and adds it to CI. It also:
* Bumps cibuildhweel to a version that can test with 3.14
* Bumps `black` to a version that supports 3.14
*  Bumps `mypy` to a version that supports 3.14 as well
*  Allowlist some `mypy` findings that are not convenient
*  Bumps `typos` & fixes some typos
*  Bumps uv to the latest version

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,7 +92,7 @@ docs = [
     "pydot",
     "pillow>=4.2.1",
     "reno>=3.4.0",
-    "qiskit-sphinx-theme==1.16.1",
+    "qiskit-sphinx-theme==2.0.0",
     "matplotlib>=3.4",
     "sphinx-reredirects",
     "sphinxemoji",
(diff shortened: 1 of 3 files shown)


releasenotes/notes/py-314-d9385aa5b77cb478.yaml
```diff
@@ -0,0 +1,7 @@
+---
+upgrade:
```
mtreinish: I'd classify this as a "feature" or under "other" since it's new support. Upgrade is normally for something that will require users to potentially take action when upgrading.

```suggestion
features:
```

pyproject.toml
```diff
@@ -100,12 +100,12 @@ docs = [
     "lxml_html_clean",
 ]
 releaseinfra = [
-    "cibuildwheel==2.23.2",
+    "cibuildwheel==3.2.1; python_version >= '3.11'",
```
mtreinish: 3.3.0 is out now so we can use that. The changelog isn't major but might be nice to use that here too.

```suggestion
    "cibuildwheel==3.3.0; python_version >= '3.11'",
```

mtreinish (approved):

PR #1523, opened 2025-11-30
Replace `black` with `ruff`

IvanIsCoding:
This PR is stacked on top of #1521, for the true diff: https://github.com/IvanIsCoding/rustworkx/compare/upgrade-314...IvanIsCoding:rustworkx:ruff-format

`ruff format` is more mature now. We save one dependency in CI. The diff from `black` to `ruff` is only 15 files with very minor changes.


IvanIsCoding:
I rushed this in #1566 because the Black security warnings were annoying. The dependency was going to be replaced soon, regardless

PR #1524, opened 2025-12-01
Fix dominance tests

IvanIsCoding:
This solves CI errors like this:
```
    AssertionError: {2: 1, 6: 2, 4: 2, 3: 2, 5: 2} != {1: 1, 2: 1, 3: 2, 4: 2, 5: 2, 6: 2}
- {2: 1, 3: 2, 4: 2, 5: 2, 6: 2}
+ {1: 1, 2: 1, 3: 2, 4: 2, 5: 2, 6: 2}
```

NetworkX changed the behaviour, I settled for checking if a dictionary is a subset of another. That let's us ignore that `1:1` is no longer on the NetworkX result.

This is required to unblock #1509


mtreinish (approved):
I'm fine with this as a workaround for the change.

Issue #1251, opened 2024-07-14
Test `rustworkx` with free-threaded Python


IvanIsCoding:
### What is the expected enhancement?

Starting with 3.13, we should test `rustworkx` with https://py-free-threading.github.io/ to make sure it at least work with a single-thread setting. I don't think we should make our data structures thread-safe (that is a lot of overhead).

Of course this will heavily depend on PyO3 and their API offerings. But so far they have made progress on removing the GIL

IvanIsCoding:
And we are blocked until `NumPy` releases a new version supporting free threaded Python anyway: https://py-free-threading.github.io/tracking/.

gulldan:
https://github.com/numpy/numpy/releases/tag/v2.1.0

IvanIsCoding:
There's a chain of PRs needing to be merged but when we get to #1528, I think we'll be very close to shipping 3.14t support.

It might come either in 0.18.x or in the 0.19.x release series.

ngoldbaum:
@IvanIsCoding hi! I've been working on ecosystem support for free-threaded Python and am one of the main developers responsible for free-threaded Python support in NumPy and PyO3. Please let me know if you'd like a hand either with development work or with code review. I'm keen to have more green entries on https://hugovk.github.io/free-threaded-wheels.

IvanIsCoding:
> [@IvanIsCoding](https://github.com/IvanIsCoding) hi! I've been working on ecosystem support for free-threaded Python and am one of the main developers responsible for free-threaded Python support in NumPy and PyO3. Please let me know if you'd like a hand either with development work or with code review. I'm keen to have more green entries on https://hugovk.github.io/free-threaded-wheels.

Currently I am waiting for review of #1527 and #1528 by @mtreinish. But if you make a copy of them together (or you have a better solution), I can merge them earlier. I have been trying to release 0.18.0 at least with 3.14t support. And maybe a 0.18.1 with 3.15t support later this year.

PR #1527, opened 2025-12-14
Migrate from `setuptools-rust` to `maturin`

IvanIsCoding:
This is a step towards #1251. Although not strictly required, `maturin` just has a nicer integration with ABI 3 and free-threaded builds.

Currently, the main blocker for free-threaded support is that we declare we use ABI 3 in `pyproject.toml`. That fails for free-threaded builds. Maturin on the other hand is programmed to ignored ABI 3 in free-threaded builds because it knows there's no ABI for it yet.

3.14t inclusion to CI will come in a separate PR.


.github/workflows/main.yml
```diff
@@ -146,7 +146,14 @@ jobs:
...
         env:
-          SETUPTOOLS_RUST_CARGO_PROFILE: "dev"
+          CARGO_PROFILE_RELEASE_OPT_LEVEL: "0"
+          CARGO_PROFILE_RELEASE_DEBUG: "true"
+          CARGO_PROFILE_RELEASE_DEBUG_ASSERTIONS: "true"
+          CARGO_PROFILE_RELEASE_OVERFLOW_CHECKS: "true"
+          CARGO_PROFILE_RELEASE_LTO: "false"
+          CARGO_PROFILE_RELEASE_PANIC: "unwind"
+          CARGO_PROFILE_RELEASE_INCREMENTAL: "true"
+          CARGO_PROFILE_RELEASE_CODEGEN_UNITS: "16"
```
mtreinish: Wouldn't it be easier to set a cargo profile to dev and get most of this for free? From the maturin docs it looks like `MATURIN_PEP517_ARGS="--profile = dev"`

mtreinish (approved):
TBH, I've never been a huge fan of maturin even dating back to when it was called pyo3-pack, especially the parallel tooling/command we already have cargo and pip/uv/etc adding a 3rd seemed unnecessary. It's also always trying to do too much, like installing the rust compiler automagically, etc. Back in the pyo3-pack days there was no integration with Python's standard tooling but at least that's improved now with PEP517 builders so this PR integrates much more cleanly and it's mostly transparent.

That being said if we need this to handle being able to declare abi3 support and also supporting free threaded builds I'm fine moving forward with this. The one thing to think about is I think we have a pattern/model for #33 now. We've been working on doing this for Qiskit the pattern for handling this depends on setuptools-rust: https://github.com/Qiskit/qiskit/blob/main/setup.py. So we might need to switch back for that at some point in the future if/when we start pursuing #33.

I have one small inline question but not a huge deal.

.github/workflows/main.yml
IvanIsCoding: I completely missed that. Indeed, `MATURIN_PEP517_ARGS` is much cleaner. I updated the `noxfile.py` too

IvanIsCoding:
Re #33: that makes sense, having a more elaborate build setup would push us back to `setup.py`. I think the issue we got could have been fixed with us reintroducing `setup.py`, but using `maturin` was easier.

PR #1528, opened 2025-12-14
Add free-threaded Python to the CI

IvanIsCoding:
This PR is stacked on #1527, see https://github.com/IvanIsCoding/rustworkx/compare/free-threaded-314...IvanIsCoding:rustworkx:ft-314-again

Related to #1251 

This adds free-threaded builds to CI for Linux x86 and macOS ARM. Later we can decide on what to do for Linux ARM, Windows and macOS Intel. This is only for 3.14t, 3.13t is not included.

cibuildwheel is not yet configured to release 3.14t, but that can be tweaked in a follow up PR.


src/generators.rs
```diff
@@ -1778,7 +1778,7 @@ pub fn karate_club_graph(py: Python, multigraph: bool) -> PyResult<graph::PyGrap
     })
 }
 
-#[pymodule]
+#[pymodule(gil_used = false)]
```
mtreinish: Reading the docs on the attribute macro now, this should probably be annotated as a submodule too right?


```suggestion
#[pymodule(gil_used = false, submodule)]
```

src/lib.rs
```diff
@@ -462,7 +462,7 @@ create_exception!(
     "Graph is not bipartite"
 );
 
-#[pymodule]
+#[pymodule(gil_used = false)]
```
mtreinish: Is this actually true? We use python to get the exception types inside the function. I'm assuming this is fine since all the tests pass, but I wanted to ask about it.

That being said reading through the pyo3 docs this was made the default in 0.28. But we still need to upgrade our PyO3 version to get that.

mtreinish (commented):
This LGTM, it's really good to see us adding this to CI and getting free threaded builds working. I had one question inline and one small suggestion but nothing major.

src/lib.rs
mtreinish: I pushed up #1603 to use pyo3 0.29.0. If this merges first I'll update that PR to remove this argument.

src/generators.rs
IvanIsCoding: From https://docs.rs/pyo3/latest/pyo3/attr.pymodule.html: I think we have been doing it wrong all this time. But I added submodule

src/lib.rs
IvanIsCoding: Indeed, this is the default and now I can remove this diff!

mtreinish (approved):

PR #1535, opened 2026-01-22
Fix bug in `rustworkx.bridges`

### LINKED ISSUE #1534: .bridges does not identify the first edge in a graph as a bridge
### Information

- **rustworkx version**: 0.17.1
- **Python version**: 3.10.10
- **Rust version**: 1.86.0
- **Operating system**: Ubuntu 24.04.3

`rustworkx.bridges()` fails to identify the first edge in a graph as a bridge.

Specifically:
- In a 2-node graph with a single edge, bridges() returns an empty set.
- In a 3-node path graph (a–b–c), only the second of the two edges is reported as a bridge.
- Adding a dummy/sentinel node does not reliably fix the issue.

AFAIK, this contradicts the standard graph-theoretic definition of a bridge (an edge whose removal increases the number of connected components).

For an undirected simple graph:
- A graph with 2 nodes and 1 edge should report that edge as a bridge.
- A 3-node path graph should report both edges as bridges.

```python
import rustworkx as rx

g = rx.PyGraph(multigraph=False)
a = g.add_node("a")
b = g.add_node("b")
g.add_edge(a, b, None)

print(rx.bridges(g))
# Current output: set()
# Expected output: {(0, 1)} (or equivalent)
```

```python
import rustworkx as rx

# Case 2: 3-node path graph
g = rx.PyGraph(multigraph=False)
a = g.add_node("a")
b = g.add_node("b")
c = g.add_node("c")
g.add_edge(a, b, None)
g.add_edge(b, c, None)

print(rx.bridges(g))
# Current output: {(1, 2)}
# Expected output: {(0, 1), (1, 2)}
```

Adding a sentinel (in case this is some issue with index 0) does not fix the problem:

```python
import rustworkx as rx

# Case 1: 2 nodes, 1 edge
g = rx.PyGraph(multigraph=False)
sentinel = g.add_node(None)
a = g.add_node("a")
b = g.add_node("b")
g.add_edge(a, b, None)

print(rx.bridges(g))
# Current output: set()
# Expected output: {(1, 2)} (or equivalent)
```

IvanIsCoding:
Closes #1534 

Of course there can be bridges connected to the root.  The bug was easy to miss.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,7 +28,7 @@ jobs:
           python-version: "3.12"
       - run: |
           pip install -U --group lint
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@1.92 # TODO: unpin clippy
         with:
           components: rustfmt, clippy
       - name: Test Build


mtreinish (approved):
Oh that's a good catch, I'd probably have missed this too

PR #1536, opened 2026-01-27
add random walk sampling algorithm

SILIZ4:
This PR adds functions to sample a random walk on directed and undirected graphs. This should probably be extended in the future to include tunable edge weights.


IvanIsCoding:
I will take a look at this. The failure is from Clippy, it should be easy to fix

Issue #1501, opened 2025-09-11
Randomized iteration order for `PyGraph.neighbors`


adamglos92:
### Information

- **rustworkx version**: 0.17.1
- **Python version**: 3.12.11
- **Rust version**: (I believe it is not installed as I'm only using python version) 
- **Operating system**: Linux aarch64 (devcontainer in macOS arm64)

### What is the current behavior?

For a fixed graph I'm getting neighbors in random order.  

### What is the expected behavior?

Arbitrary, but fixed order 

Judging on implementation, set is the problem (but I'm no Rust expert)
```
    pub fn neighbors(&self, node: usize) -> NodeIndices {
        NodeIndices {
            nodes: self
                .graph
                .neighbors(NodeIndex::new(node))
                .map(|node| node.index())
                .collect::<HashSet<usize>>()
                .drain()
                .collect(),
        }
    }
```

### Steps to reproduce the problem

```
import rustworkx.generators as gen

for _ in range(3):
    g = gen.path_graph(5)
    print(g.neighbors(2))
```

PR #1546, opened 2026-01-31
1545: Make max weight matching deterministic for graphs with multiple correct solutions

jberg5:
Addresses https://github.com/Qiskit/rustworkx/issues/1545.

`max_weight_matching` was nondeterministic for some graphs with multiple valid solutions. See the tests I added, or the linked issue, for some examples of this.

I evaluated two choices:
- calling sort_unstable() on best_edge_to.values()
- using a BTreeMap instead of a HashMap

I did some benchmarking on various problem sizes and found that sorting the values of a HashMap caused a roughly 19% slowdown overall, especially for smaller problems. BTreeMap performance though looks pretty much identical modulo some noise:

image

I think determinism is a really nice property to have for a lot of different applications, and if we can achieve it without a performance penalty, it's worth it.

**AI disclaimer**:
I had claude come up with a script to find small examples of nondeterministic graphs; the examples in the tests come from there. (The example that I originally detected nondeterminism with had 60 nodes and thousands of edges and would have made a poor test case!)

I also had claude write the benchmarking script I used. I'd be happy to share it if there's interest. Nothing fancy; it just came up with some random graphs with different numbers of nodes and edges, and used matplotlib to generate a visualization.

### DIFF
--- a/tests/graph/test_max_weight_matching.py
+++ b/tests/graph/test_max_weight_matching.py
@@ -26,17 +26,14 @@ def match_dict_to_set(match):
 
 
 class TestMaxWeightMatching(unittest.TestCase):
-    def compare_match_sets(self, rx_match, expected_match, extra_failure_text=""):
+    def compare_match_sets(self, rx_match, expected_match):
         for u, v in rx_match:
             if (u, v) not in expected_match and (v, u) not in expected_match:
-                failure_text = (
+                self.fail(
                     f"Element {(u, v)} and its reverse {(v, u)} not found in "
                     f"expected output.\nrustworkx output: {rx_match}\nexpected "
                     f"output: {expected_match}"
                 )
-                if extra_failure_text:
-                    failure_text += f"\n{extra_failure_text}"
-                self.fail(failure_text)
 
     def compare_rx_nx_sets(self, rx_graph, rx_matches, nx_matches, seed, nx_graph):
         def get_rx_weight(edge):
@@ -469,17 +466,11 @@ class TestMaxWeightMatching(unittest.TestCase):
             graph, max_cardinality=True, weight_fn=lambda x: x, verify_optimum=True
         )
         for i in range(1, 10):
-            result = rustworkx.max_weight_matching(
-                graph, max_cardinality=True, weight_fn=lambda x: x, verify_optimum=True
-            )
-            self.compare_match_sets(
-                result,
-                initial_result,
-                extra_failure_text=(
-                    "Nondeterminism detected: repeating the same "
-                    f"problem {i} times yielded a different rustworkx solution"
-                ),
-            )
+            with self.subTest(f"Nondeterminism check {i}"):
+                result = rustworkx.max_weight_matching(
+                    graph, max_cardinality=True, weight_fn=lambda x: x, verify_optimum=True
+                )
+                self.compare_match_sets(result, initial_result)
 
     def test_determinism_six_node_no_max_cardinality(self):
         graph = rustworkx.PyGraph()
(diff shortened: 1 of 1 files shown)


rustworkx-core/src/max_weight_matching.rs
```diff
@@ -18,6 +18,7 @@
 #![allow(clippy::many_single_char_names)]
 
 use std::cmp::max;
+use std::collections::BTreeMap;
```
IvanIsCoding: https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html

Please use this instead. Changing to BTreeMap could change the time complexity of the algorithm

tests/graph/test_max_weight_matching.py
```diff
@@ -450,6 +453,57 @@ def test_nested_blossom_augmented(self):
...
+                (1, 4, 2),
+                (2, 4, 1),
+                (3, 4, 1),
+            ]
+        )
+
+        initial_result = rustworkx.max_weight_matching(
+            graph, max_cardinality=True, weight_fn=lambda x: x, verify_optimum=True
+        )
+        for i in range(1, 10):
```
IvanIsCoding, line 468: Use https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

IvanIsCoding (commented):
Overall I think determinism is welcome. But I don't think `BTreeMap` is the solution I want merged.

Another exercise before merging is revisiting other points of the algorithm to check if they are deterministic. Matthew wrote this ages ago. At the time we hadn't gotten the feedback from users about determinism. From time to time we still get that feedback (#1501).

rustworkx-core/src/max_weight_matching.rs
IvanIsCoding: Or https://docs.rs/rustworkx-core/0.17.1/rustworkx_core/dictmap/type.DictMap.html I mixed set vs map. We have our own alias for map

jberg5: Good call, will do!

tests/graph/test_max_weight_matching.py
jberg5: Sadly `stestr` doesn't seem to play nicely with subtests - see https://github.com/mtreinish/stestr/issues/326

I'm happy to still add `with self.subTest(i=i)` if you'd like, but the failure output is unchanged (no `(i=1)` anywhere to be seen), e.g. 
```
{0} graph.test_max_weight_matching.TestMaxWeightMatching.test_determinism_five_node_max_cardinality [0.000153s] ... FAILED

Captured traceback:
~~~~~~~~~~~~~~~~~~~

      File "/Users/jberg/Projects/rustworkx/tests/graph/test_max_weight_matching.py", line 478, in test_determinism_five_node_max_cardinality
    self.compare_match_sets(

      File "/Users/jberg/Projects/rustworkx/tests/graph/test_max_weight_matching.py", line 33, in compare_match_sets
    self.fail("oh no!")

      File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/unittest/case.py", line 715, in fail
    raise self.failureException(msg)

    AssertionError: oh no!

{0} graph.test_max_weight_matching.TestMaxWeightMatching.test_determinism_six_node_no_max_cardinality [0.000105s] ... FAILED

Captured traceback:
~~~~~~~~~~~~~~~~~~~

      File "/Users/jberg/Projects/rustworkx/tests/graph/test_max_weight_matching.py", line 505, in test_determinism_six_node_no_max_cardinality
    self.compare_match_sets(

      File "/Users/jberg/Projects/rustworkx/tests/graph/test_max_weight_matching.py", line 33, in compare_match_sets
    self.fail("oh no!")

      File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/unittest/case.py", line 715, in fail
    raise self.failureException(msg)

    AssertionError: oh no!
```

jberg5:
Thanks for the feedback and suggestions @IvanIsCoding ! Nothing else stands out to me as a source of additional nondeterminism. Just to be sure, I've been running the script I used to come up with the nondeterministic example graphs I added in the tests, and ever since switching to IndexMap it has run through 15 million random graphs so far without finding any further examples.

tests/graph/test_max_weight_matching.py
IvanIsCoding: Still keep it, there are other test runners out there. In the release we use the built-in unittest for example.

jberg5: Cool, sounds good! Have done so, and cleaned up a little as well.

IvanIsCoding (approved):

PR #1552, opened 2026-02-16
Split `hashbrown` dependency between Python-space and `rustworkx-core`

jakelishman:
`rustworkx-core` exposes `hashbrown` types in its own public API, which requires downstream crates match their resolved `hashbrown` versions to the one used by `rustworkx-core` in its API (up to potentially using a re-export of `hashbrown` from `rustworkx-core`, but one isn't provided, and it's not really scalable for `rustworkx-core` to claim it dictates the version of such a commonly used package).

The Python-space `rustworkx` package passes `hashbrown` types between PyO3 and `petgraph` interfaces, so requires hard matches to the resolved versions used by those libraries, but that's a tighter constraint.

These separate concerns motivate splitting the allowed bounds on `hashbrown`, so that downstream Rust-space consumers of `rustworkx-core` can use newer versions of `hashbrown` without being held back by the Python-space package.  For example, with this commit, Qiskit can now upgrade its internal use to `hashbrown 0.16.1`, including extracting those objects from (a newer) PyO3 and passing them to `rustworkx-core`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,6 @@ indexmap = { version = ">=1.9, <3", features = ["rayon"] }
 ndarray = { version = "0.16.1", features = ["rayon"] }
 num-traits = "0.2"
 petgraph = "0.8"
-hashbrown = { version = ">=0.13, <0.16", features = ["rayon"] }
 numpy = "0.26"
 rand = "0.9"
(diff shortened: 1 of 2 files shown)


IvanIsCoding:
I will take a look at this shortly

PR #1555, opened 2026-02-18
Use ubuntu-24.04-ppc64le and ubuntu-24.04-s390x runners for wheels.yml

IvanIsCoding:
Related to https://github.com/IBM/actionspz/issues/82

This is not strictly required for the 0.18 release, but it would greatly improve the QoL by avoiding issues with ppc64le and s390x. I expect the same behavior that happened when we added the `ubuntu-24.04-arm` runner: the build went from over 1 hour to minutes.

We need to install the GitHub App before merging this as described in https://github.com/IBM/actionspz/issues/82#issuecomment-3915001420


IvanIsCoding:
This needs to incorporate the learnings from #1568, I will update it later

IvanIsCoding:
> This needs to incorporate the learnings from #1568, I will update it later

Ok, I will need to test this in a separate PR because this is the only repo with access to the runners but:
1. I am copying the idea of virtual environments from https://github.com/Qiskit/rustworkx/pull/1568
2. We use a specific version of Python with uv. I was afraid system Python wasn't going to be enough.
3. s390x and ppc64le publishing are done separately because I don't want to discover wheter PyPA's action runs for ppc64le/s390x. I will assume it doesn't. We can upload from x86-64.

Future releases might take more bold moves such as adding tests, but given that 0.18 is already too big that is topic for another PR. This should already significantly improve the release timing, just like aarch64 did.

IvanIsCoding:
I'll check later but https://github.com/Qiskit/rustworkx/actions/runs/23701558744/job/69046081355?pr=1571 seems to be working!

update: dry-run for #1571 was successful

mtreinish (approved):
LGTM, thanks for working on this. I think the follow-on to this is to add a CI test job that is testing on these platforms. Probably as a nightly job though because I think we'll have to build numpy from source on both.

PR #1563, opened 2026-03-11
Reduce allocation overhead in dag_algo::longest_path

mtreinish:
This commit reduces the overhead from allocations in the longest_path function. Previously for each node being iterated over we were allocating a new vec and growing it as we pushed. This results in a lot of allocations and frees as we iterate over all the nodes. This commit address this overhead by moving to a single vec outside the loop and just clearing it on every iteration. This will reuse the allocation and minimize the overhead this causes.

### DIFF
--- a/rustworkx-core/src/dag_algo.rs
+++ b/rustworkx-core/src/dag_algo.rs
@@ -302,9 +302,10 @@ where
     let mut dist: HashMap<G::NodeId, (T, G::NodeId)> = HashMap::with_capacity(nodes.len()); // Stores the distance and the previous node
 
     // Iterate over nodes in topological order
+    let mut incoming_path: Vec<(T, G::NodeId)> = Vec::new();
     for node in nodes {
         let parents = graph.edges_directed(node, petgraph::Direction::Incoming);
-        let mut incoming_path: Vec<(T, G::NodeId)> = Vec::new(); // Stores the distance and the previous node for each parent
+        incoming_path.clear();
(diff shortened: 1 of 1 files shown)


rustworkx-core/src/dag_algo.rs
```diff
@@ -313,8 +314,9 @@ where
         }
         // Determine the maximum distance and corresponding parent node
         let max_path: (T, G::NodeId) = incoming_path
-            .into_iter()
+            .iter()
             .max_by(|a, b| a.0.partial_cmp(&b.0).unwrap())
+            .copied()
             .unwrap_or((T::zero(), node)); // If there are no incoming edges, the distance is zero
```
jakelishman, line 321: I don't even think you need a `Vec` at all, no?

Something like
```rust
for node in nodes {
    let max_path = graph
        .edges_directed(node, Incoming)
        .try_fold((T::zero(), node), |((max_length, prev), edge)| -> Result<_, E> {
            let source = edge.source();
            let length = dist[&source].0 + weight_fn(edge)?;
            if length > max_length {
                Ok((length, source))
            else {
                Ok((max_length, prev))
            }
        })?;
    // ...
```

jakelishman (approved):
btw I'm fine with the change as-is, whether or not you want to refactor to avoid _ever_ having a `Vec`

mtreinish:
Superseded by: https://github.com/Qiskit/rustworkx/pull/1577

PR #1568, opened 2026-03-22
Testing PowerPC and s390x runners

IvanIsCoding:
Do not merge this PR. It's just to check if the setup for #1555 works


IvanIsCoding:
This PR is interesting for two reasons:
1. https://github.com/IBM/actionspz/issues/82 is clearly working, we have access to native runners for PowerPC and s390x
2. Some of our dependencies don't build for s390x, I'll need to strip down the tests to numpy + rustworkx + networkx and skip things like Matplotlib

Nevertheless, this is a win!

PR #1576, opened 2026-04-14
Fix pagerank convergence threshold for large sparse graphs (#1575)

### LINKED ISSUE #1575: pagerank() silently returns uniform distribution on large sparse graphs due to N*tol convergence threshold
## Summary

`rustworkx.pagerank()` uses a convergence criterion `norm < N * tol` where `norm` is the L1 distance between successive iterates. Since L1 distance between probability vectors is bounded by 2, this threshold becomes useless once `N > 2/tol`. On large sparse graphs, the first iteration's L1 diff can be smaller than the threshold simply because mass barely moves in one step — the function then returns the **initial uniform vector** and reports convergence successfully.

The user gets a PageRank result that looks valid (sums to 1, no errors raised), but every node has score `1/N`. **Silent data corruption.**

## Environment

- `rustworkx` version: **0.17.1**
- Python: 3.12.8
- OS: Linux (Ubuntu 22.04)

## Minimal Reproduction

```python
import rustworkx as rx
import numpy as np

g = rx.PyDiGraph()
for i in range(2000):
    g.add_node(i)
g.add_edge(0, 1, 1.0)
g.add_edge(1, 2, 1.0)

pr = rx.pagerank(g, alpha=0.85)
vals = np.array([pr[i] for i in range(2000)])
print(f"All uniform? {np.allclose(vals, 1/2000)}")  # True — BUG
print(f"Node 2 score: {pr[2]:.6f}")                  # 0.000500 (uniform)
print(f"Expected:     0.001285 (2.6x above uniform)")
```

Output:
```
All uniform? True
Node 2 score: 0.000500
Expected:     0.001285 (2.6x above uniform)
```

## Where It Triggers

Tested with minimal repros:

| Case | Result |
|------|--------|
| N=10, 80% dangling | OK |
| N=1,000, 99.8% dangling | OK |
| N=2,000, 2 edges (99.9% dangling) | **FAIL — uniform output** |
| N=10,000, 2 edges | **FAIL — uniform output** |
| N=10,000, 50% dangling (5K edges) | OK |
| N=100,000, 2 edges | **FAIL — uniform output** |

Bug triggers when `N >= ~2000` and the active subgraph is sparse enough that the first power-iteration step produces an L1 diff below `N * tol`. We originally hit this on an 807,253-node talent migration graph (38% dangling) where `rx.pagerank()` silently returned uniform `1/N` for all nodes.

## Root Cause

`src/link_analysis.rs` lines 192-208:

```rust
let mut has_converged = false;
for _ in 0..max_iter {
    let dangling_sum: f64 = is_dangling
        .iter()
        .zip(popularity.iter())
        .map(|(cond, pop)| if *cond { *pop } else { 0.0 })
        .sum();
    let new_popularity =
        alpha * ((&a * &popularity) + (dangling_sum * &dangling_weights)) + &damping;
    let norm: f64 = new_popularity.l1_dist(&popularity).unwrap();
    if norm < (n as f64) * tol {   // <<<<<<< BUG HERE
        has_converged = true;
        break;
    } else {
        popularity = new_popularity;
    }
}
```

The threshold `(n as f64) * tol` means:

| N | tol | threshold | % of max L1 dist (=2) |
|---|-----|-----------|------------------------|
| 1,000 | 1e-6 | 0.001 | 0.05% |
| 10,000 | 1e-6 | 0.01 | 0.5% |
| 100,000 | 1e-6 | 0.1 | **5%** |
| 1,000,000 | 1e-6 | 1.0 | **50%** |

Any iteration whose L1 diff from the previous is less than `N * tol` is called "converged" — but for large N this happens trivially before any real convergence.

## Manual Trace

N=2000, 2 edges (path 0→1→2), alpha=0.85, tol=1e-6:

```
iter 0: L1 diff = 1.698e-03, threshold = 2.000e-03  ← converged! (but it isn't)
iter 1: L1 diff = 7.214e-04                          ← would have kept improving
iter 2: L1 diff = 1.226e-06
iter 3: L1 diff = 2.598e-07
iter 4: L1 diff = 6.631e-10                          ← actual convergence
```

The true converged values are `pr[2] = 0.00128`, `pr[1] = 0.00092`, `pr[0] = 0.00050`. Rustworkx returns `0.00050` for all three.

## Proposed Fix

Change line 202 from `norm < (n as f64) * tol` to `norm < tol`:

```diff
-        if norm < (n as f64) * tol {
+        if norm < tol {
             has_converged = true;
             break;
         } else {
```

This matches the docstring semantics ("error tolerance used when checking for convergence in the power method") and treats `tol` as an absolute L1 tolerance.

Happy to open a PR with the one-line fix + regression test.

## Related: NetworkX has the same formula

NetworkX's pure-Python pagerank also uses `err < N * tol`. We haven't verified whether NetworkX exhibits the same failure mode (our environment is missing `_bz2` so we couldn't run NetworkX side-by-side). If NetworkX also fails on this case, this is a broader issue worth fixing there too, but that's a separate discussion — the rustworkx fix stands on its own.

pathway:
## Summary

Fixes #1575 — a silent bug where `pagerank()` returns the initial uniform `1/N` distribution on large sparse graphs instead of the actual PageRank.

## Root Cause

The convergence check `norm < (n as f64) * tol` scales the L1 tolerance by graph size. Since L1 distance between probability vectors is bounded by 2, this threshold becomes useless once `N > 2/tol` (e.g. any N > 2000 with the default `tol=1e-6`). The first power-iteration step's L1 diff from the uniform starting vector can trivially fall below `n * tol`, causing the algorithm to report convergence and return the initial uniform vector — with no error or warning.

**Minimal repro** (included as regression test):

```python
import rustworkx as rx
import numpy as np

g = rx.PyDiGraph()
for _ in range(2000):
    g.add_node(None)
g.add_edge(0, 1, None)
g.add_edge(1, 2, None)

pr = rx.pagerank(g, alpha=0.85)
# Before this patch: pr[0] == pr[1] == pr[2] == 0.0005 (uniform 1/N)  ← BUG
# After this patch:  pr[2] ≈ 0.00128, pr[1] ≈ 0.00092, pr[0] ≈ 0.00050 ✓
```

## Change

One-line fix: `norm < (n as f64) * tol` → `norm < tol`. The `tol` parameter is now an absolute L1 tolerance.

## NetworkX parity (verified empirically)

NetworkX's `pagerank()` uses the same `err < N * tol` formula in both its scipy and pure-Python backends, and its docstrings explicitly document `len(G) * tol` semantics. So this PR is a **deliberate divergence from NetworkX**, not a fix to match it. Empirical comparison on the 2000-node repro:

| Implementation                       | pr[0]    | pr[1]    | pr[2]    |
|--------------------------------------|----------|----------|----------|
| Ground truth (`_pagerank_numpy`, eigendecomp) | 0.000499 | 0.000924 | **0.001285** |
| NetworkX `_pagerank_scipy`           | 0.000500 | 0.000925 | 0.000925 |
| NetworkX `_pagerank_python`          | 0.000500 | 0.000925 | 0.000925 |
| rustworkx 0.17.1 (current)           | 0.000500 | 0.000500 | 0.000500 |
| rustworkx + this patch               | ≈0.000500 | ≈0.000924 | ≈0.001285 |

Both libraries trip the loose threshold on iteration 0, but rustworkx is structurally worse: the loop assigns `popularity = new_popularity` only in the `else` branch (`src/link_analysis.rs:206`), so when the threshold trips it returns the *pre-iteration* vector — i.e. literally the initial uniform `1/N`. NetworkX returns the post-iteration vector, which at least carries one step of mass propagation. Same root cause, more catastrophic manifestation.

## Prior art in NetworkX (and why we should still diverge)

The same bug was reported in NetworkX as [networkx/networkx#4792](https://github.com/networkx/networkx/issues/4792) (Greg Hogan, 2021), with the same arguments. The NetworkX maintainers explicitly considered and **declined** the behavior change, taking the position that `tol` is "the per-node error tolerance" and that scaling by `N` provides "a reasonable default that works well for any size graph." The resolution was a docstring clarification ([networkx/networkx#6212](https://github.com/networkx/networkx/pull/6212), merged), not a fix.

I think rustworkx should still flip the semantic, for three reasons:

1. **Failure mode is materially worse here.** As shown above, rustworkx returns the *pre*-iteration vector when the loose threshold trips, so on the bug case the user receives the literal initial uniform `1/N` — a result that's not just imprecise but indistinguishable from "I never ran the algorithm." NetworkX's post-iteration return is wrong but at least reflects one step of propagation. NetworkX's "per-node tolerance" justification is harder to defend when the practical output is the input.
2. **rustworkx already has loud-failure precedent.** `FailedToConverge` is raised when `max_iter` is exhausted (`src/link_analysis.rs:211-215`), so a tighter threshold that occasionally raises rather than silently miscomputing is consistent with existing rustworkx error semantics.
3. **The rustworkx docstring previously claimed "matches NetworkX"** but is the only graph library I'm aware of that returns the *pre*-iteration vector — that was already a divergence, just an unintentional one. This PR makes the convergence semantics deliberately different and documents it as such.

I've also opened a separate, narrowly-scoped issue in NetworkX proposing a runtime warning when `N * tol >= 2` (the threshold is then mathematically incapable of gating anything, since L1 distance between probability vectors is bounded by 2). That's a different ask than the rejected behavior change and doesn't affect this PR either way.

## Backward compatibility / iteration impact

The new threshold is strictly tighter than the old one, so callers will see more power-iteration steps for the same `tol`. Empirically (Erdős-Rényi `N ∈ {100..10000}`, `α ∈ {0.85..0.999}`):

* Old threshold "converged" in 4–10 iterations — but on graphs where `N * tol > 2`, that convergence was illusory.
* New threshold needs ~10–20 iterations on the same graphs, comfortably within the default `max_iter=100`.
* On graphs where the old threshold tripped early but the true distribution wasn't reached, the new code may now exhaust `max_iter` and raise `FailedToConverge`. That's intentional: a loud, recoverable failure (raise `max_iter` or relax `tol`) is strictly better than silent corruption.

The release note now includes an `upgrade:` section spelling this out, plus a workaround (pass `tol=N*1e-6` to approximate the old behavior, with the caveat that the old behavior was often wrong even when not literally uniform).

## Testing

Added `test_sparse_large_graph_does_not_return_uniform` in `tests/digraph/test_pagerank.py`. Builds the 2000-node, 2-edge graph from the repro and asserts node 2 ≥ 2× uniform, node 1 ≥ 1.5× uniform, node 0 near uniform, isolated node 500 near uniform.

Existing pagerank tests should continue to pass: the new threshold is strictly tighter, so anywhere the old check converged correctly it still converges (just at the same iteration or later, never earlier).

## Release Note

`releasenotes/notes/fix-pagerank-convergence-threshold-1575.yaml` — includes both `fixes:` (the bug) and `upgrade:` (iteration-count and possible `FailedToConverge` impact).

## Notes

Discovered while computing personalized PageRank on an 807K-node talent migration graph where `rx.pagerank()` silently returned uniform `1/N` for all nodes.


IvanIsCoding:
As discussed in https://github.com/Qiskit/rustworkx/issues/1575, this introduces other problems. It will make things slower and cause `FailedToConverge` to pop more often. The comment has the description of a PR that would be accepted. This will be rejected

PR #1577, opened 2026-04-15
Avoid inner-loop allocations in `longest_path`

jakelishman:
The only use of the per-node `Vec` is to consume it immediately into a `max_by` iterator.  Rewriting the allocating loop as a single `try_fold` removes the unnecessary allocation.

On a test 1.6 million gate circuit over on Qiskit, this sped up our use of `longest_path` from 150ms to 50ms, so a 3x improvement.

### DIFF
--- a/rustworkx-core/src/dag_algo.rs
+++ b/rustworkx-core/src/dag_algo.rs
@@ -303,19 +303,18 @@ where
 
     // Iterate over nodes in topological order
     for node in nodes {
-        let parents = graph.edges_directed(node, petgraph::Direction::Incoming);
-        let mut incoming_path: Vec<(T, G::NodeId)> = Vec::new(); // Stores the distance and the previous node for each parent
-        for p_edge in parents {
-            let p_node = p_edge.source();
-            let weight: T = weight_fn(p_edge)?;
-            let length = dist[&p_node].0 + weight;
-            incoming_path.push((length, p_node));
-        }
-        // Determine the maximum distance and corresponding parent node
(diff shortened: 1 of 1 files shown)


jakelishman:
I got a question about super slow `DAGCircuit.depth` while talking to the India team, which reminded me we never merged this.  We're investing why they saw _such_ a terrible `DAGCircuit.depth` (it was ~70s which is mental regardless of DAG size), but regardless, this is a nice improvement to take.

I'll follow-up with a PR that lets us return the depth without calculating the path as well, which will help a bunch too.

rustworkx-core/src/dag_algo.rs
```diff
@@ -303,19 +303,18 @@ where
...
-            .into_iter()
-            .max_by(|a, b| a.0.partial_cmp(&b.0).unwrap())
-            .unwrap_or((T::zero(), node)); // If there are no incoming edges, the distance is zero
+        let max_path = graph
+            .edges_directed(node, petgraph::Direction::Incoming)
+            .try_fold((T::zero(), node), |longest, p_edge| -> Result<_, E> {
+                let p_node = p_edge.source();
+                let weight: T = weight_fn(p_edge)?;
+                let length = dist[&p_node].0 + weight;
+                if length >= longest.0 {
```
mtreinish, line 312: At first I was wondering if this was a different default on equality, but I checked the `std::cmp` docs and it uses the second argument if they're equal so this is good.

mtreinish (approved):
I"ll close out 1563 as this is a better solution

Issue #438, opened 2021-09-07
Add new layout methods


mtreinish:
### What is the expected enhancement?

Right now retworkx has several layout methods available: https://qiskit.org/documentation/retworkx/api.html#layout-functions
but there are some other common examples (like a planar layout and kamada kawai) missing. We should expand the available layout functions so that they can be leveraged with the drawers. We can look at what networkx offers for this functionality to get some good examples: https://networkx.org/documentation/stable/reference/drawing.html#module-networkx.drawing.layout

enavarro51:
**Add Planar Layout to retworkx**

The goal of this PR is to add a planar layout option to the other existing layout options in retworkx using the Rust language. A planar layout should provide better visualization of coupling maps in the gate_map module than existing options.

A planar graph can be displayed in the plane with no edges that intersect except at a node. As an example, a complete 4-node graph is planar since one of the diagonals can be placed outside the square, resulting in no intersecting edges. A complete 5-node graph is not planar.

![Graf_K4_v_rovině svg (1)](https://user-images.githubusercontent.com/16268251/166708800-2e3b7cae-72af-4da4-9414-c421eaa59dca.png)

Creating planar layouts in the general case is a complex problem and has been the subject of research over several decades. The process follows these steps.

1. Determine if the graph is planar. The method used in retworkx is the Left/Right planarity test from [`Ulrik Brandes:  The Left-Right Planarity Test (2009)`](http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.217.9208).

2. If the graph is planar, create a planar embedding, which is an intermediate representation of the graph in the plane. The embedding defines the relative positions of the vertices in the plane. The version we are creating is a combinatorial embedding which is defined in terms of the cyclic ordering of the edges to a vertex. This and the following step draw heavily on the existing networkx python code.
 
3. The final stage is to convert the combinatorial embedding to an actual position list for the nodes in the graph, which can then be displayed using matplotlib or graphviz.

Step 1 has been submitted as PR #475 by @georgios-ts and is awaiting review. Step 2 is about 60% complete and should be done within the next 2 weeks. Step 3 will follow.

a-matsuo:
[kamada_kawai_layout](https://networkx.org/documentation/stable/reference/generated/networkx.drawing.layout.kamada_kawai_layout.html) is very usefull to draw a heavy-hex coupling map.
Adding it would be helpful.
[...]

PR #1583, opened 2026-05-05
Add kamada_kawai_layout function

wallies:
Adds `rustworkx.kamada_kawai_layout`, supporting both `PyGraph` and `PyDiGraph` inputs. Implements the original Kamada-Kawai (1989) algorithm: an outer loop selects the node with the largest partial-gradient norm and an inner loop applies a 2D Newton step against the local Hessian until convergence.

Disconnected graphs are laid out by running Kamada-Kawai independently on each connected component and packing the components in a horizontal row, which avoids the visual collapse seen with single-objective Kamada-Kawai on disconnected inputs. Directed graphs have their distance matrix symmetrised before optimising (Kamada-Kawai is fundamentally undirected).

Related to #438

## Example outputs

Side-by-side comparisons against `networkx.kamada_kawai_layout` (NetworkX 3.6.1, which uses L-BFGS-B). Both produce K-K minima; differences are rotation/reflection of the same energy minimum.

### Florentine families (15 nodes, connected)

Florentine families comparison

Medici is centrally placed in both. Topologically equivalent under reflection.

### 3x4 hexagonal lattice (24 nodes, connected)

Hexagonal lattice comparison

Both recover the regular hexagonal grid structure.

### Three disconnected components: K4, C5, K3

Disconnected components comparison

NetworkX 3.6.1 solves K-K globally on the full distance matrix; components overlap when solved jointly (left). This implementation solves each component independently and packs them horizontally (right).


rustworkx/__init__.py
```diff
@@ -987,6 +987,113 @@ def networkx_converter(graph, keep_attributes: bool = False):
...
+        will be dictionaries with the node attributes from the input networkx
+        graph where the ``"__networkx_node__"`` key contains the node from the
+        input networkx graph.
+
+    :returns: A rustworkx graph, either a :class:`~rustworkx.PyDiGraph` or a
+        :class:`~rustworkx.PyGraph` based on whether the input graph is directed
+        or not.
+    :rtype: :class:`~rustworkx.PyDiGraph` or :class:`~rustworkx.PyGraph`
+    """
+    if graph.is_directed():
```
IvanIsCoding: This is dead code, remove it and make it raise a type error like the other functions.

If the body of this is reached, `@_rustworkx_dispatch` has already identified it is neither a `PyGraph` nor a `PyDiGraph`. We don't want to implicitly convert NetworkX graphs either.

src/layout/kamada_kawai.rs
```diff
@@ -0,0 +1,387 @@
...
+        }
+        edge_weights[e.id().index()] = Some(w);
+    }
+
+    let edge_cost = |e: petgraph::stable_graph::EdgeIndex| -> PyResult<Nt> {
+        edge_weights[e.index()].ok_or_else(|| PyValueError::new_err("Missing edge weight"))
+    };
+
+    let nodes: Vec<NodeIndex> = graph.node_indices().collect();
+    let rows: Vec<(usize, Vec<Option<Nt>>)> = nodes
```
IvanIsCoding, line 92: I'd say reuse https://www.rustworkx.org/apiref/rustworkx.all_pairs_dijkstra_path_lengths.html instead of manually unrolling things. Then you can focus on building your matrix. It makes the code cleaner. It should also handle negative weights for you.

src/layout/kamada_kawai.rs
```diff
@@ -0,0 +1,387 @@
...
+            }
+        }
+    }
+
+    if Ty::is_directed() {
+        for i in 0..n_bound {
+            for j in (i + 1)..n_bound {
+                let d_ij = matrix[i * n_bound + j];
+                let d_ji = matrix[j * n_bound + i];
+                let d = d_ij.min(d_ji);
```
IvanIsCoding: So this line is incorrect and the counter example are weakly connected graphs.

My suggestion: https://www.rustworkx.org/apiref/rustworkx.PyDiGraph.to_undirected.html#rustworkx-pydigraph-to-undirected, call this and pass it to the Dijkstra call as an undirected graph.

https://www.rustworkx.org/apiref/rustworkx.weakly_connected_components.html has an example of a weakly connected graph. Try seeing what the layout looks like for  graph with edges like`[(0, 1), (2, 1)]` in this code vs NetworkX's code

tests/digraph/test_layout.py
```diff
@@ -474,3 +474,86 @@ def test_spiral_layout_equidistant(self):
             9: (-1.0, 0.2018583081028111),
         }
         self.assertLayoutEquiv(expected, res)
+
+
+class TestKamadaKawaiLayout(LayoutTest):
```
IvanIsCoding, line 479: For digraphs, just remember to add a test for the weakly connected case: https://www.rustworkx.org/apiref/rustworkx.weakly_connected_components.html has a simple example, you might want to reuse one of the existing test cases and flip the edges to make things weakly connected.

src/layout/kamada_kawai.rs
```diff
@@ -0,0 +1,387 @@
...
+        }
+        x_offset += width + padding;
+    }
+}
+
+#[allow(clippy::too_many_arguments)]
+pub fn kamada_kawai_layout<Ty: EdgeType + Sync>(
+    py: Python,
+    graph: &StablePyGraph<Ty>,
+    pos: Option<HashMap<usize, Point>>,
```
IvanIsCoding: Use https://docs.rs/rustworkx-core/latest/rustworkx_core/dictmap/type.DictMap.html, `DictMap` is closer to Python's dictionary behavior. I am not implying this is required, but we got so many complains about dict order not being deterministic that I just stick to it by default. This comments applies everywhere.

IvanIsCoding (commented):
This is a great start, I just left some comments about:
1. Directed graphs
2. Some conventions followed by rustworkx

This is very close to getting merged, it would be a fantastic addition to 0.18.0.

n.b: I was on vacation, that was the reason behind the delay of the feedback. this is a quality contribution.

IvanIsCoding:
There’s a CI failure but nothing too hard. `nox -eformat` fixes it.

rustworkx/__init__.py
wallies: Removed

src/layout/kamada_kawai.rs
wallies: Used edge_weights_from_callable to consolidate the edge weight extraction. Going through all_pairs_dijkstra_path_lengths literally would require constructing a  Python callable to wrap the default_weight=None case, which adds plumbing rather than removing it. Happy to switch if you'd prefer the literal interpretation.

src/layout/kamada_kawai.rs
wallies: now builds an undirected StableUnGraph from the source topology and runs Dijkstra on it. Verified with 0->1<-2: d(0,2) is now 2 instead of Inf.

tests/digraph/test_layout.py
wallies: tests the 0->1<-2->3<-4 topology with assertions that d(0,4) > d(0,1).

src/layout/kamada_kawai.rs
wallies: Swapped pos. Kept fixed as HashSet since the only operations on it are contains and is_empty (matches spring_layout's precedent at src/layout/spring.rs). Happy to also switch fixed if you'd prefer.

IvanIsCoding (approved):

PR #1589, opened 2026-05-27
Add biadjacency matrix helpers

### LINKED ISSUE #1412: Add networkx's `from_biadjacency_matrix` and `biadjacency_matrix`

In NetworkX, I can create a graph based on a bipartite adjacency matrix using the [`from_biadjacency_matrix`](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.bipartite.matrix.from_biadjacency_matrix.html) function. Rustwork currently has no such functionality. Similarly, I cannot create a biadjacency matrix from a graph.

This would be tremendously helpful in applications where I have information on edges between two set of graph nodes stored in a non-sqaure matrix (=biadjacency matrix).

spital:
Fixes https://github.com/Qiskit/rustworkx/issues/1412

## Summary

This adds dense NumPy biadjacency matrix support for both graph types:

- `PyGraph.from_biadjacency_matrix()`
- `PyDiGraph.from_biadjacency_matrix()`
- `rustworkx.graph_biadjacency_matrix()`
- `rustworkx.digraph_biadjacency_matrix()`

For an input matrix with shape `(m, n)`, the constructors create `m + n` nodes, with row nodes first and column nodes second. Non-null entries create edges between the corresponding row an
d column nodes; for `PyDiGraph`, those edges are directed from row nodes to column nodes.

The output functions take explicit row and column node-index orders. These orders must contain existing node indices, must not contain duplicates, and must be disjoint. Parallel edges supp
ort the same `sum`, `min`, `max`, and `avg` reductions as the existing adjacency matrix helpers.

## Why

Issue #1412 requested NetworkX-like biadjacency matrix functionality for non-square bipartite matrix workflows. This fills that API gap while keeping the behavior explicit for rustworkx no
de-index based graphs.

## Notes for reviewers

- Added tests for construction, non-zero and `NaN` null values, dtype validation, empty dimensions, directed edge directionality, undirected reverse-edge handling, parallel-edge reductions
, invalid `parallel_edge`, missing/removed node indices, duplicate orders, overlapping orders, and empty output orders.
- Added type stubs, API docs entries, and a release note.
- Includes a small clippy-only cleanup in `rustworkx-core/src/planar/lr_planar.rs` so current `cargo clippy -D warnings` passes.

## Checks run

- `git diff --check`
- `cargo fmt --all -- --check`
- `cargo check --workspace --all-features`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `nox -e lint`
- `nox -e stubs`
- `nox -e test -- adjacency_matrix` — 85 passed
- `nox -e test` — 2410 run, 2399 passed, 11 skipped


rustworkx-core/src/planar/lr_planar.rs
```diff
@@ -280,13 +280,11 @@ where
...
                 self.lowpt_2.insert(ei, w_height);
             }
-            DfsEvent::BackEdge(v, w, _) => {
-                // do *not* consider ``(v, w)`` as a back edge if ``(w, v)`` is a tree edge.
-                if Some(&(w, v)) != self.eparent.get(&v) {
-                    let ei = (v, w);
-                    self.lowpt.insert(ei, self.height[&w]);
-                    self.lowpt_2.insert(ei, self.height[&v]);
-                }
+            // do *not* consider ``(v, w)`` as a back edge if ``(w, v)`` is a tree edge.
```
IvanIsCoding: Remove this diff it's unrelated to the PR

releasenotes/notes/add-biadjacency-matrix-1412.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Added :meth:`~rustworkx.PyGraph.from_biadjacency_matrix`,
+    :meth:`~rustworkx.PyDiGraph.from_biadjacency_matrix`,
+    :func:`~rustworkx.graph_biadjacency_matrix`, and
+    :func:`~rustworkx.digraph_biadjacency_matrix` for dense NumPy
```
IvanIsCoding: So document the individual methods for the `PyDiGraph` and `PyGraph` methods (e.g. `from_biadjacency_matrix` have different documentation entries). But for `biadjacency_matrix`, you should only note it once in the release with the universal function you are missing.

rustworkx/__init__.pyi
```diff
@@ -100,6 +100,8 @@ from .rustworkx import graph_condensation as graph_condensation
 from .rustworkx import weakly_connected_components as weakly_connected_components
 from .rustworkx import digraph_adjacency_matrix as digraph_adjacency_matrix
 from .rustworkx import graph_adjacency_matrix as graph_adjacency_matrix
+from .rustworkx import digraph_biadjacency_matrix as digraph_biadjacency_matrix
```
IvanIsCoding, line 103: A reminder to:
1. Add a universal function with `@_rustworkx_dispatch` that works with both graph types
2. Document that function stub here

src/connectivity/mod.rs
```diff
@@ -734,6 +734,178 @@ pub fn digraph_adjacency_matrix<'py>(
...
+fn apply_biadjacency_edge(
+    matrix: &mut Array2<f64>,
+    parallel_edge_count: &mut HashMap<[usize; 2], usize>,
+    row: usize,
+    column: usize,
+    edge_weight: f64,
+    parallel_edge_fn: ParallelEdgeFn,
+) {
+    let key = [row, column];
+    let count = parallel_edge_count.entry(key).or_insert(0);
```
IvanIsCoding: Use  a `match` with the `entry()` method: https://github.com/Qiskit/rustworkx/blob/9fb91e80ddeba9a2e994f918be942c682367a867/src/toposort.rs#L266

It will save one hashmap access. Those add up. The code is also cleaner.

src/connectivity/mod.rs
```diff
@@ -734,6 +734,178 @@ pub fn digraph_adjacency_matrix<'py>(
     Ok(matrix.into_pyarray(py))
 }
 
+type ParallelEdgeFn = fn(f64, f64, usize) -> f64;
```
IvanIsCoding: For code organization:
1. create a `biadjacency.rs` in this module and import it
2. move the duplicate code from `digraph.rs` and `graph.rs` here with generics as well

src/connectivity/mod.rs
```diff
@@ -734,6 +734,178 @@ pub fn digraph_adjacency_matrix<'py>(
...
+///
+/// :return: The biadjacency matrix for the input directed graph as a numpy array
+/// :rtype: numpy.ndarray
+#[pyfunction]
+#[pyo3(
+    signature=(graph, row_order, column_order, weight_fn=None, default_weight=1.0, null_value=0.0, parallel_edge="sum"),
+    text_signature = "(graph, row_order, column_order, /, weight_fn=None, default_weight=1.0, null_value=0.0, parallel_edge=\"sum\")"
+)]
+#[allow(clippy::too_many_arguments)]
+pub fn digraph_biadjacency_matrix<'py>(
```
IvanIsCoding, line 828: After you create the other file, keep the main function here still

src/connectivity/mod.rs
```diff
@@ -734,6 +734,178 @@ pub fn digraph_adjacency_matrix<'py>(
...
+    weight_fn: Option<Py<PyAny>>,
+    default_weight: f64,
+    null_value: f64,
+    parallel_edge: &str,
+) -> PyResult<Bound<'py, PyArray2<f64>>> {
+    validate_biadjacency_node_order(&graph.graph, &row_order, &column_order)?;
+    let parallel_edge_fn = parallel_edge_fn(parallel_edge)?;
+    let row_map = biadjacency_node_map(&row_order);
+    let column_map = biadjacency_node_map(&column_order);
+    let mut matrix = Array2::<f64>::from_elem((row_order.len(), column_order.len()), null_value);
```
IvanIsCoding: So this is the fatal flaw in the PR: this shouldn't be a NumPy array. This should be a `scipy` object, one of their sparse arrays.

You'll need to follow https://docs.rs/pyo3/0.26.0/pyo3/#example-using-python-from-rust to interact with Python objects from SciPy. Import the module, throw an error if it is not there, and then proceed.

src/digraph.rs
```diff
@@ -3622,6 +3651,50 @@ where
...
+    if null_value.is_nan() {
+        !elem.is_nan()
+    } else {
+        *elem != null_value
+    }
+}
+
+fn _from_biadjacency_matrix<'p, T>(
+    py: Python<'p>,
+    matrix: PyReadonlyArray2<'p, T>,
```
IvanIsCoding: Same flaw, deal with `scipy` objects. You can check how NetworkX handles it for a reference.

tests/digraph/test_adjacency_matrix.py
```diff
@@ -313,3 +313,153 @@ def test_parallel_edge(self):
             rustworkx.digraph_adjacency_matrix(
                 graph, weight_fn=lambda x: float(x), parallel_edge="error"
             )
+
+
+class TestDiGraphBiadjacencyMatrix(unittest.TestCase):
```
IvanIsCoding, line 388: As a reminder, tests will look different. https://github.com/Qiskit/rustworkx/pull/1221 has the old code we deleted but the summary is: tests are optional, if `scipy` is not present tests should not run. We should update `pyproject.toml` to only install `scipy` for a range of python versions and preferably skip Windows as well.

tests/graph/test_adjacency_matrix.py
```diff
@@ -311,3 +311,151 @@ def test_parallel_edge(self):
             rustworkx.graph_adjacency_matrix(
                 graph, weight_fn=lambda x: float(x), parallel_edge="error"
             )
+
+
+class TestGraphBiadjacencyMatrix(unittest.TestCase):
+    def test_from_biadjacency_matrix(self):
```
IvanIsCoding, line 387: Same comment about test setup

IvanIsCoding (commented):
So there are many comments. On the most immediate part, look at https://github.com/Qiskit/rustworkx/blob/main/rustworkx/__init__.py for a reference on the universal functions. And I left a couple Rust comments.

With that being said, the biggest flaw in the PR is that we don't return  or accept any types from https://docs.scipy.org/doc/scipy/reference/sparse.html. This is what users expect when dealing with sparse arrays.

You'll need to rewrite your whole code to not use NumPy and tweak tests such that `scipy` does not become a required dependency. It should be an optional dependency.

One way of decoupling our code from `scipy` is to leverage https://github.com/Qiskit/rustworkx/pull/1515. We can serialize to matrix market and then deserialize from it with `scipy` for `rustworkx.biadjacency_matrix`. And the other way around for `from_biadjacency_matrix`, ask `scipy` to export a format we can read and then load it.

rustworkx-core/src/planar/lr_planar.rs
spital: Thank you for comments @IvanIsCoding , looking to address them.
This particular change was due to `cargo clippy --workspace --all-targets --all-features -- -D warnings` failing for cargo 1.95; Rust 1.96 released recently.
I understand CI might pin lower version, but maybe useful for the future, would you find useful if I create new PR for this little change to fix future clippy ?

IvanIsCoding: Sure, feel free to bump the Rust versions for Clippy and the formatter. Just try to keep it separate from this PR

spital: Done in #1595

releasenotes/notes/add-biadjacency-matrix-1412.yaml
spital: Updated. The note now documents `PyGraph.from_biadjacency_matrix` and `PyDiGraph.from_biadjacency_matrix` individually

rustworkx/__init__.pyi
spital: Added. `rustworkx.biadjacency_matrix` now lives in `rustworkx/__init__.py` decorated with `@_rustworkx_dispatch`

src/connectivity/mod.rs
spital: Done, the parallel-edge accumulator now matches on `HashMap::entry()` with `Vacant` / `Occupied` arms

src/connectivity/mod.rs
spital: Done, added `src/connectivity/biadjacency.rs`

src/connectivity/mod.rs
spital: Done

src/connectivity/mod.rs
spital: Reworked. The export no longer produces a NumPy array.
It imports `scipy.sparse` through PyO3, builds a `coo_array` from the triplets and returns `.asformat(format)` with a new `format` parameter defaulting to `"csr"`. `null_value` is gone

src/digraph.rs
spital: Same treatment on the input side

tests/digraph/test_adjacency_matrix.py
spital: Updated

tests/graph/test_adjacency_matrix.py
spital: Same as the digraph side.
`skipIf` guard plus the `pyproject.toml` marker above

spital:
Thanks for the thorough review.
I rewrote the matrix side to use SciPy sparse arrays instead of dense NumPy, and addressed every inline comment:

- **SciPy sparse in/out, SciPy optional.** `biadjacency_matrix` builds a `scipy.sparse.coo_array` and returns `.asformat(format)` (default `"csr"`); `from_biadjacency_matrix` accepts any SciPy sparse matrix/array via `.tocoo()`. SciPy is imported at call time and raises a clear `ImportError` if it's missing, it is **not** a required runtime dependency.
- **Universal dispatcher.** Added `rustworkx.biadjacency_matrix` with `@_rustworkx_dispatch` plus its stub.
- **Code organization.** Shared logic moved to `src/connectivity/biadjacency.rs` with a generic `stable_graph_from_biadjacency_matrix<Ty: EdgeType>`; the public `#[pyfunction]`s stay in `connectivity/mod.rs`. Parallel-edge accumulation now uses `match` on `entry()`.
- **Tests optional.** Biadjacency tests `skipIf` SciPy is absent; `pyproject.toml` installs SciPy only for `python_version >= '3.11' and platform_system != 'Windows'`.
- **Rebased onto current `main` (0.18.0).** The diff is now purely additive and no longer touches `lr_planar.rs` (that bump is in #1595).

Two small design notes I'd appreciate your take on:
1. I kept a `parallel_edge` option (`"sum"` default, plus `min`/`max`/`avg`) for consistency with the existing `adjacency_matrix`. Happy to drop it if you'd rather mirror NetworkX exactly.
2. `graph_/digraph_biadjacency_matrix` currently require `row_order` and `column_order` to be disjoint (a clear `ValueError` otherwise). That's natural for bipartite use, but it's stricter than NetworkX. Let me know if you'd prefer to allow overlap (e.g. a directed square sub-matrix).

spital:
I checked this branch against current main locally. It merges cleanly, and the full tests, lint, stubs, docs, Rust formatting/checks, and Clippy all pass

IvanIsCoding:
Sorry I haven't had time to look at this. I will visit it first sometime next week.

PR #1594, opened 2026-05-30
perf: Use BufWriter when writing json

DanielVoogsgerd:
The current `node_link_data` function does not buffer its serialized JSON before writing it to a file; because of this, every emitted JSON token will trigger a syscall. This PR adds a `BufWriter` to mitigate this.

I benchmarked the performance difference on a MacBook Pro M1. The default capacity BufWriter was ~400 times faster than without the buffer, and increasing the buffer size to 64 KiB gives another 15% speed increase over the default buffer (~460 times faster).

### DIFF
--- a/src/json/node_link_data.rs
+++ b/src/json/node_link_data.rs
@@ -13,6 +13,7 @@
 use std::collections::BTreeMap;
 use std::fs::File;
 use std::io::BufWriter;
+use std::io::Write as _;
 
 use hashbrown::HashMap;
 use indexmap::IndexSet;
@@ -223,11 +224,16 @@ pub fn node_link_data<Ty: EdgeType>(
         },
         Some(filename) => {
             let file = File::create(filename)?;
-            let buffer = BufWriter::with_capacity(64 * 1024, file);
-            match serde_json::to_writer(buffer, &output_struct) {
-                Ok(_) => Ok(None),
+            let mut buffer = BufWriter::with_capacity(64 * 1024, file);
+            let res = match serde_json::to_writer(&mut buffer, &output_struct) {
+                Ok(_) => {
+                    buffer.flush()?;
+                    Ok(None)
+                }
                 Err(e) => Err(JSONSerializationError::new_err(format!("JSON Error: {e}"))),
-            }
+            }?;
+
+            Ok(res)
         }
     }
 }
--- a/src/json/node_link_data.rs
+++ b/src/json/node_link_data.rs
@@ -224,7 +224,7 @@ pub fn node_link_data<Ty: EdgeType>(
         },
         Some(filename) => {
             let file = File::create(filename)?;
-            let mut buffer = BufWriter::with_capacity(64 * 1024, file);
+            let mut buffer = BufWriter::new(file);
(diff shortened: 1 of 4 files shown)


src/json/node_link_data.rs
```diff
@@ -222,7 +223,8 @@ pub fn node_link_data<Ty: EdgeType>(
         },
         Some(filename) => {
             let file = File::create(filename)?;
-            match serde_json::to_writer(file, &output_struct) {
+            let buffer = BufWriter::with_capacity(64 * 1024, file);
+            match serde_json::to_writer(buffer, &output_struct) {
                 Ok(_) => Ok(None),
```
mtreinish: The `BufWriter` docs it's recommended to call `flush()` explicitly before dropping: https://doc.rust-lang.org/std/io/struct.BufWriter.html


```suggestion
                Ok(_) => {
                    buffer.flush().unwrap();
                    Ok(None)
                }
```

mtreinish (commented):
This LGTM thanks for doing this. I'm not sure why we used an unbuffered writer here originally, but doing this is clearly better. I just had one small inline comment.

src/json/node_link_data.rs
```diff
@@ -222,7 +223,8 @@ pub fn node_link_data<Ty: EdgeType>(
         },
         Some(filename) => {
             let file = File::create(filename)?;
-            match serde_json::to_writer(file, &output_struct) {
+            let buffer = BufWriter::with_capacity(64 * 1024, file);
```
IvanIsCoding: I don’t think we need to force the 64 MB? Just use the default parameters like in https://github.com/Qiskit/rustworkx/blob/f5789a9df1ed140e36a1a9d665c315bf13deb207/src/graphml.rs#L1156

IvanIsCoding (commented):
We should always use a buffer, this one slipped through a long time ago it seems.

Also, with regards to Matthew’s comment: it’s not a bad idea to flush, let’s see how other parts of the library use BufferedWriter

src/json/node_link_data.rs
DanielVoogsgerd: > increasing the buffer size to 64 KiB gives another 15% speed increase over the default buffer

It's 64 KiB compared to 8 KiB by default. I'll leave it up to you if this is worth it, but it's not an insignificant performance gain according to my benchmarks, but that will probably be hardware dependent anyway.

src/json/node_link_data.rs
DanielVoogsgerd: Ah yeah, I forgot, thanks for the heads-up.

IvanIsCoding (approved):

IvanIsCoding:
So I profited from this PR to buffer other methods that wrote to disk like `to_dot`.

With regards to 64kb vs 8kb (a.k.a. default) for the buffer: I did not dig too deep, but Apple Silicon seems to have bigger pages than x86-64 on Linux? http://techboards.net/threads/apple-silicon-16kb-page-size-benefits.4150/. It probably is faster on your machine, but 8kb should be better for the average user. Most of our downloads come from Linux.

All in all, thanks for reminding us. This is a trivial fix that will save many unnecessary flushes that had been happening for years.

PR #1605, opened 2026-06-15
Add longest_path_length to rustworkx_core::dag_algo

mtreinish:
This commit builds off of #1577 which improved the performance of the `longest_path` function to avoid allocating on each iteration of the loop. It was pointed out in that PR and #1563 that storing and returning a path for applications that only care about the length is unnecessary overhead. To provide an option for such cases a new function is added to offer a better alternative for these use cases.

### DIFF
--- a/releasenotes/notes/dag-longest-path-length-no-path-c91b30519ad22fcc.yaml
+++ b/releasenotes/notes/dag-longest-path-length-no-path-c91b30519ad22fcc.yaml
@@ -1,0 +1,7 @@
+---
+features:
+  - |
+    Added a new rustworkx-core function ``longest_path_length`` to the module ``dag_algo``. This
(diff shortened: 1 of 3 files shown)


mtreinish:
I forgot to put in the commit message, but the reason for the seeming duplication here is that internally I changed the typing of the `dist` variable from a `HashMap` to a `Vec` which trades off faster access and iteration speed at the cost of potential over allocation in the presence of holes in the graph. However, I didn't want to make this change to the existing function because it requires changing the trait bounds to add `NodeIndexable` trait which would have been an API change for the existing function.

IvanIsCoding (approved):
LGTM. The tests are good, existing test pass. It should be an easy merge. It is bound to be faster than the code we had before.

PR #1612, opened 2026-06-17
Publish Python 3.14 free-threaded wheels

### LINKED ISSUE #1251: Test `rustworkx` with free-threaded Python

IvanIsCoding:
Closes #1251 

Builds upon #1528. We ship what we test, with the exception of musl versions I guess? Windows is excluded because I'd rather make small steps. 0.18.1 or 0.19.0 can have it all.

I will trigger another PR to test this works

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,7 +133,7 @@ manylinux-i686-image = "manylinux2014"
 skip = "pp* cp314t-win* cp314t-*i686 cp314t-*s390x cp314t-*ppc64le *win32 *musllinux*i686"
 test-requires = "networkx"
 test-command = "python -m unittest discover {project}/tests"
-before-build = "pip install -U maturin>=1.9.0,<2.0"
+before-build = 'pip install -U "maturin>=1.9.0,<2.0"'
 test-skip = "*linux_s390x *ppc64le *i686 *win32"
 
 [tool.cibuildwheel.linux]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,7 +130,7 @@ extend-ignore-words-re = [
 [tool.cibuildwheel]
 manylinux-x86_64-image = "manylinux2014"
 manylinux-i686-image = "manylinux2014"
-skip = "pp* cp314t* *win32 *musllinux*i686"
+skip = "pp* cp314t-win* cp314t-*i686 cp314t-*s390x cp314t-*ppc64le *win32 *musllinux*i686"
 test-requires = "networkx"
 test-command = "python -m unittest discover {project}/tests"
 before-build = "pip install -U maturin>=1.9.0,<2.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -139,7 +139,7 @@ test-skip = "*linux_s390x *ppc64le *i686 *win32"
 [tool.cibuildwheel.linux]
 before-all = "yum install -y wget && {package}/tools/install_rust.sh"
 environment = 'PATH="$PATH:$HOME/.cargo/bin" CARGO_NET_GIT_FETCH_WITH_CLI="true"'
-repair-wheel-command = "auditwheel repair -w {dest_dir} {wheel} && pipx run abi3audit==0.0.9 --strict --report {wheel}"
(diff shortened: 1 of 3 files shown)


IvanIsCoding:
We also need this to fix the regular release, I guess. `maturin` needs quotes due to the versioning.

pyproject.toml
```diff
@@ -130,24 +130,24 @@ extend-ignore-words-re = [
...
 test-command = "python -m unittest discover {project}/tests"
-before-build = "pip install -U maturin>=1.9.0,<2.0"
+before-build = 'pip install -U "maturin>=1.9.0,<2.0"'
 test-skip = "*linux_s390x *ppc64le *i686 *win32"
 
 [tool.cibuildwheel.linux]
 before-all = "yum install -y wget && {package}/tools/install_rust.sh"
 environment = 'PATH="$PATH:$HOME/.cargo/bin" CARGO_NET_GIT_FETCH_WITH_CLI="true"'
-repair-wheel-command = "auditwheel repair -w {dest_dir} {wheel} && pipx run abi3audit==0.0.9 --strict --report {wheel}"
+repair-wheel-command = 'auditwheel repair -w {dest_dir} {wheel} && if echo "{wheel}" | grep -vq cp314t; then pipx run abi3audit==0.0.9 --strict --report {wheel}; else echo "Skipping abi3audit for free-threaded wheel"; fi'
```
IvanIsCoding, line 142: Yeah this is perhaps the trickiest part. I wasn't expecting abi3audit to fail with cp314t. This is a work around but I think for 0.18.0 it will do.

It will need to be revisited with 3.15t/abi3t

IvanIsCoding:
https://github.com/Qiskit/rustworkx/actions/runs/27664585648/job/81815830269 passed, I think this is good to go.

pyproject.toml
```diff
@@ -130,24 +130,24 @@ extend-ignore-words-re = [
 [tool.cibuildwheel]
 manylinux-x86_64-image = "manylinux2014"
 manylinux-i686-image = "manylinux2014"
-skip = "pp* cp314t* *win32 *musllinux*i686"
+skip = "pp* cp314t-win* cp314t-*i686 cp314t-*s390x cp314t-*ppc64le *win32 *musllinux*i686"
 test-requires = "networkx"
 test-command = "python -m unittest discover {project}/tests"
-before-build = "pip install -U maturin>=1.9.0,<2.0"
+before-build = 'pip install -U "maturin>=1.9.0,<2.0"'
```
mtreinish, line 136: I'm surprised we need to do this, I would have expected the builder to just pull this in automatically when cibuildwheel goes to build the wheel. It doesn't hurt to do it as an explicit step I guess though.

pyproject.toml
mtreinish: Yeah it's a bit surprising that abi3audit doesn't just return a non-error and skip if the tags indicate a non-abi3 wheel. It might be because we're using an old version here. The latest release is 0.0.26, 0.0.9 is from 2023 before free threaded builds IIRC. So maybe it just does not understand the 314t tag?

We can revisit this post-release since this won't block anything right now.

mtreinish (approved):
This LGTM in it's current form. I think there is some cleanup we can/should do post release but all the wheels are being built correctly so we can visit that post release.

ngoldbaum:
Thanks for getting this done and apologies for never finding the time to review.

IvanIsCoding:
> Thanks for getting this done and apologies for never finding the time to review.

No worries, thanks for advancing PyO3 in regards to free-threading compatibility.

I might send you a message asking for what an `abi3audit`  looks like with `abi3t`. Ideally it all just works out of the box? You don’t maintain the library but it is one of those ecosystem things.

ngoldbaum:
Honestly I'm not sure offhand. It's a good question though and I'll try to dig in more.

PR #1620, opened 2026-06-22
Publish experimental wheels for Pyodide

IvanIsCoding:
Following [PEP 783](https://peps.python.org/pep-0783/), we can publish the wheels. This is super new, I think we'd be one of the first libraries to actually publish Pyodide to PyPI.

With that being said, https://github.com/Qiskit/rustworkx/pull/1451 would be unblocked by this: `jupyterlite` would have a pinned version mapping to Pyodide 3.14.x. Pyodide 3.14.x would download the `rutsworkx_wasm-0.18.1-cp314-cp314-pyemscripten_2026_0_wasm32.whl` wheel from PyPI, and things would just work.

For 0.18.1, I do not want to change cibuildwheel for other wheels. So this wheel will be treated specially. I do think there might be some long-term discussion of what Pyodide versions will be supported. But as long as we don't promise too much and our docs become interactive, I think it is a a win.


IvanIsCoding:
edit: #1621 is now passing. I am cautiously optimistic about having 0.18.1 published on PyPI with Pyodide support

https://github.com/Qiskit/rustworkx/actions/runs/27928021872/job/82634272562

docs/source/install.rst
```diff
@@ -171,7 +171,7 @@ Tier Experimental
 -----------------
 
 Tier Experimental platforms are not tested upstream as part of the development process.
-Pre-compiled binaries are built by the external community in separate repositories. Not all of rustworkx might compile for
+Pre-compiled binaries are built and published to PyPI as part of the release process. Not all of rustworkx might compile for
```
mtreinish, line 174: Is the only distinction between this tier and tier 4 is that we might need nightly rust? Do you think we should just roll this into Tier 4 now?

mtreinish (commented):
Config-wise this looks correct to me, it matches the docs from cibuildwheel on how to do pyodide builds. Just one question inline about our support tiers.

docs/source/install.rst
IvanIsCoding: Threading is also disabled! I think that’s another major difference

IvanIsCoding:
I am going to keep an experimental tier with the reasoning “we feel free to disable any feature just to make it ship”. Right now, the feature is threading but if in the future something doesn’t work it gets ignored for experimental targets.

mtreinish (approved):

mtreinish:

PR #1630, opened 2026-07-08
Setup separate rustworkx-core publish job with trusted publishers

mtreinish:
This commit splits out the release publish job for the rustworkx-core crate to crates.io from the wheels file into a separate job workflow file. This then lets us setup the cargo publish job to use the trusted publishers mechanism instead of a secret token for authentication which is more secure. The docs for using trusted publisher are located here:

https://crates.io/docs/trusted-publishing

the new cargo_publish workflow's job is updated to more closely follow the example in the docs. I've already updated the crate's configuration on crates.io to use cargo_publish.yml for the trusted publisher mechanism.

### DIFF
--- a/.github/workflows/cargo_publish.yml
+++ b/.github/workflows/cargo_publish.yml
@@ -1,0 +1,23 @@
+---
+name: Publish rustworkx-core to crates.io
+on:
+  push:
+    tags:
+      - '*'
+jobs:
+  rustworkx-core:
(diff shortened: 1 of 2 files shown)


IvanIsCoding (approved):
Awesome! I will let you authorize the workflow on the crates.io side. LGTM

IvanIsCoding:

PR #1641, opened 2026-07-19
Fix undirected_gnm_random_graph so it does not return a multigraph

### LINKED ISSUE #1640: Bug: undirected_gnm_random_graph returns parallel edges
### Information

- **rustworkx version**: latest (built from `main`)
- **Python version**: 3.12
- **Rust version**: 1.93.0
- **Operating system**: macos

`undirected_gnm_random_graph` commonly returns graphs with parallel edges when `num_edges` is less than `n(n-1)/2` (i.e. graph is not complete and random edges are drawn).

Documentation says that it does *not* return a multigraph, but it does. https://www.rustworkx.org/apiref/rustworkx.undirected_gnm_random_graph.html

```python
import rustworkx as rx

# 20 nodes, 120 edges. Max simple undirected edges = 20*19/2 = 190, so this is
# dense enough that random reversed collisions are common.
g = rx.undirected_gnm_random_graph(20, 120, seed=0)

edges = g.edge_list()
distinct = {(min(u, v), max(u, v)) for u, v in edges}
dupes = len(edges) - len(distinct)

print(f"requested edges : 120")
print(f"num_edges()     : {g.num_edges()}")
print(f"distinct pairs  : {len(distinct)}")
print(f"parallel edges  : {dupes}")

# The adjacency matrix of a simple graph must be all 0/1.
A = rx.adjacency_matrix(g)
print(f"adjacency max   : {A.max()}  (must be 1.0 for a simple graph)")

assert dupes == 0, f"BUG: {dupes} parallel edges in a graph that should be simple"
assert A.max() == 1.0, f"BUG: adjacency matrix has entry {A.max()} > 1"
print("OK: graph is simple (no parallel edges)")
```

The issue is here:
https://github.com/Qiskit/rustworkx/blob/5ff5a9f762899338c3364ab5d637a30c654d9277/rustworkx-core/src/generators/random_graph.rs#L426

Finding existing edges doesn't use a normalized ordering for undirected edges, so duplicates are possible. Incidentally this is pretty inefficient, it loops over every edge; we could easily replace this with a hashmap and bring the complexity from O(m^2) to O(m) (where m in number of edges).

jberg5:
Fixes https://github.com/Qiskit/rustworkx/issues/1640

See issue for root cause / discussion.

### DIFF
--- a/releasenotes/notes/fix-undirected-gnm-parallel-edges-7b3c1e9a4f2d8065.yaml
+++ b/releasenotes/notes/fix-undirected-gnm-parallel-edges-7b3c1e9a4f2d8065.yaml
@@ -1,0 +1,13 @@
+---
+fixes:
+  - |
+    Fixed :func:`~rustworkx.undirected_gnm_random_graph` (and the corresponding
+    ``gnm_random_graph`` generator in ``rustworkx-core``) so that it no longer
+    returns a graph containing parallel edges. Previously a randomly drawn edge
+    ``(v, u)`` was not recognized as a duplicate of an already-added edge
+    ``(u, v)``, so undirected graphs could be returned with more than one edge
+    between the same pair of nodes, contradicting the documented behavior that
+    the generator does not produce a multigraph. This is also a performance
+    improvement - O(m) scan of edges replaced with O(1) set membership lookup.
+    See `#1640 <https://github.com/Qiskit/rustworkx/issues/1640>`__.
+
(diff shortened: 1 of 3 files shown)


jberg5:
Have added both Python test and Rust test, let me know if you like both, or one, or neither.

IvanIsCoding (approved):
I made some minor edits but this is good to go. I will backport it to 0.18.1, should that get released.

You've contributed before, the timelines are loose. But I look forward to seeing the bug fix live.

IvanIsCoding:

PR #1643, opened 2026-07-20
Publish PowerPC free-threaded wheels

### LINKED ISSUE #1639: Free-threading support for PowerPC
Hi, thanks for adding free-threading support for x86 and aarch64.
However, ppc64le is not supported. Are there any plans to add this?

IvanIsCoding:
Closes #1639. With the custom Power PC runner, I am confident this will work. I will test it before the release.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,7 +131,7 @@ extend-ignore-words-re = [
 [tool.cibuildwheel]
 manylinux-x86_64-image = "manylinux2014"
 manylinux-i686-image = "manylinux2014"
-skip = "pp* cp314t-win* cp314t-*i686 cp314t-*s390x cp314t-*ppc64le *win32 *musllinux*i686"
+skip = "pp* cp314t-win* cp314t-*i686 cp314t-*s390x *win32 *musllinux*i686"
 test-requires = "networkx"
 test-command = "python -m unittest discover {project}/tests"
 before-build = 'pip install -U "maturin>=1.9.0,<2.0"'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,7 +131,7 @@ extend-ignore-words-re = [
 [tool.cibuildwheel]
 manylinux-x86_64-image = "manylinux2014"
 manylinux-i686-image = "manylinux2014"
-skip = "pp* cp314t-win* cp314t-*i686 cp314t-*s390x *win32 *musllinux*i686"
+skip = "pp* cp314t-win* c*t-*i686 cp314t-*s390x *win32 *musllinux*i686"
 test-requires = "networkx"
 test-command = "python -m unittest discover {project}/tests"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -131,7 +131,7 @@ extend-ignore-words-re = [
 [tool.cibuildwheel]
 manylinux-x86_64-image = "manylinux2014"
 manylinux-i686-image = "manylinux2014"
-skip = "pp* cp314t-win* cp314t-*i686 cp314t-*s390x cp314t-*ppc64le *win32 *musllinux*i686"
+skip = "pp* cp314t-win* cp314t-*i686 cp314t-*s390x *win32 *musllinux*i686"
```
mtreinish: Would it makes sense to change these to `c*t-win`, `c*t-*i686`, `c*t-*s390x` if the tag changes on 3.15? Also if this is targeted for 0.19 could we just enable free threading on all our supported platforms (with i686 maybe being the exception because the python ecosystem has largely moved away from supporting 32bit x86 platforms).

IvanIsCoding: So I will update this but: we should keep skipping 3.14t because it will keep being built as abi3t will never support this. With that being said, from 3.15 onwards we will always have abi3t so the tag will be constant.

IvanIsCoding:
Alright, the strategy will be as follows with this PR merged:
* For everything we published free-threaded in 0.18.0 plus PowerPC, we will keep publishing 3.14t wheels. This will last a couple years, until 3.14t is EOL. We will also publish abi3t wheels starting from 3.15+
* For Windows and Linux s390x, we will never publish 3.14t wheels. However, we will publish abi3t wheels starting at 3.15. The
* For win32 and i686: no free-threaded wheels

I hope that makes sense. The reasoning is that abi3t will keep the maintenance burden lower for all platforms, it's just one extra wheel. But 3.14t will be left in that awkward state, people are actively using this version but it will not support abi3t

mtreinish (approved):
Yeah that makes sense. I'm not super worried about the extra wheels for 3.14 free threaded vs abi3 free threaded. It's still far less package files than we had in the pre-abi3 days.

mtreinish:
Did you want to test it before I hit merge?

As an aside we should probably add a manual trigger to the wheel build job. This is something we added on Qiskit based on a label. It's quite convenient to trigger the wheel building jobs (without pypi publish obviously) for debugging cibuildwheel configuration on PRs.

IvanIsCoding:
I will try to mimic Qiskit's setup in a future PR, then.

https://github.com/Qiskit/rustworkx/actions/runs/30454648607 is underway. Once we confirm it builds all wheels, I will merge the PR and backport.

IvanIsCoding:
I can’t believe PowerPC passed and we got an x86-64 failures. Well, I will investigate it more. The current change is valid

IvanIsCoding:

PR #1653, opened 2026-07-23
Prepare 0.18.1 release

mtreinish:
We've collected a fair number of bugfixes on top of 0.18.0 that would be good to release. This commit prepares for tagging a 0.18.1 release by bumping the version numbers and adding a prelude release note. Once this merges it should be tagged as 0.18.1.


IvanIsCoding (approved):
I think this is good to go, we do need to move some release notes but we have accumulated enough small things that it is worth releasing.

With that being said, I am going to be biased and say "can we try including this":
1. #1643, this closes #1639 and it's a trivial amount of work,
2. #1620, this is less trivial but it will let us stop depending on `pyodide-recipes` (https://github.com/pyodide/pyodide-recipes/issues/599).

IvanIsCoding:
Also, I fixed it the public docs via a manual commit but: https://github.com/Qiskit/rustworkx/blob/540441810f8b9f4bba1f1f5b9212ac0d293616e2/docs/source/conf.py#L29

Let's update the docs version, we forgot about that in 0.18.0. I will send a PR to avoid that kind of mistake in the future

edit: or alternatively just merge #1655 and backport. that works too.

IvanIsCoding:
Flagging from https://github.com/Qiskit/rustworkx/actions/runs/30454648607/job/90585000226

GCC from our pinned cibuildwheel is clashing with NumPy?! For some reason we are building NumPy from source and it’s failing for x86-64

mtreinish:
It's the use of manylinux 2014. Modern numpy wheels are built on manylinux 2_27 which is obviously not compatible with the manylinux 2010, so pip is trying to build it from source. But the gcc version we have in the manylinux container is too old now too so that fails. We should be able to set the `--prefer-binary` flag on the test install step so that it will just skip the newer versions of numpy that don't have manylinux compatible wheels. Where we'll have trouble is with newer python versions that don't have any wheels on a compatible manylinux tag. Obviously numpy 2.4.6 worked though since we could release 0.18.0 (before numpy 2.5.0 was released) so maybe we just pin to numpy < 2.5 in cibuildwheel for right now and look at bumping our manylinux version in 0.19.0.

IvanIsCoding:
For sanity, I ran another dry-run on my fork: https://github.com/IvanIsCoding/rustworkx/actions/runs/30496390043/job/90726120705

I think this is good to go.

PR #1655, opened 2026-07-26
Avoid having to manually bump documentation versions

IvanIsCoding:
See https://github.com/Qiskit/rustworkx/pull/1653#issuecomment-5085327017

These days, Python ships with built-in support to read TOML files, so I think we can avoid this all together.


mtreinish (approved):
This should be fine. I am slightly concerned about assuming the docs are built from a full checkout of the repo. I feel like there might be some packaging cases where this might not be the case. But if it's an issue for downstream packaging we can revisit this again, and not having to worry about this version number is a win.

IvanIsCoding:

PR #1657, opened 2026-07-29
DNM: Test wheel builds with numpy constraints

mtreinish:


mtreinish:
This worked so I applied it to #1653

PR #1660, opened 2026-07-29
Bump manylinux version to manylinux 2_28

mtreinish:
During the run up to the rustworkx 0.18.1 we realized that the latest numpy releases were using manylinux 2_28 for their published wheels while rustworkx was building with manylinux 2014. This was forcing rustworkx's wheel build job to build numpy from source as there was no compatible precompiled version available. Additionally, starting in numpy 2.5.0 the version of GCC required to build numpy was newer than the version included in the manylinux 2014 image. This commit opts to raise the manylinux version we use for building qiskit to manylinux 2_28 to bring it inline with numpy, our only python dependency. This means functionally the precompiled packages for rustworkx just require glibc 2.28 or newer (up from glibc 2.17 in manylinux 2014). Glibc 2.28 was released in 2018 so it is sufficiently old that most of our user base should already be on a compatible linux distro.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -129,8 +129,11 @@ extend-ignore-words-re = [
 ]
 
 [tool.cibuildwheel]
-manylinux-x86_64-image = "manylinux2014"
-manylinux-i686-image = "manylinux2014"
+manylinux-x86_64-image = "manylinux_2_28"
+manylinux-i686-image = "manylinux_2_28"
+manylinux-aarch64-image = "manylinux_2_28"
+manylinux-ppc64le-image = "manylinux_2_28"
+manylinux-s390x-image = "manylinux_2_28"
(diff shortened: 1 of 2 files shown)


mtreinish:
We shouldn't backport this, because it's too onerous to make this bump in a patch release. But for 0.19.0 I think we should just bump the manylinux version to avoid the issues we hit in cibuildwheel for 0.18.1 (I worked around the packaging things in #1653 by capping the numpy version in cibuildwheel).

IvanIsCoding (approved):
LGTM. For 0.19.0 I was planning on bumping many Linux, removing Python 3.10, downgrading macOS x86-64 from CI, adding Windows aarch64, dropping win32…. This is a good start

IvanIsCoding (approved):

PR #1661, opened 2026-08-01
Remove Python 3.10 support, bump MSRV to Rust 1.89, and start testing Python 3.15

IvanIsCoding:
* Python 3.10 is reaching EoL, most likely when rustworkx 0.19.0 is released it will be past the point to drop support for it.
* Rust MSRV was bumped to 1.93.1. We advanced our MSRV in roughly one year
* Python 3.15 beta is available, I think it's worth covering it in CI

Because of the MSRV bump, Clippy started finding new warnings. So most `.rs` changes are just addressing clippy


jakelishman:
Is it possible to be a bit more conservative with the MSRV?

Rustworkx (of course) isn't bound by Qiskit's own policy, but we're at 1.89 with [an MSRV policy](https://github.com/Qiskit/qiskit/blob/25955368cf14fb4f1b614a4fcd1dabf4b3c02e36/CONTRIBUTING.md#python-and-rust-versions) that requires a Rust version at least 1y old at the start of a minor-version series.  Rust 1.93 won't reach that til late January, after both the Qiskit 2.6 and 2.7 releases are already out, so if Rustworkx bumps now and we want to contribute changes here, we won't be able to pull them in until 2.8 in April next year.

IvanIsCoding:
Ah, I forgot to check Qiskit MSRV’s because generally Qiskit was ahead. Of course, Rust 1.89 will it be

IvanIsCoding:
> Is it possible to be a bit more conservative with the MSRV?
> 
> Rustworkx (of course) isn't bound by Qiskit's own policy, but we're at 1.89 with [an MSRV policy](https://github.com/Qiskit/qiskit/blob/25955368cf14fb4f1b614a4fcd1dabf4b3c02e36/CONTRIBUTING.md#python-and-rust-versions) that requires a Rust version at least 1y old at the start of a minor-version series. Rust 1.93 won't reach that til late January, after both the Qiskit 2.6 and 2.7 releases are already out, so if Rustworkx bumps now and we want to contribute changes here, we won't be able to pull them in until 2.8 in April next year.

Well that was easy because `if let` chains were introduced in 1.88. So the feature that caused the biggest diff and was the most convenience already landed. 1.89 it is

tox.ini
```diff
@@ -1,93 +0,0 @@
-[tox]
-minversion = 4.4.0
-envlist = py310, py311, lint
```
IvanIsCoding, line 3: After py310, py311 is the only config left for the (ancient) tox.ini we have. I will just drop it all together. We don't even use `black` or `setuptools-rust` anymore.

PR #1663, opened 2026-08-01
Stop publishing i686 and win32 wheels; start publishing Windows arm64 wheels

IvanIsCoding:
Related to #1662 

This PR adds Windows arm64 wheels and removes win32 and i686 wheels.

i686 is the strongest candidate to be removed. NumPy has not published i686 wheels for NumPy 2.x, so realistically that already narrows the amount of people that benefit from a pre-compiled wheel.

win32 still has NumPy wheels, but the situation of the Rust `numpy` crate for bridging Python and Rust has been best-effort for a while. So I am ok with removing it.

Lastly, there's Windows arm64. This is a new niche platform, but I think it would be reasonable to at least start publishing wheels for it. They'd be analogous to the macOS Intel wheels after #1662: tier 4 and cross-compiled, without tests. Adding the wheel is not hard. Maybe as time goes they become more useful for end-users.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -59,7 +59,7 @@ jobs:
     environment: release
     permissions:
       id-token: write
-    needs: ["build_wheels", "build-win32-wheels"]
+    needs: ["build_wheels"]
     steps:
       - uses: actions/download-artifact@v4
         with:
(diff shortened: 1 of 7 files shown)


IvanIsCoding:
Also: http://www.winarm64wheels.com/

For an ecosystem view of win arm64 wheels.

PR #1670, opened 2026-08-31
Raise RuntimeError instead of PanicException when pruning on finish_vertex

### LINKED ISSUE #1383: Panic on PruneSearch
When raising `PruneSearch` in the `finish_vertex` traversal event, the traversal methods raise `PanicError`.

I suggest raising `RuntimeError` instead. This properly inherits from `Exception` (and not only from `BaseException`) and has a specific type a user can `try: ...; except RunimeError: ...`.

Panic also has a side effect of printing to stderr a panic message (other than the python-raised exception traceback).
I think this stderr message should be avoided if possible, but I'm not sure there is an easy solution for this (other than the breaking change of `rustworkx-core` not panicking).

This was raised in issue https://github.com/Qiskit/rustworkx/issues/1361, which was closed in PR https://github.com/Qiskit/rustworkx/pull/1375.
I guess this point got drowned in  the documentation change discussions, so I'm opening a new issue for it here.
(The raised exception in the finish event was added to the `rustworkx` docs in that PR with the exception type unmentioned).

If there is already a PR for this, or this enhancement is unwanted feel free to close this issue.

juan52878911:
Fixes #1383.

Raising `PruneSearch` from the `finish_vertex` event reached `rustworkx-core`, which panics because there is nothing left to prune once a vertex is finished. The resulting `PanicException` derives from `BaseException`, so it was not caught by a plain `except Exception`, and a panic message was printed to stderr next to the Python traceback:

```
thread '<unnamed>' panicked at rustworkx-core/src/traversal/dfs_visit.rs:270:25:
Pruning on the `DfsEvent::Finish` is not supported!
pyo3_runtime.PanicException: Pruning on the `DfsEvent::Finish` is not supported!
```

The issue proposes raising a `RuntimeError` instead, which is what this does. The event is checked in the PyO3 handlers, before the control flow value reaches the core, so the core never panics and the stderr message is gone as well. `rustworkx-core` is left untouched: its `panic!` stays as an assertion for Rust callers, where returning `Prune` on `Finish` really is a programming error.

The panic was reachable from all three traversals, not just DFS (`dfs_visit.rs:270`, `bfs_visit.rs:243`, `dijkstra_visit.rs:283`), so all three handlers are fixed.

I checked the whole event surface to make sure the change is neither too narrow nor too wide: every event of the three traversals, with both `PruneSearch` and `StopSearch`, on `PyGraph` and `PyDiGraph`. `PruneSearch` on `finish_vertex` is the only combination that panicked, pruning on every other event still prunes, and `StopSearch` is unaffected. DFS propagating `StopSearch` back to the caller is existing documented behaviour and is left alone.

### Documentation

The docstrings for the three searches already noted that an exception was raised in that event but did not name the type, which is now `RuntimeError`. That note is duplicated in `rustworkx/__init__.py`, which is what `help()` and the published docs show, so both copies are updated. Those lines also referred to `finish_vertex` with `:class:` rather than `:meth:`, fixed in passing since they were being rewritten anyway.

### Tests

24 tests across the six `test_{bfs,dfs,dijkstra}_search.py` files, in `tests/graph/` and `tests/digraph/`:

- `prune_finish_vertex` — a `RuntimeError` is raised.
- `prune_finish_vertex_mid_traversal` — the visitor prunes on its second `finish_vertex`, which pins the exact finish order up to that point and shows the error is not raised eagerly.
- `prune_finish_vertex_no_starting_point` — same with `source=None`, so more than one component is walked.
- `stop_search_finish_vertex` — a guard that `StopSearch` in the same event is unaffected.

Reverting only the three handler files to `main` and rebuilding, **18 of the 24 fail** with the old `PanicException`; the 6 `stop_search_finish_vertex` tests pass either way, which is what makes them useful as guards.

`cargo test --workspace` is 524 passed and the Python suite is 2418 passed / 32 skipped, with `cargo fmt`, `cargo clippy --workspace --all-targets -- -D warnings`, `ruff format`, `ruff check`, `find_stray_release_notes.py` and `cargo doc -p rustworkx-core` with `-D warnings` all clean.

### DIFF
--- a/releasenotes/notes/prunesearch-finish-vertex-runtimeerror-9c4a1e7b2f83d605.yaml
+++ b/releasenotes/notes/prunesearch-finish-vertex-runtimeerror-9c4a1e7b2f83d605.yaml
@@ -1,0 +1,13 @@
+---
+fixes:
+  - |
+    Fixed :func:`~rustworkx.bfs_search`, :func:`~rustworkx.dfs_search` and
+    :func:`~rustworkx.dijkstra_search` raising a
+    ``pyo3_runtime.PanicException`` when a visitor raises
+    :class:`~rustworkx.visit.PruneSearch` from its ``finish_vertex`` event.
+    Pruning is not possible once a vertex is finished, so this was always an
+    error, but ``PanicException`` inherits from ``BaseException`` instead of
+    ``Exception``, so it was not caught by an ``except Exception`` block, and a
+    panic message was printed to ``stderr`` alongside the Python traceback. A
+    ``RuntimeError`` is now raised instead. Refer to
+    `#1383 <https://github.com/Qiskit/rustworkx/issues/1383>`__ for more details.
--- a/rustworkx/__init__.py
+++ b/rustworkx/__init__.py
@@ -1834,8 +1834,8 @@ def bfs_search(graph, source, visitor):
 
     .. note::
 
-        An exception is raised if the :class:`~rustworkx.visit.PruneSearch` is
-        raised in the :class:`~rustworkx.visit.BFSVisitor.finish_vertex` event.
+        A ``RuntimeError`` is raised if :class:`~rustworkx.visit.PruneSearch` is
+        raised in the :meth:`~rustworkx.visit.BFSVisitor.finish_vertex` event.
 
 
     :param graph: The graph to be used. This can be a :class:`~rustworkx.PyGraph`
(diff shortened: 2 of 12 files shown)


juan52878911:
Friendly ping on this one. All CI is green (including the free-threaded and MSRV matrix) and the patch coverage is 100%. Happy to adjust the error wording, test layout, or anything else if reviewers prefer a different shape.

